PR #47, opened 2020-08-27
Ignore vscode files

MichaelMcDonnell:
Ignore the Visual Studio Code files. I have set up project specific settings and git keeps complaining about the added files.


panarch:
Thanks! It makes sense, merging

PR #56, opened 2020-09-03
Add support for BETWEEN and NOT BETWEEN

ryanhossain9797:
Add support for `BETWEEN` queries

```sql
SELECT * FROM users WHERE age BETWEEN 20 AND 30 
```

```sql
SELECT * FROM users WHERE age NOT BETWEEN 20 AND 30 
```

### DIFF
--- a/src/executor/filter.rs
+++ b/src/executor/filter.rs
@@ -154,6 +154,16 @@ fn check_expr<'a, T: 'static + Debug>(
                 .next()
                 .unwrap_or(Ok(negated))
         }
+        Expr::Between {
+            expr,
+            negated,
+            low,
+            high,
+        } => {
+            let target = evaluate(expr)?;
+            let result = evaluate(low)? <= target && target <= evaluate(high)?;
+            Ok(if *negated { !result } else { result })
+        }
         _ => Err(FilterError::Unimplemented.into()),
     }
 }
--- a/src/executor/filter.rs
+++ b/src/executor/filter.rs
@@ -160,15 +160,17 @@ fn check_expr<'a, T: 'static + Debug>(
             low,
             high,
         } => {
+            let negated = *negated;
             let target = evaluate(expr)?;
-            let (low_eval, high_eval) = (evaluate(low)?, evaluate(high)?);
-            let (lower, higher) = if low_eval < high_eval {
-                (low_eval, high_eval)
-            } else {
-                (high_eval, low_eval)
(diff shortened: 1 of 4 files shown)


panarch:
Wow, it looks very good :)
I'm so glad of your contribution though the project does not have enough documentation yet.
In case of test case file, I like your choice.
GlueSQL should have a proper place to put `BETWEEN` like query test cases, but it doesn't have for now.
So, in current status, `tests/between.rs` looks good.
I'll add test case stuff by registering an another issue.

Then, only thing I'd like to mention is this,
```value 10 BETWEEN 20``` is enough to have this logic, ```10 <= value AND value <= 20```.
But current code includes
```sql
value 10 BETWEEN 20

min(10, 20) <= value AND value <= max(10, 20)
```
You don't need to concern the case ```20 BETWEEN 10```, this should simply be `false`.
Then, you can also simply `filter.rs` code you wrote.

panarch (changes requested):
Wow, it looks very good :)
I'm so glad of your contribution though the project does not have enough documentation yet.
In case of test case file, I like your choice.
GlueSQL should have a proper place to put `BETWEEN` like query test cases, but it doesn't have for now.
So, in current status, `tests/between.rs` looks good.
I'll add test case stuff by registering an another issue.

Then, only thing I'd like to mention is this,
```value 10 BETWEEN 20``` is enough to have this logic, ```10 <= value AND value <= 20```.
But current code includes
```sql
value 10 BETWEEN 20

min(10, 20) <= value AND value <= max(10, 20)
```
You don't need to concern the case ```20 BETWEEN 10```, this should simply be `false`.
Then, you can also simply `filter.rs` code you wrote.

panarch:
Oh, I just briefly searched `BETWEEN` use cases, like this https://www.postgresqltutorial.com/postgresql-between/
You have any cases which supports conversion from `BETWEEN 20 AND 10` to `BETWEEN 10 AND 20`?

ryanhossain9797:
Apparently the behavior differs. I tried it out in W3School's Live Editor and it worked, However apparently it doesn't work on Postgres or MSSQL Server which I tested just now. So I'll go ahead and remove it.

ryanhossain9797:
changes made
also addressed #57 and moved `between` test to `filter.rs`

panarch (approved):
Looks good! merging, thanks 👍

Issue #50, opened 2020-08-28
IS NULL, IS NOT NULL support


panarch:
```sql
SELECT * FROM Foo WHERE name IS NULL;
SELECT * FROM Bar WHERE name IS NOT NULL;
```

leoppro:
Could I implement it this weekend?

panarch:
Sure 👍 I'm really pleased that you get interested on contribution. 
For the first contribution, here are some info good to know at beginning.

* In case of expanding SQL supports, you need to be familiar with [sqlparser-rs](https://github.com/ballista-compute/sqlparser-rs) which is the parser GlueSQL depends on.
Mainly you need to look at will be `ast/*`, and in this case, it would be `enum Expr` in `ast/mod.rs`.

* `executor/filter.rs` handles `WHERE` clause, so that is where you need to look.
you can add `IS NULL` and `IS NOT NULL` supports.

* Before editing `src/` codes, it would be better to start by adding test cases.
In this case, `src/tests/nullable.rs` is proper position to modify.
Then you can simply test using `cargo test --all-features nullable`.

* After finishing works, then..
```
cargo fmt
cargo clippy --all-features
cargo test --all-features
```
All above commands pass, then it's all done :)  

If there is anything you want to ask or request, then please feel free to write that on this issue thread.  
Good luck!

PR #59, opened 2020-09-05
filter: support is null and not null expr

leoppro:
implement #50

### DIFF
--- a/src/tests/macros.rs
+++ b/src/tests/macros.rs
@@ -7,6 +7,15 @@ macro_rules! row {
 
 #[macro_export]
 macro_rules! select {
+    () => (
+        Payload::Select(vec![])
+    );
+    ( $( $t:path )*) => (
+        select!()
+    );
+    ( $( $t:path )* ;) => (
+        select!()
+    );
     ( $( $t:path )* ; $( $v: expr )* ) => (
         Payload::Select(vec![
             row!($( $t )* ; $( $v )* )
--- a/src/executor/evaluate/evaluated.rs
+++ b/src/executor/evaluate/evaluated.rs
@@ -251,8 +251,8 @@ impl<'a> Evaluated<'a> {
         match self {
             Evaluated::ValueRef(v) => v.is_some(),
             Evaluated::Value(v) => v.is_some(),
-            Evaluated::Literal(v) => *v != AstValue::Null,
-            Evaluated::LiteralRef(v) => **v != AstValue::Null,
+            Evaluated::Literal(v) => v != &AstValue::Null,
+            Evaluated::LiteralRef(v) => v != &&AstValue::Null,
             Evaluated::StringRef(_v) => true,
         }
     }
--- a/src/executor/evaluate/evaluated.rs
+++ b/src/executor/evaluate/evaluated.rs
@@ -246,6 +246,16 @@ impl<'a> Evaluated<'a> {
             StringRef(_) => unreachable(),
         }
     }
+
+    pub fn is_some(&self) -> bool {
+        match self {
+            Evaluated::ValueRef(v) => v.is_some(),
+            Evaluated::Value(v) => v.is_some(),
+            Evaluated::Literal(v) => *v != AstValue::Null,
+            Evaluated::LiteralRef(v) => **v != AstValue::Null,
+            Evaluated::StringRef(_v) => true,
+        }
+    }
 }
 
 fn literal_add(a: &AstValue, b: &AstValue) -> Result<AstValue> {
--- a/src/executor/filter.rs
+++ b/src/executor/filter.rs
@@ -164,16 +164,8 @@ fn check_expr<'a, T: 'static + Debug>(
 
             Ok(negated ^ (evaluate(low)? <= target && target <= evaluate(high)?))
         }
-        Expr::IsNull(expr) => Ok(match evaluate(expr)? {
-            Evaluated::ValueRef(v) => !v.is_some(),
-            Evaluated::Value(v) => !v.is_some(),
-            _ => false,
-        }),
-        Expr::IsNotNull(expr) => Ok(match evaluate(expr)? {
-            Evaluated::ValueRef(v) => v.is_some(),
-            Evaluated::Value(v) => v.is_some(),
-            _ => false,
-        }),
+        Expr::IsNull(expr) => Ok(!evaluate(expr)?.is_some()),
+        Expr::IsNotNull(expr) => Ok(evaluate(expr)?.is_some()),
         _ => Err(FilterError::Unimplemented.into()),
     }
 }
--- a/src/executor/filter.rs
+++ b/src/executor/filter.rs
@@ -94,7 +94,6 @@ fn check_expr<'a, T: 'static + Debug>(
 ) -> Result<bool> {
     let evaluate = |expr| evaluate(storage, filter_context, None, expr);
     let check = |expr| check_expr(storage, filter_context, expr);
-
     match expr {
         Expr::BinaryOp { op, left, right } => {
             let zip_evaluate = || Ok((evaluate(left)?, evaluate(right)?));
@@ -154,6 +153,16 @@ fn check_expr<'a, T: 'static + Debug>(
                 .next()
                 .unwrap_or(Ok(negated))
         }
+        Expr::IsNull(expr) => Ok(match evaluate(expr)? {
+            Evaluated::ValueRef(v) => !v.is_some(),
+            Evaluated::Value(v) => !v.is_some(),
+            _ => false,
+        }),
+        Expr::IsNotNull(expr) => Ok(match evaluate(expr)? {
+            Evaluated::ValueRef(v) => v.is_some(),
+            Evaluated::Value(v) => v.is_some(),
+            _ => false,
+        }),
         _ => Err(FilterError::Unimplemented.into()),
     }
 }
--- a/src/executor/filter.rs
+++ b/src/executor/filter.rs
@@ -94,6 +94,7 @@ fn check_expr<'a, T: 'static + Debug>(
 ) -> Result<bool> {
     let evaluate = |expr| evaluate(storage, filter_context, None, expr);
     let check = |expr| check_expr(storage, filter_context, expr);
+
     match expr {
         Expr::BinaryOp { op, left, right } => {
             let zip_evaluate = || Ok((evaluate(left)?, evaluate(right)?));
--- a/src/tests/nullable.rs
+++ b/src/tests/nullable.rs
@@ -1,4 +1,5 @@
 use crate::*;
+use Value::*;
 
 pub fn nullable(mut tester: impl tests::Tester) {
     tester.run_and_print(
@@ -12,8 +13,6 @@ CREATE TABLE Test (
     tester.run_and_print("INSERT INTO Test (id, num, name) VALUES (NULL, 2, \"Hello\")");
     tester.run_and_print("INSERT INTO Test (id, num, name) VALUES (1, 9, \"World\")");
     tester.run_and_print("INSERT INTO Test (id, num, name) VALUES (3, 4, \"Great\")");
-
-    use Value::*;
 
     let found = tester
         .run("SELECT id, num, name FROM Test")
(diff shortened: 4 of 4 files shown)


leoppro:
@panarch  PTAL

panarch:
Thanks! This looks very good starting, ok.
Now I also got what kind of comments I should mention, your work helps a lot, thanks.

Before comment, here is what `Evaluated` enum does.
I write explanation in here below, and also I'll add another issue to add comment to `Evaluated` enum.

There exists 5 types.
`LiteralRef`, `Literal`, `StringRef`, `ValueRef` and `Value`.
This many different types are for avoiding unnecessary `clone()` calls.

### `Literal` vs `Value`
`Literal` is value from parser(sqlparser-rs) and `Value` is from this repo, executor. so...

* `LiteralRef`
`100` in `WHERE 100 IS NULL`.
Number `100` is required but it doesn't need to take ownership from parsed sql (AST).

* `Literal`
`8 + 3` in `WHERE 8 + 3 IS NULL`.
`8` and `3` are `LiteralRef` but executor should take ownership of calculated value `8 + 3`.
So it is `Literal`.

* `ValueRef`
`name` in `WHERE name IS NULL`.
`name` is fetched from storage so it is `Value` but no need of taking ownership.

* `Value`
`name + 1` in `WHERE name + 1 IS NULL`.
`name` is `ValueRef` and `1` is `LiteralRef` but it requires `+` operation.
Then the result is `Value`! In this case, we know that the exact type of `name + 1` so there's no need of keeping `Literal` type.

* `StringRef`
`"Something"` in `WHERE "Something" IS NULL`.
It is similar with `LiteralRef` but `sqlparser-rs` handles `"Something"` string value in different way so `Evaluated` also takes this kind of seperated type.

So...
I'd like to request 3 changes.

1. `tests/nullable.rs` test cases should reach all 10 possible match arms. (5 in `IS NULL` and 5 in `IS NOT NULL`)
It looks good to add another function in `tests/nullable.rs`.
And it is also not for testing `blend` so it would be enough by counting number of fetched rows.
You can see that counting test case pattern in `tests/nested_select.rs`.

2. 
```rust
Expr::IsNull(expr) => Ok(match evaluate(expr)? {
    Evaluated::ValueRef(v) => !v.is_some(),
    Evaluated::Value(v) => !v.is_some(),
    _ => false,
}),
```
should be something like this below, it is required to handle all arms
```rust
Expr::IsNull(expr) => Ok(match evaluate(expr)? {
    Evaluated::ValueRef(v) => !v.is_some(),
    Evaluated::Value(v) => !v.is_some(),
    Evaluated::LiteralRef(v) ...
    Evaluated::Literal(v) ...
    Evaluated::StringRef(_) => false,
}),
```

3.
And If it is possible, it's better not to expose inner types of enum `Evaluated` to outside.
`filter.rs` is outside of `Evaluated` so, let's move `match` codes into `evaluated.rs`.
You can add some method like which `Value` has `is_some()`, `Evaluated` also can have that sort of method.

panarch (changes requested):
Thanks! This looks very good starting, ok.
Now I also got what kind of comments I should mention, your work helps a lot, thanks.

Before comment, here is what `Evaluated` enum does.
I write explanation in here below, and also I'll add another issue to add comment to `Evaluated` enum.

There exists 5 types.
`LiteralRef`, `Literal`, `StringRef`, `ValueRef` and `Value`.
This many different types are for avoiding unnecessary `clone()` calls.

### `Literal` vs `Value`
`Literal` is value from parser(sqlparser-rs) and `Value` is from this repo, executor. so...

* `LiteralRef`
`100` in `WHERE 100 IS NULL`.
Number `100` is required but it doesn't need to take ownership from parsed sql (AST).

* `Literal`
`8 + 3` in `WHERE 8 + 3 IS NULL`.
`8` and `3` are `LiteralRef` but executor should take ownership of calculated value `8 + 3`.
So it is `Literal`.

* `ValueRef`
`name` in `WHERE name IS NULL`.
`name` is fetched from storage so it is `Value` but no need of taking ownership.

* `Value`
`name + 1` in `WHERE name + 1 IS NULL`.
`name` is `ValueRef` and `1` is `LiteralRef` but it requires `+` operation.
Then the result is `Value`! In this case, we know that the exact type of `name + 1` so there's no need of keeping `Literal` type.

* `StringRef`
`"Something"` in `WHERE "Something" IS NULL`.
It is similar with `LiteralRef` but `sqlparser-rs` handles `"Something"` string value in different way so `Evaluated` also takes this kind of seperated type.

So...
I'd like to request 3 changes.

1. `tests/nullable.rs` test cases should reach all 10 possible match arms. (5 in `IS NULL` and 5 in `IS NOT NULL`)
It looks good to add another function in `tests/nullable.rs`.
And it is also not for testing `blend` so it would be enough by counting number of fetched rows.
You can see that counting test case pattern in `tests/nested_select.rs`.

2. 
```rust
Expr::IsNull(expr) => Ok(match evaluate(expr)? {
    Evaluated::ValueRef(v) => !v.is_some(),
    Evaluated::Value(v) => !v.is_some(),
    _ => false,
}),
```
should be something like this below, it is required to handle all arms
```rust
Expr::IsNull(expr) => Ok(match evaluate(expr)? {
    Evaluated::ValueRef(v) => !v.is_some(),
    Evaluated::Value(v) => !v.is_some(),
    Evaluated::LiteralRef(v) ...
    Evaluated::Literal(v) ...
    Evaluated::StringRef(_) => false,
}),
```

3.
And If it is possible, it's better not to expose inner types of enum `Evaluated` to outside.
`filter.rs` is outside of `Evaluated` so, let's move `match` codes into `evaluated.rs`.
You can add some method like which `Value` has `is_some()`, `Evaluated` also can have that sort of method.

leoppro:
Thanks for the thorough review. PTAL @panarch

leoppro:
I found an incompatible behavior with MySQL. In MySQL, the behavior regarding null is as follows:

**Schema (MySQL v5.7)**

    CREATE TABLE Test (
        id INTEGER NULL,
        num INTEGER,
        name TEXT
    )

---

**Load Data**

    INSERT INTO Test (id, num, name) VALUES (NULL, 2, "Hello");
    INSERT INTO Test (id, num, name) VALUES (1, 9, "World");
    INSERT INTO Test (id, num, name) VALUES (3, 4, "Great");

---
**Query 1**

    SELECT id, num FROM Test WHERE id + 1 IS NULL;

| id  | num |
| --- | --- |
| null | 2   |

---
**Query 2**

    SELECT id + 1 FROM Test;

| id + 1 |
| ------ |
| null |
| 2      |
| 4      |

---
**Query 3**

    SELECT id, num FROM Test WHERE id = NULL;

No results.

---
**Query 4**

    SELECT id, num FROM Test WHERE id IS NULL;

| id  | num |
| --- | --- |
| null | 2   |

---

But in GlueSQL, `NULL + 1 = 1` and `NULL = NULL` is `true`. Do we need to change the behavior on null to be consistent with MySQL?

panarch:
Wow, `NULL` has lots of stuff to think about, thanks for taking this carefully.
Existing code looks very good, and if you are ok then can I request small changes for `nullable` test codes?
Now `tests/nullable.rs` has lots of test cases so it would be good to simplify little bit, like this
* [src/tests/blend.rs](https://github.com/gluesql/gluesql/blob/main/src/tests/blend.rs)

And your second comment about MySQL compatible issue looks also very interesting.
My opinion is this.
1. `NULL IS NULL`, `NULL = NULL` is ok.
I think it's better to deal `NULL` as a type, not something systemic value.
Nowadays SQL is not only for programmers but also marketers and many others.
So, forcing them to understand something systemic issue is not affordable.
Without that, GlueSQL itself already handles `NULL` as a type, not something special else.
So naturally it's ok to see `NULL IS NULL` as true.

2. `NULL + 1` should be `NULL`.
Good point, I think it should be fixed and current implementation I made is wrong.
Existing implementation was simply because I didn't care about this case, I made that work during I'm working on `aggregate` task.
This one is related to `gluesql::data::Value` and `sqlparser::ast::Value as AstValue`.
I will simply divide using `value` for `gluesql::data::Value` and `literal` for `sqlparser::ast::Value as AstValue`.
There exists 3 possible scenarios.
* `literal + literal`
* `value + value`
* `value + literal`

But the last one, we don't need to care `value + literal` because it starts by converting literal to value. so it becomes `value + value` case.

Two things required to be changed.
* `literal + literal`
Those functions are defined in [src/executor/evaluate/evaluated.rs](https://github.com/gluesql/gluesql/blob/main/src/executor/evaluate/evaluated.rs)
`literal_add`, `literal_subtract`, `literal_multiply` and `literal_divide`.
Existings only take `AstValue::Number(_)` type but it should also handle `AstValue::Null`.

* `value + value`
It's defined in [src/data/value.rs](https://github.com/gluesql/gluesql/blob/main/src/data/value.rs).
In `impl Value`, there exists 4 functions `add`, `subtract`, `multiply` and `divide`.
These have wrong implementations, so you can simply fix it to show `NULL + 1` to `NULL`.
But the problem is, after you fix that, integration tests will fail. It is because of `SUM` aggregation.
In [src/executor/aggregate.rs](https://github.com/gluesql/gluesql/blob/main/src/executor/aggregate.rs),
```rust
"SUM" => {
    let value_to_sum = get_first_value(args)?;

    match aggregated.get(func) {
        Some((current, value)) => {
            if &index <= current {
                return Ok(aggregated);
            }

            Ok(aggregated.update(func, (index, value.add(value_to_sum)?)))
        }
        None => Ok(aggregated.update(func, (index, value_to_sum.clone()))),
    }
}
```
This uses `value.add(value_to_sum)?`, but after changing `NULL + 1` behavior, it should be also fixed.
You may be able to adapt similar approach from `COUNT` implementation.

---
So my opinion is..
* `NULL = NULL` is `true` => OK.
* `NULL + 1` should be `NULL` => Change required.

How do you think of this?

src/tests/macros.rs
```diff
@@ -7,6 +7,15 @@ macro_rules! row {
...
 macro_rules! select {
+    () => (
+        Payload::Select(vec![])
+    );
+    ( $( $t:path )*) => (
+        select!()
+    );
+    ( $( $t:path )* ;) => (
+        select!()
+    );
```
panarch, line 18: By the way, thanks! this looks good, now `select!` can take empty result. 👍

leoppro:
I agree with this: 

> * `NULL = NULL` is `true` => OK.
> * `NULL + 1` should be `NULL` => Change required.

On the mysql compatibility issue, I think I can submit another PR to fix it.
In this PR, I will continue to refine the test cases, but will not modify the null related behavior

panarch:
That sound good, I also agree that it's good to divide `NULL` compatibility issue using another PR.
After you refine the test cases, then I'll merge it.

leoppro:
@panarch PTAL

panarch:
Test cases looks perfect!
And sorry for the late comment but here is something that I'd like to get your opinion.
```rust
pub fn is_some(&self) -> bool {
    match self {
        Evaluated::ValueRef(v) => v.is_some(),
        Evaluated::Value(v) => v.is_some(),
        Evaluated::Literal(v) => *v != AstValue::Null,
        Evaluated::LiteralRef(v) => **v != AstValue::Null,
        Evaluated::StringRef(_v) => true,
    }
}
```

here, `*v != AstValue::Null` and `**v != AstValue::Null` uses dereferencing, and I usually prefer using `&` than `*`.
The reason is...
In this code we cannot check whether `v` in `Evaluated::Literal(v)` implements `Copy` trait or not.
So for making sure that `*v` doesn't call unnecessary `clone()`, we should check the code of `AstValue`.

* [sqlparser-rs - src/ast/value.rs](https://github.com/ballista-compute/sqlparser-rs/blob/main/src/ast/value.rs)
```rust
/// Primitive SQL values such as number and string
#[derive(Debug, Clone, PartialEq, Eq, Hash)]
#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
pub enum Value {
    /// Numeric literal
    #[cfg(not(feature = "bigdecimal"))]
    Number(String),
   ...
...
```
In this case, it's ok. that sqlparser `Value` does not implement `Copy` trait.

But when we use `v != &AstValue::Null`, we don't need to worry about unnecessary `clone()` may happen or not, because it will not happen.
So, I think using referencing is better choice, how do you think?

leoppro:
@panarch I agree with you, and I've made the relevant changes.

panarch:
Thanks! merging 👍

panarch (approved):

PR #67, opened 2020-09-13
change the behavior about null

### LINKED ISSUE #64: NULL + 1 should be NULL
https://github.com/gluesql/gluesql/pull/59#issuecomment-687709531

Current: 
`NULL + 1 => 1`

Expected:
`NULL + 1 => NULL`

leoppro:
fix #64


leoppro:
@panarch  PTAL

panarch (approved):
Looks good!

I firstly had a different idea about aggregation works `SUM` with null columns.
* `SUM` of `[1, NULL, 2]` should be `3`, not `NULL`.
For supporting that, extra exceptional modification in `aggregate.rs` is required.

But I like the current, we define the rule `NULL + 1 = NULL`, and current looks more consistent.

Thanks! merging

Issue #53, opened 2020-08-28
CREATE TABLE IF NOT EXISTS support


panarch:
`IF NOT EXISTS` syntax is not supported yet.
Let's make it work.

panarch:
It is expected to add `src/tests/create_table.rs`.

leoppro:
Hi, could I pick this up?

panarch:
Sure! 👍

leoppro:
Hi, @panarch. 
Through testing and reading the source code, I have found that the current behavior for creating tables is as follows:

 - When Table A does not exist, Table A is created and the result is successful.
 - When Table A already exists, create Table A. The result is successful and the old Table A will be overwritten.
 - The `IF NOT EXISTS` keyword will be ignored and will not work.

So I plan to implement the following two features:

- If Table A already exists, creating Table A should return an error.
- Make the `IF NOT EXISTS` keyword work.

Both of these features need to determine whether a particular table exists first, and I have observed that I can use the `Store::fetch_schema` function to know whether a table exists, but the problem is that the `Store` interface and the `StoreMut` interface don't seem to support transactions, So I can't guarantee that no other thread will make changes to the schema between the time I call `Store::fetch_schema` and `StoreMut::insert_schema`.

panarch:
* When Table A already exists, create Table A. The result is successful and the old Table A will be overwritten.
This is obviously... yes it's a bug.
I'm totally for your implementation plan.
It's time to make a new friend to lonely `StoreError::SchemaNotFound`. 😀

And about transaction issue, you're right.
`INSERT` query also has a same issue, because it also requires `Store::fetch_schema` first.

Current implementation does not have transaction functionality.

I'm planning to provide `Transaction` as an one of traits in `store/`, and in the case of `Transaction`, not like `Store` and `StoreMut` it will be optional.
Programmers can choose whether implementing `Transaction` trait or not.
That new `Transaction` trait will provide not only SQL transactions using `BEGIN` & `COMMIT`, but also it will make existing mutation queries(`CREATE`, `INSERT`...) atomic.

So, it would be ok to ignore transaction issue.

And I'll also share major feature release plan soon.
Briefly.. I'm thinking about bumping minor version when GlueSQL has a new `Store` traits.
* v0.2 - `ALTER TABLE`
* v0.3 - `FOREIGN KEY`
* v0.4 - `TRANSACTION`
* v0.5 - `INDEX`

ryanhossain9797:
[...]

PR #68, opened 2020-09-13
support if not exists for create table and if exist for drop table

leoppro:
support #53 

- check if the table exists when creating a table
- check if the table does not exist when dropping a table
- support if not exists for creating a table
- support if exists for dropping a table

### DIFF
--- a/src/executor/execute.rs
+++ b/src/executor/execute.rs
@@ -1,19 +1,20 @@
-use serde::Serialize;
 use std::fmt::Debug;
-use thiserror::Error;
 
+use serde::Serialize;
 #[cfg(feature = "alter-table")]
 use sqlparser::ast::AlterTableOperation;
 use sqlparser::ast::{ObjectType, SetExpr, Statement, Values};
+use thiserror::Error;
+
+use crate::data::{get_name, Row, Schema};
+use crate::parse::Query;
+use crate::result::{MutResult, Result};
+use crate::store::{AlterTable, Store, StoreMut};
 
 use super::fetch::{fetch, fetch_columns};
 use super::filter::Filter;
 use super::select::select_with_aliases;
 use super::update::Update;
-use crate::data::{get_name, Row, Schema};
-use crate::parse::Query;
-use crate::result::{MutResult, Result};
-use crate::store::{AlterTable, Store, StoreMut};
 
 #[derive(Error, Serialize, Debug, PartialEq)]
 pub enum ExecuteError {
@@ -29,6 +30,11 @@ pub enum ExecuteError {
     #[cfg(feature = "alter-table")]
     #[error("unsupported alter table operation: {0}")]
     UnsupportedAlterTableOperation(String),
+
+    #[error("TableNotExists")]
+    TableNotExists,
+    #[error("TableAlreadyExists")]
+    TableAlreadyExists,
 }
 
 #[derive(Serialize, Debug, PartialEq)]
@@ -60,9 +66,21 @@ pub fn execute<T: 'static + Debug, U: Store<T> + StoreMut<T> + AlterTable>(
     };
 
     match prepared {
-        Prepared::Create(schema) => storage
-            .insert_schema(&schema)
-            .map(|(storage, _)| (storage, Payload::Create)),
+        Prepared::Create {
+            schema,
+            if_not_exists,
+        } => {
+            if storage.fetch_schema(&schema.table_name).is_ok() {
+                return if if_not_exists {
+                    Ok((storage, Payload::Create))
+                } else {
+                    Err((storage, ExecuteError::TableAlreadyExists.into()))
+                };
+            }
+            storage
+                .insert_schema(&schema)
+                .map(|(storage, _)| (storage, Payload::Create))
+        }
         Prepared::Insert(table_name, rows) => rows
             .into_iter()
             .try_fold((storage, 0), |(storage, num), row| {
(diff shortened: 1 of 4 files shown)


panarch (changes requested):
Behavior looks perfect!
You even work on `DROP SCHEMA` exception, too. That's great.

And let's just only discuss about internal implementation.
There are two stuffs to see
1. `Error` is handled as proper business logic
```rust
 Err((storage, Error::Store(StoreError::SchemaAlreadyExists))) => {
    if if_not_exists {
        Ok((storage, Payload::Create))
    } else {
        Err((storage, Error::Store(StoreError::SchemaAlreadyExists)))
     }
}
```
In this code, we encounter `StoreError::SchemaAlreadyExists`, but in specific condition `if_not_exists` then it transforms into `Ok`.
It can have some design issues, `Error` is better to be handled as "Terminal" state, not "Non-terminal" one.
But here, the role of "Error" is expanded to "Non-terminal" and I think we'd better to keep as narrow as possible, keeping error's role as terminal.
So, once we see the error then we should return, not changing into `Ok`.

2. Store trait `insert*` consistency.

Store trait has names of `insert` for `insert_schema` and `insert_data`.
`insert_data` overwrites no matter whether existing data is or not.
but now `insert_schema` doesn't.
We need to make the rule consistent and simple.
* [HashMap - insert](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.insert)
* [im_rc::HashMap - insert](https://docs.rs/im/15.0.0/im/hashmap/struct.HashMap.html#method.insert)
* [sled - insert](https://docs.rs/sled/0.34.4/sled/struct.Db.html)
(Let's just ignore returning existing value as a return parameter for now.)

And I think it's better not to consider whether there already exists data or not, simply overwritting is better to choose, how do you think of this?
And then move `SchemaAlreadyExists` logics into `execute.rs`, not in `Store`.
I'd like to say sorry about previous guide on issue, I simply thought it might be naturally `StoreError::SchemaAlreadyExists` but.. I was better not to specify the location `StoreError`.


### So...
How about this?
* Making `execute.rs` to handle that logics `SchemaAlreadyExists` and failure on dropping table for not found.

It means, it requires to take `sled_storages.rs` code changes back and apply it to `execute.rs`.
`execute` function directly handles `if_not_exists` and `if_exists` without affecting `sled_storage.rs` implementation.

leoppro:
I'm sorry I've been busy at work and haven't been able to update my PR, I will update it soon.

panarch:
Please don't feel sorry at all, I hope you enjoy writing codes here. It's fully ok because it is not your primary work.
If there exists something blocks other issues, then I will request you something.
But in the case of this issue, it's independent from any other ones so you can simply spend time by only following your own schedule.

leoppro:
> In this code, we encounter StoreError::SchemaAlreadyExists, but in specific condition if_not_exists then it transforms into Ok.
It can have some design issues, Error is better to be handled as "Terminal" state, not "Non-terminal" one.
But here, the role of "Error" is expanded to "Non-terminal" and I think we'd better to keep as narrow as possible, keeping error's role as terminal.
So, once we see the error then we should return, not changing into Ok.

hi @panarch , I don't quite understand your first comment, how should I handle the error?

src/executor/execute.rs
```diff
@@ -45,9 +46,20 @@ pub fn execute<T: 'static + Debug, U: Store<T> + StoreMut<T>>(
...
-            .insert_schema(&schema)
-            .map(|(storage, _)| (storage, Payload::Create)),
+        Prepared::Create {
+            schema,
+            if_not_exists,
+        } => match storage.insert_schema(&schema) {
+            Ok((storage, _)) => Ok((storage, Payload::Create)),
+            Err((storage, Error::Store(StoreError::SchemaAlreadyExists))) => {
+                if if_not_exists {
+                    Ok((storage, Payload::Create))
```
panarch, line 57: @leoppro 
Which I mentioned before, "Terminal" and "Non-terminal" were something more conceptual.
For now, what we need to see is simple.
On `match storage.insert_schema(&schema)`, though it returns error with `Err((storage, Error::Store(StoreError::SchemaAlreadyExists)))` but it transforms into `Ok((storage, Payload::Create))` on the specific condition `if if_not_exists`.

Once an error occurred, it's ok to change into another error message but, not good to change into non-error.

What I requested for the first one "Error is handled as proper business logic",
you can simply think as removing work of `Err` -> `Ok` cases.

leoppro: @panarch I understand what you are suggesting is this: I need to return another error and let the business logic layer decide whether to ignore that error. When `if_not_exists` is true and the schema already exists, I should return an error like `SchemaAlreadyExistsButCanIgnore`, then let the business logic layer(which is also the user) handle the error.

leoppro: I think there are some drawbacks to this: 
1. It's more unique and behaves differently from other similar databases, such as sqlite and MySQL. 
2. For the ORM framework, in order to adapt to gluesql, they need to implement special error handling logic.

leoppro: To solve this problem, I suggest adding a warning output to the `Glue::execute`, some errors that can be ignored are output to the user by warnings

panarch: @leoppro 
Ok, now I think I also got what you're not sure about.
At first, 
> I need to return another error and let the business logic layer decide whether to ignore that error.
My description wasn't enough, what I wanted to say was which "business logic layer decide whether to ignore that error" should not happen.
Once error occurs, then it's an error, it should not be ignored or accepted.

Current nested 2-level error matches can be reduced into one.
Like this,
* Replace `StoreError::SchemaAlreadyExists` to `ExecuteError::SchemaAlreadyExists`
* Revert `sled_storage.rs` changes and move that business logic codes to `execute.rs`
It means, `execute.rs` will check whether schema exists or not. (not in `storage`)

How do you think?

panarch: > It's more unique and behaves differently from other similar databases, such as sqlite and MySQL.
Your code changes are already good enough, I don't think we need to change behaviors which affects to users.
My opinion is only about internal implementation, not about outer behavior.

leoppro:
@panarch PTAL again

src/executor/execute.rs
```diff
@@ -60,9 +66,21 @@ pub fn execute<T: 'static + Debug, U: Store<T> + StoreMut<T> + AlterTable>(
...
-            .map(|(storage, _)| (storage, Payload::Create)),
+        Prepared::Create {
+            schema,
+            if_not_exists,
+        } => {
+            if storage.fetch_schema(&schema.table_name).is_ok() {
+                return if if_not_exists {
+                    Ok((storage, Payload::Create))
+                } else {
+                    Err((storage, ExecuteError::TableAlreadyExists.into()))
```
panarch, line 77: 👍 I like this.
You transformed `StoreError::SchemaNotFound` to a different and more informative error.
I haven't thought about this before but it looks necessary to do, thanks :)

panarch (approved):
Thanks :) I worried that there are too much changes I made so it may bothers you to keep work on this issue.

Looks beautiful, merging!

PR #69, opened 2020-09-13
Add support for Inserting Multiple rows

ryanhossain9797:
`INSERT INTO Table (id, age) VALUES (1, 10), (2, 15);`

### DIFF
--- a/src/data/row.rs
+++ b/src/data/row.rs
@@ -90,7 +90,7 @@ impl Row {
 
                         match literal {
                             Expr::Value(literal) => {
-                                Value::from_data_type(data_type.clone(), nullable, literal)
+                                Value::from_data_type(&data_type, nullable, literal)
                             }
                             Expr::Identifier(Ident { value, .. }) => Ok(Value::Str(value.clone())),
                             _ => Err(RowError::UnsupportedAstValueType.into()),
--- a/src/data/row.rs
+++ b/src/data/row.rs
@@ -45,7 +45,11 @@ impl Row {
             .ok_or_else(|| RowError::ConflictOnEmptyRow.into())
     }
 
-    pub fn new(column_defs: Vec<ColumnDef>, columns: &[Ident], source: &Query) -> Result<Self> {
+    pub fn new(
+        column_defs: Vec<ColumnDef>,
+        columns: &[Ident],
+        source: &Query,
+    ) -> Result<Vec<Self>> {
         let values = match &source.body {
             SetExpr::Values(Values(values)) => values,
             _ => {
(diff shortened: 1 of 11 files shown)


ryanhossain9797:
@panarch take a look?
I'm not sure if this was the right way to go about it.

src/executor/execute.rs
```diff
@@ -48,11 +48,19 @@ pub fn execute<T: 'static + Debug, U: Store<T> + StoreMut<T>>(
...
+            match rows.into_iter().try_fold(
+                (storage, Vec::<Row>::new()),
+                |(storage, mut rows), row| {
+                    let (storage, key) = storage.generate_id(&table_name)?;
+                    let (storage, row) = storage.insert_data(&key, row)?;
+                    rows.push(row);
+                    Ok((storage, rows))
+                },
+            ) {
+                Ok((storage, rows)) => Ok((storage, Payload::Insert(rows))),
```
panarch, line 61:
```rust
rows.into_iter().try_fold(...).map(|(storage, rows)| (storage, Payload::Insert(rows))
```
Changing value type is only required for `Ok`, then you can simply use `map` rather than `match`.

src/data/row.rs
```diff
@@ -45,58 +45,64 @@ impl Row {
...
-                }
+        if values.is_empty() {
+            return Err(RowError::Unreachable.into());
+        }
+
+        values
+            .iter()
+            .map(|values| {
+                column_defs
+                    .clone()
```
panarch, line 68: The only reason that `Row::new` takes ownership of `column_defs` is that it is fetched by `prepare` function so at least someone should take it.
But at the same time, for the code logic, it's actually enough to use `column_defs` by borrowing.

This `clone()` calls are unnecessary, you will be able to remove this `clone()` calls.

src/executor/execute.rs
```diff
@@ -127,9 +135,9 @@ fn prepare<'a, T: 'static + Debug>(
         } => {
             let table_name = get_name(table_name)?;
             let Schema { column_defs, .. } = storage.fetch_schema(table_name)?;
-            let row = Row::new(column_defs, columns, source)?;
+            let rows = Row::new(column_defs, columns, source)?;
```
panarch, line 138: Good details 👍

src/executor/execute.rs
```diff
@@ -48,11 +48,19 @@ pub fn execute<T: 'static + Debug, U: Store<T> + StoreMut<T>>(
...
             .map(|(storage, _)| (storage, Payload::Create)),
-        Prepared::Insert(table_name, row) => {
-            let (storage, key) = storage.generate_id(&table_name)?;
-            let (storage, row) = storage.insert_data(&key, row)?;
-
-            Ok((storage, Payload::Insert(row)))
+        Prepared::Insert(table_name, rows) => {
+            match rows.into_iter().try_fold(
+                (storage, Vec::<Row>::new()),
+                |(storage, mut rows), row| {
```
panarch, line 54: I recognized this `mut` and I also recently stuck at collecting vector using immutable api.
Even `im_rc::Vector` doesn't have sufficient immutable apis, recently I'm thinking about making immutable wrapper for `std Vector` and `std HashMap`.
Anyway for this issue, let's take a different approach.

At the beginning when I decided the result type of `INSERT` as `Row`, it wasn't because of something special reason. I just chose it. And the situation which only requires to return a single `Row`, it was ok, just for a single row.
But it looks inefficient to return all successfully inserted row.
Like `DELETE` and `UPDATE` queries, it would be just enough to return number of affected rows, not full row data.

Let's remove it at this point, but using the different issue.
https://github.com/gluesql/gluesql/issues/70

For this issue, I'll merge PR with that `mut` use.
You can remove it working on issue #70, can you please also take a look that issue?

panarch (changes requested):
Thanks! looks good and only little change looks required.

* Removing `MultiRowInsertNotSupported` error
```rust
#[derive(Error, Serialize, Debug, PartialEq)]                                                                                                                                                         
pub enum RowError {
    ....
    #[error("inserting multiple rows not supported")]
    MultiRowInsertNotSupported,
    ....
}
```
Thanks to your contribution, we don't need this error anymore.
You can simply remove it, you may not see any warning from cargo clippy because it is exposed to outside.
So like this kind of elimination work, we need to take care by ourselves.

---
Except for this one, I add comments to codes. Thanks 👍 👍

src/executor/execute.rs
ryanhossain9797: Sure, I'll get on it ASAP

src/data/row.rs
ryanhossain9797: I had already tried that already, It does work but then it requires `data_type` on line 96 to be cloned. would that be fine?

panarch: @ryanhossain9797 Good point, oh. I also saw that `Value::from_data_type`, I think it's better to change `Value::from_data_type` to borrow `data_type` rather than taking ownership.
`Value::from_data_type` is only used here so let's change the type to `&DataType`.

fyi. Though `Value::from_data_type` is used only in `row.rs` but I put this method in `Value` is the case when we think about adding a new value type.
If we consider to add a new value, for example, may be.. `I32`, 32 bit signed integer, at that case if we need to take some effort about where to change, that can be quite tough.
But I wanted to simply add new value type by only modifying `value.rs` file, without thinking about any other files.

panarch (changes requested):
https://github.com/gluesql/gluesql/pull/69#discussion_r487694840
Only except this tiny one that everything looks good :)
I'll merge this right after that.

src/data/row.rs
ryanhossain9797: Yep I can see why it's useful
Anyway, done with changes

panarch (approved):

panarch:
Awesome, thanks!

PR #71, opened 2020-09-14
Return row count instead of rows on Insert

### LINKED ISSUE #70: Chainging return type of `INSERT`
Current
```rust
Payload::Insert(Vec<Row>)
```

Expected
```rust
Payload::Insert(usize)
```

Let's simply return a number of inserted rows.

ryanhossain9797:
Return number of successfully inserted rows instead of the entire rows on INSERT queries
resolves #70


panarch (approved):
Wow you did really fast, thanks! Looks good, merging

Issue #76, opened 2020-09-19
 'SingleQuotedString' is not accepted in inserts


willy610:
Problem with insert of text.

Insert into  tableA values ("WORKS BUT NOT STANDARD" , 'STANDARD BUT IS NOT ACCEPTED');

Most exports from other databases uses SingleQuotedString. More Standard SQL.

Seems like expressions with SingleQuotedString works

willy610:
I have tested/verified  corrections on a cloned repos.

How to proceed?

Create a new Branch?

Or await something from the community ?

panarch:
You can fork this repository and then GitHub will make cloned repository, url may become https://github.com/willy610/gluesql
Then, work on that cloned repo. After that you can make pull request to this main repo, then I can review your code change.

This process is based on GitHub, so if you are familiar with git but not GitHub, then you may spend some time to get used to it.

* https://docs.github.com/en/github/getting-started-with-github/fork-a-repo

This page above looks good to see.

willy610:
It seems like the bug remains in the latest issue version = "0.1.16"

panarch:
Yes, because no one takes this issue yet.
If you'd like to do this then you can add PR and I'll review and merge.

willy610:
In ``glusql-main-3`` literals with SingleQuotedString is accepted but nothing is stored.
Works fine with double quote.
Perhaps some unit tests is to be introducde in ``/tests/basic.rs``

``INSERT INTO Test (id, num, name) VALUES (1,2,'Job'):``

``INSERT INTO Test (id, num, name) VALUES ;``

willy610:
Of cource should be with escaped strings

``INSERT INTO Test (id, num, name) VALUES (3,4,'Job''s Macintosh');``

KyGost:
Looks like it sorts itself out for every command but `INSERT`. Working on this now.

KyGost:
#145

KyGost:
@panarch close?

panarch:
@KyGost Oh, thanks :) I'll close

Issue #73, opened 2020-09-18
Testdata from file and/or stdin


willy610:
Love this project.
Have a lot of sql database dumps I want to have for verification
A separate bin project can't refer to locally libraries.

But a rough approach is to introduce tests for loading sql from a file instead of inline strings

Give me a hint where to start in my downloaded glusql-main

panarch:
Thanks!

> But a rough approach is to introduce tests for loading sql from a file instead of inline strings
This sounds really good, I agree that running tests from files can be certainly useful when we consider to run other database's sql test cases for verification.

For the beginning, it would be good to see [examples/api_usage.rs](https://github.com/gluesql/gluesql/blob/main/examples/api_usage.rs)

It would be good to add a new example code which loads sql file and runs.

willy610:
I work like an anarchist for a moment. Off group.
Align later on in some way, if it suites you

(The published gluesql = { version = "0.1.15", features = ["sled-storage"] } is not the same as what's committed on the project,
alright? So there is no multi line insert and others I need when importing more volumes)

So I stick to the downloaded source and will try to add test case for file load.

By the way I have already modified so that 'SingleQuotedString' is accepted in sql-statements. 
More SQL-standard.
(Verified at insert sofar).
Today only DoubleQuotedString is accepted.

panarch:
That sounds good,
Yes that's true, after "0.1.15" released, there were lots of improvements newly made, thanks to @ryanhossain9797 and @leoppro 
New version release is required but I didn't do that yet.

I was planning to release "0.1.16" after I add `GROUP BY` supports and I estimated `GROUP BY` issue as a single week one.
But.... it took quite a long more, so that's the reason of this delay.
I'll release new version next week, then you will be able to use using version numbers.

Also thanks a lot for reporting bug, yeah at the beginning I wasn't aware about that 'SingleQuotedString' is not SQL-standard.
If you are ok, can you please add this 'SingleQuotedString' issue to issue stack with problematic SQL samples?

willy610:
It's a pitty column expression and table expression can't be named and referenced

#80, #82

panarch:
@willy610 That issues don't mean that table expressions can't be named and referenced, it is still possible.

* #82 is an independent issue.
* #80 is a prior issue for providing name & reference functions.

panarch:
This is too far to think at this time.
Let's make it again when we can consider more details.

PR #87, opened 2020-10-02
Issue #76 SingleQuotedString

willy610:
The original SQL did use single quote as string embeddings. And most sql backups use SingleQuotedString. A good source for testdata


panarch:
related to #87

panarch:
Changes look very good, no issue on your final changes. Thanks!
But only one thing, can I request you to remove unrelated commits which is related to #73 ?
By merging commits into a single one, or you can also make another PR without that commits.

willy610:
Try to 'remove' in order to clean up

PR #89, opened 2020-10-05
Report an error for inserting more values than table columns

### LINKED ISSUE #62: INSERT with more values is accepted
```sql
CREATE TABLE Test (id INTEGER);
INSERT INTO Test VALUES (100, 200);
```

below insertion query should report an error but it is accepted.
* [src/data/row.rs](https://github.com/gluesql/gluesql/blob/main/src/data/row.rs)

Current `Row` struct cares lack of values or columns but not handling more values properly.

silathdiir:
Closes #62 .

Fixes to report an error `TooManyValues` if inserting more values than target columns of the specified table.

### DIFF
--- a/src/data/row.rs
+++ b/src/data/row.rs
@@ -17,6 +17,9 @@ pub enum RowError {
     #[error("literals does not fit to columns")]
     LackOfRequiredValue(String),
 
+    #[error("literals have more values than target columns")]
+    TooManyValues,
+
     #[error("unsupported ast value type")]
     UnsupportedAstValueType,
 
@@ -61,6 +64,10 @@ impl Row {
         values
             .iter()
             .map(|values| {
+                if values.len() > column_defs.len() {
+                    return Err(RowError::TooManyValues.into());
+                }
+
                 (&column_defs)
                     .iter()
                     .enumerate()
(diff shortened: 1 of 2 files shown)


panarch (approved):
Everything looks good 👍 Thanks! merging

panarch:
@silathdiir Oh, I just found I missed a feedback I had to leave.

https://github.com/gluesql/gluesql/commit/4c8d18d5b67bd6b982631a7ba1da5bca35e1881c

```rust
if values.len() > column_defs.len() {
...
```
This statement can be simply checked outside of iteration, I added a quick fix using the commit I mentioned above.

silathdiir:
Hi @panarch , thanks. I will try to test if it works.

I doubted that outside `values` has another mean (so I added the test case as below in this PR).

```
INSERT INTO TableA VALUES (100), (100, 200);
```

outside values = `(100), (100, 200)`
inside values 1 = `(100)`
inside values 2 = `(100, 200)`

I could not confirm that. Will test again. Thanks.

panarch:
@silathdiir  Oh!!! While I was stuck at other hard issue, I too rushed on this one.
Haha I made a quite silly mistake, please ignore my last comment.
Thanks so so much for the fast response.

I think it wasn't a good idea I set two different variable names using a same `values`.
I used same `values` for `Vec of Vec` and `Vec`, I was entirely confused of this at this point.
Your work is correct, I'll revert the commit I mentioned.

panarch:
@silathdiir FYI. I did some supplementary work related to the mistake I made.
https://github.com/gluesql/gluesql/pull/90

* Flatten nested value iteration
* Remove reachable `Unreachable` error

Issue #51, opened 2020-08-28
ALTER TABLE support


panarch:
There exists two store traits: `Store` and `StoreMut`.

Plan is adding new `Alter` store trait, but this time, `Alter` trait will be optional.

Provide 3 SQL statements
* ADD COLUMN
* DROP COLUMN
* RENAME COLUMN

PR #92, opened 2020-10-12
Implement ALTER TABLE - ADD, DROP, RENAME COLUMN & RENAME TABLE 

panarch:
Implement ALTER TABLE functionality.

* ADD COLUMN
* DROP COLUMN
* RENAME COLUMN
* RENAME TABLE
are now supported.

`AlterTable` trait is optional, who wants to use `AlterTable` should enable `alter-table` feature.


panarch:
resolve #51

Issue #82, opened 2020-09-21
SELECT foo AS bar ... support,


panarch:
mentioned in https://github.com/gluesql/gluesql/issues/80

```sql
SELECT id AS SOMEALIASCOLUMNNAME, num FROM Test
```

Current [blend.rs](https://github.com/gluesql/gluesql/blob/main/src/executor/blend.rs) is not handling `AS`.
Simple fix, `blend` does not need to work on `AS` at all, just ignore `AS` and simply evaluate column expression.

willy610:
Seems to having a solution in blend.rs as said above.
On both indents and expressions

```
SELECT ingredientid,ingredientid+2000 as PLUS2000 FROM ingredient 
WHERE energy > 500 
ORDER BY 1;

ingredientid    PLUS2000
15      2015
46      2046
128     2128
181     2181
253     2253
```

```
SELECT ingredientid AS NEWCOLNAME 
FROM   ingredient 
WHERE  ingredientname = 'mjölk';

NEWCOLNAME
169
```

PR #96, opened 2020-10-16
fix issues/82

zhangchunzhong:
first time to commit a rust code, please feel freely to give comments.Thanks in advance!

### DIFF
--- a/src/data/value.rs
+++ b/src/data/value.rs
@@ -375,10 +375,10 @@ impl Value {
     pub fn is_some(&self) -> bool {
         use Value::*;
 
-        match self {
-            Empty | OptBool(None) | OptI64(None) | OptF64(None) | OptStr(None) => false,
-            _ => true,
-        }
+        !matches!(
+            self,
+            Empty | OptBool(None) | OptI64(None) | OptF64(None) | OptStr(None)
+        )
     }
 }
 
--- a/src/executor/blend.rs
+++ b/src/executor/blend.rs
@@ -155,47 +155,48 @@ impl<'a, T: 'static + Debug> Blend<'a, T> {
                         None => err!(BlendError::TableNotFound(table_alias.to_string())),
                     }
                 }
-                SelectItem::UnnamedExpr(expr) => match expr {
-                    Expr::Identifier(ident) => match get_value(&context, &ident.value) {
-                        Some(value) => Blended::Single(once(Ok(value))),
-                        None => err!(BlendError::ColumnNotFound(ident.to_string())),
-                    },
-                    Expr::CompoundIdentifier(idents) => {
-                        if idents.len() != 2 {
-                            return err!(BlendError::FieldDefinitionNotSupported);
-                        }
-
-                        let table_alias = &idents[0].value;
-                        let column = &idents[1].value;
-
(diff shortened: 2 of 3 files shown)


panarch (changes requested):
So welcome of your first pr, thanks :)

Most of changes basically look good and I want to request small changes.
`blend.rs` has some confusing codes but you did well.
Let us begin with explaining background story on this.
What `blend.rs` does is this,
In `SELECT a, b, c FROM Foo;`
Using fetched rows from `Foo` table, in `GlueSQL` we call it as `Context`.
`blend.rs` consumes `Context` and extracts only specified columns `a, b, c`.

To make that work, as you saw `blend` parses `SelectItem`.
Though there certainly exists a module named `evalute` which takes borrowed `Context` as a param and return evaluated value, `blend.rs` `blend` function works in a little bit different way.
`evaluate` works by `clone` value from `Context`.
When we run simple `blend` work, like `a, b, c`, we actually don't need to `clone`, we can simply take ownership of that value from `Context`.
Some customized code in `blend.rs` is all for avoiding unnecessary `clone` calls.
That optimizations are only applied for simple identifiers.
Complex ones like `Expr::BinaryOp { .. } | Expr::Function(_)`, `blend` simply calls `evaluate`.

----

Now the review part.

1. Currently there are too many separated commits, can you reduce into two commits?
May be you need to use force push on this.

* The changes you made on `src/data/value.rs`
Thanks for doing this stuff, and this is independent from main work so it looks better to split into a different commit.
* And all other changes

2. Rather than mentioning an issue number, please change to use sentence format with description

Though you don't need to write too many details, It's required to be recognizable without opening the issue page.

3. Merging `SelectItem::ExprWithAlias` into `SelectItem::UnnamedExpr`

`match` in Rust is so.. beautifully powerful.
No redundant codes are required.
You can actually use like this below,
```rust
SelectItem::UnnamedExpr(expr) |
SelectItem::ExprWithAlias { expr, .. } => match expr {
..
```

And also you don't need to specify unused identifiers,
```rust
Expr::BinaryOp { left: _, op: _, right: _ }

// It is enough to use this
Expr::BinaryOp { .. }
```

4. Adding `AS` test cases to `src/tests/blend.rs` can be enough

I think it's too much to have separate test cases only for this aliased expressions.
Can you remove `select_as.rs` then add test cases using `AS` in `src/tests/blend.rs`?

---
Please feel free to mention back anything you need more, thanks 👍

zhangchunzhong:
The changes made on src/data/value.rs is just to pass the "cargo clippy" in github action:
https://github.com/gluesql/gluesql/pull/96/commits/0357243125ad64bf344f0aedb976a80f4b09e835

panarch:
@zhangchunzhong Yes, that's right. It's just because changes are made by https://github.com/gluesql/gluesql/pull/96/commits/0357243125ad64bf344f0aedb976a80f4b09e835 and https://github.com/gluesql/gluesql/pull/96/commits/858de2ed67d98207f756a689e046dd8c6bd4b5a9

panarch:
resolve #82

panarch:
@zhangchunzhong Thanks! merging

PR #98, opened 2020-10-18
Add aliased projection support - AS keyword,

### LINKED ISSUE #82: SELECT foo AS bar ... support,

panarch:
SELECT foo AS bar FROM Table1;
AS keyword in projection(blend.rs) is now supported

resolve #82

### DIFF
--- a/src/data/value.rs
+++ b/src/data/value.rs
@@ -375,10 +375,10 @@ impl Value {
     pub fn is_some(&self) -> bool {
         use Value::*;
 
-        match self {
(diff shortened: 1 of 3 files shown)


panarch:
https://github.com/gluesql/gluesql/pull/96

PR #103, opened 2020-10-24
Projection aliases support for SELECT queries

### LINKED ISSUE #83: Getting SELECT projection aliases
Related to #80 - due to long threads, it's hard to recognize what exact task is, so let's use this issue.

input:
```sql
SELECT a, b AS foo, SUM(id) AS bar FROM TableA;
```
output:
```
vec!["a", "foo", "bar"]
```

This issue is adding a new function `get_projection_aliases` to `src/parse.rs` which takes `sqlparser::ast::Query` reference and return column aliases, `Vec<String>`

```
fn get_projection_aliases(query: &sqlparser::ast::Query) -> Vec<String>
```

panarch:
Now the result of SELECT query returns both aliases and rows.
Implement support for fetching aliases.

e.g.
```sql
SELECT id, name as foo FROM TableA;
```

The query above provides not only fetched rows, but also it serve aliases like `["id", "foo"]`.

Change `Payload::SELECT` from
```rust
Payload::SELECT(Vec<Row>)
```
to
```rust
Payload::SELECT {
  aliases: Vec<String>,
  rows: Vec<Row>,
}
```

resolve #83

### DIFF
--- a/src/executor/blend.rs
+++ b/src/executor/blend.rs
@@ -41,15 +41,7 @@ impl<'a, T: 'static + Debug> Blend<'a, T> {
 
     pub fn apply(&self, context: Result<AggregateContext<'a>>) -> Result<Row> {
         let AggregateContext { aggregated, next } = context?;
-
-        let values = self
-            .blend(aggregated, next)?
-            .into_iter()
-            .map(|value| match Rc::try_unwrap(value) {
-                Ok(value) => value,
-                Err(value) => (*value).clone(),
-            })
-            .collect();
+        let values = self.blend(aggregated, next)?;
 
         Ok(Row(values))
     }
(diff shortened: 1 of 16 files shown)


panarch:
@willy610 Thanks for reporting this issue, now it works!

willy610:
Nice implementation.

I think the public sql programmer in interfacing with the select thinks in terms of ``columns`` or ``labels`` or similar more than ``alias``. 

I think ``alias`` is more ``AST``

 Could be confusing also because all in the projection does not for sure have any ``AS NAME`` but sticks to the column name from a table.

Perhaps labels is preferred??

panarch:
@willy610 That sounds... great.
I agree of the point you mentiond, `alias` sounds more.. related to `AST`, it may not fit to user-side names.

`labels` sounds good and I'll do work on this soon. thanks!

Issue #102, opened 2020-10-23
UPPER & LOWER function support


panarch:
Currently there is no support for non-aggregating functions.

`UPPER("asdfS") -> "ASDFS"` can be a first example of this non-aggregating function.

PR #106, opened 2020-10-26
Support UPPER & LOWER non-aggregating functions

panarch:
It starts with `UPPER` and `LOWER` functions.

Let's add more non-aggregating functions!
It is now easy to implement by editing `evaluate_function` function in `executor/evaluate/mod.rs`.


panarch:
resolve #106

panarch:
resolve #102

Issue #101, opened 2020-10-23
Plus(+) & Minus(-) signs support in WHERE


panarch:
Following `WHERE` clauses are not supported now.

```sql
SELECT * FROM TableA WHERE +1 = 1;
SELECT * FROM TableA WHERE -1 = -1;
```

Proper place to add test cases would be in `src/tests/filter.rs`.

yejihan-dev:
I would like to work on this issue :)

Issue #52, opened 2020-08-28
Remove panic! uses in evaluated.rs


panarch:
[src/executor/evaluate/evaluated.rs](https://github.com/gluesql/gluesql/blob/main/src/executor/evaluate/evaluated.rs)

There are a few `panic!()` uses in `evaluated.rs`.
I used `panic!()` because currently there's no way to reach that branches.
But it's better not to use panic, replace those to managed unreachable error types.

PR #111, opened 2020-10-31
[Revised] Support Plus(+) and Minus(-) signs in WHERE

yejihan-dev:
This pull request resolves issue #101 by supporting unary operations of Value. Specific modifications are the following.
- value.rs & evaluated.rs: Functions to evaluate unary operations of the expression were added.
- mod.rs: Expressions now handle unary operations.
- test.rs: Two test cases to check if WHERE supports plus(+) and minus(-) signs were added.

### DIFF
--- a/src/executor/evaluate/evaluated.rs
+++ b/src/executor/evaluate/evaluated.rs
@@ -56,7 +56,7 @@ impl<'a> PartialEq for Evaluated<'a> {
                     StringRef(r) => eq_ast(l, r),
                     ValueRef(r) => r == l,
                     Value(r) => &r == l,
-                    Literal(_) => panic!(),
+                    Literal(r) => *l == r,
                 },
                 StringRef(l) => match other {
                     LiteralRef(r) => eq_ast(r, l),
@@ -77,12 +77,14 @@ impl<'a> PartialEq for Evaluated<'a> {
                     StringRef(r) => eq_val(&l, r),
                     ValueRef(r) => &l == r,
                     Value(r) => l == r,
-                    Literal(_) => panic!(),
+                    Literal(r) => l == r,
                 },
                 Literal(l) => match other {
+                    Literal(r) => l == r,
+                    LiteralRef(r) => &l == r,
                     ValueRef(r) => r == &l,
                     StringRef(_) => false,
-                    _ => panic!(),
+                    Value(_) => panic!(),
                 },
             }
         }
@@ -99,7 +101,7 @@ impl<'a> PartialOrd for Evaluated<'a> {
                 ValueRef(r) => r.partial_cmp(l).map(|o| o.reverse()),
                 Value(r) => r.partial_cmp(*l).map(|o| o.reverse()),
                 StringRef(_) => None,
-                Literal(_) => panic!(),
+                Literal(r) => literal_partial_cmp(&l, &r),
             },
             ValueRef(l) => match other {
                 LiteralRef(r) => l.partial_cmp(r),
@@ -109,7 +111,7 @@ impl<'a> PartialOrd for Evaluated<'a> {
                     data::Value::Str(l) => (&l.as_str()).partial_cmp(r),
                     _ => None,
                 },
-                Literal(_) => panic!(),
+                Literal(r) => l.partial_cmp(&r),
             },
             Value(l) => match other {
                 LiteralRef(r) => l.partial_cmp(*r),
@@ -119,7 +121,7 @@ impl<'a> PartialOrd for Evaluated<'a> {
                     data::Value::Str(l) => (&l.as_str()).partial_cmp(r),
                     _ => None,
                 },
-                Literal(_) => panic!(),
+                Literal(r) => l.partial_cmp(r),
             },
             StringRef(l) => match other {
                 LiteralRef(_) => None,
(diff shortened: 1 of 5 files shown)


src/executor/evaluate/evaluated.rs
```diff
@@ -342,3 +378,25 @@ fn literal_divide(a: &AstValue, b: &AstValue) -> Result<AstValue> {
...
 }
+
+fn literal_positive(a: &AstValue) -> Result<AstValue> {
+    match a {
+        AstValue::Number(a) => match a.parse::<i64>() {
+            Ok(a) => Ok(AstValue::Number(a.to_string())),
+            _ => panic!(),
+        },
+        AstValue::Null => Ok(AstValue::Null),
+        _ => Err(EvaluateError::UnreachableLiteralArithmetic.into()),
```
panarch, line 389: In fact.. current `evaluated.rs` file has some problems.
`panic!` uses and branches which are named `Unreachable` but actually reachable.

It looks like you followed existing code patterns well like `fn literal_add` functions.
But you don't need to follow bad codes, :) it's my fault.
https://github.com/gluesql/gluesql/issues/52

Good thing is, not like `literal_add` , these are unary operator function so it only takes a single operand.
So it's easier to remove `panic!()`, let's try out.

* `panic!()` in here can easily reproduced by this test case which uses float type `+2.0`.
```sql
SELECT id FROM Hunter WHERE +2.0 > 1
```
You also need to consider float type uses.
In `src/data/value.rs`, `impl TryFrom<&AstValue> for Value {` TryFrom implementation would be good to see.
```rust
AstValue::Number(v) => v
     .parse::<i64>()
     .map_or_else(|_| v.parse::<f64>().map(Value::F64), |v| Ok(Value::I64(v)))
     .map_err(|_| ValueError::FailedToParseNumber.into()),
```
(I'd like to make it simpler but I didn't know what to do at that time I wrote in this kind of way)

* Reachable but named `(EvaluateError::UnreachableLiteralArithmetic.into()),`
```sql
SELECT id FROM Hunter WHERE +"asdf" = -1
```
This sql query above will show `UnreachableLiteralArithmetic`, though `Unrechable-` prefixed errors are not supposed to see.

Let's remove these two `panic!()` and `Unreachable` error.

panarch (changes requested):
One more thing,
Current names for apply plus and minus unary operators are "positive" and "negative".
But using adjective words for function names sound more like... returning current state, not converting or chaining work.

`value.positive()` looks like... it would return whether value is positive or not.
So, let's consider names using verbs or with objectives.
Interesting point in here is... it's quite tough to find proper names.

If there is a way to use verb, it would be great but...
I also tried to find more proper ones but failed.
Most of references simply say "unary plus", "unary minus", "unary negation". etc.

So, how about using simply "unary_plus" and "unary_minus" for these functions rather than "positive" and "negative"?

yejihan-dev:
@panarch I really appreciate your thorough reviews! It was my first time making a pull request to an open source project. I learning so much. :)

I finished changing function names, simplifying `match` cases, and adding codes for a float type. I'm currently working on errors regarding `SELECT id FROM Hunter WHERE +2.0 > 1`. It seems like I need to add more cases inside `fn partial_cmp(&self, other: &Evaluated<'a>) -> Option<Ordering> {` which is in `src/data/evaluated.rs`. It should be done by tonight. Again, thanks for your review!

src/executor/evaluate/evaluated.rs
```diff
@@ -99,7 +101,7 @@ impl<'a> PartialOrd for Evaluated<'a> {
                 ValueRef(r) => r.partial_cmp(l).map(|o| o.reverse()),
                 Value(r) => r.partial_cmp(*l).map(|o| o.reverse()),
```
panarch, line 102: fyi.
Reason that `reverse()` is used is that this is comparison between `gluesql::data::Value` and `sqlparser::ast::Value`.

`gluesql::data::Value` has implementation of `PartialCmp` with `sqlparser::ast::Value`.
However `sqlparser::ast::Value` does not have implementation of `PartialCmp` with `gluesql::data::Value`.
And Rust does not allow adding more `impl` in different modules.
so that's why `reverse` was required.

panarch (changes requested):
Beautiful,
I firstly thought this issue is good first issue but hmm.. it has quite more than I estimated.. but you did too good, thanks a lot :)

* Tiny one for the current commit title,

"[Revised]" is not required because.. After this PR has merged, that does not have any meaning when we see simply main branch's commits log.

* Code branches you newly added to `evaluated.rs`.

Before starting, you don't need to take care about code branches in `value.rs` (about correct behaviors), it would be enough with unit tests and let's just ignore it for now.

This is about `evaluated.rs`.
When `unary_plus` and `unary_minus` consumes `LiteralRef` then it returns `Literal`.
And... previous `evaluated.rs` had many `panic!()` uses about `Literal`.
So you also needed to work on `literal_partial_cmp`, you did very well, thanks.

I guess you might feel little bit uncomfortable about the changes in `evaluated.rs`, not sure about whether those changes might have errors or not.

However, there are no `mut` in `evaluated.rs` and what it means is that you can simply test all branches. It is not that tough.
This is one of beautiful stuffs when we do functional programming, it's quite easy to write test cases.

As a reviewer, by scanning all code branches you added, it is not hard to find missed test cases and bugs.

For example,
```sql
SELECT * FROM Boss WHERE id <= +2;
SELECT * FROM Boss WHERE +id <= +2;
SELECT id FROM Hunter WHERE 2.0 > 1.0;
```
Those will show buggy results.

So,
Let's add more test cases to cover all newly added code branches in `evaluated.rs`.
Now we have `literal_plus` and `literal_minus` so we can make `Literal` from `LiteralRef` and `Value` from `ValueRef`.
So you will be able to cover all branches without any issue.

yejihan-dev:
I found out that the error comes from unnecessary reverse operations. Thank you for pointing out! :) I changed the code and added some additional test cases in `src/tests/filter.rs`.

- Proper uses of unary operations
```
SELECT name FROM Boss WHERE -2.0 < -1.0
SELECT id FROM Hunter WHERE +2 > +1.0
SELECT name FROM Boss WHERE id <= +2
SELECT name FROM Boss WHERE +id <= 2
```
- Erroneous uses of unary operations
```
SELECT id FROM Hunter WHERE +'abcd' > 1.0 // EvaluateError::LiteralUnaryPlusOnNonNumeric
SELECT id FROM Hunter WHERE -'abcd' < 1.0 // EvaluateError::LiteralUnaryMinusOnNonNumeric
SELECT id FROM Hunter WHERE +name > 1.0 // ValueError::UnaryPlusOnNonNumeric
SELECT id FROM Hunter WHERE -name < 1.0 // ValueError::UnaryMinusOnNonNumeric
```

panarch (approved):
Great, thanks!
Looks good, merging 👍 👍 👍

PR #112, opened 2020-11-04
#76 'SingleQuotedString' is not accepted in inserts

zhangchunzhong:
I made a auto-merged commit version.


panarch:
Because it is done by @willy610, I think it would be better to ask @willy610 first about taking this issue.

If he'd like to finish https://github.com/gluesql/gluesql/pull/87 himself, then I'll take his work.
Else I will use this, thanks

Issue #63, opened 2020-09-06
RIGHT [OUTER] JOIN support


panarch:
For now, GlueSQL supports 2 join types, `INNER JOIN` and `LEFT OUTER JOIN`.
`RIGHT OUTER JOIN` Implementation can be a bit more tough, but all modifications required to change is in a single file, 198 lines of codes.
* [src/executor/join.rs](https://github.com/gluesql/gluesql/blob/main/src/executor/join.rs)

It can be a sort of quite fun programming quiz.

willy610:
I have an implementation on join (all kinds (outer,inner,left and right)) in both ``Mathematica`` and ``Javascript``

They could work on ordered input or not.
I would like to try to implement that in Rust.

I hope someone is not at the moment working in ``join.rs``

In any case the solution today is slow

panarch:
Sounds good, then I'll assign you on this issue.

willy610:
Some feedback 

Have a working joiner now.
Left, right, inner, outer etc
(Sofar 242 lines!!)

Slightly prepared for using future existing ordered indexes...

Some notes/question and/or suggestions

1. In the ``select.rs`` perhaps one should introduce a ``'from'`` function
replacing all join stuff in ``select.rs`` and perform all tasks in the ``'from' (TableWithJoins)``
And of cource ``join.rs`` becomes something else (partly moved to inside ``'from'``)

2. I think all selections producing 'relations' should have sufficient content
so that a temporary relation could be constructed and asked for 
Like in ``execute.rs Statement::CreateTable + Statement::Insert``
(So there will be an entry in fetch like ``fetch_columns_defs()`` )

3. The projection in ``select.rs picks named and columns and must find them in the result from 'from'

See alos issue #83

Made a reverse of AST for Select into an BNF and than to a railed bnf in attachment


[sql_selectsvg.pdf](https://github.com/gluesql/gluesql/files/5342404/sql_selectsvg.pdf)

panarch:
I think I don't get it, can you provide more details about `from` function suggestion?
It sounds like similar with current structs named `BlendContext`, `FilterContext` and `AggregateContext`.

willy610:
I will look more into the design pattern Context.
It is complicated I think but powerful.

I'll be back

willy610:
Well, I think I have come out with a working solution a from component.
The new ``from.fetchblended()`` will do the complete join of 0..n tables
so the old ``join`` will be replaced.
But
[...]

PR #114, opened 2020-11-13
Design of join. Issue #63

willy610:
Don't understand how to merge two joined rows into one row.
As I understand
central in a join is some kind of evalute the join constraint on two rows,
one from left scan. or earlier joins, and one from right scan.

I seems like a ``Filter.check(othertable,othercolumns,otherrows)``
 will use self.row and use all rows in the chain 'next' to evaluate the join constraint

In my tryouts a managed to use ``Filter.check(...)`` properly I think like
``Filter(leftrows).check(righttable, rightcolumns, rightrows)``

But when later on when I will chain the right row at the end of lefts.next I don't
know how to do. Like ``a_left_row.concat_into(a_right_row);``

In the ``source executor/from.rs`` is a snippet with the problem

For the record I can produce the right answer for ``LEFT JOIN`` and ``RIGHT JOIN`` but can't dela with three tabled, that is two joins as I don't know how ti use the ``concat_into``

You can look into the ``executor/from.rs`` for a more detailed program source

### DIFF
--- a/src/executor/from.rs
+++ b/src/executor/from.rs
@@ -1,0 +1,40 @@
+/*
+As I understand
+central in a join is some kind of evalute the join constraint on two rows,
+one from left scan. or earlier joins, and one from right scan.
+I seems like a Filter.check(othertable,othercolumns,otherrows)
+ will use self.row and use all rows in the chain 'next' to evaluate the join constraint
+
+In my tryouts a managed to use Filter.check(...) properly I think like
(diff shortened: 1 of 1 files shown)


panarch:
I changed the design of `FilterContext` and `BlendContext` in here, https://github.com/gluesql/gluesql/tree/async-store-api
I will merge into main branch in a few days and release a new version.

That will be helpful for you to see how `concat_into` like function works.
I'll close this PR and let's talk again after `async` work is done.

PR #115, opened 2020-11-15
Convert store traits to async

panarch:
Now all existing store traits support `async`; `Store`, `StoreMut` and `AlterTable`.

* Most of internal functions which was returning `Iterator`, now those returns `TryStream` due to `Future` support.

* Change the way of adding integration tests.
Using `test_case!` macro, we can add a new test case and `run!`, `count!` and `test!` macros are now available to use.


panarch:
close #115

PR #116, opened 2020-11-20
Allow Binary Operations Between <i64> and <f64>

yejihan-dev:
This PR would be a simple expansion of binary operations. :) `evaluated.rs` now handles binary operations between <i64> and <f64>. To check the code, I added some test cases in `filter.rs`. Examples of test cases are
```
(5, "SELECT name FROM Boss WHERE 2 = 1.0 + 1"),
(3, "SELECT id FROM Hunter WHERE -1.0 - 1.0 < -1")
```

### DIFF
--- a/src/executor/evaluate/evaluated.rs
+++ b/src/executor/evaluate/evaluated.rs
@@ -129,7 +129,7 @@ impl<'a> PartialOrd for Evaluated<'a> {
                 ValueRef(data::Value::Str(r)) => l.partial_cmp(&r.as_str()),
                 Value(data::Value::Str(r)) => l.partial_cmp(&r.as_str()),
                 StringRef(r) => l.partial_cmp(r),
-                Literal(_) => panic!(),
+                Literal(_) => None,
                 _ => None,
             },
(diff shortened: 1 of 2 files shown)


panarch (commented):
Perfect, nothing I need to comment more.
Thanks 👍 

Below here is an additional work which can be interesting to try at this point.
Feel free to append a new commit on this PR.

Now you noticed that there are 4 arithmetic op functions, `literal_add`, `literal_subtract`, `literal_multiply` and `literal_divide`.
They are all equivalent only except for `+`, `-`, `*`, `/` operators.
This is quite good chance to get familiar with Rust's declarative macro!

Let's make a macro to generate those 4 op functions.
You may need to get two params for the macro, one for the function name and the other for op closure.

yejihan-dev:
@panarch It's been a while since I made last commit. (I had multiple exams ;_;) I made a new commit that implements a declarative macros on four operations!

panarch:
@yejihan-dev :D I understand. Looks good! merging 👍

PR #123, opened 2020-11-25
Added benchmarks

Redblueflame:
Added benchmarks to constantly check the performance increase / regression while programming.

In its current state, only two tests were added:
- [x] insert_one (Inserting elements one by one in the storage)
- [x]  select_one (Selecting a s single element in the database containing 100K elements).

Before merging this pull request, I'll need to add some other tests to catch regressions as quickly as possible:
- [x]  select_one_tainted (Selecting a single element, but with multiple tables in the database, not only the one we're searching)
- [x]  select_many (Selecting a range of elements in a table)
- [x]  select_many_tainted (Selecting many elements in a table, while having other tables present in the database)
- [ ]  ~~select_one_non_primary (Select an element using something else than the primary key, but will be indexed)~~
- [ ]  ~~select_one_non_indexed (Select an element using something else than the primary key and an indexed field)~~

Cf. next comment


Redblueflame:
I'm considering this PR as ready, and for the missing benchmarks, they are described in #124.

panarch (approved):
Beautiful start, thanks 👍

PR #125, opened 2020-11-25
Fix naming

Redblueflame:
In recent IntellijRust builds, there is an issue (tracked here: https://github.com/intellij-rust/intellij-rust/issues/5249) that causes an error while using an exported type that was imported using mod privately.

This PR only have for objective to fix this.


Redblueflame:
Closed in favor of a future PR with a rebase from master.

PR #127, opened 2020-11-25
Indexing

Redblueflame:
This is the draft PR for the indexing.

Just to justify some of my choices:
- Implementing directly the operator functions makes it impossible to combine multiple conditions on the same array of elements. Let's take an example:
```sql
CREATE TABLE Testing (id INTEGER, field_one INTEGER, field_two TEXT);
CREATE INDEX field_one;
-- Skipped the insertion of the actual elements
SELECT * FROM Testing WHERE `field_one` > 5 OR `field_one` < 2;
```
In that specific case, if the Storage only implements the basic operators, it means that the index has to be crawled twice, and the cost of doing that can be significant on big tables.

The actual objective is to have a simplified condition system, that can easily be expanded afterwards to create a custom Query Builder for GlueSQL, and being read by the Storage backend.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 # GlueSQL project files
-data/
+/data/
 
 # Vim
 *.swp
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,10 @@ default = ["sled-storage", "alter-table"]
 # You can include whether ALTER TABLE support or not for your custom database implementation.
 alter-table = []
 
+# INDEXes are optionnal.
+# You can choose to import INDEXes or not for your custom implementation.
+index = []
+
 # Who wants to make a custom storage engine,
 # default storage engine sled-storage is not required.
 sled-storage = ["sled", "bincode"]
--- a/src/data/conditions.rs
+++ b/src/data/conditions.rs
@@ -2,8 +2,10 @@
 //! TODO: Add more documentation about how each comparison works.
 
 use crate::Value;
+use sqlparser::ast::{Expr, Query};
+
 /// The conditions are the base type of condition, and are used to describe all WHERE closes supported.
-enum Condition {
+pub enum Condition {
     /// Represents an expression of the type `column_name = value`
     /// For example: `country = "France"`
     Equals { column_name: String, value: Value },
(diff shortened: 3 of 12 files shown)


panarch:
Awesome! I'll keep following.

And.. I think it would be good to decide the first index specs to implement.
I'd like to suggest starting with as minimal as possible.

* Multi-column index support
To support full syntaxes, it's required to have multi-column index
```sql
CREATE INDEX index_name ON table_name (column1, column2, ...);
```
But for the first implementation, I think we can start with only supporting single-column index.
How do you think of this?

* unique support
I saw current draft codes are considering `unique` constraint.
`unique` works both with index and without index.
And currently there's no `unique` support without index.
So, how about considering `unique` later?
I think it would be complicated enough to only consider core index functionality in this pr.

panarch:
Ok, I understand you can consider to make another AST to support index. but let's think about more simple solution.

If we build another AST from `sqlparser`'s AST, then we are responsible to manage that new `AST`.
However that will add too much extra tasks to manage.
Let's keep depending on `sqlparser`.

For example,
```sql
SELECT * FROM Testing WHERE field_one > 5 AND field_two > 10;
```
Assume that sql query is converted to `sqlparser`'s AST.
Then, what our executor works is...
`scan_data` by using `SELECT * FROM Testing`;
and filtering using `WHERE field_one > 5 AND field_two > 10;`

what our index needs to do is...
`scan_something_indexed` by using `SELECT* FROM Testing WHERE field_one > 5`;
and filtering using `WHERE field_two > 10;`

So what we only need to build is "AST" for index, only for the index, not anything else.
Then there might be no need of touching any of filter.rs `check_expr` function code.


But we still need a way to store indexed result.
May be we can use `HashMap` like struct which stores built index information.
And make `select` or `fetch` which uses `scan_data` to get as a new function param.

willy610:
At least in joins perhaps one need some complement to AST like Reversed Polish Notation telling the calculation order after some optimization. Don't extend SQL AST. Risky. Introduce some simple operations and express them in RPN.

Just for the record we have CASCADE in foreign keys, derived tables, dependent and independent subquery and indexing join of two tables before matching the third one. (I did use HashMap in a try out some time ago. Nice...)

And of cource the ``where`` requirements are different than the ``join`` requirements

Lot to do next year too

KyGost:
@panarch not sure that I feel very good about using only the first column's index and filtering for the rest.

But I think that the main point, for all points in this development, is that if we can get it working first, we can tune it based on performance later; agreed?

That is to say, we should try to get it working, a simple way or a more complicated way (whatever) first and from there we can improve-- main aim is to get indexing working.

panarch:
> But I think that the main point, for all points in this development, is that if we can get it working first, we can tune it based on performance later; agreed?

I agree, index has really a lot to do.
We will be able to make the index work with quite minimal code changes.

We can even start with only supporting indexing column which only compares with literals (`sqlparser::ast::value`)
```sql
... WHERE indexed_col = 1;
```

After that we can expand supporting more indexing later.
```sql
... WHERE indexed_col = (SELECT col FROM TableA LIMIT 1)
```
(e.g. nested select but it does not use any outer context values, so this `indexed_col` can be indexed)

PR #141, opened 2021-02-02
Implement UNIQUE constraint in CREATE TABLE

### LINKED ISSUE #105: UNIQUE constraint support in CREATE TABLE
```sql
CREATE TABLE TableA (
  id INTEGER UNIQUE
);
```

When executing `INSERT` and `UPDATE`, this `UNIQUE` constraint should be checked.

silathdiir:
Closes #105 .

I add a new source file `validate.rs` which only contains the logic of unique validation for now.
The function `validate_unique_by` applies unique validation on both `insert` and `update` rows.

### DIFF
--- a/src/data/table.rs
+++ b/src/data/table.rs
@@ -45,7 +45,7 @@ impl<'a> Table<'a> {
     }
 }
 
-pub fn get_name<'a>(table_name: &'a ObjectName) -> Result<&'a String> {
+pub fn get_name(table_name: &ObjectName) -> Result<&String> {
     let ObjectName(idents) = table_name;
 
     idents
--- a/src/data/value.rs
+++ b/src/data/value.rs
@@ -49,7 +49,7 @@ pub enum ValueError {
     UnaryMinusOnNonNumeric,
 
     #[error("floating columns cannot be set to unique constraint")]
-    FloatCannotBeUnique,
+    ConflictOnFloatWithUniqueConstraint,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
@@ -242,23 +242,7 @@ impl TryInto<UniqueKey> for &Value {
             I64(v) | OptI64(Some(v)) => Ok(UniqueKey::I64(*v)),
             Str(v) | OptStr(Some(v)) => Ok(UniqueKey::Str(v.clone())),
             Empty | OptBool(None) | OptI64(None) | OptStr(None) => Ok(UniqueKey::Null),
-            F64(_) | OptF64(_) => Err(ValueError::FloatCannotBeUnique.into()),
-        }
-    }
-}
-
-impl TryInto<UniqueKey> for Value {
-    type Error = Error;
-
-    fn try_into(self) -> Result<UniqueKey> {
-        use Value::*;
-
-        match self {
-            Bool(v) | OptBool(Some(v)) => Ok(UniqueKey::Bool(v)),
-            I64(v) | OptI64(Some(v)) => Ok(UniqueKey::I64(v)),
-            Str(v) | OptStr(Some(v)) => Ok(UniqueKey::Str(v)),
-            Empty | OptBool(None) | OptI64(None) | OptStr(None) => Ok(UniqueKey::Null),
-            F64(_) | OptF64(_) => Err(ValueError::FloatCannotBeUnique.into()),
+            F64(_) | OptF64(_) => Err(ValueError::ConflictOnFloatWithUniqueConstraint.into()),
         }
     }
 }
@@ -472,7 +456,7 @@ impl Value {
 
 #[cfg(test)]
 mod tests {
-    use super::*;
+    use super::Value;
 
     #[test]
     fn eq() {
@@ -495,21 +479,5 @@ mod tests {
         assert_eq!(Value::OptStr(None), Value::OptStr(None));
 
         assert_eq!(Value::Empty, Value::Empty);
-    }
-
-    #[test]
-    fn float_cannot_be_unique() {
-        assert_eq!(
-            TryInto::<UniqueKey>::try_into(Value::F64(1.0)),
-            Err(ValueError::FloatCannotBeUnique.into())
-        );
-        assert_eq!(
-            TryInto::<UniqueKey>::try_into(Value::OptF64(Some(1.0))),
-            Err(ValueError::FloatCannotBeUnique.into())
-        );
-        assert_eq!(
-            TryInto::<UniqueKey>::try_into(Value::OptF64(None)),
-            Err(ValueError::FloatCannotBeUnique.into())
-        );
     }
 }
--- a/src/data/value.rs
+++ b/src/data/value.rs
@@ -472,7 +472,7 @@ impl Value {
 
 #[cfg(test)]
 mod tests {
-    use super::Value;
+    use super::*;
 
     #[test]
     fn eq() {
@@ -495,5 +495,21 @@ mod tests {
         assert_eq!(Value::OptStr(None), Value::OptStr(None));
 
         assert_eq!(Value::Empty, Value::Empty);
+    }
+
+    #[test]
+    fn float_cannot_be_unique() {
+        assert_eq!(
+            TryInto::<UniqueKey>::try_into(Value::F64(1.0)),
+            Err(ValueError::FloatCannotBeUnique.into())
+        );
+        assert_eq!(
+            TryInto::<UniqueKey>::try_into(Value::OptF64(Some(1.0))),
+            Err(ValueError::FloatCannotBeUnique.into())
+        );
+        assert_eq!(
+            TryInto::<UniqueKey>::try_into(Value::OptF64(None)),
+            Err(ValueError::FloatCannotBeUnique.into())
+        );
     }
 }
--- a/src/data/value.rs
+++ b/src/data/value.rs
@@ -7,7 +7,7 @@ use thiserror::Error as ThisError;
 
 use sqlparser::ast::{DataType, Expr, Ident, Value as AstValue};
 
-use crate::executor::GroupKey;
+use crate::executor::{GroupKey, UniqueKey};
 use crate::result::{Error, Result};
 
 #[derive(ThisError, Serialize, Debug, PartialEq)]
@@ -48,8 +48,8 @@ pub enum ValueError {
     #[error("unary minus operation for non numeric value")]
     UnaryMinusOnNonNumeric,
 
-    #[error("Duplicate entry '{0}' for unique column '{1}'")]
-    DuplicateEntryOnUniqueField(String, String),
+    #[error("floating columns cannot be set to unique constraint")]
+    FloatCannotBeUnique,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
@@ -227,6 +227,38 @@ impl TryInto<GroupKey> for Value {
             Str(v) | OptStr(Some(v)) => Ok(GroupKey::Str(v)),
             Empty | OptBool(None) | OptI64(None) | OptStr(None) => Ok(GroupKey::Null),
             F64(_) | OptF64(_) => Err(ValueError::FloatCannotBeGroupedBy.into()),
+        }
+    }
+}
+
+impl TryInto<UniqueKey> for &Value {
+    type Error = Error;
+
+    fn try_into(self) -> Result<UniqueKey> {
+        use Value::*;
+
+        match self {
+            Bool(v) | OptBool(Some(v)) => Ok(UniqueKey::Bool(*v)),
+            I64(v) | OptI64(Some(v)) => Ok(UniqueKey::I64(*v)),
+            Str(v) | OptStr(Some(v)) => Ok(UniqueKey::Str(v.clone())),
+            Empty | OptBool(None) | OptI64(None) | OptStr(None) => Ok(UniqueKey::Null),
+            F64(_) | OptF64(_) => Err(ValueError::FloatCannotBeUnique.into()),
+        }
+    }
+}
+
+impl TryInto<UniqueKey> for Value {
+    type Error = Error;
+
+    fn try_into(self) -> Result<UniqueKey> {
+        use Value::*;
+
+        match self {
+            Bool(v) | OptBool(Some(v)) => Ok(UniqueKey::Bool(v)),
+            I64(v) | OptI64(Some(v)) => Ok(UniqueKey::I64(v)),
+            Str(v) | OptStr(Some(v)) => Ok(UniqueKey::Str(v)),
+            Empty | OptBool(None) | OptI64(None) | OptStr(None) => Ok(UniqueKey::Null),
+            F64(_) | OptF64(_) => Err(ValueError::FloatCannotBeUnique.into()),
         }
     }
 }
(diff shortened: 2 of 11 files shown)


panarch:
@silathdiir Looks great, thanks so so much :)
Behavior looks perfect, and I only have some opinions about the internal implementation. I'll finish review and leave comments until tomorrow, thanks!

src/data/table.rs
```diff
@@ -45,7 +45,7 @@ impl<'a> Table<'a> {
     }
 }
 
-pub fn get_name<'a>(table_name: &'a ObjectName) -> Result<&'a String> {
+pub fn get_name(table_name: &ObjectName) -> Result<&String> {
```
panarch, line 48: Thanks 👍

src/executor/validate.rs
```diff
@@ -0,0 +1,204 @@
...
+) -> Result<()> {
+    validate_rows_by(storage, table_name, column_validation, rows, |r| r).await
+}
+
+pub async fn validate_rows_by<'a, T: 'static + Debug, U: 'static + Debug, F>(
+    storage: &impl Store<T>,
+    table_name: &str,
+    column_validation: ColumnValidation,
+    items: &'a [U],
+    item_to_row: F,
```
panarch, line 30: How about using `Iterator` than `&[U]` or `&[Row]`.
Then you don't need to use this kind of closure functions and you can merge `validate_rows` and `validate_rows_by` into a single function.

Here in the `validate_rows_by` function, `U` contains `(T, Row)` but `T` is not used in this function.
It means that `T` should be excluded outside of the function `validate_rows_by`, not inside.

src/executor/validate.rs
```diff
@@ -0,0 +1,204 @@
...
+}
+
+#[derive(Debug)]
+struct UniqueCheck<'a> {
+    column_index: usize,
+    values: Vec<&'a Value>,
+}
+
+fn create_unique_checks<'a, U: 'static + Debug, F>(
+    table_column_defs: &[ColumnDef],
```
panarch, line 144: Only subset of `table_column_defs` is used, and it is only limited to `unique_column_indexes`.
So, it would be better to be merged into a single param.
You can use something like... `unique_columns: &[(usize, ColumnDef)]`, then you don't need to get unnecessary extra params into this function.

Then, you can remove an unnecessary `panic!()` possible case in here below,
`163 table_column_defs[column_index].name.to_string(),`
You can simply replace from `table_column_defs[column_index]` to `column_def` which is panic safe.

src/executor/validate.rs
```diff
@@ -0,0 +1,204 @@
...
+where
+    F: Fn(&U) -> &Row,
+{
+    let item_len = items.len();
+    let mut checks = Vec::with_capacity(unique_column_indexes.len());
+    for &column_index in unique_column_indexes {
+        let mut values: Vec<&'a Value> = Vec::with_capacity(item_len);
+        for item in items {
+            let new_val = item_to_row(item).get_value(column_index).unwrap();
+            match values.binary_search_by(|v| compare_values_for_sort(v, new_val).unwrap()) {
```
panarch, line 158: This was impressive, I appreciate.
I agree that this filtering is a good choice.
But how about using sets in `std::collections` rather than using custom implementation?

panarch (changes requested):
Thanks, including comments I attached to the codes,
here is remaining ones.

1. Removing `mut` uses

It is a ground rule inside of execution codes.
It might be little bit harder to write codes without `mut` but it will make us to write clear and more readable code.

2. Removing `unwrap` uses

If the error is related to storage data conflict, then you can use `Conflict` prefix for error name.
If you think some `unwrap()`s are really unavoidable and there's no possibility to be panic, then you can use `Unreachable` prefix.

Except for the two cases, we need explicit error variables and test cases to generate those errors.

3. `UNIQUE` & `NULL` field can have multiple `NULL` values.

It panics now.

src/executor/validate.rs
silathdiir: Hi @panarch , I added this function `compare_value_for_sort` as an `internal` (just in this file) implementation of trait `Ord`. Because I need an efficient method to insert and find a Value in the sorted Vector. The function `binary_search/binary_search_by` requires `T` must implement the trait `Ord`, the function `compare_value_for_sort` is used for this.

As you mentioned the sets of `std::collections`, I found [HashSet](https://doc.rust-lang.org/std/collections/struct.HashSet.html) and [BTreeSet](https://doc.rust-lang.org/std/collections/struct.BTreeSet.html). `BTreeSet` needs T must implement the trait `Ord` (and cannot be customed with a closure like `binary_search_by`).

I try to use `HashSet` and add `#[derive(Hash)]` trait to `data::Value`, but the type `f64` cannot been hashed (since `NAN`?). I found [this-discussion](https://internals.rust-lang.org/t/f32-f64-should-implement-hash/5436), it seems that the type `f64` needs to be wrapped by `OrderedFloat` (crate [ordered-float](https://github.com/reem/rust-ordered-float)), but the code seems not good.

What's your option? Do you have any suggestions for this? Thanks.

panarch: @silathdiir Oh! I was supposed to mention this issue in the first comment but I forgot to write, thanks for the reply.
I also experienced `Eq + Hash` related issue before when I was implementing `GROUP BY` functionality.

Yes you're right. `f64`, floating point number is the problem. `f64` does not implement `Eq`, so there's no way to support `Hash`.
However, it is also reasonable for `f64` not to have `Eq` because of zero comparison issue.
If we want to support hashing for floating point numbers, then the right choice is using something like `Decimal` type, but that is not implemented in GlueSQL now.

What I decided to do was preventing the use of float in the `GROUP BY`.
If someone wants to `GROUP BY` on float fields, then they need to run some functions like `ceil`, `floor` or `round`, or.. they can use `DECIMAL` type. Though we don't have all that features, but we will have.

I also agree that using `OrderedFloat` is not a good idea. I think we don't need to use something.. tricks, we are making a new sql database, it is not copying other db's functionalities, so..


fyi.
This might be useful to see.

https://github.com/gluesql/gluesql/blob/036080b7c5155a2bb6b8a26bdaa6a26805019e53/src/executor/aggregate/hash.rs#L7

https://github.com/gluesql/gluesql/blob/036080b7c5155a2bb6b8a26bdaa6a26805019e53/src/data/value.rs#L195

I added some `GroupKey` related codes to `data/value.rs`,
I wanted to gather codes which handle the types in `Value` enum to `data/value.rs`.
Then, when we consider to add any new data type to `Value`, it would be easy to work on it because we don't need to take care of any other codes outside of `data/value.rs`. That was my intention.

---

Ok, let's make the issue simple.
Doing same approach with `GROUP BY`,
* Float type cannot have `UNIQUE` constraint

Then you may need to add a new error for `CREATE` query which is normal error type.
Then, in the `validate.rs`, one `Conflict-` prefixed error will be required for floating point + unique validation case.
How do you think of this?

silathdiir:
Hi @panarch , I update the code according to your code review. Please help review if you have time. Thanks.

src/data/value.rs
```diff
@@ -47,6 +47,9 @@ pub enum ValueError {
 
     #[error("unary minus operation for non numeric value")]
     UnaryMinusOnNonNumeric,
+
+    #[error("floating columns cannot be set to unique constraint")]
+    FloatCannotBeUnique,
```
panarch, line 52: Test case for this error is required. You can add it to `src/tests/unique.rs`.

src/executor/validate.rs
```diff
@@ -0,0 +1,245 @@
...
+    ConflictOnUniqueColumnDataType(String, String),
+
+    #[error("duplicate entry '{0}' for unique column '{1}'")]
+    DuplicateEntryOnUniqueField(String, String),
+
+    #[error("table already exists")]
+    TableAlreadyExists,
+
+    #[error("table does not exist")]
+    TableNotExists,
```
panarch, line 40: There is a test case for `ExecuteError::TableNotExists` but no `ValidateError::TableNotExists`.
Please add a test case which generates `ValidateError::TableNotExists`.

src/executor/validate.rs
```diff
@@ -0,0 +1,245 @@
...
+
+    #[error("column '{0}' of data type '{1}' is unsupported for unique constraint")]
+    UnsupportedDataTypeForUniqueColumn(String, String),
```
panarch, line 43: `UnsupportedDataTypeForUniqueColumn` also needs a test case.

btw, error message looks really kind enough to see details, thanks :)

panarch (changes requested):
Looks great, thanks so so much.
This work is really awesome and I also really appreciate that you introduced a new module `validate`, I like this.
We can expand `validate` module by adding more validations in the future, good module name to start.

For the review, very small ones left.
I found that there is no test case for some newly added errors.
Please add test cases and I'll approve and merge this.
Thanks 👍

silathdiir:
Hi @panarch , I add test cases for error `UnsupportedDataTypeForUniqueColumn` (creating table with FLOAT UNIQUE columns) in `src/tests/unique.rs`, and move the test case of `ValidateError::TableAlreadyExists` to `src/tests/unique.rs`.

But I could not confirm about adding test cases for the below errors.

1. `ValidateError::ConflictOnStorageColumnIndex`: It should be reported if the inserting/updating rows have wrong number of columns with Storage Schema (e.g. one inserting row has 2 columns but table schema has 3).

2. `ValidateError::ConflictOnUniqueColumnDataType`: It should be reported if the inserting value is type of Float for an Unique column (it is transformed from `ValueError::FloatCannotBeUnique` and adds column name and data type).

1. `ValidateError::TableNotExists`: It should be reported when fetching schema in `src/execute/validate.rs` (but it is hidden by `ExecuteError::TableNotExists` when running SQLs in test cases).

1. `ValueError::FloatCannotBeUnique`：I add Unit test cases to `src/data/value.rs`, since it is converted to `ValidateError::ConflictOnUniqueColumnDataType` in `src/execute/validate.rs`.

For these four errors, since they are hidden by other errors (`ExecuteError`) when running SQLs in test cases, I suppose to add `Unit` test cases with Fake implementation of `Store` Trait in `src/execute/validate.rs`. I could not confirm if it is the right method. Do you have any suggestions on this? Thanks.

panarch:
@silathdiir Thanks for the details, I got the issue.
And your feedback is really helpful for me to write explicit guideline about errors. now I got what I need to provide for contributors. I'll add it soon, thanks so much.

* All errors except the prefixes with `Conflict-` or `Unreachable-` should have equal or more than one **integration** test case.
* Enum `gluesql::Error` is not for unit test, these errors are exposed to users so there should be a way to generate from user side.

And..
If there exists a way to generate `Conflict-` or `Unreachable-` prefixed errors from user side (without customly modifying storage data), it is a bug.
In current integration test system, there should be no way to generate `Conflict-` or `Unreachable-`.


#### `Conflict-` type errors exist because of storage level data conflict.
If users only manipulate data using `GlueSQL` and that is properly executed, then that is not possible to see `Conflict-` type errors.
One of the cases which users can see `Conflict-` errors are.. when user manipulate the storage data manually without accessing `GlueSQL` library and the modified data structure does not fit to `GlueSQL`.

e.g. `GlueSQL` fetched schema data and schema has a column with FLOAT UNIQUE
In normal use cases, it should be filtered out when user tried to create table with FLOAT UNIQUE.

#### `Unreachable-` type errors are mainly because of parser.
Theoretically if the parsed AST is exactly as same as SQL, then we may not need `Unreachable-` errors.
However, from the point of view from GlueSQL, it looks like AST is superset of SQL.
Some ASTs don't look possible to generate from SQL.
Then, sure we also don't need to write codes for **unreachable** AST.
It is `Unreachable-`'s main role.

Other use cases are... when it is too hard to deal with hidden errors.
Good to avoid use `Unreachable-` for hidden errors, but if we get too stress from this, then use this.
It is not ideal, but it is better than giving up.

---

e.g. You recognized this as a kind of `Unreachable-`, it may look like an hidden error, but I'll provide how to resolve.
> ValidateError::TableNotExists: It should be reported when fetching schema in src/execute/validate.rs (but it is hidden by ExecuteError::TableNotExists when running SQLs in test cases).

I figured out and it was really true that `ValidateError::TableNotExist` is hidden so there's no way to generate.
`INSERT` and `UPDATE` is related to this error due to `validate_rows`.

1. `INSERT` -> `prepare` calls `fetch_schema` -> hidden by `ExecuteError::TableNotExists`
2. `UPDATE` -> `prepare` calls `fetch_columns` -> hidden by `FetchError::TableNotFound`
(Excuse for naming convention mismatch 😂)

Solution can be simple, you can remove `ValidateError::TableNotExists`.
Make `validate_rows` to accept `column_defs` as a param. (you might need borrowing or `Rc`..)
1. `INSERT` -> Update `Prepared::Insert` to carry `column_defs`.
2. `UPDATE` -> Replace `fetch_columns` and use `fetch_schema` directly, then you can get `column_defs`. because `fetch_columns` returns `Vec<Ident>` which is already filtered out.

---
I'll also take a look on
`ValidateError::ConflictOnUniqueColumnDataType`, `ValidateError::ConflictOnUniqueColumnDataType` and `ValueError::FloatCannotBeUnique` tomorrow.

---
One more thing, it is independent one I missed in the previous review.
I found that `validate_table` is only used for `CREATE TABLE`.
Current location that `validate_table` in `validate.rs` can be little bit confusing.
if the `validate_table` function is used in more than one cases, then it's ok.
However, the only use case for `validate_table` is `CREATE TABLE`.
This should be only with create table function.
It looks like you worried about these below codes to get too long.
image
Then I'd like to suggest `select.rs` approach.

There was too much to do for `SELECT`, so I made a module named `executor/select.rs`.
Let's make a module named `executor/create_table.rs` and move those existing above code + `validate_table` to `create_table.rs`.
Then, we don't need to make `validate_table` function as public.
The current codes require some effort to figure out where `validate_table` is used, but if we move to `create_table.rs` then it becomes straightforward. (because it can be private)

panarch:
@silathdiir 
image

This code above is not required.

* `impl TryInto<UniqueKey> for &Value` -> `validate.rs` uses this.
* `impl TryInto<UniqueKey> for Value` -> dead code

panarch:
> `ValidateError::ConflictOnStorageColumnIndex`: It should be reported if the inserting/updating rows have wrong number of columns with Storage Schema (e.g. one inserting row has 2 columns but table schema has 3).

This looks good.

---

> `ValidateError::ConflictOnUniqueColumnDataType`: It should be reported if the inserting value is type of Float for an Unique column (it is transformed from `ValueError::FloatCannotBeUnique` and adds column name and data type).

I saw that `ValueError::FloatCannotBeUnique` is not used and transformed to `ValidateError::ConflictOnUniqueColumnDataType`.
Let's make it simple.
It is proper to use `Conflict-` prefix.
Let's rename `ValueError::FloatCannotBeUnique` to `Conflict-` prefixed error and remove `ValidateError::ConflictOnUniqueColumnDataType`.
How do you think of this?

silathdiir:
Hi @panarch , I update the code according to your code review as below:

1. Moves function `validate_table` to a new source file `executor/create_table.rs`, and adds `CreateTableError`.

2. Passes `column_defs` as part of `enum ColumnValidation` parameter into the function `validate_rows`.

3. Renames `ValueError::FloatCannotBeUnique` to `ValueError::ConflictOnFloatWithUniqueConstraint`, and returns it directly in `executor/validate.rs`.

Please help review again if you have time. Thanks.

panarch (approved):
Beautiful, thanks again and I really, really appreciate the work you've done.

I'll release a new version soon!
Merging 👍

PR #144, opened 2021-02-27
CAST

KyGost:
Implements `CAST(`[value]` AS `[type]`)`


KyGost:
Note: bug not enhancement

KyGost:
TODO: Better errors
`[Err(Value(FailedToParseNumber))]`
Isn't quite informative enough for a query like
```SQL
INSERT INTO feeder.pathology (identity, patient_number, lab_number, item, start, value) SELECT CAST(UNIQUE_IDENT AS INTEGER), UR_Number, Lab_Number, Item, CAST(Service_Date AS INTEGER), CAST(ROUND_VALUE AS INTEGER) FROM load.csv.n2
```

panarch:
> TODO: Better errors
> `[Err(Value(FailedToParseNumber))]`
> Isn't quite informative enough for a query like
> 
> ```sql
> INSERT INTO feeder.pathology (identity, patient_number, lab_number, item, start, value) SELECT CAST(UNIQUE_IDENT AS INTEGER), UR_Number, Lab_Number, Item, CAST(Service_Date AS INTEGER), CAST(ROUND_VALUE AS INTEGER) FROM load.csv.n2
> ```

I agree, the errors early I added need to have more detail on the message.

panarch (changes requested):
Awesome, `CAST` really means a lot.
And also thanks for making a module for functions 👍

There looks something hits me to say about relation between literal (`AstValue` or `sqlparser::ast::Value`) and value (`gluesql::data::Value`).

Though `Evaluated` uses Rust `enum` for representing the structure itself, it is actually not sufficient.
It would be better to say as directed semi-acyclic graph which only allows self-loops.

Beginning nodes are `LiteralRef`, `StringRef` and `ValueRef`.
Terminal node is only one, `Value`. Because we return `Vec<Value>` as a result so it is.

Edges
* `LiteralRef` -> `LiteralRef` | `Literal` | `Value`
* `Literal` -> `Literal` | `Value`
* `StringRef` -> `StringRef` | `Literal` | `Value`
* `ValueRef` -> `ValueRef` | `Value`
* `Value` -> `Value`

However,
```rust
impl TryInto<AstValue> for Value { .. }```
```
and
```rust
impl Value {
    fn into_non_optional(self) -> Value { .. }

    ..
}
```
Those two methods add new edges from `Value` to literal (`AstValue`).

Information between `Value` and `AstValue` are not 1:1 identical, so I think this kind of conversion can be a kind of risk which may cause a problem in the future.

Some examples can be like.. these below.

e.g.
1. `LiteralRef` -> `LiteralRef` : `CAST(123 AS INTEGER)`
`123` begins as a `LiteralRef` and it is already a number so no more action required, `LiteralRef` leaves.

2. `LiteralRef` -> `Literal` : `CAST(123 AS TEXT)`
`123` is a `LiteralRef` but once after converted into `TEXT`, `Evaluated` should have its ownership, so `Literal` is required for the result.

3. `StringRef` -> `Literal` : `CAST("123" AS INTEGER)`
`StringRef` is similar with `LiteralRef` but it is from `ast::Ident`.

4. `ValueRef` -> `ValueRef` : `CAST(id AS INTEGER)`
Let's assum `id` is `OptI64` type, after the `CAST` there's no need to change anything.
So the same `ValueRef` leaves.

---

If we don't care of the performance, we can think `Evaluated` enum with two types, only `Literal` and `Value`.

We can simply begin from converting `LiteralRef` -> `Literal`, `StringRef` -> `Literal` and `ValueRef` -> `Value`.
However, the reason `Evaluated` has those 5 different types is because of the performance.
Trying to reduce unnecessary memory cloning.

Excuse for not sufficient documentation.
In fact, because of your this pr, it made me think more clearly & explicitly.
How do you think of this?

KyGost:
@panarch thoughts?

src/data/value/ast_value.rs
```diff
@@ -73,7 +73,57 @@ impl TryFrom<&AstValue> for Value {
...
+        )),
+        (DataType::Float(_), AstValue::Number(value)) => Ok(AstValue::Number(value)),
+        (DataType::Text, AstValue::Boolean(value)) => Ok(AstValue::SingleQuotedString(
+            (if value { "TRUE" } else { "FALSE" }).to_string(),
+        )),
+        (DataType::Text, AstValue::Number(value)) => Ok(AstValue::SingleQuotedString(value)),
+        (_, AstValue::Null) => Ok(AstValue::Null),
+        _ => Err(ValueError::UnimplementedCast.into()),
+    }
+}
```
panarch, line 129: Let's just move this function to `evaluated.rs`.
Hmm.. it would be better to find a different place soon.
However, currently all pure util functions which handles `AstValue` only are in `evaluated.rs`.

src/executor/evaluate/evaluated.rs
```diff
@@ -356,30 +356,13 @@ impl<'a> Evaluated<'a> {
         let cast_value = |value: &data::Value| value.cast(data_type).map(Evaluated::Value);
 
         // TODO: Avoid StringRef -> AstValue (?)
+        // Decided: Due to the explicit call, we can abandon -Ref
```
panarch, line 359: 👍

src/data/value/ast_value.rs
KyGost: Done :-)
Moved to /data/ast_value.rs

KyGost:
@panarch any ideas on what is causing:
```
thread 'function_cast' panicked at 'assertion failed: `(left == right)`
  left: `Err(Data(ImpossibleCast))`,
 right: `Err(Data(ImpossibleCast))`', C:\Users\kyran.SYRIS\Documents\GitHub\gluesql\src\tests\function\cast.rs:102:9
```

Once that's fixed PR is ready to move out of draft

src/data/value/error.rs
```diff
@@ -40,4 +40,13 @@ pub enum ValueError {
 
     #[error("floating columns cannot be set to unique constraint")]
     ConflictOnFloatWithUniqueConstraint,
+
+    #[error("impossible cast")]
+    ImpossibleCast,
+
+    #[error("unreachable impossible cast")]
+    UnreachableImpossibleCast,
```
panarch, line 48: This is unused, you can remove it.

src/data/value/mod.rs
```diff
@@ -170,6 +169,63 @@ impl Value {
...
+            (DataType::Boolean, Value::F64(value)) =>
+            {
+                #[allow(clippy::float_cmp)]
+                if *value == 1.0 {
+                    Ok(Value::Bool(true))
+                } else if *value == 0.0 {
+                    Ok(Value::Bool(false))
+                } else {
+                    Err(ValueError::ImpossibleCast.into())
+                }
```
panarch, line 199: image

This can make clippy satisfied.
Because `==` operator assumes that value implemented `Eq` but float only has `PartialEq`.
However, still `==` can be used for `PartialEq` but not preferred by strict clippy.

KyGost: Oh okay thanks

src/executor/evaluate/evaluated.rs
```diff
@@ -343,6 +346,23 @@ impl<'a> Evaluated<'a> {
...
+
+        // Decided: Due to the explicit call, we can abandon -Ref
+        match self {
+            StringRef(value) => cast_literal(&AstValue::SingleQuotedString(value.to_string())),
+            LiteralRef(value) => cast_literal(value.to_owned()),
+            Literal(value) => cast_literal(value),
+            ValueRef(value) => cast_value(value.to_owned()),
+            Value(value) => cast_value(value),
+        }
+    }
```
panarch, line 365: You can simply take ownership of the result
```rust
pub fn cast(self, data_type: &DataType) -> Result<Evaluated<'a>> {
    use Evaluated::*;

    let cast_literal =
        |value: &AstValue| cast_ast_value(value, data_type).map(Evaluated::Literal);
    let cast_value = |value: &data::Value| value.cast(data_type).map(Evaluated::Value);

    match self {
        StringRef(value) => cast_literal(&AstValue::SingleQuotedString(value.to_string())),
        LiteralRef(value) => cast_literal(value),
        Literal(value) => cast_literal(&value),
        ValueRef(value) => cast_value(value),
        Value(value) => cast_value(&value),
    }
}
```

Then 
`pub fn cast_ast_value` would be better to take `&AstValue` like `data::Value -> cast` method.
```rust
pub fn cast_ast_value(value: &AstValue, data_type: &DataType) -> Result<AstValue> {
    match (data_type, value) {
        (DataType::Boolean, AstValue::SingleQuotedString(value))
        | (DataType::Boolean, AstValue::Number(value)) => Ok(match value.to_uppercase().as_str() {
            "TRUE" | "1" => Ok(AstValue::Boolean(true)),
            "FALSE" | "0" => Ok(AstValue::Boolean(false)),
            _ => Err(EvaluateError::ImpossibleCast),
        }?),
        (DataType::Int, AstValue::Number(value)) => Ok(AstValue::Number(
            value
                .parse::<f64>()
                .map_err(|_| EvaluateError::UnreachableImpossibleCast)?
                .trunc()
                .to_string(),
        )),
        (DataType::Int, AstValue::SingleQuotedString(value))
        | (DataType::Float(_), AstValue::SingleQuotedString(value)) => Ok(AstValue::Number(value.clone())),
        (DataType::Int, AstValue::Boolean(value))
        | (DataType::Float(_), AstValue::Boolean(value)) => Ok(AstValue::Number(
            (if *value { "1" } else { "0" }).to_string(),
        )),
        (DataType::Float(_), AstValue::Number(value)) => Ok(AstValue::Number(value.clone())),
        (DataType::Text, AstValue::Boolean(value)) => Ok(AstValue::SingleQuotedString(
            (if *value { "TRUE" } else { "FALSE" }).to_string(),
        )),
        (DataType::Text, AstValue::Number(value)) => Ok(AstValue::SingleQuotedString(value.clone())),
        (_, AstValue::Null) => Ok(AstValue::Null),
        _ => Err(EvaluateError::UnimplementedCast.into()),
    }
}
```

panarch: fyi. That is for reducing `clone()` calls for `CAST`, using this way, it can handle `cast` with a single clone call.

KyGost: Ah! Thanks!

panarch (approved):
All looks good, beautiful :) Thanks so so much.

PR #145, opened 2021-02-27
Single Quotes

KyGost:
Relates to #76

### DIFF
--- a/src/data/value.rs
+++ b/src/data/value.rs
@@ -288,6 +288,7 @@ impl Value {
             (DataType::Boolean, AstValue::Boolean(v)) => {
                 Ok(nullable.into_value(Value::OptBool(Some(*v)), Value::Bool(*v)))
             }
+            (DataType::Text, AstValue::SingleQuotedString(v)) => Ok(Value::Str(v.to_string())),
             (DataType::Int, AstValue::Null) => nullable.as_result(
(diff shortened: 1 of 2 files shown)


KyGost:
Note: bug not enhancement
(Though I suppose they have arbitrary differences)

src/data/value.rs
```diff
@@ -288,6 +288,7 @@ impl Value {
             (DataType::Boolean, AstValue::Boolean(v)) => {
                 Ok(nullable.into_value(Value::OptBool(Some(*v)), Value::Bool(*v)))
             }
+            (DataType::Text, AstValue::SingleQuotedString(v)) => Ok(Value::Str(v.to_string())),
```
KyGost, line 291: I noticed after the fact that others have solved this issue too.
Though both of them changed it to
`Ok(nullable.as_value(Value::OptStr(Some(v.clone())), Value::Str(v.clone())))`

Would that be better?

panarch: Oh yes, that should be changed like you wrote :) thanks

panarch (changes requested):

src/data/value.rs
KyGost: Done :-)

panarch (approved):
Thanks 👍 👍 👍

PR #146, opened 2021-02-27
Load from file example for book

KyGost:
Low quality example for book.

Plausibly too low quality to be considered else but a draft.

Added to satisfy #73


KyGost:
Failed checks. Just going to convert to draft for now. Will work on at some point.

KyGost:
Cancel that it won't let me convert.

Ignore for now.

PR #154, opened 2021-03-01
Split data/value.rs single file module into multiple files

panarch:
`gluesql::data::Value` module has grown quite much, so it is too big to contain all codes in a single file.
So, I convert a single file `value.rs` to directory based module.
There's no structural change, only files are splitted.

@KyGost
If you are ok to merge this before [pr#144](https://github.com/gluesql/gluesql/pull/144), then I'll merge this into main before #144 merged, you might see small conflict.
Or I'm also fully ok to merge this after your #144 merged into main, then I'll do by myself. Please just let me know what you prefer.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 # GlueSQL project files
-data/
+/data/
 
 # Vim
 *.swp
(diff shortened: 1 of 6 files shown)


KyGost:
I'll make #144 fit :-) all good to merge

PR #155, opened 2021-03-02
Clean up evaluated.rs with macros

KyGost:

PR #156, opened 2021-03-02
Clean up evaluated.rs with macros

KyGost:
Rebuild of #155

### DIFF
--- a/src/executor/evaluate/evaluated.rs
+++ b/src/executor/evaluate/evaluated.rs
@@ -150,18 +150,18 @@ macro_rules! binary_op {
     ($name:ident, $op:tt) => {
         pub fn $name(&self, other: &Evaluated<'a>) -> Result<Evaluated<'a>> {
             let literal_binary_op = |l: &&AstValue, r: &&AstValue| match (l, r) {
-                (AstValue::Number(a), AstValue::Number(r)) => match (a.parse::<i64>(), r.parse::<i64>()) {
-                    (Ok(a), Ok(r)) => Ok(AstValue::Number((a $op r).to_string())),
-                    (Ok(a), _) => match r.parse::<f64>() {
-                        Ok(r) => Ok(AstValue::Number(((a as f64) $op r).to_string())),
+                (AstValue::Number(l), AstValue::Number(r)) => match (l.parse::<i64>(), r.parse::<i64>()) {
+                    (Ok(l), Ok(r)) => Ok(AstValue::Number((l $op r).to_string())),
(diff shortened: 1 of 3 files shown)


src/executor/evaluate/evaluated.rs
```diff
@@ -182,126 +146,145 @@ impl TryInto<Value> for Evaluated<'_> {
...
+                        _ => Err(EvaluateError::UnreachableLiteralArithmetic.into()),
+                    },
+                    (_, Ok(r)) => match a.parse::<f64>() {
+                        Ok(a) => Ok(AstValue::Number((a $op (r as f64)).to_string())),
+                        _ => Err(EvaluateError::UnreachableLiteralArithmetic.into()),
+                    },
+                    (_, _) => match (a.parse::<f64>(), r.parse::<f64>()) {
+                        (Ok(a), Ok(r)) => Ok(AstValue::Number((a $op r).to_string())),
+                        _ => Err(EvaluateError::UnreachableLiteralArithmetic.into()),
+                    },
```
panarch, line 166: You might want to change `a` to `l` :)

src/executor/evaluate/evaluated.rs
```diff
@@ -182,126 +146,145 @@ impl TryInto<Value> for Evaluated<'_> {
...
+}
 
-        let multiply_value = |l: &data::Value, other: &Evaluated<'a>| match other {
-            LiteralRef(r) => l.multiply(&l.clone_by(r)?).map(Evaluated::Value),
-            Literal(r) => l.multiply(&l.clone_by(&r)?).map(Evaluated::Value),
-            ValueRef(r) => l.multiply(r).map(Evaluated::Value),
-            Value(r) => l.multiply(&r).map(Evaluated::Value),
-            StringRef(_) => unreachable(),
-        };
+macro_rules! general_op {
```
panarch, line 222: This is only used for `cast`, so.. how about keeping your original one?

It is because.. the last refactoring would be not a macro, I think we can do refactoring codes again using generic function.
So, this looks like unnecessary margin.

panarch (changes requested):

src/executor/evaluate/evaluated.rs
KyGost: Ah yes, thought I saw that then I couldn't find it again! Haha

src/executor/evaluate/evaluated.rs
KyGost: Yeah, kinda unneccessary. I agree. Will undo.

panarch (approved):
Thanks!! 👍 👍 👍

Issue #150, opened 2021-02-28
LEFT and RIGHT functions


KyGost:
`LEFT` and `RIGHT` `TEXT` functions would be nice and shouldn't be too challenging.
https://docs.microsoft.com/en-us/sql/t-sql/functions/left-transact-sql?view=sql-server-ver15
https://docs.microsoft.com/en-us/sql/t-sql/functions/right-transact-sql?view=sql-server-ver15

If I don't get around to it first, it might be a good first issue for someone.

PR #157, opened 2021-03-02
LEFT and RIGHT functions

KyGost:
Relates to #150


src/executor/evaluate/mod.rs
```diff
@@ -161,6 +165,45 @@ async fn evaluate_function<'a, T: 'static + Debug>(
...
+                    };
+                    string.get(start_pos..)
+                }
+                .or(Some(&string))
+                .unwrap()
+                .to_string()
+            };
+
+            if args.len() != 2 {
+                return Err(EvaluateError::NumberOfFunctionParamsNotMatching {
```
panarch, line 186: Though this is same as `LOWER` & `UPPER`, it has different location so we'd better to have a test case for generating this, too.

src/executor/evaluate/mod.rs
```diff
@@ -161,6 +165,45 @@ async fn evaluate_function<'a, T: 'static + Debug>(
...
+                    expected: 2,
+                    found: args.len(),
+                }
+                .into());
+            }
+
+            let string = match eval(&args[0]).await?.try_into()? {
+                Value::Str(string) | Value::OptStr(Some(string)) => Ok(string),
+                _ => Err(EvaluateError::FunctionRequiresStringValue(name.to_string())),
```
panarch, line 195: Like the error `EvaluateError::NumberOfFunctionParamsNotMatching`, it would be also good to have for `LEFT` and `RIGHT`.

src/executor/evaluate/mod.rs
```diff
@@ -161,6 +165,45 @@ async fn evaluate_function<'a, T: 'static + Debug>(
...
+            }?;
+            let number = match eval(&args[1]).await?.try_into()? {
+                Value::I64(number) | Value::OptI64(Some(number)) => usize::try_from(number)
+                    .map_err(|_| EvaluateError::FunctionRequiresUSizeValue(name.to_string())), // Unlikely to occur hence the imperfect error
+                _ => Err(EvaluateError::FunctionRequiresIntegerValue(
```
panarch, line 200: A test case is required for generating this error.

KyGost:
@panarch I know realise the issue with the way I was doing it and why it was being mapped!-- `None`

src/executor/evaluate/mod.rs
```diff
@@ -161,6 +165,45 @@ async fn evaluate_function<'a, T: 'static + Debug>(
...
+            if args.len() != 2 {
+                return Err(EvaluateError::NumberOfFunctionParamsNotMatching {
+                    expected: 2,
+                    found: args.len(),
+                }
+                .into());
+            }
+
+            let string = match eval(&args[0]).await?.try_into()? {
+                Value::Str(string) | Value::OptStr(Some(string)) => Ok(string),
```
panarch, line 194: https://github.com/gluesql/gluesql/pull/161

Yup :) You need to handle `Str` and `OptStr` differently.

src/executor/evaluate/mod.rs
```diff
@@ -161,6 +165,45 @@ async fn evaluate_function<'a, T: 'static + Debug>(
...
+                _ => Err(EvaluateError::FunctionRequiresStringValue(name.to_string())),
+            }?;
+            let number = match eval(&args[1]).await?.try_into()? {
+                Value::I64(number) | Value::OptI64(Some(number)) => usize::try_from(number)
+                    .map_err(|_| EvaluateError::FunctionRequiresUSizeValue(name.to_string())), // Unlikely to occur hence the imperfect error
```
panarch, line 199: This looks good that it can cover not only for the literal but also it can handle `expr`, that's really cool. thanks 👍 
However, current tests only cover literal, `AstValue`.
Can you add these two test cases?
1. A test case which takes `Value::OptI64(Some(number))` branch.
2. An error test case which generates `EvaluateError::FunctionRequiresUSizeValue`

src/tests/function/left_right.rs
```diff
@@ -0,0 +1,65 @@
...
+            r#"SELECT RIGHT(name, 10) AS test FROM Item"#,
+            Ok(select!(
+                "test"
+                Str;
+                "op mc blee".to_owned();
+                "B".to_owned();
+                "d$ folken!".to_owned()
+            )),
+        ),
+        /*( Concatenation not currently supported
```
panarch, line 31: Sounds cool! let's just add `// TODO: --` not to forget 👍

panarch (changes requested):

src/executor/evaluate/mod.rs
```diff
@@ -161,6 +165,45 @@ async fn evaluate_function<'a, T: 'static + Debug>(
...
+                } else {
+                    let start_pos = if number > string.len() {
+                        0
+                    } else {
+                        string.len() - number
+                    };
+                    string.get(start_pos..)
+                }
+                .or(Some(&string))
+                .unwrap()
```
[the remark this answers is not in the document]
KyGost: Hmmm... Is it problematic?
It should be a guaranteed success?

How would you suggest it be done?

src/tests/function/left_right.rs
```diff
@@ -0,0 +1,179 @@
...
+                "blu".to_owned()
+            )),
+        ),
+        (
+            r#"SELECT LEFT(name, 3) AS test FROM NullName"#,
+            Ok(select!(
+                "test"
+                Str;
+                "".to_owned()
+            )),
```
panarch, line 96: Here, the type of `name` is `OptStr`, then I think it's better to keep `OptStr` and keep `None`. How do you think of this?
This `LEFT` converted `NULL` to empty string.

panarch (changes requested):
Looks really good,

src/tests/function/left_right.rs
KyGost: Ah good point, thanks, was just copy pasting :-)

src/executor/evaluate/mod.rs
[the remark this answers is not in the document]
KyGost: Ah! `unwrap_or`! Thanks.

src/tests/function/left_right.rs
KyGost: Oh oops now I see what you mean, yeah. Thanks.

src/tests/function/left_right.rs
```diff
@@ -0,0 +1,179 @@
...
+                None
+            )),
+        ),
+        (
+            r#"SELECT LEFT(name, 1) AS test FROM NullableName"#,
+            Ok(select!(
+                "test"
+                Str;
+                "n".to_owned()
+            )),
```
panarch, line 120: The last one, I saw that you changed to use `OptStr`, and this would be also good to use `OptStr`.
If it doesn't, then
`OptStr(Some(..))` becomes `Str` and `OptStr(None)` becomes `OptStr`.
Then column type can be different for each row.
Thought current tests only have 1 row, but using 2 rows with `OptStr(Some("a".to_owned()))` and `OptStr(None)`, we can break the column type consistency.

panarch (changes requested):

src/tests/function/left_right.rs
KyGost: Good point, just left them there.

src/executor/evaluate/mod.rs
```diff
@@ -175,6 +179,51 @@ async fn evaluate_function<'a, T: 'static + Debug>(
 
             Ok(Evaluated::Value(value))
         }
+        name @ "LEFT" | name @ "RIGHT" => {
+            let return_wrap = |value| Ok(Evaluated::Value(Value::OptStr(value)));
```
panarch, line 183: Oh, let's don't too rush

This code converts not only `Value::OptStr` to `OptStr`, but also it converts `Value::Str` to `Value::OptStr`

KyGost: I figured there wasn't much need for it, especially considering we'll be changing soon.

Do you think we need to return Str?

panarch: This is not because of any special reason.
For example, this is the code above about `LOWER` and `UPPER`.
image

LEFT | RIGHT of `Value::Str` becomes `Value::Str`.
LEFT | RIGHT of `Value::OptStr` becomes `Value::OptStr`.

Once, `LEFT` and `RIGHT` converts `Value::Str` to `Value::OptStr`, then `None` value of `Value::OptStr` becomes unreachable.
Using normal input sql, it is not possible to generate `None` from the result of `LEFT` + `Value::Str`.

KyGost: I suppose that's fair though it would require duplicating the code as the convert function takes two arguments.

Something to consider:
`LEFT` & `RIGHT` aren't just manipulations of a `Str`/`OptStr` column, they're their own values, determined by a `Str`/`OptStr` and `I64`/`OptI64`.

As per this, it *is* possible to generate `None` from the result of `LEFT` + `Str` **+** `OptI64(None)`.

panarch: > As per this, it is possible to generate None from the result of LEFT + Str + OptI64(None).
Oh, thanks, then it make sense,

* LEFT("abcd", 2) => "ab"
* LEFT(NULL, 3) => NULL
* LEFT(NULL, NULL) => NULL
* LEFT("abcd", NULL) => NULL

I thought the result of `LEFT("abcd", NULL)` as "abcd".

`LEFT("abcd", NULL) => NULL` can convert `Str` to `OptStr`.
Ok, after you fix the small clippy issue and I'll merge this :) Thanks! 👍 👍

panarch (approved):
Beautiful, thanks!! 👍 👍 👍

PR #162, opened 2021-03-05
Change literal casting function to return from AstValue to Value,

panarch:
Add `Value::TryFromLiteral` to replace `cast_ast_value` in `evaluated.rs`.
Now it can handle `CAST(1 AS FLOAT)`.
Test cases in `function::cast::cast_literal` reach all branches in `Value::TryFromLiteral` match.

@KyGost You even left a comment about `CAST(1 AS FLOAT)` issue but I missed to check it.
I just figured out and that was because of... function return value type.
Previous: `AstValue -> AstValue`
Current: `AstValue -> Value`
This was possible because `CAST` provides `DataType` information.

And I modified `cast_literal` test cases little bit to reach every branches in `match`
```rust
impl TryFromLiteral for Value {
    fn try_from_literal(data_type: &DataType, literal: &AstValue) -> Result<Value> {
       match ...
    }
}
```
Order of match branches (`try_from_literal`) and test cases (`function::cast::cast_literal`) is same, you will be able to check simply following the code lines.
`CAST` were too big task so my review was.... not sufficient, sorry for that. Writing test cases to reach every branches would be helpful for you to work on https://github.com/gluesql/gluesql/pull/157

I also found some issues on handling `cast_value`, I might work for casting value on Sunday, then I'll also share with you. :)

### DIFF
--- a/src/data/mod.rs
+++ b/src/data/mod.rs
@@ -1,7 +1,7 @@
 mod row;
 mod schema;
 mod table;
-mod value;
+pub mod value;
 
 pub use {
     row::{Row, RowError},
--- a/src/data/value/ast_value.rs
+++ b/src/data/value/ast_value.rs
@@ -1,7 +1,7 @@
 use {
     super::{error::ValueError, Value},
     crate::result::{Error, Result},
-    sqlparser::ast::Value as AstValue,
+    sqlparser::ast::{DataType, Value as AstValue},
     std::{cmp::Ordering, convert::TryFrom},
 };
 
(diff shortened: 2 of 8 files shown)


KyGost:
Oh, we've gone back to always converting `AstValue` to `Value` I thought we were avoiding this?

I figured we'd be able to fix the float issue with natural conversion.

panarch:
@KyGost
`AstValue` -> `Value` is good and it's right direction because our execute's return data is `Vec<Value>`.
`Value` -> `AstValue` is the thing that we need to avoid, it can make cycle in `Evaluated`.

Though `Evaluated` has 5 types, basically you can simplify into two, `Literal (AstValue)` and `Value`.

* `Literal` -> `Literal` or `Value`
* `Value` -> `Value`

It's all :)

https://github.com/gluesql/gluesql/blob/d642c75a84267f1a60e4505119ca6fdcde516e1f/src/data/value/mod.rs#L132
```rust
pub fn from_data_type(
        data_type: &DataType,
        nullable: bool,
        literal: &AstValue,
    ) -> Result<Self> {
```
Here, the condition to make `Value` requires three params, data type, nullable and literal.
It's little bit different from `TryFromLiteral` because it only takes two, there's no nullable.
However, it is ok for `TryFromLiteral` not to use nullable.
This is due to...
1. `CAST` cannot provide that.
2. All compare & calculation methods in `Value` (`PartialEq`, `PartialOrd`, `add`, `subtract`.. ) supports operation between non-nullable enums and `Opt-` prefixed nullable enums.

e.g.
```
Value::I64(100) + Value::OptI64(Some(200)) => Value::OptI64(Some(300))
```

PR #163, opened 2021-03-07
Update Value->cast to support nullable types,

panarch:
Provide OptBool, OptI64, OptF64 and OptStr CAST supports.
Add unit tests to CAST on Value.

@KyGost I added casting values to support `Opt-` types.

`Value` is a bit different from other modules.
For other executor codes, `Value` works like a blackbox.
It means, not like other codes, `Value` need to prove the correctness itself by unit tests.

However, I didn't take care enough about unit tests in `Value`... not a good thing.
I just tried the first unit test on `Value`, it takes all normal branches which can be reached in `cast` function.
image

### DIFF
--- a/src/data/value/mod.rs
+++ b/src/data/value/mod.rs
@@ -174,20 +174,37 @@ impl Value {
         match (data_type, self) {
             // Same as
             (DataType::Boolean, Value::Bool(_))
+            | (DataType::Boolean, Value::OptBool(_))
+            | (DataType::Text, Value::Str(_))
+            | (DataType::Text, Value::OptStr(_))
             | (DataType::Int, Value::I64(_))
+            | (DataType::Int, Value::OptI64(_))
             | (DataType::Float(_), Value::F64(_))
-            | (DataType::Text, Value::Str(_)) => Ok(self.clone()),
+            | (DataType::Float(_), Value::OptF64(_)) => Ok(self.clone()),
+
             // Boolean
-            (DataType::Boolean, Value::Str(value)) => Ok(match value.to_uppercase().as_str() {
+            (DataType::Boolean, Value::Str(value)) => match value.to_uppercase().as_str() {
                 "TRUE" => Ok(Value::Bool(true)),
                 "FALSE" => Ok(Value::Bool(false)),
-                _ => Err(ValueError::ImpossibleCast),
-            }?),
-            (DataType::Boolean, Value::I64(value)) => Ok(match value {
+                _ => Err(ValueError::ImpossibleCast.into()),
+            },
+            (DataType::Boolean, Value::OptStr(Some(value))) => {
+                match value.to_uppercase().as_str() {
+                    "TRUE" => Ok(Value::OptBool(Some(true))),
+                    "FALSE" => Ok(Value::OptBool(Some(false))),
+                    _ => Err(ValueError::ImpossibleCast.into()),
+                }
+            }
+            (DataType::Boolean, Value::I64(value)) => match value {
                 1 => Ok(Value::Bool(true)),
                 0 => Ok(Value::Bool(false)),
-                _ => Err(ValueError::ImpossibleCast),
-            }?),
+                _ => Err(ValueError::ImpossibleCast.into()),
+            },
+            (DataType::Boolean, Value::OptI64(Some(value))) => match value {
+                1 => Ok(Value::OptBool(Some(true))),
+                0 => Ok(Value::OptBool(Some(false))),
+                _ => Err(ValueError::ImpossibleCast.into()),
+            },
             (DataType::Boolean, Value::F64(value)) => {
                 if value.eq(&1.0) {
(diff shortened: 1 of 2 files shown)


KyGost:
Is there any situation where `Opt`- isn't supported but non opt is?

Would it be okay to always return an Opt?

Then we can clean up this code with stuff like

```rust
(Boolean, Str(value)) => ...
Bool(value)
(Boolean, OptStr(Some(value))) => ...
OptBool(Some(Value))
```
->
```rust
(Boolean, Str(value))
|  (Boolean, OptStr(Some(value))) => ... OptBool(Some(value))
```

Is there extra computation needed?

Also if not, why do we need seperate `Value::Bool` and `Value::OptBool`?

KyGost:
Thinking about it further, I know it makes sense as it is now from a database perspective but what if we changed it to `Result<Value>` instead?

panarch:
@KyGost
In the case of returning only non `Opt-`, it can make column types not consistent per row.
The opposite is different. If we always return `Opt-` then you might be right. At least it can work.

However, I think we don't need to put unreachable information to existing values.
e.g.
```sql
CREATE TABLE Foo (id INTEGER NOT NULL);
INSERT INTO Foo VALUES (1), (0);

SELECT CAST(id AS BOOLEAN) FROM Foo;
/*
 * vec![
 *   vec![ Value::OptBool(Some(true)) ],
 *   vec![ Value::OptBool(Some(false)) ],
 * ]
 */
```
Above query still works by returning `OptBool`.
Though it is obvious that column `id` cannot have `NULL`, returning value implies it can have `NULL` due to `OptBool`.
Unnecessary information is injected, and that `NULL` in the output value is unreachable.

---
> Thinking about it further, I know it makes sense as it is now from a database perspective but what if we changed it to Result<Value> instead?

If I understand well, then.. I'm not sure because `NULL` is not an error, it's one of normal value types.

---
`Value::Empty`, `Value::Opt-` can be confusing because it is designed for providing more information to users. (it might be only for Rust users).

SQL handles empty value from outer join and null value in a same way, both simply have `NULL`, same value.
However, it is obviously different.
And... when we only see the result `NULL`, then we cannot figure out that `NULL` is non-existence of what.

In here GlueSQL,
* By `Value::Opt-`, users can check `None` is absence of what, `None` of `INT`, `None` of `TEXT`, etc..
* `Value::Empty` represents `NULL` by outer join result is empty.

KyGost:
Hmm..
It seems like double the matching and a bunch of additional testing and maintainance to maintain `Value<(Option)>` over `Option<Value>` is the data type information worth it?
Would we not expect users to know what columns are? And it's not like `NULL` *has* a data type?

panarch:
@KyGost Thanks for mentioning this, it really makes me think a lot.
The motivation is making users and us not to worry about null checks, this is what Rust provides.
Though SQL itself is a language having `NULL`, I thought it would be possible to specify explicit types to column.

In this kind of example,

```rust
"SELECT id, num, name FROM Test",      
select!(                               
    id      | num | name               
    OptI64  | I64 | Str;               
    None      2     "Hello".to_owned();
    Some(1)   9     "World".to_owned();
    Some(3)   4     "Great".to_owned() 
),                                  
```

Because we have `OptI64` optional type, so we can set a type for the first column.
However, it still has limitation on JOIN query, because gluesql still uses... `Value::Empty`.
```rust
let sql = "                       
    SELECT p.id, i.id             
    FROM Player p                 
    LEFT JOIN Item i              
    ON p.id = i.player_id         
";                                
let found = run!(sql);            
let expected = select_with_empty!(
    id     | id;                  
    I64(1)   I64(101);            
    I64(2)   I64(102);            
    I64(3)   Empty;               
    I64(4)   I64(103);            
    I64(5)   Empty                
);                                
assert_eq!(expected, found);
```
We cannot ensure that same column of each row will have same type.

There exists 3 elements.
* data types such as `Bool`, `I64`, `F64` and `Str`.
* nullable : NULL or NOT NULL (`Opt-`)
* outer join failed

Currently we have 9 different types by doubling `Bool`, `I64`, --`.
However, to support non-null functionality... we actually do doubling once more... then 4 data types cause -> 4 * 2 * 2 => 16 types...
As you already experienced enough, it is already quite tough to handle 9 different types, so I gave up and used `Value::Empty` for join condition failure cases.

The time I decided to use `Opt-` combinations, I thought there might be a better way to write `Value` codes more readable and efficiently in the future. current codes is just because I was not good enough at using Rust.
However, I'm still not sure how to catch two rabbits,
1. values without nulls like Rust 
2. making easier to write `Value` codes

If we give up non null rules, then we can actually simply remove `Opt` value, and add `Value::Null`.
Then, for now we only need 5 types,
* Value::Bool, Value::I64, Value::F64, Value::Str and Value::Null

If we find an elegance way to strict and somewhat right way, then
We can have 16 different types, but it will be as convenient as writing with 5 types.

Issue #159, opened 2021-03-03
AUTO_INCREMENT


KyGost:
It would be quite nice to have an `AUTO_INCREMENT` feature for `INTEGER` columns.

There's a few approaches to how this could work though:

SQLite automatically acts on `INTEGER PRIMARY KEY` columns, backfilling where values are removed while the keyword specifies that backfill should not occur.
https://sqlite.org/autoinc.html

MySQL acts based on the current largest value in the column when the keyword is used.
https://dev.mysql.com/doc/refman/8.0/en/example-auto-increment.html

MSSQL stores a value and uses that. The value can be altered.
https://docs.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql-identity-property?view=sql-server-ver15

Anyone have any preference?

I'm not sure that I like how SQLite increments by default.

I kinda like the MySQL method; MSSQL sounds a bit messy and I worry for the potential issues of backfilling (SQLite).

panarch:
I'm also familiar with MySQL's `AUTO_INCREMENT` syntax.

There can be something to consider about the implementation plan.
I'm not sure that current `Store` trait methods are enough to support `AUTO_INCREMENT`.
Then, we might need to add another optional store trait for this?

KyGost:
If you were using the MySQL or SQLite method with backfill, I don't think you would need to store anything in the schema

KyGost:
Hmm
PostGreSQL does:
https://www.postgresql.org/docs/13/ddl-generated-columns.html

KyGost:
I tried a number of ways writing it how I might expect it to be writable and with the current sqlparser, `id INTEGER CONSTRAINT AUTO_INCREMENT NOT NULL` seems to be the only reasonable way.

(With the NOT NULL always required)

Thoughts on `AUTO_INCREMENT` vs `AUTOINCREMENT`?

KyGost:
nb:
```sql
id INTEGER CONSTRAINT AUTO_INCREMENT NOT NULL
```
Gives us:
```
[ColumnDef { name: Ident { value: "id", quote_style: None }, data_type: Int, collation: None, options: [ColumnOptionDef { name: Some(Ident { value: "AUTO_INCREMENT", quote_style: None }), option: NotNull }] }]
```

KyGost:
`CONSTRAINT AUTO_INCREMENT UNIQUE` & `CONSTRAINT AUTO_INCREMENT PRIMARY KEY` also works

[...]

PR #164, opened 2021-03-09
AUTO_INCREMENT

KyGost:
Relates: #159


KyGost:
Quite yuck and not yet working for rows inserted without `NULL` but working nonetheless.

@panarch how do you feel about giving `StoreMut` to `Execute::prepare` and `Row::new`?

panarch:
@KyGost 
1. `Execute::prepare`
I actually think `Execute::prepare` is... not meaningful to keep, first intention was dividing `Store` uses and `Store + StoreMut` uses but now I'm not sure that is really meaningful.
It might be better to re-combine `prepare` and `execute`, how do you think?

2. `Row::new`
You mean that providing `Store` to `Row::new`? then I think it would be good to keep `data` modules not to depend on `Store`. making it lightweight, containing "data" and some useful utility methods.

KyGost:
@panarch 
> 1. `Execute::prepare`
>    I actually think `Execute::prepare` is... not meaningful to keep, first intention was dividing `Store` uses and `Store + StoreMut` uses but now I'm not sure that is really meaningful.
>    It might be better to re-combine `prepare` and `execute`, how do you think?

That sounds like a good idea.


> 2. `Row::new`
>    You mean that providing `Store` to `Row::new`? then I think it would be good to keep `data` modules not to depend on `Store`. making it lightweight, containing "data" and some useful utility methods.

Yup. I agree, it felt off to give it. How would you suggest it be done?
I need to run a `tree.get` & `tree.set` for each row.

KyGost:
resolve #159

PR #167, opened 2021-03-09
Merge Prepared & Execute in execute.rs

KyGost:


panarch (changes requested):
I'll leave this comment, and you can check it tomorrow :)

It is not necessary to change behavior, `Clone`. because this should be a merging codes, it should not cause any behavioral changes.

```rust
 macro_rules! try_into {                          
     ($storage: expr, $expr: expr) => {           
         match $expr {                            
             Err(e) => {                          
                 return Err(($storage, e.into()));
             }                                    
             Ok(v) => ($storage, v),              
         }                                        
     };                                           
 }        
```
I saw that you used `clone` because of `try_into!`.

Here, more than one `try_into!` is used, so you need to change it slightly.
```rust
             Ok(v) => ($storage, v),              
```

Then, you can keep ownership of `Store`, you don't need to clone.

KyGost:
Hmm that seems like it might be a mess. I see now why they were separated.

src/executor/execute.rs
```diff
@@ -189,177 +162,168 @@ pub async fn execute<T: 'static + Debug, U: Store<T> + StoreMut<T> + AlterTable>
...
-            Ok(Prepared::Insert(table_name, column_defs, rows))
+            let column_defs = Rc::from(column_defs);
+            let validated = validate_rows(
+                &storage,
+                &table_name,
+                ColumnValidation::All(Rc::clone(&column_defs)),
+                rows.iter(),
+            )
+            .await;
+            try_into!(storage, validated);
```
panarch, line 222:
```rust
let (rows, table_name) = try_block!(storage, {                             
    let table_name = get_name(table_name)?;                                
    let Schema { column_defs, .. } = storage                               
        .fetch_schema(table_name)                                          
        .await?                                                            
        .ok_or::<Error>(ExecuteError::TableNotExists.into())?;             
                                                                           
    let rows = match &source.body {                                        
        SetExpr::Values(Values(values_list)) => values_list                
            .iter()                                                        
            .map(|values| Row::new(&column_defs, columns, values))         
            .collect::<Result<Vec<Row>>>()?,                               
        SetExpr::Select(select_query) => {                                 
            let query = || sqlparser::ast::Query {                         
                with: None,                                                
                body: SetExpr::Select(select_query.clone()),               
                order_by: vec![],                                          
                limit: None,                                               
                offset: None,                                              
                fetch: None,                                               
            };                                                             
                                                                           
            select(&storage, &query(), None)                               
                .await?                                                    
                .try_collect::<Vec<_>>()                                   
                .await?                                                    
        }                                                                  
        set_expr => {                                                      
            return Err(ExecuteError::UnreachableUnsupportedInsertValueType(
                set_expr.to_string(),                                      
            )                                                              
            .into());                                                      
        }                                                                  
    };                                                                     
                                                                           
    let column_defs = Rc::from(column_defs);                               
    validate_rows(                                                         
        &storage,                                                          
        &table_name,                                                       
        ColumnValidation::All(Rc::clone(&column_defs)),                    
        rows.iter(),                                                       
    )                                                                      
    .await?;                                                               
                                                                           
    Ok((rows, table_name))                                                 
});                                                                        
```

src/executor/execute.rs
```diff
@@ -189,177 +162,168 @@ pub async fn execute<T: 'static + Debug, U: Store<T> + StoreMut<T> + AlterTable>
...
+                match select_with_labels(&storage, &query, None, true).await {
+                    Ok((labels, rows)) => Ok((labels, rows.try_collect::<Vec<_>>().await)),
+                    Err(err) => Err(err),
+                }
+            };
 
-            Ok(Prepared::AlterTable(table_name, operation))
-        }
+            let (labels, rows) = try_into!(storage, select_with_labels().await);
+            let rows = try_into!(storage, rows);
```
panarch, line 323:
```rust
let (labels, rows) = try_block!(storage, {                                       
    let (labels, rows) = select_with_labels(&storage, &query, None, true).await?;
    let rows = rows.try_collect::<Vec<_>>().await?;                              
                                                                                 
    Ok((labels, rows))                                                           
});                                                                              
```

panarch (changes requested):
During the review, I got that what you might get stuck at.
Awesome, you solved all of that.

Let's make the codes more simple.
In this circumstance, keeping `try_into!` does not look like a good idea.

```rust
macro_rules! try_block {                     
    ($storage: expr, $block: block) => {{    
        match (|| async { $block })().await {
            Err(e) => {                      
                return Err(($storage, e));   
            }                                
            Ok(v) => v,                      
        }                                    
    }};                                      
}
```
Above one `try_block!` macro would be a good alternative.
You will be able to replace all `try_into!` to `try_block!`, let's remove `try_into!`.

In `try_block!` macro, we can use try operator `?` without any issue.
No tricky closure and messy `try_into!` are required
I added a few examples of conversion. :)

src/executor/execute.rs
```diff
@@ -71,88 +71,61 @@ pub async fn execute<T: 'static + Debug, U: Store<T> + StoreMut<T> + AlterTable>
...
+                    if !if_exists {
+                        try_into!(storage, schema.ok_or(ExecuteError::TableNotExists));
+                    }
 
-            storage
-                .update_data(rows)
+                    match storage.delete_schema(table_name).await {
+                        Ok((storage, _)) => Ok(storage),
+                        Err(error) => Err(error),
+                    }
```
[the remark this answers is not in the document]
KyGost: But I want it to stay a tuple for error case?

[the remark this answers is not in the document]
KyGost: Oh weird, it works

src/executor/execute.rs
KyGost: Oh wow! Nice!

KyGost:
Much nicer!

KyGost:
I left AlterTable because it would be a bit of a mess to use `try_block` there, `try_into` works nicer.

src/executor/execute.rs
```diff
@@ -189,177 +178,152 @@ pub async fn execute<T: 'static + Debug, U: Store<T> + StoreMut<T> + AlterTable>
...
-            let names = names
-                .iter()
-                .map(|name| get_name(name).map(|n| n.as_str()))
-                .collect::<Result<_>>()?;
+                let keys = fetch(&storage, table_name, columns, filter)
+                    .await?
+                    .map_ok(|(_, key, _)| key)
+                    .try_collect::<Vec<_>>()
+                    .await?;
+                Ok(keys)
```
panarch, line 305: Very tiny one, 
```rust
fetch(&storage, table_name, columns, filter)
    .await?
    .map_ok(|(_, key, _)| key)
    .try_collect::<Vec<_>>()
    .await
```

Now no need of `?` and re-wrap by `Ok`.

src/executor/execute.rs
```diff
@@ -70,89 +71,77 @@ pub async fn execute<T: 'static + Debug, U: Store<T> + StoreMut<T> + AlterTable>
...
-                    .await
-                    .map(|(storage, _)| storage)
-            })
-            .await
-            .map(|storage| (storage, Payload::DropTable)),
-        Prepared::Select { labels, rows } => Ok((storage, Payload::Select { labels, rows })),
 
         #[cfg(feature = "alter-table")]
-        Prepared::AlterTable(table_name, operation) => {
+        Statement::AlterTable { name, operation } => {
```
panarch, line 142: Your decision to keep `try_into!` make sense, it is still useful in `AlterTable`.
However, there are no other use cases except for `AlterTable`. so, Let's move `try_into!` to here.

Then you can remove `#[allow(unused_macros)]`

panarch (changes requested):
Beautiful :) Thanks so much!

I'd like to request only very tiny changes to remove `#[allow(unused_macros)]`

src/executor/execute.rs
KyGost: Hmmm it doesn't let me put it there
But I've just realised I could just use the feature item for the macro

panarch (approved):
Thanks! :) merging!

Issue #166, opened 2021-03-09
Refactoring crate::data::Value and removing Value::Empty


panarch:
Thanks to @KyGost 👍 

Prior threads,
- https://github.com/gluesql/gluesql/pull/163
- https://discord.com/channels/780298017940176946/780298017940176949/818732114782060555

Goals,
1. Removing `Value::Empty` and replace it to use `Opt-` prefixed values.
`Value::Empty` was for representing `NULL` values which is filter out by outer join condition failures.
However, it looks not worth enough to distinguish between normal `NULL`s and join failure `NULL`s.
After we replace `Value::Empty` to `Value::Opt-` based one, then we can also remove
![image](https://user-images.githubusercontent.com/2025065/110453253-51ae5380-8109-11eb-8374-2a1c21b89ec2.png)
`select_with_empty!` test helper macro.
What it means is that each column per row can really have a same type.

2. Refactoring `Value`
Current implementation of `Value` handles nullable types and non-nullable types in a same depth.
This causes internal `Value` methods to quite hard to read and write codes.
Current:
```rust
pub enum Value {           
    Bool(bool),            
    I64(i64),              
    F64(f64),              
    Str(String),           
    OptBool(Option<bool>), 
    OptI64(Option<i64>),   
    OptF64(Option<f64>),   
    OptStr(Option<String>),
    Empty,  // will be removed first                
}                          
```
Planning to do: (enum names might be changed)
```rust

enum Value {
    NotNullable(InnerValue),
    Nullable(Option<InnerValue>),
}

enum InnerValue {
    Bool(bool),
    I64(i64),
    F64(f64),
    Str(String),
}
```

panarch:
@KyGost type alias sounds nice, we might use well.
However,
```rust
type Value = Option<InnerValue>
```
This looks different because `Value`
```rust
enum Value {
    NotNullable(InnerValue),
    Nullable(Option<InnerValue>),
}
```
here `Value` needs to have both `Optional` and Non-nullable. but using `Option<InnerValue>` it can only represent nullable value.

KyGost:
Why do we want `NonNullable`?

panarch:
@KyGost 
Because if we don't use `NotNullable` then every column will become optional.
This is a simple translation from the current 1-step `Value` to 2-step `Value-InnerValue` codes.

First intention I decided to keep non-opt and `Opt-` values in 1-step `Value`, it was because I also worried about the case which it would cause more code, processing.
[...]

PR #169, opened 2021-03-10
Replace Value::Opt- and Empty to Value::Null!

panarch:
Remove `Value::OptBool`, `OptI64`, `OptF64`, `OptStr` and `Empty`.
Now all nullable values are changed to use `Value::Null`.

- **Fix null comparison - now `Value::Null != Value::Null`.**
Due to this change, I need to change `test!` macro to compare `Null` properly.
https://github.com/gluesql/gluesql/compare/add-null-value?expand=1#diff-0332555274be05c9e038aa9524f43b29f55c1b112aea18edcea3a1dbffb4c8cdR94-R165

- **Fix `update.rs` to use `TryFromLiteral` rather than `value.clone_by`.**
It was unavoidable because `Value::Null` cannot provide original type info.

resolve https://github.com/gluesql/gluesql/issues/166

@KyGost yeaaaaah 🚀🚀🚀🚀🚀

### DIFF
--- a/src/data/value/ast_value.rs
+++ b/src/data/value/ast_value.rs
@@ -8,30 +8,23 @@ use {
 impl PartialEq<AstValue> for Value {
     fn eq(&self, other: &AstValue) -> bool {
         match (self, other) {
-            (Value::Bool(l), AstValue::Boolean(r))
-            | (Value::OptBool(Some(l)), AstValue::Boolean(r)) => l == r,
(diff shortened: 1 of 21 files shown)


KyGost:
Oh wow! Are you already mostly done?!

panarch:
@KyGost yup :) Now I'll review the codes again to find something I might miss, then it's done!

KyGost:
@panarch 
Wow!! Awesome!!!

PR #177, opened 2021-03-13
Add EvaluatedRef to clean Evaluated binary operation codes

panarch:
That was quite messy to write binary operation codes in `Evaluated`
image

However, that binary operations only needs borrowed ones, so `EvaluatedRef` only takes borrowed ones to make the codes simple.

@KyGost How do you think of this?


KyGost:
Looks good :-)

Issue #178, opened 2021-03-13
Replacing `sqlparser::ast::Value` to custom enum


panarch:
Currently we use `sqlparser::ast::Value` by renaming to `Literal`.
However, in the codes we need only partial part of `ast::Value`.
* `Boolean`, `Number`, `SingledQuotedString` and `Null`
Above four types are all which are used in GlueSQL for now.

It would be good to consider making our own `enum Literal`.

```rust
enum Literal {
    Boolean(bool),
    Number(String),
    Text(String),
    Null,
}
```
the above might be enough.

KyGost:
Would we then just do a function which does a match each time we work with ast values?

panarch:
@KyGost yup, I also expect to do that.
Then we can simply use `data/literal.rs - enum Literal` to take all of them.

KyGost:
That's a good idea :-)

PR #180, opened 2021-03-15
Add custom enum Literal, it replaces ast::Value

panarch:
Now Literal means crate::data::Literal, no more sqlparser::ast::Value.

`sqlparser::ast::Value as Literal` was superset of literals what GlueSQL needs.
By using `enum Literal`, we can define `Literal` more clearly.

resolve https://github.com/gluesql/gluesql/issues/178

@KyGost it was quite a tough work, yeah! 🚀


KyGost:
Good work!! Looks good!

PR #183, opened 2021-03-19
Fix type validation to check all possible branches from INSERT & UPDATE

panarch:
Issues
* Three different points were doing type validation: `Row::new`, `value::from_expr` and `validate::validate_types`.
* There were several missing points, no null check for validation or missing type check.

I decided to merge into `Value` rather than `validate` because `TryFromLiteral` also does type validation but it was not detachable to add to `validate` module.
Previous `validate_types` wasn't able to work with some cases, e.g. `Update` -> update requires type validation only using `Value` and `DataType` but `validate_types` runs based on `Row`.

I also divided `validate` into `Value::validate_type` and `Value::validate_null`.
If the codes make `Value` using `Value::from_expr` or `Value::try_from_literal`, those also check types so only null validation is required.
The other cases requires validation of both null and types.


panarch:
@KyGost Now I think.. `tests/validate/types.rs` has enough test cases to check all possible branches from `INSERT` & `UPDATE`.

KyGost:
Looks good!

PR #186, opened 2021-03-24
Auto Increment Columns

### LINKED ISSUE #159: AUTO_INCREMENT

KyGost:
Resolve #159
Remake of #171


KyGost:
Uhhh clippy failed for network reasons

panarch:
@KyGost Oh, got it, I just pushed re-run button :)

panarch:
@KyGost Is this ready to be reviewed? For now, I'll just leave a few comments which would be helpful.

src/executor/column_options/auto_increment.rs
```diff
@@ -0,0 +1,96 @@
...
+        let storage = rows.into_iter().fold(storage, |storage, row| {
+            let mut row = row.0;
+            let storage =
+                auto_increment_columns
+                    .iter()
+                    .fold(storage, |storage, (index, column)| {
+                        if matches!(row[*index], Value::Null) {
+                            let name = &column.name.value;
+
+                            match executor::block_on(storage.generate_value(table_name, name)) {
```
panarch, line 51: You cannot use `futures::executor`, it is only supported in limited environments.
e.g. WebAssembly does not support this.

src/storages/sled_storage/auto_increment.rs
```diff
@@ -0,0 +1,66 @@
...
+            .unwrap_or(ONE);
+        let next_value = try_into!(self, value.add(&ONE));
+
+        let (self, _) = transaction!(self, |tree| {
+            let key = f!("generator/{table_name}/{column_name}");
+            let key = key.as_bytes();
+            let next_value = bincode::serialize(&next_value)
+                .map_err(err_into)
+                .map_err(ConflictableTransactionError::Abort)?;
+            tree.insert(key, next_value)?;
```
panarch, line 60: This is a single method call, so you don't need to use sled transaction.

src/data/schema.rs
```diff
@@ -22,13 +31,39 @@ impl ColumnDefExt for ColumnDef {
             .any(|ColumnOptionDef { option, .. }| option == &ColumnOption::Null)
     }
 
+    #[cfg(feature = "auto-increment")]
+    fn is_auto_incremented(&self) -> bool {
```
panarch, line 35: Good to see this in `ColumnDefExt` :)
However I saw there are still a few redundant codes in other files. You might be able to replace them to use this function.

KyGost:
> @KyGost Is this ready to be reviewed? For now, I'll just leave a few comments which would be helpful.

Yup

src/storages/sled_storage/auto_increment.rs
KyGost: Good point

src/executor/column_options/auto_increment.rs
KyGost: What can I use instead?

panarch: You don't need to block, you can simply use `async` :)

KyGost: Nevermind, it is only used in an async function. Not sure why I didn't do that before.

src/storages/sled_storage/auto_increment.rs
```diff
@@ -0,0 +1,81 @@
...
+    ) -> MutResult<Self, Vec<Row>> {
+        // FAIL: No-mut
+        let mut storage = self;
+        let mut rows = rows;
+        for column in columns.iter() {
+            let result = generate_column_values(storage, table_name, column, rows).await?;
+            storage = result.0;
+            rows = result.1;
+        }
+        Ok((storage, rows))
```
panarch, line 37: fyi. here you can use :)
```rust
use futures::stream::{self, TryStreamExt};                             
                                                                       
stream::iter(columns.iter().map(Ok))                                   
    .try_fold((self, rows), |(storage, rows), column| async move {     
        generate_column_values(storage, table_name, column, rows).await
    })                                                                 
    .await                                                             
```

KyGost: Cheater!
I couldn't get that to work hahaha! Thanks.

KyGost: Looks like it works. Thanks.

KyGost: I think I wrote nearly that exact code a few times, just missing the `move` and I think I mightn't have had the `use TryStreamExt`.

panarch:
```rust
#[async_trait(?Send)]
pub trait AutoIncrement
where
    Self: Sized,
{
    async fn generate_values(
        self,
        table_name: &str,
        columns: Vec<(usize, &ColumnDef)>,
        rows: Vec<Row>,
    ) -> MutResult<Self, Vec<Row>>;
}
```

I saw that you changed the interface to this, this seems too much for store trait.
I think.. when we consider the essential of this..
What we need to get is generated values, `Vec<Value>`. `Value` will include auto generated id.
We can also specify how many number of ids is required for the INSERT query.

Then, would it be enough with this?
```rust
async fn generate_values(self, table_name: &str, column_name: &str, size: usize) -> MutResult<Self, Vec<Value>>;
```

In the store trait's view, other params do not matter, essentially `generate_values` only need to know `table_name`, `column_name` and `size`.
Using returned `Vec<Value>` we can blend it into rows we are trying to insert.

One more,
And also, no matter how many ids are requested, we can generate multiple ids by using a single `tree.insert` call.
Rather than repeating `add(ONE)` we can add `size` at once.

src/storages/sled_storage/auto_increment.rs
panarch: Yeah, both are important.
`try_fold` is in `TryStreamExt` so it is necessary to be imported. :)

Because async requires ownership of values, so `move` should be there for ensuring the values inside are moved, not borrowed.

fyi. https://github.com/rust-lang/rust/issues/62290
The reason we use `async move { .. }` block statement inside closure is current stable rust does not support async closure.
I'm really looking forward to see async closure in stable rust.. :)

KyGost:
@panarch good point!
Will move some code back out of storage.

KyGost:
I want it to have storage to have both get and insert again and use transactions but can't get it working.

That *could* be added later thus this is ready for merge.

It would be nice to get it working though.

Giving up for now; here's the gist of what I'm trying:
```rust
#[async_trait(?Send)]
impl AutoIncrement for SledStorage {
    async fn generate_increment_values(
        self,
        table_name: &str,
        column_names: Vec<&str>,
        rows_len: usize,
    ) -> MutResult<Self, Vec<i64>> {
        transaction!(self, |tree| {
            let results: ConflictableTransactionResult<Vec<i64>> = column_names
                .iter()
                .map(|column_name| {
                    let old = tree
                        .get(f!("generator/{table_name}/{column_name}").as_bytes())?
                        .map(|data| bincode::deserialize(&data).ok())
                        .flatten()
                        .unwrap_or(1);
                    let new = old + (rows_len as i64);
                    let new = bincode::serialize(&new).unwrap_or(vec![]); // TODO: Fix

                    let key = f!("generator/{table_name}/{column_name}");
                    let key = key.as_bytes();

                    tree.insert(key, new).map(|_| old)
                })
                .collect;
            results
        })
    }
}
```

src/data/row.rs
```diff
@@ -56,6 +56,11 @@ impl Row {
                     _ => columns.iter().position(|target| target.value == name),
                 };
 
+                #[cfg(feature = "auto-increment")]
+                if column_def.is_auto_incremented() {
+                    return Ok(Value::Null);
+                }
```
panarch, line 62: Though auto increment is applied to column, it should not block adding custom value.

panarch:
We need to determine what the specification of AUTO INCREMENT is.
AUTO INCREMENT can be complicated but due to the its ambiguity.
There can be really a lot different strategy to implement this.

MySQL - `AUTO_INCREMENT`
SQLite - `AUTOINCREMENT`
Their implementation details and behaviors are quite different.
This pr cannot be ready for merge yet because we didn't even decide what to implement.

https://dev.mysql.com/doc/refman/8.0/en/example-auto-increment.html
https://sqlite.org/autoinc.html

panarch:
Oh, btw please don't get too stressed about this.
Rushing too hard can really hurt our patience, you know... our project is basically combination of `Rust` + `async / await` + `functional programming`.
After getting enough relax and then, let's talk about what would be good to do later. :)

KyGost:
@panarch 
My current thinking for method was seed/reseed style.

https://github.com/gluesql/gluesql/issues/159#issuecomment-791124320

We store the latest insert value and then increment based on the number of inserts.
If someone wants to reset it or set it to a different value they can (not implemented in this PR).
I've also not it up so it is easy to add an optional feature which uses a different method. Perhaps one that checks for heightest value.

src/data/row.rs
KyGost: Looks like you're right! I **do** need to take it more slowly.

Not sure how I managed to let this happen.

(This should only be happening if value unset)

KyGost:
Something that might need to be thought about, though, is what happens if the value is already taken?

Should it error?
Should it skip taken values?
Should it try again from the highest value only when it encounters a token value?

panarch:
> We store the latest insert value and then increment based on the number of inserts.
If someone wants to reset it or set it to a different value they can (not implemented in this PR).
I've also set it up so it is easy to add an optional feature which uses a different method. Perhaps one that checks for heighest value.

Sounds so good :)


> Something that might need to be thought about, though, is what happens if the value is already taken?

Good point to think about

* > Should it error?

I think this can be a good starting implementation.
It might be not a good option if we consider to use this in concurrent inserting situation.
However, as an simple embedded db, it can be simple and good enough.

> Should it skip taken values?

I'm not for this idea.
It would be easy to make a kind of worst case scenario.
```sql
CREATE TABLE Test (id NOT NULL AUTO_INCREMENT);

INSERT INTO TEST VALUES (NULL);
INSERT INTO TEST VALUES (100000);

-- then if we try to insert (1)..
INSERT INTO TEST VALUES (1);
```

> Should it try again from the highest value only when it encounters a taken value?

This sounds reasonable to have.

---
Then, how about this?

For this pr, we can simply take the first option - generating error.
It's simple to implement and also useful enough to use.

Later using another pr, we may additionally support the third option - try again by picking the highest value + 1 when it encounters a taken value.

KyGost:
@panarch 
> Then, how about this?
> 
> For this pr, we can simply take the first option - generating error.
> It's simple to implement and also useful enough to use.
> 
> Later using another pr, we may additionally support the third option - try again by picking the highest value + 1 when it encounters a taken value.


I agree. I think this is the right approach.

src/tests/column_options/auto_increment.rs
```diff
@@ -0,0 +1,175 @@
...
+            r#"SELECT * FROM Test"#,
+            Ok(select!(
+            id  | name
+            I64 | Str;
+            1    "test1".to_owned();
+            2    "test2".to_owned();
+            3    "test3".to_owned();
+            4    "test4".to_owned();
+            6    "test5".to_owned();
+            5    "test6".to_owned())),
```
panarch, line 78: This might be `7` instead of `5`?
`6` is manually inserted then auto increment value also should be updated.

src/store/auto_increment.rs
```diff
@@ -0,0 +1,19 @@
...
+    Self: Sized,
+{
+    async fn get_increment_value(&self, table_name: &str, column_name: &str) -> Result<i64>;
+
+    async fn set_increment_value(
+        self,
+        table_name: &str,
+        column_name: &str,
+        end: i64,
+    ) -> MutResult<Self, ()>;
```
panarch, line 18: In your previous commit, I remember you used `generate` and return `Range` type.
I thought the decision to use `Range` was really good.
Is there any special reason to divide `generate` one function into two?

It looks like that for `AutoIncrement` trait, it is enough to have 2 methods.

* `async fn generate_increment_values` which contains `get_increment_value` and `set_increment_values`

* `async fn update_increment_value` or maybe `values`? for manual update on auto increment column.

src/tests/column_options/auto_increment.rs
KyGost: Why should it be? It isn't an *auto* value?

KyGost: Say for example you are only up to 5 and someone inserts 100, should it go to 6, 100 or stay on 5?

src/store/auto_increment.rs
KyGost: I was working on a function which does it all efficienctly with one call but I couldn't get it to work and now my computer has been reset.

The reason I split it back up is for efficiency and niceness, how it is now is notably better.

I'll reply again in a minute with specifics.

src/data/schema.rs
```diff
@@ -22,13 +31,55 @@ impl ColumnDefExt for ColumnDef {
...
+                        ..
+                    }),
+                    ..
+                ] | [
+                    Token::Word(Word {
+                        keyword: Keyword::AUTOINCREMENT,
+                        ..
+                    }),
+                    ..
+                ]
```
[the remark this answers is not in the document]
KyGost: Which one? I think you made a mistake with your comment.

If we were to support one I would think that `AUTO_INCREMENT` would be better.

I added them both for ease of use, they mean the same thing and it would be easier to use for more people.
Saying that, however, I'm indifferent to whether we keep it or not, happy to change it to just one.

src/store/auto_increment.rs
KyGost: Oops! I Didn't end up replying.

The reason I moved away from range is that it is entirely unnessary, we know the end value will be start + length.

That said, not quite sure what my reasoning for moving away from the single function is.
Having two functions like this makes it so that if, in future, we want to be able set the value to a specific value we can.

That said, should we perhaps add a third which combines the two to simplify?

I was trying to merge them into one as well as making it so there's one function call for all columns which we can then put into a transaction but then lost that work.

I'll redo that.

KyGost: The issue, however, with doing it as the same function is that we can't undo the the updates if there's an error.
But oh well, not super important at them moment.

src/data/schema.rs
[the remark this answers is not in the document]
panarch: Neither `AUTOINCREMENT` and `AUTO_INCREMENT` is ANSI SQL, so it's all up to us to decide whatever to use.

I thought it would be better to keep only one, because in anyway when we need to explain this optional auto increment feature, it's simple to say using a single keyword.
e.g. MySQL -> `AUTO_INCREMENT`,  SQLite -> `AUTOINCREMENT`

I slightly prefer to use `AUTOINCREMENT` because I'm not familiar with using `_` in `CREATE TABLE` syntax.
It is usually more common to simply use `PRIMARY KEY`, `NOT NULL` -> spaces rather than `_`.
e.g. also in PostgreSQL syntax, there no `_` use. - https://www.postgresql.org/docs/9.1/sql-createtable.html

However, MySQL is too famous and it uses `_` to `AUTO_INCREMENT` so I'm ok with choosing either `AUTO_INCREMENT` or `AUTOINCREMENT`.

---
ok. let's take only one, either `AUTO_INCREMENT` or `AUTOINCREMENT`.
For what to choose, you choose and I'll follow.

panarch:
> @panarch clippy failed :-(
> I thought you weren't going to update it until it was fixed?
> 
> Regardless, the build errors for this PR can be safely ignored as they are the same as what the main branch should be getting.

@KyGost 
Oh GitHub Action is quite fast on update :)
Ok, I'll quickly fix this clippy issue right now and merge to main, then you can merge main into here. is it ok?

KyGost:
@panarch 
Sounds good :-)

panarch:
@KyGost you can merge now :) https://github.com/gluesql/gluesql/pull/192

src/tests/column_options/auto_increment.rs
panarch: someone insert 100, then the next value need to become 101.

Ok, I'm basically for not distinguishing values which are whether inserted by user or system(auto-increment).
This is also approaches of both MySQL and SQLite.

Current implementation handle values differently, for auto increment value, it ignores user entered value.

src/data/schema.rs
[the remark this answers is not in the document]
KyGost: I'll make it `AUTO_INCREMENT` for now then, we can always change it later :-)

src/tests/column_options/auto_increment.rs
panarch: The functionality for the different pr was this,
We decided to generate error when redundant value comes to `UNIQUE` column.
For example,

```sql
CREATE TABLE Test (id INTEGER UNIQUE AUTO_INCREMENT);

INSERT INTO Test VALUES (NULL);
INSERT INTO Test VALUES (NULL);
INSERT INTO Test VALUES (NULL);
-- 1, 2, 3

INSERT INTO Test VALUES (10);
-- 1, 2, 3, 10

INSERT INTO Test VALUES (NULL);
INSERT INTO Test VALUES (NULL);
-- 1, 2, 3, 10, 11, 12

INSERT INTO Test VALUES (8);
-- 1, 2, 3, 10, 11, 12, 8
INSERT INTO Test VALUES (NULL);
-- 1, 2, 3, 10, 11, 12, 8, 9

INSERT INTO Test VALUES (NULL);
-- Error! (for this pr)

-- using the future pr, we can provide support not to generate error
-- then, 1, 2, 3, 10, 11, 12, 8, 9, 13
```

> Note that this sort of feature would add notable weight to inserts.

For this, not special work is required.

MySQL
```sql
create table Test(id integer primary key auto_increment, title varchar(100));
insert into Test(id, title) values(1, "Hello");
insert into Test(id, title) values(NULL, "Good");
insert into Test(id, title) values(100, "Good");
insert into Test(id, title) values(NULL, "Good");

-- Your code here!

update Test set id = 200 where id = 100;

insert into Test(id, title) values(NULL, "Good");

delete from Test where id = 201;

insert into Test(id, title) values(NULL, "Good");
select * from Test;
```

Then, execution result is this below
| id | title |
| ----- | ----- |
| 1 | Hello |
| 2 | Good |
| 200 | Good |
| 101 | Good |
| 202 | Good |

I think MySQL's strategy is quite straightfoward.
It checks and store the biggest value when `INSERT` and `UPDATE` occurs which is for avoiding redundant number input occurs.
However, it does not care whatever `DELETE` does. because `DELETE` cannot make bigger value, it can only remove values.

How do you think?

src/data/schema.rs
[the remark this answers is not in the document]
panarch: I also like `AUTO_INCREMENT`, at least I like the MySQL's auto increment approach than SQLite's one. MySQL's one is more simple, straightforward and easy to implement, I think.
sounds good :)

src/tests/column_options/auto_increment.rs
KyGost: I, personally, don't like that.

I am familiar with MSSQL's method which is what I've implemented (we just don't have `RESEED` yet) and the way I've implemented it is more efficient than the ways you are mentioning.

These will require checking each inserted value into any AI column.

Under my method we do:
1. Filter AI columns - `n = columns` iters
2. Filter `NULL` rows for count - `n = AI_columns * rows` iters
3. Get from sled - `n = AI_columns` times
4. Add the filter count to the gets  - `n = AI_columns` times
5. Insert into sled - `n = AI_columns` times 
6. Check if rows are `NULL` - `n = AI_columns * rows` iters
7. If `NULL`, set value (and increment) - `n = rows(NULL)` iters

Using that method we also do:
- Sort rows - `n = AI_columns * rows * ?` iters
- Insert into sled - `n = AI_columns` times
or:
- Filter to rows that aren't `NULL` - `n = AI_columns * rows` iters
- Insert into sled - `n = AI_columns * rows(NOT NULL)` times

And if we want to keep highest value:
- Sort rows - `n = AI_columns * rows * ?` iters
- Get from sled - `n = AI_columns` times
- Compare highest and sled - `n = AI_columns` times
- Insert into sled - `n = AI_columns * ?` times
or:
- Filter to rows that aren't `NULL` - `n = AI_columns * rows` iters
- Get from sled - `n = AI_columns * rows(NOT NULL)` times
- Compare with sled - `n = AI_columns * rows(NOT NULL)` times
- Insert into sled - `n = AI_columns * rows(more than sled)` times

Results per insert would be the same, only difference would be consecutive inserts.

If one wanted to enable changing per insert, notably more processing would be required.

KyGost: My opinion is that, for an AI column, you mostly just leave the database to do it unless you want to manually intervene, in which case you can also run `RESEED`.

If for some reason you were regularly inserting values into an AI column, yes, I can see why that form is preferrable, this is why I think it would be better to have this as an optional mechanic.

panarch: Oh! ok btw I think now I can understand... I was wondering what causes some kind of communication miss in here.
I only considered the implementations of MySQL and SQLite, they are two main examples of `AUTO_INCREMENT` and `AUTOINCREMENT`.

I didn't care about behavior of MSSQL at all, and that is what you are familiar with.

Ok, so you are considering to implement subset of MSSQL `IDENTITY` behavior using `AUTO_INCREMENT` keyword?

KyGost: Yup, a bit of a mix of the behaviours of various database but leaning towards MSSQL's `IDENTITY` for the fast method.
I've left it really open to be able to add optional implementations that expand to be like MySQL or SQLite though :-)

panarch: Good, then can you share some brief design specification of this auto incrementation? good place would be in discussion tab.

I think that current most of our miss communication was because we simply had different idea about the AI feature itself.
If it is implementing simply `MySQL`'s one or `SQLite`'s one, then it's just ok to see the document from them.
Or if you use `IDENTITY` for this auto increment feature, then it would be also ok to simply read the document from MSSQL.

> a bit of a mix of the behaviours of various database 

However, it looks like that you are doing something different. Then I cannot figure out what's the feature which is supposed be implemented, it means that I cannot review the codes because I don't know the spec.

KyGost: https://github.com/gluesql/gluesql/discussions/194

panarch:
Closing PRs which are merged and working well in the fork.

Issue #149, opened 2021-02-27
String only output API


KyGost:
In the same vein as #148, it'd be nice to have some API.
Same questions re method of implementation.

It would be good if there was a API side way of getting results as `Vec<Vec<String>>` instead of `Vec<Vec<Value>>`.

panarch:
Sounds good :) Thanks for the suggestion.
This one, we can easily support it by [trait.ToString](https://doc.rust-lang.org/std/string/trait.ToString.html) for `gluesql::data::Value`.

Implementing `ToString` trait and adding some unit tests to `Value` would be enough.

KyGost:
Same thing with JSON output too

panarch:
@KyGost JSON output sounds also good, in this case, it would be good to customize `serde` serialize, https://serde.rs/impl-serialize.html

KyGost:
@panarch gave a useful link for this: https://github.com/gluesql/gluesql-js/blob/main/src/convert.rs

KyGost:
I think that JSON should be represented as:
For:
`SELECT name, fruit_id, rating FROM favourite_fruits`
Get:
```JSON
[
  {"name": "John", "fruit_id": 7, "rating": 9.3},
  {"name": "Bob", "fruit_id": 2, "rating": 1.0}
]
```

KyGost:
TODO:
- [x] `parse_single`
- [x] `select_as_string`
- [ ] `select_as_json`
- [ ] `insert_vec` resolve #148
- [ ] `insert_csv` resolve #153 (alternative to)

PR #187, opened 2021-03-24
Select as string & parse single API

KyGost:


src/data/value/into.rs
```diff
@@ -0,0 +1,91 @@
...
+                .map_err(|_| ValueError::ImpossibleCast)?,
+            Value::Null => return Err(ValueError::ImpossibleCast.into()),
+        })
+    }
+}
+impl TryInto<String> for Value {
+    fn try_into(&self) -> Result<String> {
+        Ok(self.to_owned().into())
+    }
+}
```
panarch, line 91:
```rust
impl Into<String> for Value {
    fn into(self) -> String { .. }
}
```
I think this is enough which you already added well.

`Value` -> `String` does not make any `Error`.
Let's simply use `Into`, without `TryInto<String>`

src/glue.rs
```diff
@@ -32,4 +32,67 @@ impl Glue {
             }
         }
     }
+
+    pub fn select_as_string(&mut self, query: &Query) -> Result<Vec<Vec<String>>> {
```
panarch, line 36: Hmm... it sounds so good to have `select_as_string` function but not sure about the return type.

Is there any special reason to merge labels and values?
image

How about simply keeping both?

```rust
Result<(Vec<String>, Vec<Vec<String>>)>
```

src/glue.rs
```diff
@@ -32,4 +32,67 @@ impl Glue {
...
+
+#[cfg(feature = "sled-storage")]
+#[cfg(test)]
+mod tests {
+    use crate::{parse_sql::parse, Glue, Payload, Row, SledStorage, Value};
+    #[test]
+    fn eq() {
+        let path = "test";
+        std::fs::remove_dir_all(&path);
+        let sled = SledStorage::new(path).unwrap();
```
panarch, line 65: Wow this looks really a proper test in the perfect location, thanks!

```rust
let config = sled::Config::default()
    .path("data/using_config")
    .temporary(true);

let storage = SledStorage::try_from(config).unwrap();
```
Let's simply use `sled`s `temporary` option :)
Then we can delete two `std::fs::remove_dir_all(&path);` in here.

panarch (changes requested):
Wow, thanks! This is brilliant. Not only this is useful for internal codes, but also it has a good role for outer library uses.

src/data/value/into.rs
```diff
@@ -0,0 +1,91 @@
...
+            Value::I64(value) => value.to_string(),
+            Value::F64(value) => value.to_string(),
+            Value::Null => String::from("NULL"),
+        }
+    }
+}
+
+pub trait TryInto<T>: Sized {
+    fn try_into(&self) -> Result<T>;
+}
```
panarch, line 20: Let's simply use `std::convert::TryInto` rather than the custom one.

Then, you might need to change a bit but it will be ok.

image

panarch (changes requested):
Wow, thanks! This is brilliant. Not only this is useful for internal codes, but also it has a good role for outer library uses.

src/glue.rs
KyGost: Ah perfect! Thanks!

src/glue.rs
KyGost: I was thinking of doing that. Unsure.
I suppose this idea is having just one very makes it more like a CSV.

It this case you don't really _need_ to separate the labels and the rows because you know [0] will always be labels.

Thoughts?

src/data/value/into.rs
KyGost: My thinking was that everything else has `try_into()` but string doesn't. I figured it might be a bit confusing if it is missing.

Any ideas on how to solve that? Or is it reasonable to let it be missing?

src/data/value/into.rs
KyGost: Oh oops! Thanks haha

src/glue.rs
KyGost: A large component to the `select_as_string` idea was that it was kinda for display more than usage.

As an example usage, the user does `select_as_string` then displays it as a table.

I can't really think of many other ways that it would be useful.
Can you?

KyGost:
Resolve #149

src/data/value/into.rs
```diff
@@ -0,0 +1,107 @@
...
+        }
+        impl Into<$type> for &Value {
+            fn into(self) -> $type {
+                match self {
+                    $($matcher => $result),+
+                }
+            }
+        }
+    };
+}
```
panarch, line 22: This macro looks unnecessary.
`Into<String> for Value` and `Into<String> for &Value` also have different logics.

```rust
impl Into<String> for &Value {
    fn into(self) -> String {
        match self {
            Value::Str(value) => value.to_string(),
            Value::Bool(value) => (if *value { "TRUE" } else { "FALSE" }).to_string(),
            Value::I64(value) => value.to_string(),
            Value::F64(value) => value.to_string(),
            Value::Null => String::from("NULL"),
        }
    }
}

impl Into<String> for Value {
    fn into(self) -> String {
        match self {
            Value::Str(value) => value,
            Value::Bool(value) => (if value { "TRUE" } else { "FALSE" }).to_string(),
            Value::I64(value) => value.to_string(),
            Value::F64(value) => value.to_string(),
            Value::Null => String::from("NULL"),
        }
    }
}
```

from `&Value` it needs `Value::Str(value) => value.to_string(),`, but from `Value` we can simply pass the ownership `Value::Str(value) => value,`

src/data/value/into.rs
```diff
@@ -0,0 +1,107 @@
...
+
+macro_rules! impl_try_into {
+    ($type:ident, $($matcher:pat => $result:expr),+) => {
+        impl TryInto<$type> for Value {
+            type Error = Error;
+            fn try_into(self) -> Result<$type> {
+                (&self).try_into()
+            }
+        }
```
panarch, line 31: Though this might be convenient for making users not to think about their codes is using whether `Value` or `&Value`, I also concern that this also can make users confused by providing two different ways to do a same task.

In the case of `impl Into<String> for Value` and `impl into<String> for &Value`, if user can use `Value` then it is always better to use `impl Into<String> for Value`.

However, these `TryInto` cases are different, both are identical.
If `impl TryInto<$type> for Value` can do anything better than `impl Try<$type> for &Value`, then it can be worth enough to have both. However, it is not in this case.

Let's have only `impl TryInto<$type> for &Value`, and then `impl_try_into`  macro is also unnecessary.
How do you think?

src/glue.rs
panarch: > It this case you don't really need to separate the labels and the rows because you know [0] will always be labels.

This sounds reasonable.
`select_as_string` is for providing convenience to users and in this assumption, it looks good to have current return type. Let's keep the current :) thanks.

panarch (changes requested):

src/data/value/into.rs
KyGost: I've just added the owned values for now to make API easier. I would consider additional matching a performance boost but not necessary. Is this okay?


The macros make things cleaner. Are you saying the `Into` one is unneccesary because it is only used once? If so, there shouldn't be a performance drop from it, it makes it took cleaner and more understandable next to the `IntoIter` ones and enables future expansions.

src/data/value/into.rs
KyGost: See https://github.com/gluesql/gluesql/pull/187#discussion_r601050195

src/glue.rs
KyGost: Awesome. Sounds good.

src/data/value/into.rs
panarch: Oh, sorry I missed the comment, I think it's reasonable not to have TryInto.
I think we can simply consider this kind of work in systemic way first.

By adding `TryInto<String> for Value` can make code might look slightly better but it causes the system confusing because `Value -> String` should not cause error but `TryInto<String> for Value` implies that failure is possible.

> My thinking was that everything else has try_into() but string doesn't.

Simple to think about this, because everything else can fail but string cannot.

src/data/value/into.rs
panarch: Ok, let's divide this into two different topic.
1. Using macro is good in here
2. Having both `impl TryInto<$type> for Value` and `impl TryInto<$type> for &Value`.

---
1. Using macro --
Macro can make thinks cleaner but I'm not sure that works in this case.

1.1. additional link to understand
`impl_try_into!` macro requires user to read what is `impl_try_into!`.
However, if we simply add `impl TryInto<...> --` then, though it may require a few more code lines, we don't need to read anything else, `TryInto` trait itself is just everything we need to know.

Macro can make codes simpler by adding kind of symbolic link to codes.

1.2. `cargo fmt` cannot work properly the code inside macro
image

Even I changed the code like this above, `cargo fmt` does not do any correction.

panarch: 2. Having both `impl TryInto<$type> for Value` and `impl TryInto<$type> for &Value`.

I think there can be a conflict between the point of systemic view and user-convenient view.

For users, I agree that it can be useful to have both, because as users, they are not supposed to understand the internal design.

However, systemically `impl TryInto<$type> for Value` is just a slight wrapper.
With simply using referencing `&` like `&value`, it can be avoided.
At least for the internal code, I prefer not to provide multiple identical ways to do same work.
If there exists both `impl TryInto<$type> for Value` and `impl TryInto<$type> for &Value`, then `&Value` version obviously implies that kind of `clone` will be used, but we may expect something different for `Value` version (passing ownership or..).
Though two look similar, but they should have different logics.

---
How about this?
Gathering codes which are not supposed to be used in internal implementation but can be useful for library users.

For example, (not sure about whether `addon` is the best name for this)
1. make a module named `addon` to root directory, `src`.
2. add `TryInto<$type> for Value` codes to `addon` module.

Then, we can provide extra convenience to external library users, at the same time we can also prevent these to be used in the internal codes.
It is even possible by systemic way using feature.
If we add a feature named `addon` and export it as default-features but not for internal `cargo test`, then we can ensure that is not used internally.

src/data/value/into.rs
KyGost: Yeah. I suppose it's not like you will write code that sometimes gets a String, sometimes gets an i64. This doesn't prevent loops from being able to happen.

Though it may have some macro implications. I see your point and it is fair enough.

src/data/value/into.rs
KyGost: **1.**
---
I see your point with the macros, they just cloud the code if they don't add anything, if it is just TryInto then yeah the macros aren't needed.

---

**2.**
---
I see what you mean with the usage/system point. That makes sense.
In this case though, wouldn't it be fair to say that `(value) try_into()` is just a function that does `&(value) try_into()` for us?
It isn't a big function but it still has functionality.

Just as each time I wanted to run `try_into()` I could write:
```rust
match (value) {
    Value::Str(value) => value.to_string(),
    Value::Bool(value) => (if *value { "TRUE" } else { "FALSE" }).to_string(),
    Value::I64(value) => value.to_string(),
    Value::F64(value) => value.to_string(),
    Value::Null => String::from("NULL")
}
```
Instead we shorten it into `(value).try_into()`.

Every time we want `(owned value).try_into()` we could write `(&(owned value)).try_into()` instead this shortens it to `(owned value).try_into()`.

If one wanted to make the point however that this function isn't worth it, when instead we could just write one extra character each time, that might be reasonable.

---

**3.**
---

**3.1.**
I suppose that the `glue.rs` and `parse_sql.rs` function as just for library uses as it stands. Perhaps we could move them there and add anything else we want to give users?

**3.2.**
Refer to above.

If it doesn't make sense to add the `(owned value).try_into()` for internal use it probably doesn't make sense to add it for library users.

KyGost: Ah! I think, as far as ownership goes, I see (part of) where I've gone wrong.

```(owned value).try_into() -> (&(owned value)).try_into()```
and
```(&(owned value)).try_into()```
behave differently; the first one will give ownership to the function.

Nonetheless, the point still stands, if one doesn't want to give away ownership they can use the second option.

A further note to above, a seperate, perhaps meaningless, point is that by having it so that `(owned value).try_into() -> (&(owned value)).try_into()` exists, you can set up other code using it. Later if we want a bit of additional performance we can do it properly and match it as owned rather than sending a reference to the reference `try_into`; this will make it so we don't have to go back through all points where this could be used, the code will already be using the improved function.

Comment (panarch) [reply src/data/value/into.rs:22]: > 3.1.
I suppose that the glue.rs and parse_sql.rs function as just for library uses as it stands. Perhaps we could move them there and add anything else we want to give users?

This really, really makes sense, rather than making a custom new module, it can be just enough to use `glue` module.
Ok, how about this, let's do like this.
In the case of `parse_sql.rs`, this might be also useful for custom database maker to simply import and use it.
So, from now when we consider to add some other functionality which is for library user-convenience, let's use `glue` module.
Then, we can also feel free to add more features to `glue` without worrying about internal code integration.

> 3.2.
Refer to above.

If it doesn't make sense to add the (owned value).try_into() for internal use it probably doesn't make sense to add it for library users.

>  Later if we want a bit of additional performance we can do it properly and match it as owned rather than sending a reference to the reference try_into; this will make it so we don't have to go back through all points where this could be used, the code will already be using the improved function.

This sounds reasonable.
When we need to have this below in the future,
```rust
impl TryInto<$type> for Value {         
    type Error = Error;                 
    fn try_into(self) -> Result<$type> {
        (&self).try_into()              
    }                                   
}                                       
```
Then let's use this rather than `(&(owned value)).try_into()`.
However, currently there does not exists any use case of `TryInto<$type> for Value`, only `for &Value` exists.

---
So, combining those two.. how about this?

* Make a `glue` as a directory by changing `glue.rs` to `glue/mod.rs`.
* Then, add `TryInto<$type> for Value` to... `glue/value.rs` or somewhere else in `glue/` module.

Then, library user can use `TryInto<$type> for Value` if they want, and we also don't need to include unused trait impl in execution layer codes.
When it comes that we need to use `TryInto<$type> for Value`, then we can simply move this to `data/value/into.rs`.

KyGost: Sounds good!

panarch (approved):
Thanks!! merging 👍 👍 👍

PR #188, opened 2021-03-24
Remove `cargo bench` from GitHub Action

panarch:
Previous item size was unnecessary too big and bench library itself also warned about it.

Now reduce the size to 5k and someone who wants to increase can easily do it by changing the `ITEM_SIZE` constant.

@KyGost Thanks :)

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -35,5 +35,3 @@ jobs:
       run: cargo test --features sled-storage --verbose
     - name: Run tests with all features
       run: cargo test --all-features --verbose
-    - name: Run benchmarks
-      run: cargo bench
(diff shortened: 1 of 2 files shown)


panarch:
Ok, I removed `cargo bench` from GitHub Action.
Reducing sample size makes CI slightly faster but most of time was because of setup, not running bench.

https://bheisler.github.io/criterion.rs/book/faq.html

> How Should I Run Criterion.rs Benchmarks In A CI Pipeline?
You probably shouldn't (or, if you do, don't rely on the results). The virtualization used by Cloud-CI providers like Travis-CI and Github Actions introduces a great deal of noise into the benchmarking process, and Criterion.rs' statistical analysis can only do so much to mitigate that. This can result in the appearance of large changes in the measured performance even if the actual performance of the code is not changing. A better alternative is to use Iai instead. Iai runs benchmarks inside Cachegrind to directly count the instructions and memory accesses. Iai's measurements won't be thrown off by the virtual machine slowing down or pausing for a time, so it should be more reliable in virtualized environments.

Issue #148, opened 2021-02-27
Non-string API


KyGost:
It would be nice to add a non-string api, that is to say a way to (mostly just) `INSERT` data and provide a `vec` rather than having to concatenate or work around GlueSQL.

Would this be best as a seperate repo, an optional component just a regular component?

panarch:
This sounds interesting to have, but not sure what would be a good way to add this feature.
If that is for supporting full sql functionality, then we can consider this as an.. ORM and it's better to use a seperate repo.
However, current main needs looks limited to inserting data.
Then optional component approach might work well.

KyGost:
I think API should work like:
`insert_vec(table_name: String, columns: Vec<String>, rows: Vec<Vec<Value>>)`

Maybe even also:
`insert_string_vec(table_name: String, columns: Vec<String>, rows: Vec<Vec<String>>)`

PR #193, opened 2021-04-01
API: Insert Vec

KyGost:
Resolve: #148

### DIFF
--- a/src/glue/mod.rs
+++ b/src/glue/mod.rs
@@ -65,6 +65,17 @@ impl Glue {
         rows: Vec<Vec<Value>>,
     ) -> Result<Payload> {
         // TODO: Make this more efficient and nicer by checking the way we execute
+        let table_name = ObjectName(vec![Ident {
+            value: table_name,
+            quote_style: None,
+        }]);
+        let columns = columns
+            .into_iter()
+            .map(|name| Ident {
+                value: name,
+                quote_style: None,
+            })
+            .collect();
         let sqlparser_rows: Vec<Vec<Expr>> = rows
             .into_iter()
             .map(|row| {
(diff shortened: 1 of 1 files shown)


src/glue/mod.rs
```diff
@@ -1,7 +1,10 @@
 #![cfg(feature = "sled-storage")]
 use {
-    crate::{execute, storages::SledStorage, ExecuteError, Payload, Query, Result},
+    crate::{execute, storages::SledStorage, ExecuteError, Payload, Query, Result, Value},
     futures::executor::block_on,
+    sqlparser::ast::{
+        Expr, Ident, ObjectName, Query as AstQuery, SetExpr, Statement, Value as Literal, Values,
```
panarch, line 6: Ok, lets use `Value as Literal` to `Value as AstValue` like `Query as AstValue` because we now have `data::Literal`.

src/glue/mod.rs
```diff
@@ -49,10 +52,66 @@ impl Glue {
...
+                            Value::Null => Literal::Null,
+                            Value::Bool(value) => Literal::Boolean(value),
+                            Value::I64(value) => Literal::Number(value.to_string(), false),
+                            Value::F64(value) => Literal::Number(value.to_string(), false),
+                            Value::Str(value) => Literal::SingleQuotedString(value),
+                        })
+                    })
+                    .collect()
+            })
+            .collect();
```
panarch, line 94: Is this an intended behavior that you decide to get `Vec<Vec<Value>>` rather than `Vec<Vec<sqlparser::ast::Value>>`?
If it is because of specific use cases, then I'll accept.

src/glue/mod.rs
KyGost: Oopsie, good point.
Thanks.

src/glue/mod.rs
KyGost: Yup, my thinking is that we don't want GlueSQL users to have to play around with `sqlparser`, we want them to interface with `Value`.

panarch: Ok :)

panarch (approved):
Thanks! 👍

Issue #196, opened 2021-04-01
Function: UUID()


KyGost:
`UUID()` is a function which generates a random, unique ID.
[T-SQL Spec for NEWID()](https://docs.microsoft.com/en-us/sql/t-sql/functions/newid-transact-sql)
[MySQL UUID](https://dev.mysql.com/doc/refman/5.7/en/miscellaneous-functions.html#function_uuid)
[MySQL UUID Short](https://dev.mysql.com/doc/refman/5.7/en/miscellaneous-functions.html#function_uuid-short)

Functions as `UUID_SHORT()` from MySQL.

Future PR?: `BYTES` data type and 128 byte UUID function

Note that, compared to performance of AUTO_INCREMENT, this will be *very* slow.

panarch:
However, is `NEWID()` designed only for `uniqueidentifier`?

More general one to have might be `UUID()`, then we can simply use https://docs.rs/uuid/0.8.2/uuid/

KyGost:
@panarch
Ah! Hmm, not sure whether it would be better to use sled's `generate_id()` or that; thoughts?

KyGost:
And if we did that should the function be called `UUID()`?

I mean we *could* add both

panarch:
Oh this was a different issue, let's forget about `UUID`.
Why are you thinking about replacing `generate_id()`...?

panarch:
`generate_id` is not public api, it should not be exposed as a store trait.

KyGost:
Huh??
Store traits are public?

So what, I'd just do `storage.tree.generate_id()` instead?

KyGost:
> Oh this was a different issue, let's forget about `UUID`.

They'd serve the same purpose, very unique values.

panarch:
For now, I cannot get the point about your intention.
I saw the pr codes and I cannot understand what is purpose of `NEWID`.
Why does we need `NEWID` which is only for `uniqueidentifier` data type in sql server?

KyGost:
It is a random ID, kinda like `AUTO_INCREMENT` but is meant to be cross database.
Functions like UUID.

Enables stuff like creating two parallel tables and unioning them while keeping unique IDs.

panarch:
Then, we can just use `UUID` rather than `NEWID`?
More importantly, implementing uuid functionality and applying to store trait are independent issues. These should be in seperated prs.

KyGost:
Sounds good.

panarch:
ref #364

PR #198, opened 2021-04-01
Function: UUID() and RAND()

KyGost:
Resolve: #196


src/tests/function/uuid.rs
```diff
@@ -0,0 +1,45 @@
...
+        ),
+        ("INSERT INTO test_default (id) VALUES (NULL)", Ok(Payload::Insert(1))),
+        (
+            "SELECT id FROM test_default", // Unreliable result
+            Ok(select!(
+                id
+                Value::I64;
+                1
+            )),
+        ),*/
```
panarch, line 39: It is not proper to add this kind of comments to codes.
We cannot track what we should do using comments.
Rather than using comments, please register issues by generalizing it.
Then not only for us, someone can work on this issue.

panarch (changes requested):
https://en.wikipedia.org/wiki/Universally_unique_identifier

UUID is not just a random number.
You can use uuid v4 for this case.
https://docs.rs/uuid/0.8.2/uuid/

KyGost:
Already logged and already working on it in a PR (#199), hence the comments.

src/tests/function/uuid.rs
KyGost: Already logged and already working on it in a PR (#199), hence the comments.

KyGost:
@panarch
> https://en.wikipedia.org/wiki/Universally_unique_identifier
> 
> UUID is not just a random number.
> You can use uuid v4 for this case.
> https://docs.rs/uuid/0.8.2/uuid/

Are you referring to 128 byte ones?
Else how do they differ?

KyGost:
Should I perhaps rename this to `UUID_SHORT`? (See linked issue)

src/tests/function/uuid.rs
panarch: Ok good, then we don't have to include this commens in this pr.

KyGost: Why not?
They are tests that I want to have but cannot yet have as there is bugs in the way?

panarch: > They are tests that I want to have but cannot yet have as there is bugs in the way?

However, you said they are already included in existing issues. Then we don't need to have using comments which are hard to track changes.

KyGost: I'll change them to state which issue they relate to.

panarch:
image

What I mean is, it is more common to use text representation for `UUID`.
You can really simply use uuid 4 from the crate.

KyGost:
We would need a new data type if we want to do that reasonably.
It would be unreasonable to waste space and enable a lot of functions which we don't want to allow on UUID; it isn't text, it is a u128, which is often formatted, for readability, as groups of hexadecimal text.

panarch:
UUID data type and UUID function is just a different independent issue.
In this pr, what we are talking about is `UUID` function, so let's focus on the issue only.

KyGost:
Exactly, which is why I have opted to make an implmentation of `UUID_SHORT` for now which we *can* reasonably represent with an existing datatype, specifically `i64` (short one bit for the signing, hence the wrap around of allowing negatives as values) or `INTEGER`.

panarch:
Was this issue for implementing UUID?
And I don't think that is a good idea to use name `UUID_SHORT` which generates `i64`.

https://tools.ietf.org/html/rfc4122
> A UUID is 128 bits long, and can guarantee uniqueness across space and time.

KyGost:
> Was this issue for implementing UUID?
> And I don't think that is a good idea to use name `UUID_SHORT` which generates `i64`.
> 
> https://tools.ietf.org/html/rfc4122
> 
> > A UUID is 128 bits long, and can guarantee uniqueness across space and time.

As mentioned in #196, MySQL implements `UUID_SHORT` which is 64 bits long.

As we have agreed this PR isn't for adding another data type thus it will not include a full length (128 bit) UUID.

I am opposed to implementing UUID as `TEXT`.

Thus, shall I rename to `UUID_SHORT` or shall we temporarily accept a short form of `UUID` until we implement a `UUID`, `Bytes`, `Binary` or other data type?

panarch:
Even MySQL `UUID` generates text format.
image

> I am opposed to implementing UUID as TEXT.

That's the point.
I'm definitely for `UUID()` to generate text.
And I'm opposed to use `UUID_SHORT` like non standard names.
For generating random numbers, I think it is enough to implement and use `RAND - generate floating point number between 0 - 1` function which is also quite common in many other languages.

KyGost:
Storing it as text seems very silly to me. If it is the standard perhaps GlueSQL should do it though.

> And I'm opposed to use `UUID_SHORT` like non standard names.

Why?
As in you intend to stick strictly to ANSI?

src/tests/function/rand.rs
```diff
@@ -0,0 +1,53 @@
...
+        (
+            "SELECT id FROM test WHERE RAND() IS NOT NULL", // Unreliable result
+            Ok(select!(
+                id
+                Value::I64;
+                1
+            )),
+        ),
+        (
+            "SELECT id FROM test WHERE RAND() > 0 AND RAND() < 1", // Unreliable result
```
panarch, line 19: `RAND()` can be zero so it needs to be `RAND() >= 0`

panarch (changes requested):
Thanks, codes look really good.
Once you remove all comments then I'll approve and merge right after.

Except for user documentation purpose, I usually hate to leave comments to codes.
Both compiler and human checks the codes whether it is ok or not.
However, comments are just extra information.
Programmer needs to manage comments not to be outdated, but it is quite tough.
Codes just run and pass tests, it does not care whether comments is correct or not.

A lot of mistake happens on this comments.
Usual case is, someone update existing codes but forgets to update comments to fit to the code changes.
Then, comments contain wrong information and it becomes just a mess.

In here, the comments which you wrote are already outdated.
* inserting `RAND()` float value to `INTEGER` type.
* `id INT` with `DEFAULT UUID()`
These are not to blame, I just wanted to say how dangerous comments can be.

I hope that this can be helpful enough for you.
It is ok to do more talk in here.
Or if you are ok with this opinion, then you can remove all comments including `// Unreliable result`.

panarch:
> Storing it as text seems very silly to me. If it is the standard perhaps GlueSQL should do it though.
> 
> > And I'm opposed to use `UUID_SHORT` like non standard names.
> 
> Why?
> As in you intend to stick strictly to ANSI?

No, because `UUID_SHORT` is not `UUID`, it is just a different type of `f64` random number.

KyGost:
> Except for user documentation purpose, I usually hate to leave comments to codes.
> Both compiler and human checks the codes whether it is ok or not.
> However, comments are just extra information.
> Programmer needs to manage comments not to be outdated, but it is quite tough.
> Codes just run and pass tests, it does not care whether comments is correct or not.
> 
> A lot of mistake happens on this comments.
> Usual case is, someone update existing codes but forgets to update comments to fit to the code changes.
> Then, comments contain wrong information and it becomes just a mess.
> 
> In here, the comments which you wrote are already outdated.
> 
> * inserting `RAND()` float value to `INTEGER` type.
> * `id INT` with `DEFAULT UUID()`
>   These are not to blame, I just wanted to say how dangerous comments can be.
> 
> I hope that this can be helpful enough for you.
> It is ok to do more talk in here.
> Or if you are ok with this opinion, then you can remove all comments including `// Unreliable result`.

The issue is, the alternative to leaving comments is just to remove stuff. This way when the issues are fixed, how do we know what to re incorportate?

panarch:
> The issue is, the alternative to leaving comments is just to remove stuff. This way when the issues are fixed, how do we know what to re incorportate?

If you care about this, you can register the issue on this.

KyGost:
How would make such an issue, referring to an unmerged PR, reference it?
Surely issued would get messy if they included these sorts of things?

panarch:
```rust
 /*
( Cannot test, see #197
  "INSERT INTO test (id) VALUES (RAND()), (RAND()), (RAND()), (RAND()), (RAND())",
  Ok(Payload::Insert(5)), // Should error if UUID isn't working
),
(
    "INSERT INTO test (id) VALUES (RAND())",
    Ok(Payload::Insert(1)), // Should error if UUID isn't working
),*/

/*( Cannot test, see #197
    "CREATE TABLE test_default (id INT UNIQUE NOT NULL DEFAULT RAND())",
    Ok(Payload::Create),
),
("INSERT INTO test_default (id) VALUES (NULL)", Ok(Payload::Insert(1))),
(
    "SELECT 1 FROM test_default WHERE id > 0 AND id < 1", // Unreliable result
    Ok(select!(
        id
        Value::I64;
        1
    )),
),*/
```
btw, I cannot even understand what is these comments for.
Why does the comment say `UUID` which is `RAND` test case?
`id` is `INT` and `RAND()` generates float, it is not a valid query.

`"CREATE TABLE test_default (id INT UNIQUE NOT NULL DEFAULT RAND())",` ,this also cannot happen. `INT` for `DEFAULT RAND()`.

If it is clear enough then you will be able to register the issue.
However, for now I cannot say how to make an issue because I don't know what you want to do after.

KyGost:
That code is essentially pseudo-code, as you said with comments, you don't get them compiled so nothing tells you to fix them.

They're just showing tests that I want in very crude ways.

What I'm saying is that these issues relate wholly to this PR, they would not make sense as seperate issues in the gluesql github, as there is nothing that can be done for them as the repository stands.

Having said that however, once this PR is merged there will be actions that could be taken.
Hence my asking whether it would be reasonable to add such issues to this GitHub repository.

PR #223, opened 2021-05-13
New AST for execution layer

panarch:
GlueSQL was using AST from sqlparser-rs which parses raw sql text to parsed AST.
However, feature requirements for parser and executor are quite different.

To support index, transaction, query planning, etc..., now it is required for us to have own execution layer AST.

New module named `translate` converts `sqlparser::ast` to `gluesql::ast`.


KyGost:
Good point, completely untying from SQLParser removes a number of repeated hassles, adds more flexibility and enables effective possiblity of modular parsers.

panarch:
@KyGost That's right. I was unnecessarily stuck at keeping parser AST.
What you've done in [MultiSQL](https://github.com/SyRis-Consulting/multisql) using `Recipe` gave me good inspiration and motivation. Thanks :)

PR #252, opened 2021-06-30
Add support for like operaror

### LINKED ISSUE #251: LIKE operator support
It looks nice to has an `LIKE` operator support, which most other sql database supports
It is quite useful to lookup rows by simple pattern

For simpler implementation maybe support just like postgres in one :
[PostgreSQL: Documentation: 8.3: Pattern Matching](https://www.postgresql.org/docs/8.3/functions-matching.html)

Maybe good to use [regex](https://github.com/rust-lang/regex) for starter

euiko:
Support for like operator such in postgres :
[PostgreSQL: Documentation: 8.3: Pattern Matching](www.postgresql.org/docs/8.3/functions-matching.html)

Implemented using [regex](https://github.com/rust-lang/regex) library
Take a look on filter test for example
Should close #251
Feel free to review it :smile:

### DIFF
--- a/src/data/literal.rs
+++ b/src/data/literal.rs
@@ -7,16 +7,13 @@ use {
     std::{borrow::Cow, cmp::Ordering, convert::TryFrom, fmt::Debug},
     thiserror::Error,
     Literal::*,
-    regex::{Regex},
+    super::StringExt,
 };
 
 #[derive(Error, Serialize, Debug, PartialEq)]
 pub enum LiteralError {
     #[error("unsupported literal binary arithmetic between {0} and {1}")]
     UnsupportedBinaryArithmetic(String, String),
-
-    #[error("failed to parse pattern: {0}")]
-    FailedToParsePattern(String),
 
     #[error("literal unary operation on non-numeric")]
     UnaryOperationOnNonNumeric,
@@ -273,17 +270,7 @@ impl<'a> Literal<'a> {
 
     pub fn like(&self, other: &Literal<'a>) -> Result<Self> {
         match (self, other) {
-            (Text(l), Text(r)) => Ok(Boolean(Regex::new(
-                &format!(
-                    "^{}$",
-                    regex::escape(r)
-                        .replace("%", ".*")
-                        .replace("_", ".")
-                )
-            )
-            .map_err(|e| LiteralError::FailedToParsePattern(e.to_string()))?
-            .is_match(l)
-        )),
+            (Text(l), Text(r)) => Ok(Boolean(l.as_ref().like(&r)?)),
             _ => Ok(Boolean(false)),
         }
     }
--- a/src/data/literal.rs
+++ b/src/data/literal.rs
@@ -7,12 +7,16 @@ use {
     std::{borrow::Cow, cmp::Ordering, convert::TryFrom, fmt::Debug},
     thiserror::Error,
     Literal::*,
+    regex::{Regex},
 };
 
 #[derive(Error, Serialize, Debug, PartialEq)]
 pub enum LiteralError {
     #[error("unsupported literal binary arithmetic between {0} and {1}")]
     UnsupportedBinaryArithmetic(String, String),
+
+    #[error("failed to parse pattern: {0}")]
+    FailedToParsePattern(String),
 
     #[error("literal unary operation on non-numeric")]
     UnaryOperationOnNonNumeric,
@@ -264,6 +268,23 @@ impl<'a> Literal<'a> {
                 format!("{:?}", other),
             )
             .into()),
+        }
+    }
+
+    pub fn like(&self, other: &Literal<'a>) -> Result<Self> {
+        match (self, other) {
+            (Text(l), Text(r)) => Ok(Boolean(Regex::new(
+                &format!(
+                    "^{}$",
+                    regex::escape(r)
+                        .replace("%", ".*")
+                        .replace("_", ".")
+                )
+            )
+            .map_err(|e| LiteralError::FailedToParsePattern(e.to_string()))?
+            .is_match(l)
+        )),
+            _ => Ok(Boolean(false)),
         }
     }
 }
--- a/src/data/literal.rs
+++ b/src/data/literal.rs
@@ -1,4 +1,5 @@
 use {
+    super::StringExt,
     crate::{
         ast::AstLiteral,
         result::{Error, Result},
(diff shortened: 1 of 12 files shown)


panarch:
Thanks a lot! Behavior looks perfect, you can simply attach a new commit after running `cargo fmt`, then GitHub Action will pass.
I'll leave other feedbacks until today.

panarch (changes requested):
Thanks for the contribution! A few ones I'd like to request.

1. Combining redundant `regex` checking codes.
Let's combine redundant code in `data/value/mod.rs` and `data/literal.rs`.
You might be not sure where to put shared `like` function, and that's right.
It looks that we need to make a new module for this.
However, neither `literal` or `value` does not look a proper place to have.
Let's add a new module.
In this case, we need to expand `String`, so `trait` approach would be good.

`src/data/string_ext.rs`
```rust
pub trait StringExt {                             
    fn like(&self, pattern: &str) -> Result<bool>;
}

impl StringExt for String {
    ..
}                                               
```
This kind of stuff can work.
Then, we can import `StringExt` in `data/literal.rs` and `data/value/mod.rs` and use `like` function.

In `src/data/string_ext.rs`,
If you can make `FailedToParsePattern`, then you need to add an integration test for this error. Or.. if this is kind of unreachable error, then you can use `Unreachable-` prefix for this error.
Feel free to add a new error type for `StringExt`, `StringExtError` would be good enough.

2. Add tests to reach all branches which are newly added.
Current tests does not hit any of `fn like` in `data/literal.rs`.
```rust
// in data/literal.rs
pub fn like(&self, other: &Literal<'_>) -> Result<Self> { 
    panic!();
}
```
Even if you change the content to `panic!()`, all tests still pass.
You can simply reach this `like` by adding 
```sql
SELECT name FROM Boss WHERE "ABC" LIKE "A_C";
```

src/data/value/mod.rs
```diff
@@ -283,6 +284,25 @@ impl Value {
...
+                    "^{}$",
+                    regex::escape(b)
+                        .replace("%", ".*")
+                        .replace("_", ".")
+                )
+            )
+            .map_err(|e| ValueError::FailedToParsePattern(e.to_string()))?
+            .is_match(a)
+        )),
+            _ => Ok(Bool(false)),
```
panarch, line 303: `_ => Ok(Bool(false)),` branch needs to be tested.
```sql
SELECT name FROM Boss WHERE name LIKE 10;
```
will work

euiko: I realize that for other data types currently are can give mislead behaviour, when any other non text use of `LIKE` operator are return a `false` result.
Maybe instead of gives false result, it is more appropriate gives error such like `unsupported operator with data type ...`
Again, this as observation in postgres behaviour on case such like this, which gives errors :
```
ERROR: operator does not exist: text ~~ integer
```

euiko:
Thanks for the review

Has been updated for :
1. Yeah firstly i'm quite not sure where to place the new like function.
Has been moved out the redundant codes as traing `StringExt` in `string_ext` module as your suggest.
And also add new kind of error related with this module as `StringExtErr`.
2. Has added some other test cases to handle cases to reach uncovered case for literal to literal and non text operation.
But I realize for non text operation, maybe behaviour can be changed to become more reasonable, please take a look at my comment on your [code's review](https://github.com/gluesql/gluesql/pull/252#discussion_r662407116)

Please take a look for this again

src/data/value/mod.rs
panarch: That really makes sense, I missed it. I'm also for postgres approach.
Also for the consistency, because current `Value` and `Literal` returns error for unsupported binary or unary operations, it would be also better to return error than returning `false` result.
Thanks for mentioning this!

Ok, then how about adding something like `ValueError::LikeOnNonString(Value, Value)` and `LiteralError::LikeOnNonString(String, String)`, and also adding `tests/filter.rs` have these two error test cases?

src/data/literal.rs
```diff
@@ -273,17 +270,7 @@ impl<'a> Literal<'a> {
...
-                    "^{}$",
-                    regex::escape(r)
-                        .replace("%", ".*")
-                        .replace("_", ".")
-                )
-            )
-            .map_err(|e| LiteralError::FailedToParsePattern(e.to_string()))?
-            .is_match(l)
-        )),
+            (Text(l), Text(r)) => Ok(Boolean(l.as_ref().like(&r)?)),
```
panarch, line 273: Tiny comment, you can simply use `l.like(b).map(Boolean)` thanks to smart Rust compiler :)

src/data/value/mod.rs
```diff
@@ -289,17 +289,7 @@ impl Value {
...
-                    "^{}$",
-                    regex::escape(b)
-                        .replace("%", ".*")
-                        .replace("_", ".")
-                )
-            )
-            .map_err(|e| ValueError::FailedToParsePattern(e.to_string()))?
-            .is_match(a)
-        )),
+            (Str(a), Str(b)) => Ok(Bool(a.like(&b)?)),
```
panarch, line 292: Another tiny one, you can use `a.like(b).map(Bool)`

src/result.rs
```diff
@@ -92,6 +94,7 @@ impl PartialEq for Error {
             (Value(e), Value(e2)) => e == e2,
             (Literal(e), Literal(e2)) => e == e2,
             (Interval(e), Interval(e2)) => e == e2,
+            (StringExt(e), StringExt(e2)) => e == e2,
```
panarch, line 97: I.. felt this even from the first commit which in your pr, I appreciate your sense and skill to figure out structure and patterns in the code. It really looks that you are already familiar enough with the code base, awesome :)

panarch (changes requested):
As you suggested, changing `false` case to error sounds really good.

1. `false` to error
2. `cargo fmt`
Then, let's merge to the main :) Thanks!

src/data/literal.rs
euiko: still amazed how amazing rust compiler is :D
anyway thanks for the knowledge

src/result.rs
euiko: Thanks :), these well organized codes that help a lot to understand and make familiar with

euiko:
Updated with last feedback @panarch, please take a look.
Sorry for late update quite busy on weekend

panarch (approved):
You can feel comfortable to do it at your own pace :)
Everything looks really good! Thanks a lot 👍 👍 👍 
Approve and merging!

Issue #259, opened 2021-08-02
README.md code example is outdated


panarch:
Current README.md provides..
```rust
...

for query in parse(sqls).unwrap() {
    glue.execute(&query).unwrap();
}
```
This is outdated, not working now in v0.7.

`Glue` module now accepts a raw sql string and execute.
It is enough to use this below,
```rust
glue.execute(sqls).unwrap();
```

devgony:
Let me edit README.md.
I will create Pull-request as soon as finish testing execution of multiple queries!

Issue #258, opened 2021-08-02
Fix Glue::execute to run multiple sql query string


panarch:
In `src/glue.rs`, `Glue` module is a kind of wrapper interface which provides users to execute sql queries in a simple way.

It is firstly designed to run multiple sql query string.
e.g. In `https://docs.rs/gluesql/0.7.0/gluesql/`,
```rust
   let sqls = "
         DROP TABLE IF EXISTS Glue;
         CREATE TABLE Glue (id INTEGER);
         INSERT INTO Glue VALUES (100);
         INSERT INTO Glue VALUES (200);
         SELECT * FROM Glue WHERE id > 100;
         DROP TABLE Glue;
     ";

     glue.execute(&sqls).unwrap();
}
```

However, current `Glue` module only execute the first query, (`DROP TABLE IF ..` is only executed)
https://github.com/gluesql/gluesql/blob/b03a0587c22fac5880737026656f0f71e322b500/src/glue.rs#L18-L23

Let's fix this to run multiple sql queries from a single string input.

earney:
Can we close this issue?  It seems to have been resolved in issue #415

ever0de:
> Can we close this issue? It seems to have been resolved in issue #415

This issue has not been resolved.
Please refer to <https://github.com/gluesql/gluesql/issues/415#issuecomment-983711473>.

ever0de:
Killed by <https://github.com/gluesql/gluesql/pull/562>

@ming535  thank you for your contribution

PR #260, opened 2021-08-03
renew Usage at README.md and Doc: executing multiple queries

devgony:
## 1. Renew Usage to run multiple queries

Current `execute()` method can execute the first parsed query only.  
I changed both src/lib.rs and README.md with for loop like below.

```rs
use gluesql::*;
fn main() {
    let storage = SledStorage::new("data/doc-db").unwrap();
    let sqls = vec![
        "DROP TABLE IF EXISTS Glue;",
        "CREATE TABLE Glue (id INTEGER);",
        "INSERT INTO Glue VALUES (100);",
        "INSERT INTO Glue VALUES (200);",
        "SELECT * FROM Glue WHERE id > 100;",
    ];
    for sql in sqls {
        let output = glue.execute(sql).unwrap();
        println!("{:?}", output)
    }
}
``` 

## What to do next?

Improve `execute` to be compatible with mutiple queries.

## Outdated PR

~~## Tried to fix issue #258 to execute multiple queries at glue.rs: Glue.execute~~

~~A return type is changed from `Result<Payload>` to `(Vec<Result<Payload>>, Option<Result<Payload>>)` which means we can store executed results and return the results and error(or None).~~

```rs
// src/glue.rs
pub fn execute(&mut self, sql: &str) -> (Vec<Result<Payload>>, Option<Result<Payload>>) {
        let mut results: Vec<Result<Payload>> = Vec::new();
        let parsed = parse(sql);
        for a_sql in parsed.unwrap().iter() {
            let statement = translate(a_sql).unwrap();
            let storage = self.storage.take().unwrap();
            match block_on(execute(storage, &statement)) {
                Ok((storage, payload)) => {
                    self.storage = Some(storage);
                    results.push(Ok(payload));
                }
                Err((storage, error)) => {
                    self.storage = Some(storage);
                    return (results, Some(Err(error)));
                }
            }
        }
        (results, None)
    }
```

~~## According to above fixed Glue.execute method, I renewed README.md to fix issue #259~~

```rs
// README.md
use gluesql::{Glue, SledStorage};

fn main() {
    let storage = SledStorage::new("data/doc-db").unwrap();
    let sqls = "
        DROP TABLE IF EXISTS Glue;
        CREATE TABLE Glue (id INTEGER);
        INSERT INTO Glue VALUES (100);
        INSERT INTO Glue VALUES (200);
        SELECT * FROM Glue WHERE id > 100;
    ";
    let (results, error) = glue.execute(&sqls);
    println!("Results: {:?}\nError: {:?}", results, error);
    // Results: [Ok(DropTable), Ok(Create), Ok(Insert(1)), Ok(Insert(1)), Ok(Select { labels: ["id"], rows: [Row([I64(200)])] })]
    // Error: None
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # GlueSQL
+
 [![crates.io](https://img.shields.io/crates/v/gluesql.svg)](https://crates.io/crates/gluesql)
 [![docs.rs](https://docs.rs/gluesql/badge.svg)](https://docs.rs/gluesql)
 [![LICENSE](https://img.shields.io/crates/l/gluesql.svg)](https://github.com/gluesql/gluesql/blob/main/LICENSE)
@@ -6,20 +7,25 @@
 [![Chat](https://img.shields.io/discord/780298017940176946)](https://discord.gg/C6TDEgzDzY)
 
 ## SQL Database Engine as a Library
+
 GlueSQL is a SQL database library written in Rust. It provides a parser ([sqlparser-rs](https://github.com/ballista-compute/sqlparser-rs)), execution layer, and optional storage ([sled](https://github.com/spacejam/sled)) packaged into a single library.
-Developers can choose to use GlueSQL to build their own SQL database, or as an embedded SQL database using the default storage engine.  
+Developers can choose to use GlueSQL to build their own SQL database, or as an embedded SQL database using the default storage engine.
 
 ## Standalone Mode
+
 You can use GlueSQL as an embedded SQL database. GlueSQL provides [sled](https://github.com/spacejam/sled "sled") as a default storage engine.
 
 ### Installation
+
 In your `Cargo.toml`:
+
 ```toml
 [dependencies]
 gluesql = "0.7"
 ```
 
 ### Usage
+
 ```rust
 use gluesql::{parse, Glue, SledStorage};
 
@@ -34,7 +40,7 @@ fn main() {
         SELECT * FROM Glue WHERE id > 100;
         DROP TABLE Glue;
     ";
-    
+
     for query in parse(sqls).unwrap() {
         glue.execute(&query).unwrap();
     }
@@ -42,22 +48,29 @@ fn main() {
 ```
 
 ## SQL Library Mode (For Custom Storage)
+
 ### Installation
+
 `sled-storage` is optional. So in `Cargo.toml`:
+
 ```toml
 [dependencies.gluesql]
 version = "0.7"
 default-features = false
 features = ["sorter", "alter-table", "index"]
 ```
+
 #### Three features above are optional, too.
-* `sorter` - ORDER BY support for non-indexed expressions.
-* `alter-table` - ALTER TABLE query support
-* `index` - CREATE INDEX & DROP INDEX, index support
+
+- `sorter` - ORDER BY support for non-indexed expressions.
+- `alter-table` - ALTER TABLE query support
+- `index` - CREATE INDEX & DROP INDEX, index support
 
 ### Usage
+
 There are two required 2 traits for using GlueSQL: `Store` and `StoreMut`.
 In `src/store/mod.rs`,
+
 ```rust
 pub trait Store<T: Debug> {
     async fn fetch_schema(..) -> ..;
@@ -75,6 +88,7 @@ pub trait StoreMut<T: Debug> where Self: Sized {
 
 There is also optional store traits
 In `src/store/alter_table.rs` & `src/store/index.rs`
+
 ```rust
 pub trait AlterTable where Self: Sized {
     async fn rename_schema(..) -> ..;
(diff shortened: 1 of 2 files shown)


src/glue.rs
```diff
@@ -15,24 +15,24 @@ impl Glue {
...
-            Ok((storage, payload)) => {
-                self.storage = Some(storage);
-
-                Ok(payload)
-            }
-            Err((storage, error)) => {
-                self.storage = Some(storage);
-
-                Err(error)
+    pub fn execute(&mut self, sql: &str) -> (Vec<Result<Payload>>, Option<Result<Payload>>) {
```
panarch, line 18: I agree that the new return types can provide more info to users.
However, for this purpose, the return value needs to be a bit different.
Current `Vec<Result<Payload>>` cannot have `Err`, so it is simply enough to use `Vec<Payload>`.
The second param, `Option<Result<Payload>>` also only can be `Ok`, so `Option<Error>` is just enough.
Therefore, to represent what you wanted to say would be this below,
```rust
// return value types
pub fn execute(..) -> (Vec<Payload>, Option<Error>)
```

However, let's be more careful about the interface change.
`execute` in `Glue` is just a simple wrapper for users to run queries easily.
We usually expect function return type as a form of `Result<OkValue, ErrorValue>`.
Then, result can be simply evaluated by try operator `?`.
In this case, try operator cannot be applied to tuple form so it would be a bit inconvenient.

How about keeping the previous return type `Result<Payload>` and let `execute` to return only the last payload?
If someone needs to get all in-between `Payload` values, then they can simply run each query separately.

src/glue.rs
```diff
@@ -15,24 +15,24 @@ impl Glue {
...
+        let mut results: Vec<Result<Payload>> = Vec::new();
+        let parsed = parse(sql);
+        for a_sql in parsed.unwrap().iter() {
+            let statement = translate(a_sql).unwrap();
+            let storage = self.storage.take().unwrap();
+            match block_on(execute(storage, &statement)) {
+                Ok((storage, payload)) => {
+                    self.storage = Some(storage);
+                    results.push(Ok(payload));
+                }
```
panarch, line 28: You can declare `let mut storage = self.storage.take().unwrap();` outside of `for` loop and re-use it.
Then, you don't need to repeat `self.storage = Some(storage)` -> `let storage = self.storage.take().unwrap();`

panarch (changes requested):
Awesome, thanks for the contribution. really a good start 👍 👍 👍 

I mentioned a few comments, main one is the return type of `execute` function.
Thanks!

src/glue.rs
euiko: How about using fail-fast system, so when multiple queries executed when there is error preceding another queries, it will return it immediately, otherwise the last query result will be returned.
I think it is more common sense that way, such as in db client like DBeaver you will experience that kind behaviour when you executed multiple query at once

panarch: @euiko
In the prev comment, I think I was too focused on the side story so the main message was not clear.
> How about using fail-fast system, so when multiple queries executed when there is error preceding another queries, it will return it immediately, otherwise the last query result will be returned.

Thanks! I definitely agree with this.

devgony: I **agree** to keep previous return type `Result<Payload>` and let `execute` to return only the last payload.👍
I just thought it would be nice if there was a function for execution history log just in case for people who want to trace.

src/glue.rs
devgony: I'm struggling with this but have two issues.
1. `exeucte` does not take `storage` as a reference which means we can not `borrow` (`move` fails cuz of iteration).
2. It seems like `self.storage` is mutated at each iteration. I wonder it is okay not to use `self.storage.take().unwrap()` at each iteration as well.

I hope you to give me some advice !

panarch: In this case, you don't need to concern about borrowing.
`execute` takes the ownership of `storage`, not mutable borrowing.
This is why the result of `execute` includes `Storage` instance rather than simply returning `Payload`.
It is for making execution layer codes to be pure functional without using `mut`.

`self.storage = Some(storage)` assignment is required only at the end of execution.

for example,
```rust
let parsed = parse(sql).unwrap();
let mut storage = self.storage.take().unwrap();
let mut payload: Option<Payload> = None;

for statement in parsed.iter() {
  let statement = translate(statement).unwrap();
  match block_on(execute(storage, &statement)) {
    Ok((s, p)) => {
      storage = s;
      payload = Some(p);
    }
    Err((s, e)) => {
      self.storage = Some(storage);
      return Err(e);
    }
  }
}

self.storage = Some(storage);
return Ok(payload.unwrap());
```
Something this kind of codes are which you can think about to use.

fyi.
This is just for extra info, it's ok to use `mut` at this time.
If you want something more fun stuff, then you can try iterator `try_fold` to solve this issue.
Then you don't even need to use `mut`.
`parsed.iter().try_fold(storage, ...)`
In the execution layer codes, `mut` is not allowed so `try_fold` is being quite commonly used.
https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.try_fold

README.md
```diff
@@ -1,63 +1,87 @@
...
+### Output
+
+```
+Ok(DropTable)
+Ok(Create)
+Ok(Insert(1))
+Ok(Insert(1))
+Ok(Select { labels: ["id"], rows: [Row([I64(200)])] })
+```
+
```
panarch, line 60: I think we don't need to include extra info in this.
Code example in `README.md` is not tested by `cargo`, so it's easy to be outdated.
It is better to be in `examples/` or code comments, then `cargo test` verifies that all code comments and `examples/` working.

Let's include minimal info as possible in here.
Users who need more details, they can try `examples/`.
e.g. `cargo run --example hello_world`
`examples/` directory is kind of standard which is supported by `cargo`.

README.md
```diff
@@ -1,63 +1,87 @@
...
         INSERT INTO Glue VALUES (200);
         SELECT * FROM Glue WHERE id > 100;
-        DROP TABLE Glue;
     ";
-    
     for query in parse(sqls).unwrap() {
-        glue.execute(&query).unwrap();
+        let statement = translate(&query).unwrap();
+        let output = glue.execute_stmt(statement);
+        println!("{:?}", output);
```
panarch, line 46: You can still use `glue.execute` though it can only run a single query for a sql string which includes multiple queries.
Code example here is for providing a simple and straightforward way to run the project almost without thinking.
Someone who wants more detail, they can figure it out `docs`, `tests` or `examples`.

In here, if you convert `sqls` into `Vec<_>`, you can keep using `execute`. like this
```rust
let sqls = vec![
    "DROP TABLE IF EXISTS Glue;",
    "CREATE TABLE Glue (id INTEGER);",
    ..
];

for sql in sqls {
    glue.execute(sql).unwrap();
}
```

src/lib.rs
```diff
@@ -16,19 +16,21 @@
 //!
 //! #[cfg(feature = "sled-storage")]
 //! fn main() {
-//!     let storage = SledStorage::new("data/doc-db").unwrap();
+//!     let storage = SledStorage::new("data.db").unwrap();
```
panarch, line 19: It would be better to be `data/doc-db`.
Currently, `/data` is in `.gitignore` but not `data.db`.
It can be bit inconvenient because `data.db` will appear per each `cargo test` and we need to ignore it.

panarch (changes requested):
Looks so good, now the codes pass GitHub Action successfully!
Let's just add a few modifications and merge it into the main, thanks 👍

panarch (approved):
Awesome 👍 👍 👍 
Thanks and congrats for the first contribution!

PR #270, opened 2021-08-07
Chang unreachable code that was being treated as "unwrap" to return unreachable error

ever0de:
- @panarch reviewed part `src/executor/aggregate/mod.rs` once
- Chang unreachable code that was being treated as "unwrap" to return unreachable error
- Clear TODO


panarch (approved):
Looks good! Thanks 👍 👍 👍

PR #271, opened 2021-08-07
Replace filter_map and next to find_map

ever0de:
- If you want the next value after the filter, I think can handle it with find_map


panarch:
Really a good catch, thanks a lot 👍 👍 👍

PR #274, opened 2021-08-08
Remove unnecessary IntoIterator call syntax

ever0de:
- check this: [std/keyword.for](https://doc.rust-lang.org/stable/std/keyword.for.html)
```
for-in-loops, or to be more precise, iterator loops, are a simple syntactic sugar over a common practice within Rust,
which is to loop over anything that implements IntoIterator until the iterator returned by .into_iter() returns None (or the loop body uses break).
```
- Calling IntoIterator to default, so I think I can remove it.


panarch (approved):
Good clean up, thanks a lot 👍 👍 👍

PR #289, opened 2021-08-16
Support SIN, COS, TAN functions.

### LINKED ISSUE #268: Support SIN, COS, TAN function
support to SIN, COS, TAN functions.

---
### SIN Function : Returns the sine of the given value

```SQL
SELECT SIN(0) AS result
```
|result|
|---|
|0|

```SQL
SELECT SIN(1) AS result
```
|result|
|---|
|0.8414709848078965|

---
### COS Function : Returns the cosine of the given value

```SQL
SELECT COS(0) AS result
```
|result|
|---|
|1|

```SQL
SELECT COS(1) AS result
```
|result|
|---|
|0.5403023058681398|

---
### TAN Function : Returns the tangent of the given value

```SQL
SELECT TAN(0) AS result
```
|result|
|---|
|0|

```SQL
SELECT TAN(1) AS result
```
|result|
|---|
|1.557407724654902|

maruoovv:
Support SIN, COS, TAN functions.
This functions support numeric string literals and boolean and floating numbers.
resolve #268


panarch (changes requested):
Looks good! I also really appreciate to add `name` function to `ast::Function`.
Let's change a bit and merge it to the main 👍 👍 👍

src/ast/function.rs
```diff
@@ -9,6 +9,38 @@ pub enum Function {
...
     Right { expr: Expr, size: Expr },
+    Sin(Expr),
+    Cos(Expr),
+    Tan(Expr),
+}
+
+impl Function {
+    pub fn name(&self) -> &str {
+        match self {
+            Function::Lower(_) => "LOWER",
```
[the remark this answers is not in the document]
ever0de: Adding `Display` is very good
It will be easy to use: <https://docs.rs/strum_macros/0.21.1/strum_macros/derive.Display.html>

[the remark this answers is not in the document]
panarch: Wow i didnt know that there is strum macro! Awesome, thanks for the comment :):)

src/executor/evaluate/mod.rs
```diff
@@ -320,6 +320,22 @@ async fn evaluate_function<'a, T: 'static + Debug>(
             Nullable::Null => Ok(Value::Null),
         }
         .map(Evaluated::from),
+        Function::Sin(expr) | Function::Cos(expr) | Function::Tan(expr) => {
+            let float_number = eval_to_float(func.name(), expr).await?;
```
maruoovv, line 324: I tried to adding display to Function, but i'm failed due to a lifetime problem.
The return value of to_string() is String, but the eval_to_float functions parameter is &static(now i'm changed to 'a) str, so temporary value is freed in the statement.
I tried a lot but failed, how can i solve this problem?

panarch: What you've done is actually more meaningful :)
Then, you don't even need to pass `func.name()` at all.

1. Applying `strum` crate
You can use https://crates.io/crates/strum_macros
```rust
use {
    ..
    strum::Display,                                                          
};                                                                           
                                                                             
pub enum Function {                                                          
    #[strum(to_string = "LOWER")]                                            
    Lower(Expr),                                                             
                                                                             
    #[strum(to_string = "UPPER")]                                            
    Upper(Expr),                                                             
                                                                             
    #[strum(to_string = "LEFT")]                                             
    Left { expr: Expr, size: Expr },                                         
                                                                             
    #[strum(to_string = "RIGHT")]                                            
    Right { expr: Expr, size: Expr },                                        
                                                                             
    #[strum(to_string = "CEIL")]                                             
    Ceil(Expr),                                                              
                                                                             
    #[strum(to_string = "ROUND")]                                            
    Round(Expr),                 

    ..
}                                      
```

Then, `func.to_string()` will work.

2. Better `eval_to_float`
Thanks to `Display`, now `eval_to_float` does not require function name.

```rust
let eval_to_float = |expr| async move {                                                          
    match eval(expr).await?.try_into()? {                                                        
        Value::I64(v) => Ok(Nullable::Value(v as f64)),                                          
        Value::F64(v) => Ok(Nullable::Value(v)),                                                 
        Value::Null => Ok(Nullable::Null),                                                       
        _ => Err::<_, Error>(EvaluateError::FunctionRequiresFloatValue(func.to_string()).into()),
    }                                                                                            
};                                                                                               
```

src/executor/evaluate/mod.rs
```diff
@@ -320,6 +320,22 @@ async fn evaluate_function<'a, T: 'static + Debug>(
...
+
+            let trigonometric = |func, value| match func {
+                Function::Sin(_) => f64::sin(value),
+                Function::Cos(_) => f64::cos(value),
+                Function::Tan(_) => f64::tan(value),
+                _ => panic!("Unexpected function: {:?}", func.name()),
+            };
```
panarch, line 331:
```rust
let trigonometric = |func, value| match func {
    Function::Sin(_) => f64::sin(value),
    Function::Cos(_) => f64::cos(value),
    _ => f64::tan(value),
};
```
the last branch `_ => panic!(...)` is not reachable.
For now, let's simplify by replacing `Function::Tan(_)` to `_`.

src/translate/function.rs
```diff
@@ -81,6 +81,30 @@ pub fn translate_function(sql_function: &SqlFunction) -> Result<Expr> {
...
         "ROUND" => func_with_one_arg!(Function::Round),
         "FLOOR" => func_with_one_arg!(Function::Floor),
+        "SIN" => {
+            check_len(name, args.len(), 1)?;
+
+            translate_expr(args[0])
+                .map(Function::Sin)
+                .map(Box::new)
+                .map(Expr::Function)
+        }
```
panarch, line 91: Now you can use `func_with_one_arg!(Function::Sin)`

panarch (changes requested):
Adding `Display` trait to `ast/function` suggestion is really cool, that can makes `evaluate_function` code a lot better.
Thanks a lot!

Let's apply a few changes and `strum` stuff and merge it into main!

src/executor/evaluate/mod.rs
maruoovv: Thanks for your helps! It was a simple solution by removing unnecessary parameters.
I also removed the parameter of other eval functions in the same way.

panarch (approved):
Beautiful :)
I'd also appreciate that you apply `Display` to not just for `eval_to_float` you used but also you've done for all related closures `eval_to_str` and `eval_to_integer`.

I've found that `PRIMARY KEY` is not supported yet but it is used in test cases. ```"CREATE TABLE SingleItem (id INTEGER PRIMARY KEY)",```
And.. this was what I was supposed to check but I've also recognized that I already accepted and merged other PRs which used `PRIMARY KEY`.
I'll fix it by myself and
* https://github.com/gluesql/gluesql/issues/308
later, this issue will resolve it clearly.
No further action required for you to do, all looks great.
Thanks! 👍 👍 👍 merging!!

PR #290, opened 2021-08-16
Add support for gcd and lcm operator

### LINKED ISSUE #279: Support `gcd`, `lcm`
### Descriptions
- `gcd ( numeric_type, numeric_type ) → numeric_type`
  - Greatest common divisor (the largest positive number that divides both inputs with no remainder); returns 0 if both inputs are zero; available for integer, bigint, and numeric.
- `lcm ( numeric_type, numeric_type ) → numeric_type`
  - Least common multiple (the smallest strictly positive number that is an integral multiple of both inputs); returns 0 if either input is zero; available for integer, bigint, and numeric.

#### GCD
``` sql
SELECT gcd(1071, 462) AS result;
```

|result|
|-|
|21|

#### LCM
``` sql
SELECT lcm(1071, 462) AS result;
```

|result|
|-|
|23562|

genieCS:
Add support for gcd and lcm operator.

resolve #279 .

### DIFF
--- a/src/executor/evaluate/mod.rs
+++ b/src/executor/evaluate/mod.rs
@@ -237,6 +237,14 @@ async fn evaluate_function<'a, T: 'static + Debug>(
         }
     };
 
+    let eval_to_f64 = |name: &'static str, expr| async move {
+        match eval(expr).await?.try_into()? {
+            Value::F64(number) => Ok(number),
+            Value::I64(number) => Ok(number as f64),
+            _ => Err::<_, Error>(EvaluateError::FunctionRequiresF64Value(name.to_owned()).into()),
+        }
+    };
+
     match func {
         Function::Lower(expr) => match eval_to_str("LOWER", expr).await? {
             Nullable::Value(v) => Ok(Value::Str(v.to_lowercase())),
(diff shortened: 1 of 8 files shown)


src/executor/evaluate/mod.rs
```diff
@@ -228,6 +228,14 @@ async fn evaluate_function<'a, T: 'static + Debug>(
         }
     };
 
+    let eval_to_f64 = |name: &'static str, expr| async move {
+        match eval(expr).await?.try_into()? {
+            Value::F64(number) => Ok(number),
+            Value::I64(number) => Ok(number as f64),
+            _ => Err::<_, Error>(EvaluateError::FunctionRequiresF64Value(name.to_owned()).into()),
```
panarch, line 235: 1. Is this `GCD` only works for integer values...?
We may not even need to accept `f64` in here.

2. Let's get `Value::Null`
It is sad but SQL language supports `NULL` value so...
Usual behavior is `NULL` comes into functions params, then result also becomes `NULL`.

src/tests/function/gcd.rs
```diff
@@ -0,0 +1,91 @@
...
+            Ok(Payload::Insert(1)),
+        ),
+        (
+            r#"INSERT INTO GcdF64 VALUES (2.0, 4.0);"#,
+            Ok(Payload::Insert(1)),
+        ),
+        (
+            r#"INSERT INTO GcdF64 VALUES (6.0, 8.0);"#,
+            Ok(Payload::Insert(1)),
+        ),
```
panarch, line 26: You can simply insert at once :)
```rust
    r#"INSERT INTO GcdF64 VALUES (3.0, 0.0), (2, 4), (6, 8);"#,
```

src/tests/function/gcd.rs
```diff
@@ -0,0 +1,91 @@
...
+            Ok(select!("test"; F64; 3.0)),
+        ),
+        (
+            r#"SELECT GCD(left, right) AS test FROM GcdF64 WHERE left = 2.0"#,
+            Ok(select!("test"; F64; 2.0)),
+        ),
+        (
+            r#"SELECT GCD(left, right) AS test FROM GcdF64 WHERE left = 6.0"#,
+            Ok(select!("test"; F64; 2.0)),
+        ),
```
panarch, line 42: Let's merge these four individual tests into a single one. `select!` supports multi columns

panarch (changes requested):
Really a good start 👍 👍 👍 
I left some comments, and... let's do `GCD` and `LCM` together in the same PR.

panarch (approved):
Everything looks good :) Thanks!! 👍 👍 👍 👍

Issue #264, opened 2021-08-07
Add Floor, Ceil, Round Function


tmdgusya:
I am going to add the functions below.

Ceil Func : Nearest integer greater than or equal to argument 
    - example ) 11.2 -> 12
    - example ) -11.3 -> 11

Floor Func : Nearest integer less than or equal to argument
   - example ) 11.3 -> 11
   - example ) -11.3 -> -12
   
Rounds Func : v to s decimal places. Ties are broken by rounding away from zero.
   - example ) 11.6 -> 12
   - exmaple ) 11.2 -> 11

PR #291, opened 2021-08-16
Add Floor, Ceil, Round Function

tmdgusya:
I'm done to add the functions below.

Ceil Func : Nearest integer greater than or equal to argument
- example ) 11.2 -> 12
- example ) -11.3 -> 11

Floor Func : Nearest integer less than or equal to argument

example ) 11.3 -> 11
example ) -11.3 -> -12
Rounds Func : v to s decimal places. Ties are broken by rounding away from zero.

example ) 11.6 -> 12
exmaple ) 11.2 -> 11

### DIFF
--- a/src/executor/evaluate/mod.rs
+++ b/src/executor/evaluate/mod.rs
@@ -230,6 +230,15 @@ async fn evaluate_function<'a, T: 'static + Debug>(
         }
     };
 
+    let eval_to_float = |name: &'static str, expr| async move {
+        match eval(expr).await?.try_into()? {
+            Value::I64(v) => Ok(Nullable::Value(v as f64)),
+            Value::F64(v) => Ok(Nullable::Value(v)),
+            Value::Null => Ok(Nullable::Null),
+            _ => Err::<_, Error>(EvaluateError::FunctionRequiresFloatValue(name.to_owned()).into()),
+        }
+    };
+
     match func {
         Function::Lower(expr) => match eval_to_str("LOWER", expr).await? {
             Nullable::Value(v) => Ok(Value::Str(v.to_lowercase())),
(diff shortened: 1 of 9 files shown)


src/tests/function/floor.rs
```diff
@@ -0,0 +1,85 @@
...
+            )),
+        ),
+        (
+            "SELECT FLOOR(10) AS floor FROM SingleItem",
+            Ok(select!(
+                "floor";
+                F64;
+                f64::from(10).floor()
+            )),
+        ),
```
panarch, line 38: These test cases are identical from the view of our project.
It would be good to consider merging into a one. `SELECT FLOOR(0.3), FLOOR(-0.8), FLOOR(10) FROM SingleItem`

ref https://github.com/gluesql/gluesql/pull/289#discussion_r690050617

src/executor/evaluate/mod.rs
```diff
@@ -281,5 +283,53 @@ async fn evaluate_function<'a, T: 'static + Debug>(
...
+                },
+                Value::I64(number) => f64::from_i64(number),
+                _ => None,
+            };
+
+            match number {
+                Some(number) => Ok(Evaluated::from(Value::F64(number.round()))),
+                None => Err(EvaluateError::FunctionRequiresFloatValue("ROUND".to_owned()).into()),
+            }
+        }
```
panarch, line 317: Good start! I think you also recognize there are some redundant codes which could be wrapped by a function.
Let's reduce redundancy in `Ceil`, `Round` and `Floor`.
Not just less redundancy helps programmers to write less codes, but also it is better because there's also less stuffs to be tested.

src/executor/evaluate/mod.rs
```diff
@@ -281,5 +283,53 @@ async fn evaluate_function<'a, T: 'static + Debug>(
...
+        Function::Floor(expr) => {
+            let number = match eval(expr).await?.try_into()? {
+                Value::F64(number) => Some(number),
+                Value::Str(s) => match f64::from_str(&s) {
+                    Ok(f) => Some(f),
+                    Err(_) => None,
+                },
```
panarch, line 324: Let's handle `Str` input as an error `FunctionRequiresFloatValue`.
We can accept, I64 to F64 because it is also natively supported by Rust lang itself, but Str is not.
Users who need to run `FLOOR` using `Str` value, then they can use `CAST` function.

It is also better not to convert `Err` into non `Err` types.
ref https://github.com/gluesql/gluesql/pull/289#discussion_r690053201

panarch (changes requested):
Awesome, clean and minimal 👍 👍 👍 
Let's change little bit and merge it into the main!

src/translate/function.rs
```diff
@@ -81,6 +81,30 @@ pub fn translate_function(sql_function: &SqlFunction) -> Result<Expr> {
...
+                .map(Expr::Function)
+        }
+        "FLOOR" => {
+            check_len(name, args.len(), 1)?;
+
+            translate_expr(args[0])
+                .map(Function::Floor)
+                .map(Box::new)
+                .map(Expr::Function)
+        }
```
ever0de, line 107: I made a macro called func_with_one_arg.
You can add it comfortably.
Rebase and use the master.
Below is an example of usage.

```rust
func_with_one_arg!(Function::Ceil)
```

ever0de: Can be used when it's a function and only has one argument

tmdgusya: Oh thank you, i'll use that macro that can remove redundant code

panarch (approved):
Awesome, `eval_to_float` looks really, really good :)
Thanks a lot! 🥇

panarch:
resolve https://github.com/gluesql/gluesql/issues/264

PR #292, opened 2021-08-16
Implement TRIM function

### LINKED ISSUE #267: Implement TRIM function
- Implement TRIM function
- This function is multibyte safe
- Implementing this feature allows to perform the following actions

1.  Remove whitespaces from prefixes and suffixes
    - If only str parameter
    - [rust/std/str/trim](https://doc.rust-lang.org/stable/std/primitive.str.html#method.trim)
    - In the example above, consideration of `\t` should also be considered
2. If the specifier is `BOTH`, Remove the specifier parameter to the prefix and suffix
    - @panarch Please check the contents below
    - `SELECT TRIM(BOTH 'xyz' FROM 'xyzybarxzxxz');`
    -  The result values for this query are different and which method do you want to adopt?
    1. PostgreSQL
        - `bar` 
    2. MySQL
        - `ybarxzxxz`
3. If the specifier is `LEADING`, Remove the specifier parameter to the prefix
    - `SELECT TRIM(LEADING 'xyz' FROM 'xyzybarxzxxz');`
    1. PostgreSQL
        - `barxzxxz`
    2. MySQL
        - `ybarxzxxz`
4. If the specifier is `TRAILING`, Remove the specifier parameter to the suffix
    - `SELECT TRIM(TRAILING 'xyz' FROM 'xyzybarxzxyz');`
    1. PostgreSQL
        - `xyzybar`
    2. MySQL
        - `xyzybarxz`

## A point of caution
- If none of the specifiers BOTH, LEADING, or TRAILING is given, `BOTH` is assumed
- Which adopt, PostgreSQL or MySQL

### PostgreSQL
```sql
trim(both 'xyz' from 'yxTomxx') → Tom
```

- This is a non-standard syntax for `trim()`
    - @panarch Don't have to support?

```sql
trim(both from 'yxTomxx', 'xyz') → Tom
```

### MySQL
```sql
SELECT TRIM('  bar   ')  -> bar
SELECT TRIM(LEADING 'x' FROM 'xxxbarxxx') -> barxxx
SELECT TRIM(BOTH 'x' FROM 'xxxbarxxx') -> bar
SELECT TRIM(TRAILING 'xyz' FROM 'barxxyz') -> barx
```

## Reference

- [PostgreSQL](https://www.postgresql.org/docs/13/functions-string.html)

image

- [MySQL](https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_trim)

image

ever0de:
resolved #267 

### ISSUE Reopen
- Other features are not implemented, so the issue must be reopened in the near future.

### Implemented
- TRIM function
- macro that can be used when function has one argument

### Check List
- [x] `cargo check --all`
- [x] `cargo fmt`
- [x] `cargo clippy --release --all --all-targets`

### DIFF
--- a/src/executor/evaluate/mod.rs
+++ b/src/executor/evaluate/mod.rs
@@ -281,5 +281,10 @@ async fn evaluate_function<'a, T: 'static + Debug>(
 
             Ok(Evaluated::from(Value::Str(converted)))
         }
+        Function::Trim(expr) => match eval_to_str("TRIM", expr).await? {
+            Nullable::Value(string) => Ok(Value::Str(string.trim().to_owned())),
+            Nullable::Null => Ok(Value::Null),
+        }
+        .map(Evaluated::from),
     }
 }
--- a/src/ast/function.rs
+++ b/src/ast/function.rs
@@ -9,6 +9,7 @@ pub enum Function {
     Upper(Expr),
     Left { expr: Expr, size: Expr },
     Right { expr: Expr, size: Expr },
+    Trim(Expr),
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
--- a/src/tests/function/mod.rs
+++ b/src/tests/function/mod.rs
@@ -1,3 +1,4 @@
 pub mod cast;
 pub mod left_right;
+pub mod trim;
 pub mod upper_lower;
--- a/src/tests/function/trim.rs
+++ b/src/tests/function/trim.rs
@@ -39,6 +39,15 @@ test_case!(trim, async move {
             "SELECT TRIM(a => 2) FROM Item;",
             Err(TranslateError::NamedFunctionArgNotSupported.into()),
         ),
+        (
+            "CREATE TABLE NullName (name TEXT NULL)",
+            Ok(Payload::Create),
+        ),
+        ("INSERT INTO NullName VALUES (NULL)", Ok(Payload::Insert(1))),
+        (
+            "SELECT TRIM(name) AS test FROM NullName;",
+            Ok(select_with_null!(test; Value::Null)),
+        ),
     ];
 
     for (sql, expected) in test_cases {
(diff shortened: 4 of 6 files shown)


src/executor/evaluate/mod.rs
```diff
@@ -281,5 +281,10 @@ async fn evaluate_function<'a, T: 'static + Debug>(
 
             Ok(Evaluated::from(Value::Str(converted)))
         }
+        Function::Trim(expr) => match eval_to_str("TRIM", expr).await? {
+            Nullable::Value(string) => Ok(Value::Str(string.trim().to_owned())),
+            Nullable::Null => Ok(Value::Null),
```
panarch, line 286: It seems that missing test case is only this one.
```rust
Nullable::Null => {
    panic!("lol");
}
```

It still pass all tests.
Except for this, it's.. perfect. 👍 

You may feel test cases are not enough but it's really enough.
Many string functions uses `eval_to_str` and it is based on `evaluate`, so it is not necessary to cover all ast expr branches in here.

This can be an one good reason of why closure or function is better than using macro.
Macro shows similar behavior with function in the view of.. reducing redundant codes, but it is only for the human, not the system.

To verify all codes, we need to reach all code branches generated by macros, but we don't need to do that if we use functions.
However, as we all know, there are also many advantages of using macros, so not easy to say what's better I think..

panarch (changes requested):
Awesome work!
You've done really well, almost nothing to say.
There is only a single one to mention.
Thanks 👍 👍 👍

src/executor/evaluate/mod.rs
ever0de: Oh no... I'll add a test case for Null.

ever0de:
Modified it! Please review it again

panarch (approved):
Awesome, everything looks great :)
Thanks!!! 👍 👍 👍

Issue #282, opened 2021-08-10
Implement Function: LTRIM, RTRIM


devgony:
# Implement Function: LTRIM, RTRIM

### LTRIM

```sql
-- postgresql ltrim
ltrim ( string text [, characters text ] ) → text
ltrim('zzzytest', 'xyz') → 'test'
```

### RTRIM

```sql
-- postgresql rtrim
rtrim ( string text [, characters text ] ) → text
rtrim('testxxzx', 'xyz') → 'test'
```

### `trim_start_matches` for LTRIM

```rust
pub fn trim_start_matches<'a, P>(&'a self, pat: P) -> &'a str
where
    P: Pattern<'a>,
```

Returns a string slice with all prefixes that match a pattern repeatedly removed.

The [pattern](https://doc.rust-lang.org/std/str/pattern/index.html) can be a `&str`, [char](https://doc.rust-lang.org/std/primitive.char.html), a slice of [char](https://doc.rust-lang.org/std/primitive.char.html)s, or a function or closure that determines if a character matches.

⇒ **_Should use `a slice of chars` as the pattern to implement postgresql LTRIM_**

## This issue will cover...

1. Add LTRIM(), RTRIM() at `ast/function.rs` 
2. Add LTRIM(), RTRIM() at `translate/function.rs`
3. Implement LTRIM(), RTRIM() at `executor/evaluate/mod.rs`
4. Implement test cases at `src/tests/function/`
5. Should pass `cargo check`, `cargo fmt`, `cargo clippy —all-targets`, `cargo test`

## References

1. [https://www.postgresql.org/docs/13/functions-string.html](https://www.postgresql.org/docs/13/functions-string.html)
2. [https://doc.rust-lang.org/std/string/struct.String.html#method.trim_start_matches](https://doc.rust-lang.org/std/string/struct.String.html#method.trim_start_matches)

ever0de:
Shouldn't it be closed?

panarch:
@ever0de Thanks!

PR #294, opened 2021-08-18
Implement Function: LTRIM, RTRIM

devgony:
## These functions will remove the longest string containing only characters in characters (a space by default) from the start(LTRIM) or end(RTRIM) of string. #282 

```rs
// src/executor/evaluate/mod.rs
Function::Ltrim { expr, chars } | Function::Rtrim { expr, chars } => {
            let name = if matches!(func, Function::Ltrim { .. }) {
                "LTRIM"
            } else {
                "RTRIM"
            };
            let pattern: Result<Vec<char>> = match chars {
                Some(chars) => match eval_to_str(chars).await? {
                    Nullable::Value(v) => Ok(v.chars().collect::<Vec<char>>()),
                    Nullable::Null => {
                        return Ok(Evaluated::from(Value::Null));
                    }
                },
                None => Ok(" ".chars().collect::<Vec<char>>()),
            };
            match eval_to_str(expr).await? {
                Nullable::Value(v) => {
                    if name == "LTRIM" {
                        Ok(Value::Str(v.trim_start_matches(&pattern?[..]).to_string()))
                    } else {
                        Ok(Value::Str(v.trim_end_matches(&pattern?[..]).to_string()))
                    }
                }
                Nullable::Null => Ok(Value::Null),
            }
            .map(Evaluated::from)
        }
```

## Create `func_with_two_arg` macro to handle optional second argument.
```rs
// src/translate/function.rs
    macro_rules! func_with_two_arg {
        ($func: ident) => {{
            check_len_range(stringify!($func).to_owned(), args.len(), 1, 2)?;
            let expr = translate_expr(args[0])?;
            let chars = if args.len() == 1 {
                None
            } else {
                Some(translate_expr(args[1])?)
            };
            Ok(Expr::Function(Box::new(Function::$func { expr, chars })))
        }};
    }
```

### DIFF
--- a/src/ast/function.rs
+++ b/src/ast/function.rs
@@ -66,6 +66,10 @@ pub enum Function {
     Degrees(Expr),
     #[strum(to_string = "PI")]
     Pi(),
+    #[strum(to_string = "LTRIM")]
+    Ltrim { expr: Expr, chars: Option<Expr> },
+    #[strum(to_string = "RTRIM")]
+    Rtrim { expr: Expr, chars: Option<Expr> },
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
--- a/src/executor/evaluate/mod.rs
+++ b/src/executor/evaluate/mod.rs
@@ -528,6 +528,33 @@ async fn evaluate_function<'a, T: 'static + Debug>(
 
             Ok(Evaluated::from(Value::I64(lcm(left, right))))
         }
+        Function::Ltrim { expr, chars } | Function::Rtrim { expr, chars } => {
+            let name = if matches!(func, Function::Ltrim { .. }) {
+                "LTRIM"
+            } else {
+                "RTRIM"
+            };
+            let pattern: Result<Vec<char>> = match chars {
+                Some(chars) => match eval_to_str(chars).await? {
+                    Nullable::Value(v) => Ok(v.chars().collect::<Vec<char>>()),
+                    Nullable::Null => {
+                        return Ok(Evaluated::from(Value::Null));
+                    }
+                },
+                None => Ok(" ".chars().collect::<Vec<char>>()),
+            };
+            match eval_to_str(expr).await? {
+                Nullable::Value(v) => {
+                    if name == "LTRIM" {
+                        Ok(Value::Str(v.trim_start_matches(&pattern?[..]).to_string()))
+                    } else {
+                        Ok(Value::Str(v.trim_end_matches(&pattern?[..]).to_string()))
+                    }
+                }
+                Nullable::Null => Ok(Value::Null),
+            }
+            .map(Evaluated::from)
+        }
     }
 }
 
--- a/src/tests/function/ltrim_rtrim.rs
+++ b/src/tests/function/ltrim_rtrim.rs
@@ -1,0 +1,99 @@
+use crate::*;
+
+test_case!(ltrim_rtrim, async move {
+    use Value::Str;
+    let test_cases = vec![
+        ("CREATE TABLE Item (name TEXT)", Ok(Payload::Create)),
+        (
+            r#"INSERT INTO Item VALUES (" zzzytest"), ("testxxzx ")"#,
+            Ok(Payload::Insert(2)),
+        ),
+        (
+            r#"SELECT LTRIM(name) AS test FROM Item"#,
+            Ok(select!(
+                "test"
+                Str;
+                "zzzytest".to_owned();
+                "testxxzx ".to_owned()
+            )),
+        ),
+        (
+            r#"SELECT LTRIM(name, ' xyz') AS test FROM Item"#,
+            Ok(select!(
+                "test"
+                Str;
+                "test".to_owned();
+                "testxxzx ".to_owned()
+            )),
+        ),
+        (
+            r#"SELECT RTRIM(name) AS test FROM Item"#,
+            Ok(select!(
+                "test"
+                Str;
+                " zzzytest".to_owned();
+                "testxxzx".to_owned()
+            )),
+        ),
+        (
+            r#"SELECT RTRIM(name, 'xyz ') AS test FROM Item"#,
+            Ok(select!(
+                "test"
+                Str;
+                " zzzytest".to_owned();
+                "test".to_owned()
+            )),
+        ),
+        (
+            r#"SELECT LTRIM(1) AS test FROM Item"#,
+            Err(EvaluateError::FunctionRequiresStringValue("LTRIM".to_owned()).into()),
+        ),
+        (
+            r#"SELECT LTRIM(name, 1) AS test FROM Item"#,
+            Err(EvaluateError::FunctionRequiresStringValue("LTRIM".to_owned()).into()),
+        ),
+        (
+            r#"SELECT RTRIM(1) AS test FROM Item"#,
+            Err(EvaluateError::FunctionRequiresStringValue("RTRIM".to_owned()).into()),
+        ),
+        (
+            r#"SELECT RTRIM(name, 1) AS test FROM Item"#,
+            Err(EvaluateError::FunctionRequiresStringValue("RTRIM".to_owned()).into()),
+        ),
+        (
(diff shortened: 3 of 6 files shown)


src/executor/evaluate/mod.rs
```diff
@@ -286,5 +286,27 @@ async fn evaluate_function<'a, T: 'static + Debug>(
...
         .map(Evaluated::from),
+        Function::Ltrim { expr, chars } | Function::Rtrim { expr, chars } => {
+            let name = if matches!(func, Function::Ltrim { .. }) {
+                "LTRIM"
+            } else {
+                "RTRIM"
+            };
+            let pattern: Result<Vec<char>> = match eval_to_str(name, chars).await? {
+                Nullable::Value(v) => Ok(v.chars().collect::<Vec<char>>()),
+                Nullable::Null => Ok(" ".chars().collect::<Vec<char>>()),
```
devgony, line 297: I just realized that if the 2nd arg `pattern` is Null, the result should be Null 😞

devgony: And empty string should do nothing..

panarch: Oh, I also didn't catch that, good point!
Then, AST would be.. like this?
```rust
pub enum Function {                           
    ...
    Ltrim { expr: Expr, chars: Option<Expr> },
    Rtrim { expr: Expr, chars: Option<Expr> },
}
```

devgony: I guess so. But It's weird current code works with the test cases without Option<>

src/tests/function/ltrim_rtrim.rs
```diff
@@ -0,0 +1,83 @@
...
+            r#"SELECT Rtrim(name) AS test FROM Item"#,
+            Ok(select!(
+                "test"
+                Str;
+                " zzzytest".to_owned();
+                "testxxzx".to_owned()
+            )),
+        ),
+        (
+            r#"SELECT Rtrim(name, 'xyz ') AS test FROM Item"#,
```
panarch, line 39: You are currently still working on it, so I'll simply leave as a comment.
As other SQL keywords are used as capital words (`SELECT`, `AS`, `FROM`...), it would be also good to use `LTRIM` and `RTRIM`.

src/executor/evaluate/mod.rs
panarch: That sounds like good time to add more test cases :)

src/tests/function/ltrim_rtrim.rs
devgony: Thank you for tidying up.
I may still be confused with asf function name 😰

src/translate/function.rs
```diff
@@ -78,6 +78,34 @@ pub fn translate_function(sql_function: &SqlFunction) -> Result<Expr> {
...
+                expr = translate_expr(args[0])?;
+                chars = Expr::Literal(AstLiteral::Null);
+            } else if let Ok(()) = check_len(name, args.len(), 2) {
+                expr = translate_expr(args[0])?;
+                chars = translate_expr(args[1])?;
+            } else {
+                return Err(TranslateError::UnsupportedExpr(String::from("LTRIM")).into());
+            }
+            Ok(Expr::Function(Box::new(Function::Ltrim { expr, chars })))
+        }
```
panarch, line 94: Here, `match` would be good to simplify codes and reduce redundant `check_len` calls.

fyi. below codes are two different examples, I think the second one is better in this case.
1.
```rust
"LTRIM" => match args.len() {                                                     
    1 => translate_expr(args[0])                                                  
        .map(|expr| Function::Ltrim { expr, chars: None })                        
        .map(Box::new)                                                            
        .map(Expr::Function),                                                     
    2 => translate_expr(args[0])                                                  
        .and_then(|expr| translate_expr(args[1]).map(|chars| (expr, Some(chars))))
        .map(|(expr, chars)| Function::Ltrim { expr, chars })                     
        .map(Box::new)                                                            
        .map(Expr::Function),                                                     
    _ => Err(TranslateError::UnsupportedExpr(String::from("LTRIM")).into()),      
},                                                                                
```

2.
```rust
"LTRIM" => {                                                                
    let (expr, chars) = match args.len() {                                  
        1 => (translate_expr(args[0])?, None)                               
        2 => (translate_expr(args[0])?, translate_expr(args[1]).map(Some)?),
        n => {                                                              
            return Err(TranslateError::FunctionArgsLengthRangeNotMatching { 
                name: "LTRIM".to_owned(),                                   
                min: 1,                                                     
                max: 2,                                                     
                found: n,                                                   
            });                                                             
        }                                                                   
    };                                                                      
                                                                            
    Ok(Expr::Function(Box::new(Function::Ltrim { expr, chars })))           
}                                                                           
```

And.. you don't need to hesitate about adding errors, please feel free to add errors if you feel that there is a new case which existing errors don't cover.

In your case the existing one `FunctionArgsLengthNotMatching` does not fit to your case because `LTRIM` and `RTRIM` accepts one or two params. Then, you can add a new one something like this.
```rust
#[error("function args.length not matching: {name}, expected: {expected}, found: {found}")]    
FunctionArgsLengthNotMatching {                                                                
    name: String,                                                                              
    expected: usize,                                                                           
    found: usize,                                                                              
},                                                                                             
                                                                                               
#[error("function args.length range not matching: {name}, range: {min}-{max}, found: {found}")]
FunctionArgsLengthRangeNotMatching {                                                           
    name: String,                                                                              
    min: usize,                                                                                
    max: usize,                                                                                
    found: usize,                                                                              
},                                                                                             
```

panarch: @devgony
Thanks to @No-YE, he added `check_len_range` closure to `translate/function` module.
ref. https://github.com/gluesql/gluesql/pull/311

It would good to apply in here :)
https://github.com/gluesql/gluesql/blob/83db4959ecf884c0fcfc5bf7b129add3a44760b5/src/translate/function.rs#L40-L52

src/translate/function.rs
```diff
@@ -73,6 +73,19 @@ pub fn translate_function(sql_function: &SqlFunction) -> Result<Expr> {
...
+    macro_rules! func_with_two_arg {
+        ($bound: ident, $func: ident) => {{
+            check_len_range(stringify!($func).to_owned(), args.len(), 1, 2)?;
+            let expr = translate_expr(args[0])?;
+            let chars = if args.len() == 1 {
+                None
+            } else {
+                Some(translate_expr(args[1])?)
+            };
+            Ok(Expr::Function(Box::new($bound::$func { expr, chars })))
```
devgony, line 85: It looks a bit ugly, but the best way to pass struct in macro i've found so far.
If there's better way i will fix real quick.

panarch: Here you can simply try this
```rust
 macro_rules! func_with_two_arg {
        ($func: ident) => {{
            check_len_range(stringify!($func).to_owned(), args.len(), 1, 2)?;
            let expr = translate_expr(args[0])?;
            let chars = if args.len() == 1 {
                None
            } else {
                Some(translate_expr(args[1])?)
            };
            Ok(Expr::Function(Box::new(Function::$func { expr, chars })))
        });
}
```

src/ast/function.rs
```diff
@@ -59,7 +77,18 @@ pub enum Function {
...
+        power: Expr,
+    },
+    Ltrim {
+        expr: Expr,
+        chars: Option<Expr>,
+    },
+    Rtrim {
+        expr: Expr,
+        chars: Option<Expr>,
+    },
```
panarch, line 91: Let's `#[strum(to_string = "LTRIM")]` and `#[strum(to_string = "RTRIM")]` to capitalize function names for the consistency.

panarch (changes requested):
Awesome 👍
There are only very tiny comments, let's quick fix and merge it into main branch!

src/translate/function.rs
devgony: Thank you for simplifying!
All your advices were adjusted and New PR was submitted!

panarch:
Looks all good, thanks a lot 👍 👍 👍

PR #296, opened 2021-08-18
Add functions: ASIN(), ACOS(), ATAN() 

### LINKED ISSUE #285: Implement Function: ASIN, ACOS, ATAN
## ACOS(X)

Returns the arc cosine of X, that is, the value whose cosine is X. Returns NULL if X is not in the range -1 to 1.
```
-- mysql reference
mysql> SELECT ACOS(1);
        -> 0
mysql> SELECT ACOS(1.0001);
        -> NULL
mysql> SELECT ACOS(0);
        -> 1.5707963267949
``` 
## ASIN(X)

Returns the arc sine of X, that is, the value whose sine is X. Returns NULL if X is not in the range -1 to 1.

```
-- mysql reference
mysql> SELECT ASIN(0.2);
        -> 0.20135792079033
mysql> SELECT ASIN('foo');

+-------------+
| ASIN('foo') |
+-------------+
|           0 |
+-------------+
1 row in set, 1 warning (0.00 sec)

mysql> SHOW WARNINGS;
+---------+------+-----------------------------------------+
| Level   | Code | Message                                 |
+---------+------+-----------------------------------------+
| Warning | 1292 | Truncated incorrect DOUBLE value: 'foo' |
+---------+------+-----------------------------------------+
``` 

## ATAN(X)

Returns the arc tangent of X, that is, the value whose tangent is X.

```
-- mysql reference
mysql> SELECT ATAN(2);
        -> 1.1071487177941
mysql> SELECT ATAN(-2);
        -> -1.1071487177941
``` 

## Reference

https://dev.mysql.com/doc/refman/8.0/en/mathematical-functions.html#function_acos
https://www.postgresql.org/docs/9.1/functions-math.html

boomkim:
Add `ASIN()`, `ACOS()`, `ATAN()` functions. 
These functions support numeric string literals and boolean and floating numbers.
resolve #285 

referenced @maruoovv 's [pull request](https://github.com/gluesql/gluesql/pull/289) a lot.


boomkim:
Let me fix clippy error and pull request again.

boomkim:
Resolved clippy error.

src/executor/evaluate/error.rs
```diff
@@ -42,4 +45,7 @@ pub enum EvaluateError {
 
     #[error("unreachable empty aggregate value: {0:?}")]
     UnreachableEmptyAggregateValue(Aggregate),
+
+    #[error("out of range: {0}")]
+    OutOfRange(String),
```
panarch, line 50: Ok, here let's simply depend on Rust compiler.
Rust compiler will return `NaN` and we can use that rather than returning explicit error.

src/executor/evaluate/mod.rs
```diff
@@ -281,6 +283,72 @@ async fn evaluate_function<'a, T: 'static + Debug>(
...
         }
+        Function::ASin(expr) => {
+            let number = match eval(expr).await?.try_into()? {
+                Value::F64(v) => Some(v),
+                Value::Str(v) => match f64::from_str(&v) {
+                    Ok(f) => Some(f),
+                    Err(_) => None,
+                },
+                Value::I64(v) => f64::from_i64(v),
+                _ => None,
```
panarch, line 294: Now, let's make the spec clear.

You can only accept three types; `F64`, `I64` and `Null`.
Except for these three types, `FunctionReuqiresFloatValue` can be the one to return.
`F64` - we can use directly.
`I64` - `as f64` to cast and we can use that.
`Null` -> then return value becomes also null.

panarch (changes requested):
Changes are simple and straightforward to look around.
Only change requests are about the specifications.
Thanks! 👍 👍 👍

panarch (approved):
Everything looks great, thanks! 👍 👍 👍

Issue #276, opened 2021-08-09
Support timeout based GC to SledStorage snapshot


panarch:
`SledStorage` supports MVCC transaction using snapshots to manage internal data.
When data manipulation occurs, snapshot always grows even if `delete` op is executed.
Snapshot records all history run by transactions.
However, transaction instances usually exist only for short period, expired snapshot data can exist.
`gc` in SledStorages cleans up that garbage snapshot records.

Currently `gc` simply removes all expired data right after it figures out that there is no related transaction(`txid`).
New timeout based GC mode manages data based on duration from transaction creation time to gc time.

### `SledStorage` is daemonless, so transaction can last forever.
Not like some other daemon based databases, `SledStorage` works as an embedded SQL database without daemon.
It means that there's no one who manages transactions.
In many cases, outdated transaction can exist in the storage data.

e.g.
If user runs `BEGIN` to begin transaction and they simply forget to run `COMMIT` or `ROLLBACK`, then this transaction simply lasts forever so gc cannot clean this data.

By limiting max lifetime of transaction, this kind of garbage accumulation scenario can be resolved.

### Preparing time-travel queries.
Thanks to snapshot based structure, it is easy to add support of time-travel queries.
Users can access historical data using this feature.
ref.
https://www.cockroachlabs.com/blog/time-travel-queries-select-witty_subtitle-the_future/

To support time-travel queries, `gc` must not delete snapshot data until the configured time expires.

PR #297, opened 2021-08-19
Implement SledStorage transaction timeout support,

panarch:
`SledStorage` users can set transaction timeout.
Transaction terminates when either `ROLLBACK` | `COMMIT` are executed by users or transaction time limit has exceeded.

`SledStorage` now supports `set_transaction_timeout` method to set transaction timeout in milliseconds.
```rust
let storage = SledStorage:new(filename).unwrap();

storage.set_transaction_timeout(Some(3600 * 1000)); // in milliseconds
```

resolve https://github.com/gluesql/gluesql/issues/276

### DIFF
--- a/src/storages/sled_storage/alter_table.rs
+++ b/src/storages/sled_storage/alter_table.rs
@@ -1,13 +1,19 @@
 #![cfg(feature = "alter-table")]
 
 use {
-    super::{error::err_into, error::StorageError, fetch_schema, key, lock, SledStorage, Snapshot},
+    super::{
+        error::err_into,
+        fetch_schema, key,
+        lock::{self, LockAcquired},
+        transaction::TxPayload,
+        SledStorage, Snapshot,
+    },
     crate::{
         ast::ColumnDef, executor::evaluate_stateless, schema::ColumnDefExt, utils::Vector,
         AlterTable, AlterTableError, MutResult, Result, Row, Schema, Value,
     },
     async_trait::async_trait,
-    sled::transaction::{ConflictableTransactionError, TransactionError},
+    sled::transaction::ConflictableTransactionError,
     std::{iter::once, str},
 };
 
@@ -15,25 +21,11 @@ macro_rules! try_self {
     ($self: expr, $expr: expr) => {
         match $expr {
             Err(e) => {
-                return Err(($self, e.into()));
+                return Err(($self, e));
             }
             Ok(v) => v,
         }
     };
-}
-
-macro_rules! transaction {
-    ($self: expr, $expr: expr) => {{
-        let result = $self.tree.transaction($expr).map_err(|e| match e {
-            TransactionError::Abort(e) => e,
-            TransactionError::Storage(e) => StorageError::Sled(e).into(),
-        });
-
-        match result {
-            Ok(_) => Ok(($self, ())),
-            Err(e) => Err(($self, e)),
-        }
-    }};
 }
 
 #[async_trait(?Send)]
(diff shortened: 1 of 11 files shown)


src/storages/sled_storage/transaction.rs
```diff
@@ -76,73 +88,88 @@ impl Transaction for SledStorage {
...
+        };
 
-                        tree.insert(value_key, snapshot)?;
-                    }
-                    None => {
-                        tree.remove(value_key)?;
-                    }
-                };
-            }
+        match gc() {
```
ever0de, line 151: Can it be implemented in this way?
```rust
gc().map(|_| (storage, ())).map_err(|e| (storage, e))
```

src/storages/sled_storage/transaction.rs
```diff
@@ -76,73 +88,88 @@ impl Transaction for SledStorage {
...
+            Ok(_) => Ok((storage, ())),
+            Err(e) => Err((storage, e)),
+        }
```
ever0de, line 154: Can it be implemented in this way?

```rust
gc().map(|_| (storage, ())).map_err(|e| (storage, e))
```

ever0de: This solution does not work.
Storage needs to be copied...

ever0de:
```rust
gc().map(|_| (storage.clone(), ()))
            .map_err(|e| (storage, e))
```
hmm..

panarch: Yeah, I'd also really like to make this part more simple, but closures takes ownership so it cannot work.
image

In the repo, you may recognize lots of macro uses named `try_self!` or `try_into!` and all that `try_{..}` macro is due to this issue.

src/storages/sled_storage/alter_table.rs
```diff
@@ -134,8 +131,14 @@ impl AlterTable for SledStorage {
...
-            Ok(())
-        })
+            Ok(TxPayload::Success)
+        });
+
+        match self.check_retry(tx_result) {
+            Ok(true) => self.rename_schema(table_name, new_table_name).await,
+            Ok(false) => Ok((self, ())),
+            Err(e) => Err((self, e)),
+        }
```
ever0de, line 141: Make decl macro..

```rust
macro_rules! check_retry {
    ($self:expr, $tx_result:expr, $method:ident ($($params:expr),*)) => {{
        match $self.check_retry($tx_result) {
            Ok(true) => $self.$method($($params),*).await,
            Ok(false) => Ok(($self, ())),
            Err(e) => Err(($self, e)),
        }
    }};
}
```

Why don't you write it this way?
The pattern seems to repeat itself.

`check_retry!(self, tx_result, rename_schema(table_name, new_table_name))`

panarch: I agree that could be a good choice to use.
However, I'm actually trying to reduce using macros now because decl macros are convenient but it can also make hard to track code coverage.
If we use Rust closures or functions, Rust compiler can verify its behavior, but macro is more likely a simple tool for generating codes.
To verify codes, it is enough to verify each function or closure, but for the case of macros, we still need to check all the codes generated by macros.

Here in transaction implementation, it was especially tough due to a lot of concurrency issues.
I depend a lot on [tarpaulin](https://github.com/xd009642/tarpaulin), it provide code coverage test for Rust.
e.g.
image

image

It is very convenient, I can easily check which codes are covered or not.
(fyi. I usually run `cargo tarpaulin --out Html --output-dir {dir-to_export}` to generate HTML format report and review it).

However, if I use macro to pack these codes, then I'm still supposed to test each code branches but I cannot get proper support from `tarpaulin`.

It can be really cool if we can make `match -- check_retry` part as a shared function, but because of `$self.$method($($params),*).await,` this part... it's a bit tough.
I didn't know what to do so just left like now. but... maybe I can make the function.. by wrapping `$self.$method($($params),*).await,` as a closure and passing it to a shared function to handle it.
Sounds... quite interesting

ever0de: I agree that it is difficult to track code. 
I will make it a shared function and deliver it to you.

panarch: You can also upload as a separate PR. This becomes really a good motivation to do something.
I think we can also apply similar approach to remove `try_`s, `try_self!`, `try_block!` and `try_into!`. Now.. I think I can replace it into normal functions.

ever0de: Okay, I'll put it in a separate PR

PR #304, opened 2021-08-20
Replace binary op macro to normal function,

panarch:
Replace binary_op! macro to binary_op function

### DIFF
--- a/src/executor/evaluate/evaluated.rs
+++ b/src/executor/evaluate/evaluated.rs
@@ -83,34 +83,46 @@ impl<'a> PartialOrd for Evaluated<'a> {
     }
 }
 
-macro_rules! binary_op {
-    ($name:ident, $op:tt) => {
-        pub fn $name<'b>(&self, other: &Evaluated<'a>) -> Result<Evaluated<'b>> {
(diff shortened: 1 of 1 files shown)


src/executor/evaluate/evaluated.rs
```diff
@@ -83,34 +83,46 @@ impl<'a> PartialOrd for Evaluated<'a> {
...
-            }
+fn binary_op<'a, 'b, T, U>(
+    l: &Evaluated<'a>,
+    r: &Evaluated<'b>,
+    value_op: T,
+    literal_op: U,
+) -> Result<Evaluated<'b>>
+where
+    T: Fn(&Value, &Value) -> Result<Value>,
+    U: Fn(&Literal<'a>, &Literal<'b>) -> Result<Literal<'b>>,
```
ever0de, line 94: It looks like a function used only once.
I recommend [FnOnce](https://doc.rust-lang.org/stable/std/ops/trait.FnOnce.html).

panarch: Oh, I missed that, thanks!

ever0de (approved):
LGTM!

PR #307, opened 2021-08-21
Implement TRIM function - Other features

### LINKED ISSUE #267: Implement TRIM function

ever0de:
resolved #267 

- [x] If the specifier is BOTH, Remove the specifier parameter to the prefix and suffix
- [x] If the specifier is LEADING, Remove the specifier parameter to the prefix
- [x] If the specifier is TRAILING, Remove the specifier parameter to the suffix

I left a comment because I thought it would be neat to implement it with From.

### DIFF
--- a/src/executor/evaluate/error.rs
+++ b/src/executor/evaluate/error.rs
@@ -49,9 +49,6 @@ pub enum EvaluateError {
     #[error("unreachable empty aggregate value: {0:?}")]
     UnreachableEmptyAggregateValue(Aggregate),
 
-    #[error("unreachable iterator unwrap")]
-    UnreachableIteratorUnwrap,
-
     #[error("the divisor should not be zero")]
     DivisorShouldNotBeZero,
 }
--- a/src/executor/evaluate/error.rs
+++ b/src/executor/evaluate/error.rs
@@ -49,6 +49,9 @@ pub enum EvaluateError {
     #[error("unreachable empty aggregate value: {0:?}")]
     UnreachableEmptyAggregateValue(Aggregate),
 
+    #[error("unreachable iterator unwrap")]
+    UnreachableIteratorUnwrap,
+
     #[error("the divisor should not be zero")]
     DivisorShouldNotBeZero,
 }
--- a/src/executor/evaluate/mod.rs
+++ b/src/executor/evaluate/mod.rs
@@ -432,8 +432,14 @@ async fn evaluate_function<'a, T: 'static + Debug>(
             };
 
             let mut expr_chars = expr_str.chars();
-            let first_char_is_empty = expr_chars.next().unwrap().is_whitespace();
-            let last_char_is_empty = expr_chars.last().unwrap().is_whitespace();
+            let first_char_is_empty = expr_chars
+                .next()
+                .ok_or(EvaluateError::UnreachableIteratorUnwrap)?
+                .is_whitespace();
+            let last_char_is_empty = expr_chars
+                .last()
+                .ok_or(EvaluateError::UnreachableIteratorUnwrap)?
+                .is_whitespace();
 
             let both = || match (first_char_is_empty, last_char_is_empty) {
                 (true, true) => {
--- a/src/executor/evaluate/mod.rs
+++ b/src/executor/evaluate/mod.rs
@@ -419,54 +419,58 @@ async fn evaluate_function<'a, T: 'static + Debug>(
                 return Ok(Value::Str(expr_str.to_owned())).map(Evaluated::from);
             }
 
-            match trim_where {
-                Some((trim_where_field, filter_chars)) => {
-                    let filter_chars = match eval_to_str(filter_chars).await? {
-                        Nullable::Value(str) => str.chars().collect::<Vec<_>>(),
-                        Nullable::Null => return Ok(Evaluated::from(Value::Null)),
-                    };
+            let (trim_where_field, filter_chars) = match trim_where {
+                Some(trim_where) => trim_where,
+                None => {
+                    return Ok(Evaluated::from(Value::Str(expr_str.trim().to_owned())));
+                }
+            };
 
-                    let mut expr_chars = expr_str.chars();
-                    let first_char_is_empty = expr_chars.next().unwrap().is_whitespace();
-                    let last_char_is_empty = expr_chars.last().unwrap().is_whitespace();
+            let filter_chars = match eval_to_str(filter_chars).await? {
+                Nullable::Value(str) => str.chars().collect::<Vec<_>>(),
+                Nullable::Null => return Ok(Evaluated::from(Value::Null)),
+            };
 
-                    let result = match trim_where_field {
-                        TrimWhereField::Both => match (first_char_is_empty, last_char_is_empty) {
-                            (true, true) => {
-                                let expr = expr_str.trim();
-                                expr.trim_matches(&filter_chars[..])
-                            }
-                            (true, false) => {
-                                let expr = expr_str.trim_start();
-                                expr.trim_end_matches(&filter_chars[..])
-                            }
-                            (false, true) => {
-                                let expr = expr_str.trim_end();
-                                expr.trim_start_matches(&filter_chars[..])
-                            }
-                            (false, false) => expr_str.trim_matches(&filter_chars[..]),
-                        },
-                        TrimWhereField::Leading => {
-                            if first_char_is_empty {
-                                expr_str.trim_start()
-                            } else {
-                                expr_str.trim_start_matches(&filter_chars[..])
-                            }
-                        }
-                        TrimWhereField::Trailing => {
-                            if last_char_is_empty {
-                                expr_str.trim_end()
-                            } else {
-                                expr_str.trim_end_matches(&filter_chars[..])
-                            }
+            let mut expr_chars = expr_str.chars();
+            let first_char_is_empty = expr_chars.next().unwrap().is_whitespace();
+            let last_char_is_empty = expr_chars.last().unwrap().is_whitespace();
+
+            let both = || match (first_char_is_empty, last_char_is_empty) {
+                (true, true) => {
+                    let expr = expr_str.trim();
+                    expr.trim_matches(&filter_chars[..])
+                }
+                (true, false) => {
+                    let expr = expr_str.trim_start();
+                    expr.trim_end_matches(&filter_chars[..])
+                }
+                (false, true) => {
+                    let expr = expr_str.trim_end();
+                    expr.trim_start_matches(&filter_chars[..])
+                }
+                (false, false) => expr_str.trim_matches(&filter_chars[..]),
+            };
+
+            Ok(Value::Str(
+                match trim_where_field {
+                    TrimWhereField::Both => both(),
+                    TrimWhereField::Leading => {
+                        if first_char_is_empty {
+                            expr_str.trim_start()
+                        } else {
+                            expr_str.trim_start_matches(&filter_chars[..])
                         }
                     }
-                    .to_owned();
-
-                    Ok(Value::Str(result))
+                    TrimWhereField::Trailing => {
+                        if last_char_is_empty {
+                            expr_str.trim_end()
+                        } else {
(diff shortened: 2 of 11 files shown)


src/translate/expr.rs
```diff
@@ -16,10 +18,8 @@ use {
 pub fn translate_expr(sql_expr: &SqlExpr) -> Result<Expr> {
     match sql_expr {
         SqlExpr::Identifier(ident) => match ident.quote_style {
-            Some(_) => Ok(Expr::Literal(AstLiteral::QuotedString(
-                ident.value.to_owned(),
-            ))),
-            None => Ok(Expr::Identifier(ident.value.to_owned())),
+            Some(_) => Ok(Expr::Literal(AstLiteral::QuotedString(ident.value.clone()))),
```
panarch, line 21: Wow, that's really good catch to replace `ident.value.to_owned()` to `ident.value.clone()`!
`ident.value` type is `String` so.. `clone` is better in here. Thanks!

ever0de:
I Thinking about code coverage and refactoring...

src/executor/evaluate/mod.rs
```diff
@@ -315,11 +315,73 @@ async fn evaluate_function<'a, T: 'static + Debug>(
...
+                                }
+                            }
+                            "TRAILING" => {
+                                if last.is_whitespace() {
+                                    expr.trim_end()
+                                } else {
+                                    expr.trim_end_matches(&filter_chars[..])
+                                }
+                            }
+                            _ => unimplemented!(),
```
ever0de, line 370: Should I use `UnsupportedCompoundIdentifier` here?

src/tests/function/trim.rs
```diff
@@ -48,6 +35,56 @@ test_case!(trim, async move {
...
+                "TRIM(TRAILING 'xyz' FROM name)"
+                Value::Str;
+                "     blank".to_owned();
+                "xxxyzblank".to_owned();
+                "xxxyzblank".to_owned();
+                "     blank".to_owned();
+                "  xyzblank".to_owned();
+                "xxxyzblankxyz".to_owned()
+            )),
+        ),
```
ever0de, line 87: Add error cases

panarch:
You can convert this PR as a draft :)

ever0de:
I'm waiting for [PR](https://github.com/sqlparser-rs/sqlparser-rs/pull/334)

ever0de:
Reopening PR due to the release of version 0.10 of `sqlparser`.

ever0de:
Sorry, I pushed after minor code refactoring.

panarch:
No need to rush :) I'll review it during daytime tomorrow.

src/executor/evaluate/mod.rs
```diff
@@ -395,11 +395,67 @@ async fn evaluate_function<'a, T: 'static + Debug>(
...
+                Nullable::Null => return Ok(Value::Null).map(Evaluated::from),
+            };
+            let expr_str = expr_str.as_str();
+
+            if expr_str.is_empty() {
+                return Ok(Value::Str(expr_str.to_owned())).map(Evaluated::from);
+            }
+
+            match trim_where {
+                Some((trim_where_field, filter_chars)) => {
```
panarch, line 410: Less indent would be better for readability.
How about this?

```rust
let ((trim_where_field, filter_chars)) = match trim_where {
    Some(trim_where) => trim_where,
    None => {
        return Ok(Evaluated::from(Value::Str(expr_str.trim().to_owned())));
    }
};

Comment (ever0de) [reply src/executor/evaluate/mod.rs:410]: The indentation is reduced and it looks very good!!!

Comment (ever0de) [reply src/executor/evaluate/mod.rs]: fixed it!

File: src/translate/ast_literal.rs [hunk 5]
```diff
@@ -29,6 +31,7 @@ pub fn translate_ast_literal(sql_value: &SqlValue) -> Result<AstLiteral> {
     })
 }
 
+// TODO: Repalce the functions below with From
```
panarch, line 34: May be.. this comment is no longer needed?

ever0de: Ah! posted wrong
I'll fix it right away

src/executor/evaluate/mod.rs
```diff
@@ -408,11 +408,77 @@ async fn evaluate_function<'a, T: 'static + Debug>(
...
+                        if last_char_is_empty {
+                            expr_str.trim_end()
+                        } else {
+                            expr_str.trim_end_matches(&filter_chars[..])
+                        }
+                    }
+                }
+                .to_owned(),
+            ))
+            .map(Evaluated::from)
```
panarch, line 480: Ok, let me have some time to think about this part.
Here the Unreachable- case is what all caused by ourselves, sqlparser-rs and gluesql.
I think.. there must be a way to resolve it.
I'll also try, it looks like quite interesting puzzle.

ever0de: You should think of a way to check if it's whitespace without using an iterator.

ex) `str[0..1] == " "`
...Should I do this?

panarch: Awesome realtime, currently.. I just began to try..
from something like this
image

src/executor/evaluate/error.rs
```diff
@@ -49,6 +49,9 @@ pub enum EvaluateError {
     #[error("unreachable empty aggregate value: {0:?}")]
     UnreachableEmptyAggregateValue(Aggregate),
 
+    #[error("unreachable iterator unwrap")]
+    UnreachableIteratorUnwrap,
```
panarch, line 53: Quick comment, now you can remove this :)

src/executor/evaluate/mod.rs
```diff
@@ -408,11 +408,42 @@ async fn evaluate_function<'a, T: 'static + Debug>(
...
+        Function::Trim { expr, trim_where } => {
+            let expr_str = match eval_to_str(expr).await? {
+                Nullable::Value(str) => str,
+                Nullable::Null => return Ok(Value::Null).map(Evaluated::from),
+            };
+            let expr_str = expr_str.as_str();
+
+            if expr_str.is_empty() {
+                return Ok(Value::Str(expr_str.to_owned())).map(Evaluated::from);
+            }
```
panarch, line 420: Result of `TRIM(NULL FROM "")` is expected to be `NULL` but, it returns empty string.
This looks not necessary to have.

src/executor/evaluate/mod.rs
```diff
@@ -408,11 +408,42 @@ async fn evaluate_function<'a, T: 'static + Debug>(
...
+                    return Ok(Evaluated::from(Value::Str(expr_str.trim().to_owned())));
+                }
+            };
+
+            let filter_chars = match filter_chars {
+                Some(expr) => match eval_to_str(expr).await? {
+                    Nullable::Value(str) => str.chars().collect::<Vec<_>>(),
+                    Nullable::Null => return Ok(Evaluated::from(Value::Null)),
+                },
+                None => vec![' '],
```
panarch, line 434: These two cases need to be covered in the integration tests.

panarch (changes requested):
Looks good!
Tiny comments are here, now it is soon to be merged. Thanks!

src/executor/evaluate/error.rs
ever0de: I made a mistake. Thank you

src/executor/evaluate/mod.rs
ever0de: Fixed it

src/executor/evaluate/mod.rs
ever0de: To test this you need to modify sqlparser.

panarch: Oh, I see.
Is there a reason to change filter_chars type from `Expr` to `Option<Expr>`?
It looks like that conversion is not necessary.
If we keep `Expr` then,
```rust
let filter_chars = match eval_to_str(filter_chars).await? {   
    Nullable::Value(str) => str.chars().collect::<Vec<_>>(),  
    Nullable::Null => return Ok(Evaluated::from(Value::Null)),
};                                                            
```
No unreachable case is required to be covered.

panarch (changes requested):
A single more `NULL` test case is required.
```rust
(                                                                
    r#"SELECT TRIM(TRAILING NULL FROM name) FROM Test LIMIT 2;"#,
    Ok(select_with_null!(                                        
        "TRIM(TRAILING NULL FROM name)";                         
        Value::Null;                                             
        Value::Null                                              
    )),                                                          
),                                                               
```

src/executor/evaluate/mod.rs
ever0de: This is to add only translate and error cases after modification in sqlparser in the future.

If you feel you don't need it, I'll restore it.

panarch: Ok,
At first, we don't need to prepare future changes which cannot be covered by us.
Second, I think this is why I was confused.
If we aim to set the proper AST, then... 

Not this,
```rust
Trim {                                         
    expr: Expr,                                
    trim_where: Option<(TrimWhereField, Expr)>,
},       
```           

Here would be the right one..?
```rust
Trim {                                         
    expr: Expr,                                
    filter_chars: Option<Expr>,
    trim_where: Option<TrimWhereField>,
},       
```
Because `filter_chars` and `trim_where` are independent to each other.

ever0de: I think that's right. I relied too much on existing types

panarch: It's fully understandable. We really know that there were lots of changes you've done on `TRIM`, not only the changes in here, but also contributions to the `sqlparser-rs`. :)

ever0de: Fixed it

panarch (approved):
Awesome, also hope to see the update is applied to `sqlparser-rs` :)
Thanks!

Issue #306, opened 2021-08-21
Update binary operator `/` to return custom error when the divisor is zero.


MRGRAVITY817:
Current implementation for binary operator `/` panics out when the divisor is `0`. 
```console
[SQL] SELECT * FROM Arith WHERE 10 / 0 = id;
thread 'arithmetic' panicked at 'attempt to divide by zero', src/data/literal.rs:236:42
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
I think it would be nice to return a custom error for this.

This should also be considered when implementing `%` operator #299

panarch:
Sounds so good! and it would be good to make a separate PR for this issue then I'll quickly review and merge it. Thanks for reporting.

MRGRAVITY817:
@panarch Since this issue is resolved, I think it should be closed.

panarch:
Oh I missed it, thanks a lot!

PR #309, opened 2021-08-21
Support `DivisorShouldNotBeZero` error and Remove `{number} / {interval}` implementations

MRGRAVITY817:
As mentioned on issue #306 , current `/` operator doesn't handle divisor zero error, but panic out.

So I added two new errors: `ValueError::DivisorShouldNotBeZero` and `LiteralError::DivisorShouldNotBeZero`,
and also changed some code to handle error.

### DIFF
--- a/src/data/literal.rs
+++ b/src/data/literal.rs
@@ -233,6 +233,7 @@ impl<'a> Literal<'a> {
     }
 
     pub fn divide<'b>(&self, other: &Literal<'a>) -> Result<Literal<'b>> {
+        use super::Interval as I;
         match (self, other) {
             (Number(l), Number(r)) => {
                 if let (Ok(l), Ok(r)) = (l.parse::<i64>(), r.parse::<i64>()) {
@@ -252,6 +253,9 @@ impl<'a> Literal<'a> {
                 }
             }
             (Number(l), Interval(r)) => {
+                if (r == &I::Microsecond(0)) | (r == &I::Month(0)) {
+                    return Err(LiteralError::DivisorShouldNotBeZero.into());
+                }
                 if let Ok(l) = l.parse::<i64>() {
                     Ok(Interval(l / *r))
                 } else if let Ok(l) = l.parse::<f64>() {
--- a/src/data/literal.rs
+++ b/src/data/literal.rs
@@ -343,4 +343,38 @@ mod tests {
         matches!(Null.concat(Boolean(true)), Null);
         matches!(Null.concat(Null), Null);
     }
+
+    #[test]
+    fn divide() {
+        use crate::data::interval::Interval as I;
+
+        macro_rules! num {
+            ($num: expr) => {
+                Number(Cow::Owned($num.to_owned()))
+            };
+        }
+
+        macro_rules! itv {
+            ($itv: expr) => {
+                Interval(I::Microsecond($itv))
+            };
+        }
+
+        let num_divisor = |x: &str| Number(Cow::Owned(x.to_owned()));
+        let itv_divisor = |x: i64| Interval(I::Microsecond(x));
+
+        assert_eq!(num!("12").divide(&num_divisor("2")).unwrap(), num!("6"));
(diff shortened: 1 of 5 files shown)


panarch (changes requested):
Thanks so much!

This is not a problem of the changes in here.
I've found that `number / interval` division operation was allowed!
That's my mistake and obviously it's a bug.
* https://github.com/gluesql/gluesql/issues/310

In here, you can simply remove codes of `Interval` divisor check codes because it will be removed in that #310 issue.

MRGRAVITY817:
> Thanks so much!
> 
> This is not a problem of the changes in here.
> I've found that `number / interval` division operation was allowed!
> That's my mistake and obviously it's a bug.
> 
>     * [`{number} / {interval}` should not be allowed #310](https://github.com/gluesql/gluesql/issues/310)
> 
> 
> In here, you can simply remove codes of `Interval` divisor check codes because it will be removed in that #310 issue.

Aha 😃 Sure I'll update it

src/data/literal.rs
```diff
@@ -233,36 +236,41 @@ impl<'a> Literal<'a> {
...
+                    } else {
+                        Ok(Interval(*l / r))
+                    }
                 } else if let Ok(r) = r.parse::<f64>() {
-                    Ok(Interval(*l / r))
+                    if r == 0.0 {
+                        Err(LiteralError::DivisorShouldNotBeZero.into())
+                    } else {
+                        Ok(Interval(*l / r))
+                    }
```
panarch, line 266: Two `Err` in here are only the cases not covered by tests.
Let's add tests for generating those :)

src/data/literal.rs
```diff
@@ -320,4 +326,34 @@ mod tests {
...
+        assert_eq!(num!("12").divide(&num_divisor("2.0")).unwrap(), num!("6"));
+        assert_eq!(num!("12.0").divide(&num_divisor("2")).unwrap(), num!("6"));
+        assert_eq!(num!("12.0").divide(&num_divisor("2.0")).unwrap(), num!("6"));
+        assert_eq!(itv!(12).divide(&num_divisor("2")).unwrap(), itv!(6));
+        assert_eq!(itv!(12).divide(&num_divisor("2.0")).unwrap(), itv!(6));
+        matches!(num!("12").divide(&Null).unwrap(), Null);
+        matches!(itv!(12).divide(&Null).unwrap(), Null);
+        matches!(Null.divide(&num_divisor("2")).unwrap(), Null);
+        matches!(Null.divide(&Null).unwrap(), Null);
+    }
```
panarch, line 358: New unit tests for `Literal`! Thanks!

panarch (changes requested):
Awesome, and also thanks that you removed some `{number} / {interval}` cases :)
The only request here is that adding two test cases which generate `{interval} / {f64 literal}` and `{interval} / {i64 literal}`.

panarch (approved):
Looks so good! Thanks a lot for the contribution.

fyi. It's also nice that you also add fix for removing `{number} / {interval}`.
However, it would be better to contain a single task for the single PR.
You can simply think about the PR as a bundle of work which can be summarized as an one context.

src/data/literal.rs
ever0de: Shouldn't we use `assert` macro where we use 'matches'?

panarch: There's actually a tricky issue.
Because.. `Value::Null != Value::Null`.
In SQL `NULL` can be only compared by `IS NULL`.
`Value` does not implement `Eq`, it only implements `PartialEq` because of this `Null` value.

MRGRAVITY817: I was also curious about this. Thanks for the question @ever0de  and thanks for the answer @panarch 👍

MRGRAVITY817:
> Looks so good! Thanks a lot for the contribution.
> 
> fyi. It's also nice that you also add fix for removing `{number} / {interval}`.
> However, it would be better to contain a single task for the single PR.
> You can simply think about the PR as a bundle of work which can be summarized as an one context.

Thanks for the feedback. I should definetely persue one PR per issue/task.

Issue #299, opened 2021-08-20
Implement modulo(`%`) binary operator


panarch:
e.g.
```
12 % 3 = 4
10 % 12 = 10
..
```

Following tasks are expected to be done
* add `fn modulo` to `data/value/mod.rs`
You can also add unit tests of this new `modulo` fn, there's a existing test case named `arithmetic` and you can use it.

* add `fn modulo` to `data/literal.rs`
Similar to `Value`, but.. there's lack of unit tests though it is required.
You need to add a new one.

* add `fn modulo` to `execute/evaluate/evaluated.rs`

* add integration test cases of modulo to `src/tests/arithmetic.rs`
Here you can also cover code branches of `evaluated.rs` by adding `%` test cases to `arithmetic`.

ref https://github.com/gluesql/gluesql/pull/295

MRGRAVITY817:
I'll do those tasks 👍

PR #319, opened 2021-08-24
Implement binary operator modulo `%`

MRGRAVITY817:
### Resolving issue #299.
I added some code to execute modulo operator `%` to get arithmetic remainder.

Basically `%` operator supports `Integer` and `Float` value as dividend and divisor.
(Nice explanation about division rule is in PR #295  )
Here are several use cases for `%`:
```sql
# int % int => int
SELECT 12 % 5 AS mod;
 mod
-----
   2

# float % float => float
# float % int => float
# int % float => float
SELECT 12.0 % 5 AS mod;
 mod
-----
 2.0
```

Operation with at least one `Null` value will return error.
By the division rule, you will get an `DivisorShouldNotBeZero` error if you try to divide with zero.

### Here are the materials I've looked for this PR
- PostgreSQL math operators and functions: https://www.postgresql.org/docs/13/functions-math.html
- MySQL arithmetic operators: https://dev.mysql.com/doc/refman/8.0/en/arithmetic-functions.html#operator_mod


panarch (approved):
Beautiful, looks good! thanks a lot 👍  merging!

Issue #300, opened 2021-08-20
Update `DIV` & `MOD` to use `divide` & `modulo` functions in `Evaluated`


panarch:
Current `DIV` and `MOD` implementation works separately with `Evaluated`'s `divide` and `modulo`.
Let's merge these to use `Evaluated`'s implementation.

Prior issue to be done first - https://github.com/gluesql/gluesql/issues/299

ref https://github.com/gluesql/gluesql/pull/295

MRGRAVITY817:
I will update those functions with issue #299

MRGRAVITY817:
One thing to point out, that the behavior of `/` operator and `DIV` aren't same.
Tested on postgreSQL,
```sql

SELECT 12/5.0 AS div;
#   Result
        div
--------------------
 2.4000000000000000


SELECT DIV(12, 5.0) AS div;
#  Result
 div
-----
   2
```
Seems like `/` operator **can** return `float` when dividend or divisor is `float`, whereas `DIV` returns `integer` quotient for any numeric type.

Well of course, this ain't a big deal since all I need to do is to convert the result of `/` to `i64` for `DIV` 🍰 .

Reference
---
https://www.postgresql.org/docs/13/functions-math.html

panarch:
Thanks, that's really a good detail.
PostgreSQL's way that you suggested sounds reasonable.
I also agree about `DIV` function to return only `i64` type.

MRGRAVITY817:
> Thanks, that's really a good detail.
> PostgreSQL's way that you suggested sounds reasonable.
> I also agree about `DIV` function to return only `i64` type.

Oh, and also `DIV` cannot do `{interval} / {numeric}` as well. Only `numeric`s are supported.
Since there are quite a lot of differences ... maybe I should leave the `DIV` as it is, and implement only `%` -> `MOD`.

panarch:
> Oh, and also DIV cannot do {interval} / {numeric} as well. Only numerics are supported.

That also makes sense.

> Since there are quite a lot of differences ... maybe I should leave the DIV as it is, and implement only % -> MOD.

Nice details you've catched. I also agree, thanks!
Sounds reasonable to handle `DIV` and `/` separately.

PR #321, opened 2021-08-24
Update math function `MOD` to use `modulo` in `Evaluated`

MRGRAVITY817:
### This resolves issue #300.

Major updates are
- Update `MOD` to use `modulo()` in `Evaluated`
- Change the name of divisor zero error


panarch (approved):
Good merge and looks great!
Thanks :) merging!

Issue #140, opened 2020-12-26
CASE expression support


panarch:
https://www.sqlitetutorial.net/sqlite-case/

```sql
CASE case_expression
     WHEN when_expression_1 THEN result_1
     WHEN when_expression_2 THEN result_2
     ...
     [ ELSE result_else ] 
END
```

This is for evaluating `sqlparser::ast::Expr::Case`

KyGost:
Working on this

KyGost:
I take it back, maybe at some other time hahaha

KyGost:
@MRGRAVITY817 
If desired I have a working solution in https://github.com/SyRis-Consulting/multisql

panarch:
https://github.com/SyRis-Consulting/multisql/blob/0d4b7953add8509fae11dc259deca88e0ae23e27/src/executor/recipe/resolve.rs#L126-L148

@KyGost Sounds good to look around :)

Current status in this PR is also working well, we are now trying to make the code quality better and easier to read.
@MRGRAVITY817 has successfully removed `unwrap()` uses, and the only remaining stuff is eliminating `clone()`.
It would be also nice that you can also participate by sharing ideas here.

PR #330, opened 2021-08-27
Support conditional function `CASE`

MRGRAVITY817:
## Resolves issue #140 
I added some codes to support conditional functional `CASE`.
The `CASE` expression evaluates a list of conditions and returns an expression based on the result of the evaluation.
It is similar to the `IF-THEN-ELSE` statement in other programming languages.

### Basic structure of the `CASE` function
```sql
/* expressions in [ ] are empty-able */
CASE [ {case_expression} ]
  WHEN {condition_1} THEN {result_1}
  WHEN {condition_2} THEN {result_2}
  ...
  WHEN {condition_n} THEN {result_n}
  [ ELSE {else_result} ]
END;
```
- If `{case_expression}` is empty, `CASE` performs `searched_case`, otherwise `simple_case`.
- If `ELSE {else_result}` is empty, it will return `Null` when it meets the condition. 

### Use cases
1. Simple Case: Using case with `case_expression`. It will match with `condition` and return `result`. 
```sql
SELECT CASE id         /* CASE case_expression */
  WHEN 1 THEN "Tony"   /* WHEN condition THEN result */
  WHEN 2 THEN "Steve"
  WHEN 3 THEN "Natasha"
  ELSE "Bruce"        /* if nothing matches, default */
END;
```
2. Searched Case: Using case *without* `case_expression`. It will return `result` only for true `condition`. 
```sql
SELECT CASE                  /* CASE case_expression */
  WHEN id = 1 THEN "Tony"    /* WHEN condition THEN result */
  WHEN id = 2 THEN "Steve"
  WHEN id = 3 THEN "Natasha"
  ELSE "Bruce"               /* if nothing matches, default */
END;
```
### References
- Sqlite `CASE`: https://www.sqlitetutorial.net/sqlite-case/
- PostgreSQL `CASE`: https://www.postgresql.org/docs/current/functions-conditional.html#FUNCTIONS-CASE

### DIFF
--- a/src/executor/evaluate/error.rs
+++ b/src/executor/evaluate/error.rs
@@ -57,7 +57,4 @@ pub enum EvaluateError {
 
     #[error("negative substring length not allowed")]
     NegativeSubstrLenNotAllowed,
-
-    #[error("the result types should be equal: {0}")]
-    UnequalResultTypes(String),
 }


src/ast/expr.rs
```diff
@@ -50,4 +50,10 @@ pub enum Expr {
     Aggregate(Box<Aggregate>),
     Exists(Box<Query>),
     Subquery(Box<Query>),
+    Case {
+        operand: Option<Box<Expr>>,
+        conditions: Vec<Expr>,
+        results: Vec<Expr>,
```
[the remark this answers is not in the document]
MRGRAVITY817: I merged `conditions` and `results` to `when_then: Vec<(Expr, Expr)>`.
For some reason `tarpaulin` fails to cover `match` while it covers the branches inside of it, rest of the code that I added are fully covered. 
I am getting this PR out of the `draft`. Thanks for that advice again 😄

src/translate/expr.rs
```diff
@@ -113,7 +113,7 @@ pub fn translate_option_expr(sql_option_expr: &Option<Box<SqlExpr>>) -> Result<O
     match sql_option_expr {
         Some(expr) => match translate_expr(expr).map(Box::new) {
             Ok(expr) => Ok(Some(expr)),
-            Err(_) => Err(TranslateError::UnsupportedExpr(expr.to_string()).into()),
+            Err(_) => unreachable!(),
```
MRGRAVITY817, line 116: Here I change this part to unreachable, because `sqlparser-rs`'s `ParseError` will never let the program to reach that branch.

src/executor/evaluate/mod.rs
```diff
@@ -197,6 +197,26 @@ pub async fn evaluate<'a, T: 'static + Debug>(
...
+            }
+
+            let else_result = match else_result {
+                Some(result) => Some(eval(result).await?),
+                None => None,
+            };
+
+            match operand {
+                Some(o) => expr::simple_case(eval(o).await?, wt, else_result),
+                None => expr::searched_case(wt, else_result),
```
panarch, line 217: I checked that `simple_case` and `searched_case` are only being used in here, it means that you don't need to split logic codes into several different ones.
This can simply make code harder to read.

If the reason is that `Case` expression handler codes are too long, then you can split it by making private function like.. `evaluate_case` (ref. `evaluate_function`)
However, it looks just ok to have all in here.

panarch (changes requested):
This was an issue that a lot more to consider, but you really did well, thanks!
There are a few comments.
It looks that it's a good time to get familar with `StreamExt` and `TryStreamExt`.
https://docs.rs/futures/0.3.16/futures/stream/trait.StreamExt.html
https://docs.rs/futures/0.3.16/futures/stream/trait.TryStreamExt.html

src/executor/evaluate/mod.rs
```diff
@@ -197,6 +197,26 @@ pub async fn evaluate<'a, T: 'static + Debug>(
...
         }
+        Expr::Case {
+            operand,
+            when_then,
+            else_result,
+        } => {
+            let mut wt = Vec::new();
+            for (when, then) in when_then.iter() {
+                wt.push((eval(when).await?, eval(then).await?))
+            }
```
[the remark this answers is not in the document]
MRGRAVITY817: Oh, this was the code I was looking for 😂 
I tried not to use `mut` but didn't have the perfect solution... thanks for the advice!

src/executor/evaluate/mod.rs
MRGRAVITY817: Aha that makes sense... I'd better merge those into one `fn case()`. Thanks for feedback!

src/translate/expr.rs
```diff
@@ -79,10 +79,46 @@ pub fn translate_expr(sql_expr: &SqlExpr) -> Result<Expr> {
...
         SqlExpr::Exists(query) => translate_query(query).map(Box::new).map(Expr::Exists),
         SqlExpr::Subquery(query) => translate_query(query).map(Box::new).map(Expr::Subquery),
+        SqlExpr::Case {
+            operand,
+            conditions,
+            results,
+            else_result,
+        } => Ok(Expr::Case {
+            operand: translate_option_expr(operand)?,
+            when_then: translate_and_zip(conditions, results)?,
```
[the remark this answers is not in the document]
MRGRAVITY817: Oh, good point. I will remove `pub` and edit the function names right away :D

src/translate/expr.rs
```diff
@@ -79,10 +79,46 @@ pub fn translate_expr(sql_expr: &SqlExpr) -> Result<Expr> {
...
+        .zip(second.into_iter())
+        .collect::<Vec<_>>();
+    Ok(result)
+}
+
+pub fn translate_option_expr(sql_option_expr: &Option<Box<SqlExpr>>) -> Result<Option<Box<Expr>>> {
+    match sql_option_expr {
+        Some(expr) => match translate_expr(expr).map(Box::new) {
+            Ok(expr) => Ok(Some(expr)),
+            Err(_) => Err(TranslateError::UnreachableExpr("CASE".to_owned()).into()),
```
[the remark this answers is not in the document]
MRGRAVITY817: Oh, thanks. Now I've editted that part and coverage test passes! Thanks a lot

panarch (changes requested):
Now we can look around behavioral issues :)

```rust
(                                 
    r#"                           
    SELECT CASE id                
        WHEN 1 THEN name          
        WHEN 2 THEN 2             
        WHEN 3 THEN name          
        ELSE "Malfoy" END         
    AS case FROM Item;            
    "#,                           
    Ok(select_with_null!(         
        case;                     
        Str("Harry".to_owned());  
        I64(2);                   
        Str("Hermione".to_owned())
    )),                           
),                                
```

In `CASE`, every `result` expressions need to have a same type, but current executor allows this kind of test example.

MRGRAVITY817:
Gosh I totally misunderstood about `CASE` results! Gonna fix right away

MRGRAVITY817:
> In `CASE`, every `result` expressions need to have a same type, but current executor allows this kind of test example.

@panarch To achieve this, I think it would be nice to create a implemented function for `Value` enum
```rust
impl Value {
    pub fn get_type(&self) -> Result<DataType> {
        // Code that returns data type of a value
    }
}
```
I think, this might be useful for other future developments.
Tell if there's already any similar implementations, also whether this is an good 👍  or bad 👎  idea

panarch:
> > In `CASE`, every `result` expressions need to have a same type, but current executor allows this kind of test example.
> 
> @panarch To achieve this, I think it would be nice to create a implemented function for `Value` enum
> 
> ```rust
> impl Value {
>     pub fn get_type(&self) -> Result<DataType> {
>         // Code that returns data type of a value
>     }
> }
> ```
> 
> I think, this might be useful for other future developments.
> Tell if there's already any similar implementations, also whether this is an good 👍 or bad 👎 idea

I think that is not necessary to have `get_type` because `Value`'s type and `DataType` can be 1:1 matched except `Null`.
As you added `modulo` to `Evaluated`, `Value` and `Literal`, this looks like that we also need to have a new function for checking types between them.

src/executor/evaluate/expr.rs
```diff
@@ -87,3 +89,78 @@ pub fn between<'a>(
...
+                Value::Str(_) => Ok(6),
+                Value::Time(_) => Ok(7),
+                Value::Timestamp(_) => Ok(8),
+                Value::Null => Ok(9),
+            }
+        })
+        .collect::<Vec<_>>();
+
+    if result_types.iter().any(|t| t != &result_types[0]) {
+        Err(EvaluateError::UnequalResultTypes("CASE".to_owned()).into())
```
panarch, line 123: You can simplify the code by using `Iterator`.

1. `windows` + `all` or `any`
https://doc.rust-lang.org/std/primitive.slice.html#method.windows
2. `try_fold`

Many other approaches may work.
I'd recommend you to try the first option.

panarch: fyi. `matches!` - https://doc.rust-lang.org/std/macro.matches.html

MRGRAVITY817: Oh, I never knew about `windows` and `try_fold`. Thanks for help!

src/executor/evaluate/expr.rs
```diff
@@ -95,31 +97,22 @@ pub fn case<'a>(
...
-                Value::Str(_) => Ok(6),
-                Value::Time(_) => Ok(7),
-                Value::Timestamp(_) => Ok(8),
-                Value::Null => Ok(9),
-            }
-        })
-        .collect::<Vec<_>>();
+        .map(|r| -> Value { r.to_owned().try_into().unwrap() })
+        .tuple_windows()
+        .any(|(left, right)| disc(&left) != disc(&right));
```
MRGRAVITY817, line 113: @panarch 
I've simplified this part with a function called `std::mem::discriminant()`, which returns a variant of Enum. 
And I also used `tuple_windows()` which I found out while I was searching about the `windows()` function you recommended me to use on recent feedback. The basic functionality of those are simillar.
Thankfully, those aweful `Value::Variant(_) => Ok(i32)`s are gone now! 

Tell me if this is a good idea, or should I try another better method. Learning a lot from this issue! 😄

MRGRAVITY817: @panarch 
I've simplified this part with a function called `std::mem::discriminant()`, which returns a variant of Enum. 
And I also used `tuple_windows()` which I found out while I was searching about the `windows()` function you recommended me to use on recent feedback. The basic functionality of those are simillar.
Thankfully, those aweful `Value::Variant(_) => Ok(i32)`s are gone now! 

Tell me if this is a good idea, or should I try another better method. Learning a lot from this issue! 😄

panarch (changes requested):
Nice! Huge improvements to see :)

1. Merge `pub fn case` function into `evaluate`
Currently, `Case` handling codes are divided into two: `evaluate` and `case`.
However, there's no certain reason to divide a single logic into two different function.
`case` function is declared as `pub` but there's no other one which use this `case` function, too.

2. Remove `unwrap` uses.
Neither `unwrap()` or `Unreachable-` error is required in here to handle `CASE` statement.
Once after you've removed all `unwrap()` codes, then the codes will also become more simple and easier to read.

3. Make `case` test a bit more informative
Current test codes are also good enough in the view of system.
Let's make this little bit more informative.
`CASE` basically takes any `Expr`, not only for simple `Literal` or `Expr::Identifier`.
```rust
 r#"                             
 SELECT CASE id                  
     WHEN 1 THEN name            
     WHEN 2 THEN name            
     WHEN (1 + 2) THEN name      
     ELSE "Malfoy" || "hello" END
 AS case FROM Item;              
 "#,            
```
Whatever the `Expr` can work,
`"Malfoy" || "hello"` or `(1 + 2)`, etc.
Our integration tests basically have two roles; one is for the verifying systemic completeness and the other is for providing usage guide to our users.
So it would be good to let users to explicitly recognize that `CASE` support `Expr`.

src/executor/evaluate/expr.rs
```diff
@@ -1,3 +1,7 @@
+use itertools::Itertools;
+
+use crate::EvaluateError;
+
```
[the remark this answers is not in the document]
MRGRAVITY817: Affirmative 👍

MRGRAVITY817:
>     2. Remove `unwrap` uses.
>        Neither `unwrap()` or `Unreachable-` error is required in here to handle `CASE` statement.
>        Once after you've removed all `unwrap()` codes, then the codes will also become more simple and easier to read.

@panarch Hmm... I cannot think out a better solution to improve this part when removing `unwrap()`.
```rust
let type_diff = results
        .iter()
        .map(|r| -> Value { r.to_owned().try_into().unwrap() })  
        .tuple_windows()       
        .any(|(left, right)| disc(&left) != disc(&right));
```
Any suggestions?

panarch:
Ok I would recommend you to do more aggressive rearrangement of codes.
You can even remove unnecessary clone use.

src/executor/evaluate/mod.rs
```diff
@@ -197,6 +197,88 @@ pub async fn evaluate<'a, T: 'static + Debug>(
...
             Err(EvaluateError::UnreachableWildcardExpr.into())
         }
+        Expr::Case {
+            operand,
+            when_then,
+            else_result,
+        } => {
+            use super::EvaluateError;
+            use itertools::Itertools;
+            use std::mem::discriminant as disc;
```
panarch, line 207: Let's merge this into the top `use`.
If top `use` is too verbose, then that would be the proper timing for splitting file.
However, let's simply merge it for now.

src/executor/evaluate/mod.rs
```diff
@@ -197,6 +197,88 @@ pub async fn evaluate<'a, T: 'static + Debug>(
...
+                .unzip::<_, _, Vec<_>, Vec<_>>();
+
+            let results = match &else_result {
+                Some(er) => [thens, [er.to_owned()].to_vec()].concat(),
+                None => thens,
+            };
+
+            if !results
+                .into_iter()
+                .filter_map(|result| result.try_into().ok())
```
panarch, line 237: You should not ignore `Err`, this is even not an unreachable case.

src/executor/evaluate/mod.rs
```diff
@@ -197,6 +197,88 @@ pub async fn evaluate<'a, T: 'static + Debug>(
...
+                            .collect::<Vec<_>>();
+
+                        if thens.iter().any(|then| then.is_err()) {
+                            Err(EvaluateError::BooleanTypeRequired("CASE".to_owned()).into())
+                        } else {
+                            match thens
+                                .into_iter()
+                                .map(|then| match then {
+                                    Ok(t) => t,
+                                    _ => None,
```
panarch, line 267: This is the other case that `Err` is not handled.

src/executor/evaluate/mod.rs
```diff
@@ -197,6 +197,88 @@ pub async fn evaluate<'a, T: 'static + Debug>(
...
+                        let thens = when_then
+                            .into_iter()
+                            .map(|(when, then)| match when.try_into() {
+                                Ok(Value::Bool(condition)) => Ok(condition.then(|| then)),
+                                _ => Err(()),
+                            })
+                            .collect::<Vec<_>>();
+
+                        if thens.iter().any(|then| then.is_err()) {
+                            Err(EvaluateError::BooleanTypeRequired("CASE".to_owned()).into())
```
panarch, line 261: There's no reason to split `Err` handler into two separate codes.

In `collect`, you can use `.collect::<Result<Vec<_>>()?`

ref. https://doc.rust-lang.org/rust-by-example/error/iter_result.html

src/executor/evaluate/mod.rs
```diff
@@ -197,6 +197,88 @@ pub async fn evaluate<'a, T: 'static + Debug>(
...
+                                    _ => None,
+                                })
+                                .find(|then| then.is_some())
+                            {
+                                Some(Some(result)) => Ok(result),
+                                _ => match else_result {
+                                    Some(result) => Ok(result),
+                                    _ => Ok(Evaluated::from(Value::Null)),
+                                },
+                            }
```
panarch, line 276: You can simplify codes by using `flatten` and `unwrap_or` methods in `Option`.

https://doc.rust-lang.org/std/option/enum.Option.html#method.flatten

panarch (changes requested):
Great progress!
I found that there are two cases which convert `Err` to `Option`.
Let's fix those, then I'll provide more syntax sugar to simplify code logic.

src/executor/evaluate/mod.rs
MRGRAVITY817: Alrighty ✅

src/executor/evaluate/mod.rs
MRGRAVITY817: Aha, that's a good point. I'll fix it.

src/executor/evaluate/mod.rs
MRGRAVITY817: I'll try with your method! Thanks for the feedback 😃

src/executor/evaluate/mod.rs
```diff
@@ -197,6 +199,72 @@ pub async fn evaluate<'a, T: 'static + Debug>(
...
+                .into_iter()
+                .unzip::<_, _, Vec<_>, Vec<_>>();
+
+            let results = match &else_result {
+                Some(er) => [thens.clone(), [er.to_owned()].to_vec()].concat(),
+                None => thens.clone(),
+            }
+            .into_iter()
+            .map(|result| result.try_into())
+            .collect::<Result<Vec<Value>>>()?;
```
ever0de, line 234:
```suggestion
            let results = else_result
                .as_ref()
                .map(|er| [thens.clone(), vec![er.to_owned()]].concat())
                .unwrap_or_else(|| thens.clone())
                .into_iter()
                .map(TryInto::try_into)
                .collect::<Result<Vec<Value>>>()?;
```

src/executor/evaluate/mod.rs
```diff
@@ -197,6 +199,72 @@ pub async fn evaluate<'a, T: 'static + Debug>(
...
+
+            if !results
+                .iter()
+                .map(|result| discriminant(result))
+                .all_equal()
```
ever0de, line 239:
```suggestion
            if !results.iter().map(discriminant).all_equal()
```

src/executor/evaluate/mod.rs
```diff
@@ -197,6 +199,72 @@ pub async fn evaluate<'a, T: 'static + Debug>(
...
+            let operand = match operand {
+                Some(expr) => Some(eval(expr).await?),
+                None => None,
+            };
+
+            let when_then: Vec<(Evaluated, Evaluated)> = stream::iter(when_then.iter())
+                .map(Ok::<_, Error>)
+                .and_then(|(when, then)| async move { Ok((eval(when).await?, eval(then).await?)) })
+                .try_collect::<Vec<_>>()
+                .await?;
```
ever0de, line 216:
```suggestion
            let when_then: Vec<(Evaluated, Evaluated)> = stream::iter(when_then)
                .map(Ok::<_, Error>)
                .and_then(|(when, then)| async move { Ok((eval(when).await?, eval(then).await?)) })
                .try_collect()
                .await?;
```

src/executor/evaluate/mod.rs
```diff
@@ -197,6 +199,72 @@ pub async fn evaluate<'a, T: 'static + Debug>(
...
+                .map(|result| discriminant(result))
+                .all_equal()
+            {
+                Err(EvaluateError::UnequalResultTypes("CASE".to_owned()).into())
+            } else {
+                match operand {
+                    Some(op) => Ok(when_then.iter().find(|(when, _)| when.eq(&op)).map_or(
+                        else_result.unwrap_or_else(|| Evaluated::from(Value::Null)),
+                        |(_, then)| then.to_owned(),
+                    )),
```
ever0de, line 247:
```suggestion
                    Some(op) => Ok(when_then.into_iter().find(|(when, _)| when.eq(&op)).map_or(
                        else_result.unwrap_or_else(|| Evaluated::from(Value::Null)),
                        |(_, then)| then,
                    )),
```

ever0de (commented):
Refactor minor code!

src/executor/evaluate/mod.rs
```diff
@@ -197,6 +199,72 @@ pub async fn evaluate<'a, T: 'static + Debug>(
...
+            .collect::<Result<Vec<Value>>>()?;
+
+            if !results
+                .iter()
+                .map(|result| discriminant(result))
+                .all_equal()
+            {
+                Err(EvaluateError::UnequalResultTypes("CASE".to_owned()).into())
+            } else {
+                match operand {
```
panarch, line 243: `operand` is only being used in here, not above. So, you don't need to split `operand` handling codes into two different positions.

You can also use `find_map` here to merge `find` and `map_or`.

```rust
match operand {                                                 
    Some(op) => {                                               
        let op = eval(op).await?;                               
                                                                
        Ok(when_then                                            
            .into_iter()                                        
            .find_map(|(when, then)| when.eq(&op).then(|| then))
            .unwrap_or_else(|| Evaluated::from(Value::Null)))           
    }    
```

panarch (changes requested):
Thanks to @ever0de , comments are really helpful.

Now, we got a big step forward from prev commit, it's so nice that you removed all unnecessary `unwrap` calls 👍 
It looks that two steps are remaining.
1. Removing `clone`
2. Removing unnecessary `zip` and `unzip`.

If we achieve to remove all unnecessary `clone` calls, then I think `zip` & `unzip` will automatically become clear.

For now, it would be good to apply @ever0de comments first.
Then, let's try to solve remaining `clone` puzzle.
I'll also try and leave instant comment if I found something.

src/executor/evaluate/mod.rs
MRGRAVITY817: Lots of useful new things in here! thanks

MRGRAVITY817:
@panarch @ever0de Thanks for both of you!

src/executor/evaluate/mod.rs
MRGRAVITY817: Ahah thanks! `find_map` looks very useful :D

src/translate/expr.rs
```diff
@@ -79,10 +79,38 @@ pub fn translate_expr(sql_expr: &SqlExpr) -> Result<Expr> {
...
+        SqlExpr::Case {
+            operand,
+            conditions,
+            results,
+            else_result,
+        } => Ok(Expr::Case {
+            operand: translate_option_expr(operand)?,
+            when_then: translate_when_then(conditions, results)?,
+            else_result: translate_option_expr(else_result)?,
+        }),
```
panarch, line 91:
```rust
SqlExpr::Case {                                                 
    operand,                                                    
    conditions,                                                 
    results,                                                    
    else_result,                                                
} => Ok(Expr::Case {                                            
    operand: operand                                            
        .as_ref()                                               
        .map(|expr| translate_expr(expr.as_ref()).map(Box::new))
        .transpose()?,                                          
    when_then: conditions                                       
        .iter()                                                 
        .zip(results)                                           
        .map(|(when, then)| {                                   
            let when = translate_expr(when)?;                   
            let then = translate_expr(then)?;                   
                                                                
            Ok((when, then))                                    
        })                                                      
        .collect::<Result<Vec<_>>>()?,                          
    else_result: else_result                                    
        .as_ref()                                               
        .map(|expr| translate_expr(expr.as_ref()).map(Box::new))
        .transpose()?,                                          
}),                                                             
```

Here, `Case` might be too much for splitting into functions.
For you to concentrate on `evaluate`, here would be good to use.

panarch (changes requested):

src/translate/expr.rs
MRGRAVITY817: Thanks. This gives me the better point of view 🦅

ever0de:
Can you tell me the current status of this PR?
Can I do a review?

MRGRAVITY817:
> Can you tell me the current status of this PR?
> Can I do a review?

I am currently working on checking whether the `THEN` value from `WHEN THEN` is an `Evaluated::Literal` or `Evaluated::Value`. But I'd gladly have any helpful reviews 😄 since I've started this just now

src/executor/evaluate/error.rs
```diff
@@ -54,4 +54,7 @@ pub enum EvaluateError {
 
     #[error("negative substring length not allowed")]
     NegativeSubstrLenNotAllowed,
+
+    #[error("the result types should be equal: {0:?}")]
```
ever0de, line 58:
```suggestion
    #[error("the result types should be equal: {0}")]
```
`:?` is Debug. Since Value is String, how about using Display?

ever0de (commented):
Single comment

src/executor/evaluate/error.rs
MRGRAVITY817: Good point. Thanks for the feedback 👍

src/executor/evaluate/mod.rs
```diff
@@ -170,6 +172,74 @@ pub async fn evaluate<'a, T: 'static + Debug>(
...
+                    .map(|(_, t)| (discriminant(t)))
+                    .unique()
+                    .collect_vec();
+
+                if let Some(then_type_first) = then_type.first() {
+                    if then_type_first != &discriminant(&else_result) || then_type.len() != 1 {
+                        return Err(EvaluateError::UnequalResultTypes("CASE".to_owned()).into());
+                    }
+                }
+            }
```
MRGRAVITY817, line 218: This part checks `when` result type equality only when `else_result` exists, but it should check also when the `else_result` is `None`. Just a self-review for my mistake!

src/executor/evaluate/mod.rs
```diff
@@ -170,6 +172,100 @@ pub async fn evaluate<'a, T: 'static + Debug>(
...
+                }
+            };
+
+            Ok(when_then
+                .into_iter()
+                .find_map(|(when, then)| when.eq(&operand).then(|| then))
+                .unwrap_or_else(|| match else_result {
+                    Some(er) => er,
+                    None => Evaluated::from(Value::Null),
+                }))
```
panarch, line 267: `Literal` and `Value` can be mixed in `then`, so
```rust
let (when_then, else_result) = match when_then
    .iter()
    .map(|(_, then)| match then {
        Evaluated::Literal(l) => Some(l),
        _ => None,
    })
    .collect::<Option<Vec<_>>>()
```
this above looks unnecessary.

fyi.
 ```rust
 if !when_then                                 
     .iter()                                                    
     .map(|(_, then)| then.try_into().map(discriminant))        
     .all_equal()                                               
 {                                                              
     return Err(EvaluateError::UnequalResultTypes("CASE".to_owned()).into());
 }                                                          
                                                            
 for (when, then) in when_then.into_iter() {                
     if when == operand {                                   
         return Ok(then);                                   
     }                                                      
 }                                                                           
                                                           
 Ok(else_result.unwrap_or_else(|| Evaluated::from(Value::Null)))
```

This is still not complete code, but it passes the current test `case`.

MRGRAVITY817: > `Literal` and `Value` can be mixed in `then`, so
> 
> ```rust
> let (when_then, else_result) = match when_then
>     .iter()
>     .map(|(_, then)| match then {
>         Evaluated::Literal(l) => Some(l),
>         _ => None,
>     })
>     .collect::<Option<Vec<_>>>()
> ```
> 
> this above looks unnecessary.

They can be mixed, but then we have to use `try_into()` for every `then` values, no matter the enum variance. When every `then` values are `Evaluated::Literal` type, there's no need to use `try_into()` so I thought it would be better split the conditions as:
1. When all the `then` values are `Evaluated::Literal(l)` -> Get the variant value `l` and type check
2. When there's at least one `Evaluated::Value(v)` -> Squeeze the value out with `try_into()` and type check

Well, I'd be glad if using `try_into()` for every case isn't a bad idea. Because my logic here has produced bunch of ugly condition branched codes :< Please let me know if there's something wrong about my opinion 👍

panarch: Oh.. ok now I got why you wanted to handle `Literal` separately, thanks for sharing the details.
Then.. let's think about merge these `then` comparing logic into a single statement :)
How about using window function?
Rather than using `all_equal` in here, we can use `try_fold` for the alternative for window.
Or.. simply you can try

MRGRAVITY817: Thanks for the feedback! I'll try shrinking code a lot 📄

MRGRAVITY817:
@panarch Hooray!! Thanks for all advices. I've learned so much from this!

Issue #329, opened 2021-08-26
Add `PartialEq` and `PartialOrd` unit tests to `Literal`


panarch:
https://github.com/gluesql/gluesql/blob/23e0e875f307cf5d8187e70eac1a3765d6a3ea5e/src/data/literal.rs#L68-L103

There is no unit test for `PartialEq` and `PartialOrd` impl in `Literal`.
Let's add it :)

panarch:
ref. https://github.com/gluesql/gluesql/pull/332

PR #332, opened 2021-08-29
Feature test partial

zmrdltl:
Add unit test code `PartialEq`, `PartialOrd`
`PartialEq` - only same data type allowed to compare
`PartialOrd` - only same data type allowed to compare except number. number allowed to compare between
integer and float point

### DIFF
--- a/src/data/literal.rs
+++ b/src/data/literal.rs
@@ -434,4 +434,80 @@ mod tests {
         assert_eq!(PartialEq::eq(&itv!(123), &itv!(1234)), false);
         assert_eq!(PartialEq::eq(&itv!(123), &Null), false);
     }
+    #[test]
+    fn partial_ord() {
+        use crate::data::interval::Interval as I;
+        use std::cmp::Ordering;
+        macro_rules! text {
+            ($text: expr) => {
+                Text(Cow::Owned($text.to_owned()))
+            };
+        }
+        macro_rules! itv {
+            ($itv: expr) => {
+                Interval(I::Microsecond($itv))
+            };
+        }
+        macro_rules! num {
+            ($num: expr) => {
+                Number(Cow::Owned($num.to_owned()))
+            };
+        }
+        //Boolean
+        let result = &Boolean(false).partial_cmp(&Boolean(true));
+        assert_eq!(result, &Some(Ordering::Less));
+        let result = &Boolean(true).partial_cmp(&Boolean(true));
+        assert_eq!(result, &Some(Ordering::Equal));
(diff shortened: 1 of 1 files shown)


src/data/literal.rs
```diff
@@ -393,4 +394,164 @@ mod tests {
...
+        macro_rules! num {
+            ($num: expr) => {
+                Number(Cow::Owned($num.to_owned()))
+            };
+        }
+        //Boolean
+        assert_eq!(
+            &Boolean(PartialEq::eq(&Boolean(true), &Boolean(true))),
+            &Boolean(true)
+        );
```
panarch, line 419: You don't need to use `Literal::Boolean` and `PartialEq::eq`.
`Literal::Boolean` wrapping is unnecessary.
`PartialEq::eq` is called implicitly by `==` comparison.
Therefore, this would be enough.
```rust
assert_eq!(Boolean(true), Boolean(true))
```

You can apply this to the codes below.

src/data/literal.rs
```diff
@@ -393,4 +394,164 @@ mod tests {
...
+            };
+        }
+        macro_rules! num {
+            ($num: expr) => {
+                Number(Cow::Owned($num.to_owned()))
+            };
+        }
+        //Boolean
+        let result = &Boolean(false).partial_cmp(&Boolean(true));
+        assert_eq!(result, &Some(Ordering::Less));
```
panarch, line 502:
```rust
assert_eq!(Boolean(false).partial_cmp(&Boolean(true)), Some(Ordering::Less));
```
`&` is unnecessary, you can simply compare values.

panarch (changes requested):
Looks good an that's also great that you carefully checked int and float in `partial_cmp` tests.
There's a tiny request about simplifying codes.
Thanks!

src/data/literal.rs
zmrdltl: image
i don't know how to resolve this. Without casting to Boolean, i got this error

panarch: You can simply follow `clippy`'s suggestion.

```rust
assert_eq!(Boolean(true), Boolean(true));
```
or
```rust
assert!(Boolean(true) == Boolean(true));
```

This clippy error is because you tried using `assert_eq!` macro for comparing literal boolean values.

zmrdltl: Finally i understood!! thanks

panarch (approved):
Awesome, thanks!

resolve https://github.com/gluesql/gluesql/issues/329

PR #335, opened 2021-09-02
CAST from string to INTERVAL

devgony:
### Case1. Interval literal
```rs
// src/data/value/literal.rs
(DataType::Interval, Literal::Text(v)) => {
    Ok(Value::Interval(Interval::try_from(v.as_str())?))
}
```

### Case2. Interval value
```rs
// src/data/value/mod.rs
(DataType::Interval, value) => value.try_into().map(Value::Interval),
```
```rs
// src/data/value/into.rs
impl TryInto<Interval> for &Value {
    type Error = Error;

    fn try_into(self) -> Result<Interval> {
        Ok(match self {
            Value::Str(value) => Interval::try_from(value.as_str())?,
            _ => return Err(ValueError::ImpossibleCast.into()),
        })
    }
}
```

### DIFF
--- a/src/data/value/literal.rs
+++ b/src/data/value/literal.rs
@@ -2,7 +2,7 @@ use {
     super::{error::ValueError, Value},
     crate::{
         ast::DataType,
-        data::Literal,
+        data::{Interval, Literal},
         result::{Error, Result},
     },
     chrono::{offset::Utc, DateTime, NaiveDate, NaiveDateTime, NaiveTime},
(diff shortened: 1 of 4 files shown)


src/data/value/literal.rs
```diff
@@ -198,6 +200,9 @@ impl Value {
 
                 Ok(Value::Str(v.to_owned()))
             }
+            (DataType::Interval, Literal::Text(v)) => {
+                Ok(Value::Interval(Interval::try_from(v.as_str())?))
```
panarch, line 204: It would be good to use `map` in `Result`.

```rust
Interval::try_from(v.as_str()).map(Value::Interval)
```

panarch (changes requested):
Thanks! There's nothing to comment about the behavior.
A few requests here are just code style tweaks.
Looks great, integration tests also look very informative, thanks 👍

devgony:
> Thanks! There's nothing to comment about the behavior.
> A few requests here are just code style tweaks.
> Looks great, integration tests also look very informative, thanks 👍

Thank you for reviewing and advising nice code style!
I just fixed and re-pushed it!

panarch (approved):
Awesome, merging!

PR #339, opened 2021-09-05
Raise error when column option `PRIMARY KEY` is used.

### LINKED ISSUE #308: `PRIMARY KEY` not supported error is required
`PRIMARY KEY` is not supported yet.
However GlueSQL allows `CREATE TABLE SingleItem (id INTEGER PRIMARY KEY)` but do nothing for `PRIMARY KEY` attr.
Until the feature is implemented, it should be an error.

It is expected to play with codes in https://github.com/gluesql/gluesql/tree/main/src/executor/alter

heka1024:
Motivation:
- `PRIMARY KEY` is not supported in `GlueSQL`. However `GlueSQL` allows `PRIMARY KEY` attr. Until the feature is implemented, it should be an error.

Modification:
- Modify pattern matching of `translate_column_option_def`
- Add TC about primary key

Result:
- Raise error when user try to make column with `PRIMARY KEY`
- Closes #308

### DIFF
--- a/src/tests/alter/create_table.rs
+++ b/src/tests/alter/create_table.rs
@@ -49,6 +49,10 @@ test_case!(create_table, async move {
             Err(TranslateError::UnsupportedColumnOption("CHECK (true)".to_owned()).into()),
         ),
         (
+            "CREATE TABLE Glue (id INTEGER PRIMARY KEY)",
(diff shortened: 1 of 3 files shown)


panarch (approved):
Awesome :) Thanks a lot, merging!

Issue #293, opened 2021-08-17
Implement Function: SUBSTR


hyewoni:
## Implement Function: SUBSTR

`substr ( string text, start integer [, count integer ] ) → text
`

Extracts the substring of string starting at the start'th character, and extending for count characters if that is specified. (Same as substring(string from start for count).

## Examples
```
postgres=# select substr('ABC', 0, 3);
 substr 
--------
 AB
(1 row)

postgres=# select substr('ABC', 1, 3);
 substr 
--------
 ABC
(1 row)

postgres=# select substr('ABC', -1, 3);
 substr 
--------
 A
(1 row)

postgres=# select substr('ABC', -1, 4);
 substr 
--------
 AB
(1 row)

```

## Reference
https://www.postgresql.org/docs/13/functions-string.html
https://github.com/postgres/postgres/blob/18bac60ede44359a1e577df80aef196e371c902e/src/backend/utils/adt/varlena.c#L872

PR #341, opened 2021-09-07
Implement substr function

vbbono:
## Resolve https://github.com/gluesql/gluesql/issues/293

substr ( string text, start integer [, count integer ] ) → text

Extracts the substring of string starting at the start'th character, and extending for count characters if that is specified. (Same as substring(string from start for count).

### DIFF
--- a/src/executor/evaluate/mod.rs
+++ b/src/executor/evaluate/mod.rs
@@ -601,6 +601,45 @@ async fn evaluate_function<'a, T: 'static + Debug>(
             }
         }
         .map(Evaluated::from),
+
+        Function::Substr { expr, start, count } => {
+            let string = match eval_to_str(expr).await? {
+                Nullable::Value(v) => v,
+                Nullable::Null => {
+                    return Ok(Evaluated::from(Value::Null));
+                }
+            };
+            let start = match eval_to_integer(start).await? {
+                Nullable::Value(v) => v,
+                Nullable::Null => return Ok(Evaluated::from(Value::Null))
+            };
+
+            let count = match count {
+                Some(expr) => match eval_to_integer(expr).await? {
+                    Nullable::Value(v) => match v {
+                        x if x < 0 => return Err(EvaluateError::NegativeSubstrLenNotAllowed.into()),
+                        _ => v,
+                    },
+                    Nullable::Null => -1,
+                },
+                None => -1,
+            };
+
+            let s: i64 = if start <= 0 { 0 } else { start - 1 };
+            let e: i64 = start + count - 1;
+            let s = s as usize;
+            let e = if count == -1 || e > string.len() as i64 { string.len() } else { e as usize };
+
+            let result = if count == 0 || s >= string.len() {
+                String::from("")
+            } else if count == -1 {
+                String::from(&string[s..])
+            } else {
+                String::from(&string[s..e])
+            };
+
+            Ok(Evaluated::from(Value::Str(result)))
+        }
     }
 }
 
--- a/src/executor/evaluate/mod.rs
+++ b/src/executor/evaluate/mod.rs
@@ -611,7 +611,7 @@ async fn evaluate_function<'a, T: 'static + Debug>(
             };
             let start = match eval_to_integer(start).await? {
                 Nullable::Value(v) => v,
-                Nullable::Null => return Ok(Evaluated::from(Value::Null))
+                Nullable::Null => return Ok(Evaluated::from(Value::Null)),
             };
 
             let count = match count {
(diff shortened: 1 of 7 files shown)


src/translate/function.rs
```diff
@@ -254,6 +254,23 @@ pub fn translate_function(sql_function: &SqlFunction) -> Result<Expr> {
...
+        "SUBSTR" => {
+            check_len_range(name, args.len(), 2, 3)?;
+
+            let expr = translate_expr(args[0])?;
+            let start = translate_expr(args[1])?;
+            let count = if args.len() == 2 {
+                None
+            } else {
+                Some(translate_expr(args[2])?)
+            };
```
panarch, line 266: fyi. Let's get familiar with using methods provided by `bool` and `Option` :)
https://doc.rust-lang.org/std/primitive.bool.html#method.then
https://doc.rust-lang.org/std/option/enum.Option.html#method.transpose

```rust
let count = (args.len() > 2)         
    .then(|| translate_expr(args[2]))
    .transpose()?;
```

src/executor/evaluate/mod.rs
```diff
@@ -601,6 +601,49 @@ async fn evaluate_function<'a, T: 'static + Debug>(
...
+            let count = match count {
+                Some(expr) => match eval_to_integer(expr).await? {
+                    Nullable::Value(v) => match v {
+                        x if x < 0 => return Err(EvaluateError::NegativeSubstrLenNotAllowed.into()),
+                        _ => v,
+                    },
+                    Nullable::Null => -1,
+                },
+                None => -1,
+            };
```
panarch, line 626: Current `count` variable has two independent roles
1. whether `count` exists or not
2. number of chars to substr.

`ast` already properly provides as a form of `Option<Expr>` and you'll be able to convert it into `Option<usize>`.
Then, you can handle `Option<usize>` directly without merging two info into `i64` type.

panarch (changes requested):
Looks really good!
Let's make it a bit more Rust's way :)
Then, current codes will become even more simple to read.

src/translate/function.rs
vbbono: Thanks for the information.
Edited as you said.

src/executor/evaluate/mod.rs
vbbono: I don't know if it's okay to change type of `count` to` Option<usize>`,, :joy_cat: 
Because, to calculate the variable `e`, `count` may have to be added to the `start`, which causes type casting.
Is there any simpler way? XD

src/executor/evaluate/mod.rs
```diff
@@ -601,6 +601,52 @@ async fn evaluate_function<'a, T: 'static + Debug>(
...
+                    }
+                }
+                None => string.len(),
+            };
+
+            let result = if s >= string.len() {
+                String::from("")
+            } else {
+                String::from(&string[s..e])
+            };
```
panarch, line 646: Ok, here type casting to `usize` looks inevitable.
However, we can simplify the codes by doing a bit of rearrangement.
`std::cmp::{min, max}` would be useful.

```rust                                                                    
let start = match eval_to_integer(start).await? {                             
    Nullable::Value(v) => v - 1,                                              
    Nullable::Null => {                                                       
        return Ok(Evaluated::from(Value::Null));                              
    }                                                                         
};                                                                            
                                                                              
let end = match count {                                                       
    Some(expr) => match eval_to_integer(expr).await? {                        
        Nullable::Value(count) => {                                           
            if count < 0 {                                                    
                return Err(EvaluateError::NegativeSubstrLenNotAllowed.into());
            }                                                                 
                                                                              
            min(max(start + count, 0) as usize, string.len())                 
        }                                                                     
        Nullable::Null => {                                                   
            return Ok(Evaluated::from(Value::Null));                          
        }                                                                     
    },                                                                        
    None => string.len(),                                                     
};                                                                            
                                                                              
let start = min(max(start, 0) as usize, string.len());                        
let string = String::from(&string[start..end]);                               
```

panarch: I mainly focused on two points.
1. handling `start` codes were splitted into two.
```rust
let s: usize = if start <= 0 { 0 } else { (start - 1) as usize };
```
```rust
let result = if s >= string.len() {
    String::from("")
} else {
    String::from(&string[s..e])
};
```

2. Also the `count` handling match codes
```rust
let count = match count {
    Some(expr) => match eval_to_integer(expr).await? {
         ..
```
```rust
let e = match count {
    Some(v) => {
        if (start - 1 + v) < 0 {
    ..
```

panarch (changes requested):
Now it's really close to the merge, thanks!

src/tests/function/substr.rs
```diff
@@ -0,0 +1,151 @@
...
+                "AB".to_owned()
+            )),
+        ),
+        (
+            r#"SELECT SUBSTR("ABC", -1, NULL) AS test FROM SingleItem"#,
+            Ok(select!(
+                "test"
+                Str;
+                "ABC".to_owned()
+            )),
```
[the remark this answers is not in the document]
vbbono: Thanks I'll remove tha tc !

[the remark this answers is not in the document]
vbbono: Thanks! I'll change TC to return NULL!

vbbono:
I think this PR is ready for re-review :rofl: Please take a look. Thanks !

PR #347, opened 2021-09-12
Simplify evaluate module codes by replacing from take(1).. to try_next()

panarch:
In `executor/evaluate/mod.rs`,
* Replace `take(1)` .. `collect` codes to `try_next`
* Remove redundant `negated` uses


ever0de (approved):
Awesome!

PR #352, opened 2021-09-15
Repeatfunction

inhwa1025:


src/tests/function/repeat.rs
```diff
@@ -0,0 +1,59 @@
+use crate::*;
+
+test_case!(repeat, async move {
+    //use Value::{Null, Str, I64};
+
```
panarch, line 5: You can remove this comment :)

src/executor/evaluate/mod.rs
```diff
@@ -573,6 +573,22 @@ async fn evaluate_function<'a, T: 'static + Debug>(
...
+                Nullable::Value(v) => v,
+                Nullable::Null => {
+                    return Ok(Evaluated::from(Value::Null));
+                }
+            };
+            let num = match eval_to_integer(num).await? {
+                Nullable::Value(v) => v,
+                Nullable::Null => {
+                    return Ok(Evaluated::from(Value::Null));
+                }
```
panarch, line 587: Let's just have a single more test case.
Precisely, this `Nullable::Null` is not tested by current `repeat` integration test

panarch (changes requested):
Looks really good!
I left just tiny comments here.
Let's merge this into the main soon 👍

PR #353, opened 2021-09-16
[Fix, Update] - Partial_Eq error, unit test

zmrdltl:
Fix bug Partial_Eq function compensated for the shortcomings in comparing integers and floating point.
ex) `12.0 != 12`

Update unit test comparing between float and int


panarch (approved):
Thanks :) merging 👍

PR #355, opened 2021-09-16
Add MemoryStorage integration tests to reach all storage impl codes

panarch:


src/storages/memory_storage/mod.rs
```diff
@@ -38,8 +38,7 @@ impl Store<Key> for MemoryStorage {
     async fn fetch_schema(&self, table_name: &str) -> Result<Option<Schema>> {
         self.items
             .get(table_name)
-            .map(|item| item.schema.clone())
-            .map(Ok)
+            .map(|item| Ok(item.schema.clone()))
```
ever0de, line 41: Q: Is there any reason to change this?

tests/memory_storage.rs
```diff
@@ -24,3 +24,98 @@ impl Tester<Key, MemoryStorage> for MemoryTester {
...
+            .as_ref()
+            .map(Vec::len),
+        Ok(0),
+    );
+
+    assert_eq!(
+        block_on(storage.scan_indexed_data("Idx", "hello", None, None)).map(|_| ()),
+        Err(Error::StorageMsg(
+            "[MemoryStorage] index is not supported".to_owned()
+        ))
```
ever0de, line 93:
```suggestion
        block_on(storage.scan_indexed_data("Idx", "hello", None, None)).unwrap_err(),
        Error::StorageMsg(
            "[MemoryStorage] index is not supported".to_owned()
        )
```
how about this change?

ever0de: If you change it, you have to change the other conditions as well.

ever0de (commented):
Minor comment (nitpicking)

src/storages/memory_storage/mod.rs
panarch: Nothing special reason, `cargo-tarpaulin` marks two `map` as not reached codes, so I just merged two independent lines into an one.

tests/memory_storage.rs
panarch: The content `Ok` of `storage.scan_indexed_data` is `Iterator` and it does not implement `Debug`.
`unwrap_err` requires `Debug` impl for the `Ok(T)`, so I simply replaced `Iterator` to `()`.

ever0de: oh, need a debug implementation
It's good to leave it like this.

ever0de (approved):
No other changes seem necessary

PR #356, opened 2021-09-19
Handle do not match columns and values

### LINKED ISSUE #189: Throw on too many values
I've just noticed that this doesn't error, that's a problem:
```sql
INSERT INTO TestUnique (name) VALUES ('test1'), ('test2', 3)
```
Instead it acts like:
```sql
INSERT INTO TestUnique (name) VALUES ('test1'), ('test2')
```

ever0de:
fix #189

### DIFF
--- a/src/data/row.rs
+++ b/src/data/row.rs
@@ -15,8 +15,8 @@ pub enum RowError {
     #[error("lack of required column: {0}")]
     LackOfRequiredColumn(String),
 
-    #[error("literals does not fit to columns")]
-    LackOfRequiredValue(String),
+    #[error("do not match columns and values")]
+    DoNotMatchColumnsAndValues,
 
     #[error("literals have more values than target columns")]
     TooManyValues,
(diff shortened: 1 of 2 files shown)


src/data/row.rs
```diff
@@ -41,34 +41,40 @@ impl Row {
     }
 
     pub fn new(column_defs: &[ColumnDef], columns: &[String], values: &[Expr]) -> Result<Self> {
+        if !columns.is_empty() && values.len() != columns.len() {
+            return Err(RowError::DoNotMatchColumnsAndValues.into());
+        }
         if values.len() > column_defs.len() {
```
panarch, line 47: It would be good to use `else if`

src/data/row.rs
```diff
@@ -41,34 +41,40 @@ impl Row {
...
             return Err(RowError::TooManyValues.into());
         }
 
+        let columns: Box<dyn Iterator<Item = &String>> = if columns.is_empty() {
+            Box::new(column_defs.iter().map(|ColumnDef { name, .. }| name))
+        } else {
+            Box::new(columns.iter())
+        };
```
panarch, line 55: `iter_enum::Iterator` would be quite a good choice to take it here.

```rust
#[derive(Iterator)]                                                    
enum Columns<I1, I2> {                                                 
    All(I1),                                                           
    Specified(I2),                                                     
}                                                                      
                                                                       
let columns = if columns.is_empty() {                                  
    Columns::All(column_defs.iter().map(|ColumnDef { name, .. }| name))
} else {                                                               
    Columns::Specified(columns.iter())                                 
};                                                                     
```

src/data/row.rs
```diff
@@ -15,8 +15,8 @@ pub enum RowError {
     #[error("lack of required column: {0}")]
     LackOfRequiredColumn(String),
 
-    #[error("literals does not fit to columns")]
-    LackOfRequiredValue(String),
+    #[error("do not match columns and values")]
+    DoNotMatchColumnsAndValues,
```
panarch, line 19: Error name or message is for describing what's wrong happened. so, imperative sentence might not be a good choice.
How about simply `ColumnAndValuesNotMatched`?

panarch (changes requested):
Looks really nice, thanks!
Just a few tiny comments.

src/data/row.rs
ever0de: A very nice change! thank you.

src/data/row.rs
ever0de: That's a good comment. changed it

src/data/row.rs
ever0de: It seems like a good change. I'll fix it

src/data/row.rs
```diff
@@ -25,6 +25,13 @@ pub enum RowError {
     ConflictOnEmptyRow,
 }
 
+#[derive(iter_enum::Iterator)]
+
+enum Columns<I1, I2> {
```
panarch, line 30: Let's just remove this blank line, then all looks good!

ever0de: Uh.. why did this go in? will do fixup

panarch (approved):
Thanks! merging 👍

PR #359, opened 2021-09-19
Support binary operator `XOR`, update data-type `Int`

zmrdltl:
1. Adding a `XOR` operator to the `Sqlparser` and then adding a function to the current project. Add unit test code to determine `XOR` of two Boolean values.
2. Modify data-type `Int` factors according to the updated data-type from `Sqlparser`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,7 @@ iter-enum = "1"
 itertools = "0.10"
 pin-project = "1.0"
 serde = { version = "1", features = ["derive"] }
-sqlparser = { version = "0.10", features = ["serde"] }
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -57,7 +57,7 @@ iter-enum = "1"
 itertools = "0.10"
 pin-project = "1.0"
 serde = { version = "1", features = ["derive"] }
-sqlparser = { version = "0.10", features = ["serde"] }
+sqlparser = { git = "https://github.com/sqlparser-rs/sqlparser-rs/", rev = "014b82f03d82d6f36b5271376e33461002709a55" }
```
panarch, line 60: Now we no more need `serde` in `sqlparser`, I wasn't aware of it.
Thanks that you removed it.

panarch (approved):
Awesome!
Thanks a lot 👍

PR #360, opened 2021-09-20
add uuid data type

maruoovv:
- for adding uuid type, add uuid crate dependency
 - uuid data format support simple, hyphenated, urn type
 - https://crates.io/crates/uuid

### DIFF
--- a/src/data/literal.rs
+++ b/src/data/literal.rs
@@ -47,7 +47,10 @@ impl<'a> TryFrom<&'a AstLiteral> for Literal<'a> {
         let literal = match ast_literal {
             AstLiteral::Boolean(v) => Boolean(*v),
             AstLiteral::Number(v) => Number(Cow::Borrowed(v)),
-            AstLiteral::QuotedString(v) => Text(Cow::Borrowed(v)),
+            AstLiteral::QuotedString(v) => {
+                println!("AstLiteral::QuotedString : {}", v);
+                Text(Cow::Borrowed(v))
+            }
             AstLiteral::Interval {
                 value,
                 leading_field,
--- a/src/data/literal.rs
+++ b/src/data/literal.rs
@@ -47,10 +47,7 @@ impl<'a> TryFrom<&'a AstLiteral> for Literal<'a> {
         let literal = match ast_literal {
             AstLiteral::Boolean(v) => Boolean(*v),
             AstLiteral::Number(v) => Number(Cow::Borrowed(v)),
-            AstLiteral::QuotedString(v) => {
-                println!("AstLiteral::QuotedString : {}", v);
-                Text(Cow::Borrowed(v))
-            }
+            AstLiteral::QuotedString(v) => Text(Cow::Borrowed(v)),
             AstLiteral::Interval {
                 value,
                 leading_field,
--- a/src/data/value/literal.rs
+++ b/src/data/value/literal.rs
@@ -1,4 +1,3 @@
-use uuid::Uuid;
 use {
     super::{error::ValueError, Value},
     crate::{
@@ -8,6 +7,7 @@ use {
     },
     chrono::{offset::Utc, DateTime, NaiveDate, NaiveDateTime, NaiveTime},
     std::{cmp::Ordering, convert::TryFrom},
+    uuid::Uuid,
 };
 
 impl PartialEq<Literal<'_>> for Value {
@@ -42,10 +42,7 @@ impl PartialEq<Literal<'_>> for Value {
                 Err(_) => false,
             },
             (Value::Interval(l), Literal::Interval(r)) => l == r,
-            (Value::UUID(l), Literal::Text(r)) => match parse_uuid(r) {
-                Ok(r) => l == &r,
-                Err(_) => false,
-            },
+            (Value::UUID(l), Literal::Text(r)) => parse_uuid(r).map(|r| l == &r).unwrap_or(false),
             _ => false,
         }
     }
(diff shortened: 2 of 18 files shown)


src/data/literal.rs
```diff
@@ -47,7 +47,10 @@ impl<'a> TryFrom<&'a AstLiteral> for Literal<'a> {
         let literal = match ast_literal {
             AstLiteral::Boolean(v) => Boolean(*v),
             AstLiteral::Number(v) => Number(Cow::Borrowed(v)),
-            AstLiteral::QuotedString(v) => Text(Cow::Borrowed(v)),
+            AstLiteral::QuotedString(v) => {
+                println!("AstLiteral::QuotedString : {}", v);
```
ever0de, line 51: i think i should delete it

src/data/value/literal.rs
```diff
@@ -1,3 +1,4 @@
+use uuid::Uuid;
 use {
     super::{error::ValueError, Value},
     crate::{
```
ever0de, line 4: I think you can include it in use `{..}`

```suggestion
use {
    uuid::Uuid;
    super::{error::ValueError, Value},
    crate::{
```

src/data/value/literal.rs
```diff
@@ -265,6 +281,13 @@ fn parse_time(v: &str) -> Result<NaiveTime> {
     Err(ValueError::FailedToParseTime(v).into())
 }
 
+fn parse_uuid(v: &str) -> Result<u128> {
+    match Uuid::parse_str(v) {
+        Ok(u) => Ok(u.as_u128()),
+        _ => Err(ValueError::FailedToParseUUID(v.to_owned()).into()),
```
ever0de, line 287: This error type is always crushed.

```rust
        _ => Err(ValueError::FailedToParseUUID(v.to_owned()).into()),
```

src/data/value/literal.rs
```diff
@@ -41,6 +42,10 @@ impl PartialEq<Literal<'_>> for Value {
                 Err(_) => false,
             },
             (Value::Interval(l), Literal::Interval(r)) => l == r,
+            (Value::UUID(l), Literal::Text(r)) => match parse_uuid(r) {
+                Ok(r) => l == &r,
+                Err(_) => false,
+            },
```
ever0de, line 48: This is a comment on `parse_uuid`.
How about improving it this way?

```suggestion
            (Value::UUID(l), Literal::Text(r)) => Uuid::parse_str(v).map(Uuid::as_u128).map(|r| l == &r).unwrap_or(false)
```

src/data/value/literal.rs
```diff
@@ -77,6 +82,10 @@ impl PartialOrd<Literal<'_>> for Value {
                 Err(_) => None,
             },
             (Value::Interval(l), Literal::Interval(r)) => l.partial_cmp(r),
+            (Value::UUID(l), Literal::Text(r)) => match parse_uuid(r) {
+                Ok(r) => l.partial_cmp(&r),
+                Err(_) => None,
+            },
```
ever0de, line 88: I think you can change this to something similar to the above

ever0de (commented):
I left a few minor comments.

src/data/value/literal.rs
maruoovv: I don't understand this comment, can you explain more detail?

ever0de: Ummm... it seems deprecated
For example, if an error occurs, the function is being used and changed to a different default value

ever0de: ex) For example, chang from error type to default in this way
```rust
match parse_uuid(r) {
                Ok(r) => l == &r,
                Err(_) => false,
}
```

Q: Should it default to when parsing fails? Or do you just want to spit out an error?

maruoovv: This function is not deprecated and still use in here.
[PartialEq](https://github.com/gluesql/gluesql/pull/360/files#diff-06ff4b2f6cc1883c8536124b289f4e2beb48ceae7a80ce995f036000fa7a2229R45
)
[PartialOrd](https://github.com/gluesql/gluesql/pull/360/files#diff-06ff4b2f6cc1883c8536124b289f4e2beb48ceae7a80ce995f036000fa7a2229R82)

I think this function spit out an error when parsing fails, and other clients using this function decide how to handle this error. ex) also spit out error, using default value...

So, like the above links, i used this function that handle errors in the clients.
Is this anti pattern in rust?

ever0de: Already in `Uuid::parse_str` there is an error type for parsing failure.
I don't think it's necessary to crush this type twice.
For example in Value
```rust
            (DataType::UUID, Literal::Text(v)) => parse_uuid(v)
                .map(Value::UUID)
                .map_err(|_| ValueError::FailedToParseUUID(v.to_string()).into()),
```
This code encounters a type called `FailedToParseUUID` twice. (in parse_uuid, now)
I don't quite understand this part

It would be nice to get help from @panarch
I will request a review.

maruoovv: The reason I created this function is because our UUID data type is u128, so I wanted to collect the string parsing and converting logic in one place rather than in several places. (But i missed several places using Uuid::parse_str in tests)

But I agree your comments It is unnecessary wrapping function and crush error twice.
If @panarch agree this function unnecessary, I will change this.

src/data/value/mod.rs
```diff
@@ -130,6 +136,7 @@ impl Value {
             (DataType::Date, value) => value.try_into().map(Value::Date),
             (DataType::Timestamp, value) => value.try_into().map(Value::Timestamp),
             (DataType::Interval, value) => value.try_into().map(Value::Interval),
+            (DataType::UUID, value) => value.try_into().map(Value::UUID),
```
panarch, line 139: let's add an unit test to the below. There is an `cast` unit test.

src/tests/data_type/uuid.rs
```diff
@@ -0,0 +1,73 @@
+use {
+    crate::{ast::DataType, data::Value},
+    std::borrow::Cow,
+    uuid::Uuid,
+};
+
+use crate::*;
```
panarch, line 7: You can also merge this into a single `use`.

panarch (changes requested):
Sorry for the late review.
Most of them looks really good enough. I just left some tiny comments on this.
Thanks!

panarch (approved):
Awesome 👍 👍 👍 
Merging!

PR #368, opened 2021-10-02
Support aggregation `AVG`

zmrdltl:
# Changes
#### Added code to test for columns with and without null, respectively.


src/executor/aggregate/state.rs
```diff
@@ -90,13 +89,20 @@ impl<'a> State<'a> {
...
-            .map(|(k, (_, v))| (k, v))
+            .map(|(k, v)| (k, v))
             .chunks(size)
             .into_iter()
             .enumerate()
             .map(|(i, entries)| {
                 let aggregated = entries
-                    .map(|((_, aggr), value)| (aggr, value))
+                    .map(|((_, aggr), v)| {
+                        //v.2 : exists to store the intermediate value of the avg function.
```
ever0de, line 99: Format comment 

```suggestion
                        // v.2 : exists to store the intermediate value of the avg function.
```
ref: <https://doc.rust-lang.org/reference/comments.html>

ever0de: Resolved commit: <506f34fc51e77e060450984253e7244c76d4cf8f>

zmrdltl:
only cargo build success. test is failed. just for the structure of mod.rs, state.rs

zmrdltl:
Refactored mod.rs, state.rs structure with private enum AggrValue. 
Thanks for the help with systax and interfaces @panarch.

panarch (approved):
Awesome!!
All looks good, merging 👍 👍 👍

PR #370, opened 2021-10-03
Add LIST data type,

panarch:
User can access internal data by `UNWRAP` function like `MAP` data type.
e.g. `UNWRAP(column, "0")`, `UNWRAP(column, "3.foo.bar.0")`
Like `MAP`, `LIST` also only accepts JSON string literal for `INSERT`.


src/data/value/selector.rs
```diff
@@ -0,0 +1,48 @@
...
+
+        let value = match result {
+            Continue(Selectable::Map(map)) => Value::Map(map.clone()),
+            Continue(Selectable::List(list)) => Value::List(list.clone()),
+            Continue(Selectable::Other(value)) => value.clone(),
+            Break(_) => Value::Null,
+        };
+
+        Ok(value)
+    }
```
panarch, line 47: @ever0de @24seconds 
I succeeded to remove `is_last`!
It would be fun to see how it becomes clear compared to the previous one.

https://github.com/gluesql/gluesql/blob/9ee565e6cff606047870f8ee3201e042008e07cf/src/data/map.rs#L38-L64

src/data/value/json.rs
```diff
@@ -0,0 +1,57 @@
...
+    super::{Value, ValueError},
+    crate::result::Result,
+    serde_json::Value as JsonValue,
+    std::collections::HashMap,
+};
+
+impl Value {
+    pub fn parse_json_map(value: &str) -> Result<Value> {
+        let value: JsonValue =
+            serde_json::from_str(value).map_err(|_| ValueError::InvalidJsonString)?;
```
ever0de, line 11: No problem, but you can exclude the type specification by replacing from_str with [from_value](https://docs.rs/serde_json/1.0.68/serde_json/fn.from_value.html)

src/data/value/json.rs
```diff
@@ -0,0 +1,57 @@
...
+        if !matches!(value, JsonValue::Object(_)) {
+            return Err(ValueError::JsonObjectTypeRequired.into());
+        }
+
+        parse_json_value(value)
+    }
+
+    pub fn parse_json_list(value: &str) -> Result<Value> {
+        let value: JsonValue =
+            serde_json::from_str(value).map_err(|_| ValueError::InvalidJsonString)?;
```
ever0de, line 22: same as above

src/data/value/selector.rs
ever0de: This is really neat. admire.

src/data/value/selector.rs
```diff
@@ -0,0 +1,48 @@
...
+                Some(value) => Continue(Selectable::Other(value)),
+                None => Break(()),
+            }
+        });
+
+        let value = match result {
+            Continue(Selectable::Map(map)) => Value::Map(map.clone()),
+            Continue(Selectable::List(list)) => Value::List(list.clone()),
+            Continue(Selectable::Other(value)) => value.clone(),
+            Break(_) => Value::Null,
```
ever0de, line 43: How about throwing an error rather than returning Null when calling an index that is larger than the size of a `List`?

src/data/value/json.rs
panarch: `JsonValue` here looks unnecessary, which can be obviously inferred by the compiler.
I removed it, thanks :)
However, it does not seem possible to replace `from_str` to `from_value`. It needs to get `&str` and return `JsonValue`.

src/data/value/selector.rs
panarch: For the initial version, I just wanted `UNWRAP` selector not to generate error.
You can think that we are using `?` optional chaining for every key accesses.
It would be good to expand `selector` spec to support `?` for optional chaining which is supported by JS.

e.g. `a?.b?.c?`

src/data/value/json.rs
ever0de: Oh sorry!!
to_value, not from_value.

ever0de: Ignore the above comment
enough for now

panarch: This is a tiny one, but I also think it would be slightly better to use `serde_json::to_value`.
`serde_json::from_str` clearly shows that the input type is `&str`, but result type can be more than `JsonValue`.
In the case of serde_json::to_value`, it is obvious that the return type is `JsonValue`, but it can accept other input formats.
However, function param clearly shows that the input type of our function is `value: &str`, so.. :)

ever0de: 👍 change to match intent

panarch: Oh, it looks the behavior of `to_value` and `from_str` is different.
serde_json::to_value converts `r#"{ "a": 10 }"#` into `String`, not `Map`.

I'll revert the last commit for now.

ever0de: checked. Wrong change... sorry

panarch: No problem at all :)
It is very interesting, I'm just quite curious the reason why they have different logics.

ever0de: Naturally, I thought deserialization was necessary.
It would be interesting to know about this.

ever0de (approved):

PR #371, opened 2021-10-04
Support function `LOG`

slhmy:
# CHANGE Explain

### src/ast/function.rs
Add a enum type `Log`, it requires two args (`antilog` & `base`).
So the function represents fromula like: log<sub>base</sub>antilog.
### src/translate/function.rs
Args count check & `Expr::Function` return. (Actually I copied similiar codes from other two args translation 😳)
### src/executor/evaluate/mod.rs
The realization of `Log` referenced the usage of [f64::log](https://doc.rust-lang.org/std/primitive.f64.html#method.log).

## Tests
Refered tests from other `expr_log`.
‘Type error’ or ‘Null value‘ of either args will cause corresponding error.

### DIFF
--- a/src/ast/function.rs
+++ b/src/ast/function.rs
@@ -48,6 +48,8 @@ pub enum Function {
     Exp(Expr),
     #[strum(to_string = "LN")]
     Ln(Expr),
+    #[strum(to_string = "LOG")]
+    Log { antilog: Expr, base: Expr },
     #[strum(to_string = "LOG2")]
     Log2(Expr),
     #[strum(to_string = "LOG10")]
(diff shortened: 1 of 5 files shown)


ever0de (approved):
Thanks for the feature implementation.
Waiting for @panarch's review.

panarch (approved):
Perfect :)

In the last few months, thanksfully lot's of new contributors have added many new SQL functions.
So there would exist slightly different code styles in `translate/function.rs` and `evaluate/mod.rs`.
You can also participate if it fits your interest, or I can also introduce another issue which would be good to look around a different part of GlueSQL.

What you've done in this issue are
* Expanding AST
* Update translate module
* Update evaluate - evaluate_function
* Add integration tests

The next one I recommend is
[MemoryStorage TRANSACTION support #373](https://github.com/gluesql/gluesql/issues/373)
You will be able to see how GlueSQL's storage layer works.

If you are ok, you can add a comment to the issue. Then, I'll set assignee to you.

---
Thanks for the contribution, merging! 👍 👍 👍

PR #372, opened 2021-10-04
add function random generate uuid value

maruoovv:
add RandomUuid function to generate random uuid value


maruoovv:
I have no idea how to test random generated value.
If there are any good way please tell me

MRGRAVITY817:
> I have no idea how to test random generated value. If there are any good way please tell me

Normally random values can be tested by checking if the generated random values matches the generation rules.
For example, when you made a function that generates an integer typed random value that ranges between 0 to 10, you should make a test code that checks if the generated value follows 2 rules; 1) is integer 2) is between 0 to 10.

I'm sure testing uuid value can be tough, since we're not the one who constructed the algorithm. Although it won't be sufficient, but I've brought a useful link for it.
https://stackoverflow.com/questions/7905929/how-to-test-valid-uuid-guid

I'll try to find more. Keep up! 👍

src/data/value/uuid.rs
```diff
@@ -0,0 +1,57 @@
+use {
+    super::error::ValueError,
+    crate::{Error, Error::Value},
+    uuid::Uuid,
+};
+
+pub fn parse_uuid(v: &str) -> Result<u128, Error> {
```
ever0de, line 7: Is there any reason the function should be shared externally?

ever0de: If you only use gluesql inside, you can replace it with `pub(crate)`.

src/data/value/uuid.rs
```diff
@@ -0,0 +1,57 @@
...
+    match Uuid::parse_str(v) {
+        Ok(u) => Ok(u.as_u128()),
+        _ => Err(Value(ValueError::FailedToParseUUID(v.to_owned()))),
+    }
+}
+
+pub fn generate_random_uuid() -> u128 {
```
ever0de, line 14: Same as above.

src/tests/function/random_uuid.rs
```diff
@@ -0,0 +1,26 @@
...
+        ),
+        (
+            "SELECT random_uuid(0) as pi FROM SingleItem",
+            Err(TranslateError::FunctionArgsLengthNotMatching {
+                name: "RANDOM_UUID".to_owned(),
+                expected: 0,
+                found: 1,
+            }
+            .into()),
+        ),
```
ever0de, line 18: What do you think about adding test code for uuid pattern matching?

src/ast/function.rs
```diff
@@ -92,6 +92,8 @@ pub enum Function {
     },
     #[strum(to_string = "UNWRAP")]
     Unwrap { expr: Expr, selector: Expr },
+    #[strum(to_string = "RANDOM_UUID")]
+    RandomUuid(),
```
ever0de, line 96: Q: Do you use random and uuid together?

Looking at the [documentation](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)), it seems that v4(in UUID) already contains the meaning of random.

src/tests/function/random_uuid.rs
maruoovv: I will add test code for pattern matching.
@panarch 
Are there any macro in gluesql can use for pattern matching? Or should I create a new one?

ex) 
```
(
    "SELECT RANDOM_UUID() as UUID FROM SingleItem",
    Ok(type_match!(
        UUID
        Value::UUID;
    )),
),
```

PR #374, opened 2021-10-04
Replace eval_to_* func to macro

vbbono:


vbbono:
I have a question about how to write code.
Which of the two methods is better?

```
Function::Ceil(expr) => Ok(eval_to_float!(expr).ceil())
            .map(Value::F64)
            .map(Evaluated::from),

Function::Ceil(expr) => Ok(Evaluated::fromValue::F64(eval_to_float!(expr).ceil()))

```

ever0de:
> I have a question about how to write code. Which of the two methods is better?
> 
> ```
> Function::Ceil(expr) => Ok(eval_to_float!(expr).ceil())
>             .map(Value::F64)
>             .map(Evaluated::from),
> 
> Function::Ceil(expr) => Ok(Evaluated::fromValue::F64(eval_to_float!(expr).ceil()))
> ```

This is a matter of taste, but it will be easier to see if the depth of () is minimized.

src/executor/evaluate/mod.rs
```diff
@@ -563,57 +510,39 @@ async fn evaluate_function<'a, T: 'static + Debug>(
...
-                        }
-                        min(max(start + count, 0) as usize, string.len())
-                    }
-                    Nullable::Null => return Ok(Evaluated::from(Value::Null)),
-                },
-                None => string.len(),
+            let end = if count < 0 {
+                return Err(EvaluateError::NegativeSubstrLenNotAllowed.into());
+            } else {
+                min(max(start + count, 0) as usize, string.len())
```
ever0de, line 545: How about this code?

```suggestion
            let end = (count >= 0)
                .then(|| min(max(start + count, 0) as usize, string.len()))
                .ok_or(EvaluateError::NegativeSubstrLenNotAllowed)?;
```

ever0de: It is not related to the current PR, but it was seen during the code change.

PR #376, opened 2021-10-05
Support AlterTable for MemoryStorage

### LINKED ISSUE #351: MemoryStorage `ALTER TABLE` query support
Current `MemoryStorage` has only implemented `Store` and `StoreMut` traits.
This issue is for adding `AlterTable` store trait impl for `MemoryStorage`.

ref.
https://github.com/gluesql/gluesql/blob/main/src/storages/memory_storage/alter_table.rs

ever0de:
Support AlterTable For MemoryStorage, resolved #351 

I checked the factors below
- [x] `cargo clippy --release --all --all-targets`
- [x] `cargo build --release`
- [x] `cargo check --all`
- [x] `cargo test`

### DIFF
--- a/src/storages/memory_storage/alter_table.rs
+++ b/src/storages/memory_storage/alter_table.rs
@@ -4,18 +4,115 @@ use {
     super::MemoryStorage,
     crate::{
         ast::ColumnDef, result::MutResult, schema::ColumnDefExt, store::AlterTable,
-        AlterTableError, Value,
+        AlterTableError, Result, Value,
     },
     async_trait::async_trait,
 };
 
-macro_rules! unwrap_or_return_err {
-    ($storage:expr, $expr:expr) => {
-        match $expr {
-            Ok(success) => success,
-            Err(err) => return Err(($storage, err.into())),
+impl MemoryStorage {
+    fn rename_column_sync(
+        &mut self,
+        table_name: &str,
+        old_column_name: &str,
+        new_column_name: &str,
+    ) -> Result<()> {
+        let item = self
+            .items
+            .get_mut(table_name)
+            .ok_or_else(|| AlterTableError::TableNotFound(table_name.to_owned()))?;
+
+        let mut column_def = item
+            .schema
+            .column_defs
+            .iter_mut()
+            .find(|column_def| column_def.name == old_column_name)
+            .ok_or(AlterTableError::RenamingColumnNotFound)?;
+
+        column_def.name = new_column_name.to_owned();
+
+        Ok(())
+    }
+
+    fn add_column_sync(&mut self, table_name: &str, column_def: &ColumnDef) -> Result<()> {
+        let item = self
+            .items
+            .get(table_name)
+            .ok_or_else(|| AlterTableError::TableNotFound(table_name.to_owned()))?;
+
+        if item
+            .schema
+            .column_defs
+            .iter()
+            .any(|ColumnDef { name, .. }| name == &column_def.name)
+        {
+            let adding_column = column_def.name.to_owned();
+
+            return Err(AlterTableError::AddingColumnAlreadyExists(adding_column).into());
         }
-    };
+
+        let ColumnDef { data_type, .. } = column_def;
+        let nullable = column_def.is_nullable();
+        let default = column_def.get_default();
+        let value = match (default, nullable) {
+            (Some(expr), _) => {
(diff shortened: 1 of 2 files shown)


ever0de:
Cases that were not taken into account were corrected and uploaded.

Since it is very different from the existing implementation, I organized the commit and force-pushed it.
Please understand this advantage.

src/storages/memory_storage/alter_table.rs
```diff
@@ -3,50 +3,169 @@
...
+            storage,
+            item.schema
+                .column_defs
+                .iter_mut()
+                .find(|column_def| column_def.name == old_column_name)
+                .ok_or(AlterTableError::RenamingColumnNotFound)
+        );
+        column_def.name = new_column_name.to_owned();
+
+        Ok((storage, ()))
```
panarch, line 67: It looks that even though Try v2 has come into the stable Rust, we cannot replace `unwrap_or_return_err!` macro to the normal function which has exactly same functionality.
Then, it is not `try_..` which you mentioned is right.

Now, how about this?
We don't actually need `async` and `MutResult` for implementing internal logic.

If we add a new internal method for `AlterTable`, then we might not need this macro.

Let's suppose that we add a `rename_column_sync` function.
Then...
```rust
async fn rename_column(                                                             
    self,                                                                           
    table_name: &str,                                                               
    old_column_name: &str,                                                          
    new_column_name: &str,                                                          
) -> MutResult<Self, ()> {                                                          
    let mut storage = self;                                                         
                                                                                    
    match storage.rename_column_sync(table_name, old_column_name, new_column_name) {
        Ok(()) => Ok((storage, ())),                                                
        Err(e) => Err((storage, e)),                                                
    }                                                                               
}                                                                                   
```

```rust
impl MemoryStorage {                                                               
    fn rename_column_sync(                                                         
        &mut self,                                                                 
        table_name: &str,                                                          
        old_column_name: &str,                                                     
        new_column_name: &str,                                                     
    ) -> Result<()> {                                                              
        let item = self                                                            
            .items                                                                 
            .get_mut(table_name)                                                   
            .ok_or_else(|| AlterTableError::TableNotFound(table_name.to_owned()))?;
                                                                                   
        let mut column_def = item                                                  
            .schema                                                                
            .column_defs                                                           
            .iter_mut()                                                            
            .find(|column_def| column_def.name == old_column_name)                 
            .ok_or(AlterTableError::RenamingColumnNotFound)?;                      
                                                                                   
        column_def.name = new_column_name.to_owned();                              
                                                                                   
        Ok(())                                                                     
    }                                                                              
}                                                                                  
```


Like this above, actual implementation can even become more simple.

ever0de: There was '&mut self'.
I'm only thinking about avoiding mutable things, but I forgot that Storage allows it.
This is a good solution. I'll fix it right away

ever0de: How about making this a closure?

ever0de: MemoryStorage can be an example of a storage implementation.
So you have to write explicit code.
I will edit as per your feedback.

ever0de: Done.

src/storages/memory_storage/alter_table.rs
```diff
@@ -3,50 +3,176 @@
...
-        Err((
-            self,
-            Error::StorageMsg("[MemoryStorage] alter-table is not supported".to_owned()),
-        ))
+        let mut storage = self;
+
+        result_into(
+            storage.drop_column_sync(table_name, column_name, if_exists),
+            storage,
+        )
```
panarch, line 169: Oh, there already exists a solution for this case!

```rust
use crate::result::TrySelf;
```
Then, 
```rust
storage                                                   
    .drop_column_sync(table_name, column_name, if_exists) 
    .try_self(storage)                                    
```


ref. https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv
Maybe it would be better to use name of `into_mut` rather than `try_self`... hmm..
In anyway, we can use `TrySelf` in here for now.

ever0de: There was something good about `result.rs`.
I'll also think about the naming.

panarch (approved):
Awesome 👍 👍 👍
Looks all good, merging!

Issue #337, opened 2021-09-03
Apply `bigdecimal` feature to `sqlparser-rs`


panarch:
`sqlparser-rs` provides `bigdecimal` as an optional feature.
Currently, `bigdecimal` is not being used in here.
So `Literal::Number` contains `String`, but it would be better to use `bigdecimal`.
Then, we don't need to parse `String` to `i64` or `f64` in execution layer.

In https://github.com/gluesql/gluesql/blob/main/src/data/literal.rs
We can clean up all binary operations between number values.

devgony:
Let me try it!

PR #380, opened 2021-10-11
Apply bigdecimal feature (solved issue)

devgony:
# Apply bigdecimal feature (#337) 

1. Add bigdecimal
```rs
// Cargo.toml
sqlparser = { version = "0.11", features = ["serde", "bigdecimal"] }
bigdecimal = { version = "0.3", features = ["serde"]}
```
2. Add serde_str to solve `DeserializeAnyNotSupported` error.
```rs
// Cargo.toml
serde_str="0.1.0"
```
```rs
// src/ast/ast_literal.rs
pub enum AstLiteral {
    Boolean(bool),
    #[serde(with = "serde_str")]
    Number(BigDecimal),
...
```

3. Bigdecimal has larger floating point than f64 
    => fixed with rectified float.
```rs
// sqrt_power.rs
            "SELECT POWER(32,0.3) as power_with_float FROM SingleItem",
            Ok(select!(
                power_with_float
                F64;
                f64::from(32).powf(0.3) // expected: [F64(2.82842712474619)], found: [F64(2.8284271247461907)
            )),
```

### DIFF
--- a/src/tests/function/cast.rs
+++ b/src/tests/function/cast.rs
@@ -23,7 +23,7 @@ test_case!(cast_literal, async move {
         ),
         (
             r#"SELECT CAST(3 AS BOOLEAN) AS cast FROM Item"#,
-            Err(ValueError::LiteralCastToBooleanFailed("3".to_owned()).into()),
+            Ok(select!(cast Bool; true)),
         ),
         (
             r#"SELECT CAST(NULL AS BOOLEAN) AS cast FROM Item"#,
--- a/src/tests/function/cast.rs
+++ b/src/tests/function/cast.rs
@@ -59,7 +59,7 @@ test_case!(cast_literal, async move {
         ),
         (
             r#"SELECT CAST("foo" AS FLOAT) AS cast FROM Item"#,
-            Err(ValueError::LiteralCastToFloatFailed("foo".to_owned()).into()),
+            Err(ValueError::LiteralCastFromTextToFloatFailed("foo".to_owned()).into()),
         ),
         (
             r#"SELECT CAST(TRUE AS FLOAT) AS cast FROM Item"#,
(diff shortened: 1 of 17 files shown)


panarch (changes requested):
Sorry for the late review.
Looks nice :)
Let's just try to remove unnecessary `unwrap()`s.

panarch:
We can now replace `bigdecimal::ToPrimitive` to `data::bigdecimal_ext`.
Then, we can remove `BigDecimal::is_integer` checks.

```rust
use bigdecimal::BigDecimal;                               
                                                          
pub trait BigDecimalExt {                                 
    fn to_i64(&self) -> Option<i64>;                      
                                                          
    fn to_f64(&self) -> Option<f64>;                      
}                                                         
                                                          
impl BigDecimalExt for BigDecimal {                       
    fn to_i64(&self) -> Option<i64> {                     
        match self.is_integer() {                         
            true => bigdecimal::ToPrimitive::to_i64(self),
            false => None,                                
        }                                                 
    }                                                     
                                                          
    fn to_f64(&self) -> Option<f64> {                     
        bigdecimal::ToPrimitive::to_f64(self)             
    }                                                     
}                                                         
```

src/tests/function/cast.rs
```diff
@@ -23,7 +23,7 @@ test_case!(cast_literal, async move {
         ),
         (
             r#"SELECT CAST(3 AS BOOLEAN) AS cast FROM Item"#,
-            Err(ValueError::LiteralCastToBooleanFailed("3".to_owned()).into()),
+            Ok(select!(cast Bool; true)),
```
panarch, line 26: This issue is not supposed to change the `CAST` behavior.
`CAST` from `Value::Str` to `Value::Bool` only accept `1` as `TRUE`.
Let's keep the original behavior.

panarch (approved):
Beautiful, everything looks good!
Thanks a lot 👍 👍 👍

Issue #378, opened 2021-10-09
Implement `prelude` for userside APIs


MRGRAVITY817:
## `lib.rs` is exporting relentlessly...
Currently all modules are exported like this from `lib.rs`.
![image](https://user-images.githubusercontent.com/41038636/136666994-3cc18d99-b331-4d31-842a-638860152507.png)

This can cause situations like importing `Value` from `crate::Value` or `crate::data::Value`, which can be confusing.
Maybe it's time to: 

- [ ]  Extract userside APIs (`data::Value`, `Result`, `Error`, `Glue` etc) into the `prelude`
- [ ] Refactor `lib.rs` to remove unnecessary exports.

## Reference
https://doc.rust-lang.org/std/prelude/index.html

ever0de:
ref #388

PR #388, opened 2021-10-15
 Implement prelude pattern to organize userside APIs 

MRGRAVITY817:
# This resolves issue #378 
## Current problem in `lib.rs`
Current `lib.rs` is exporting most of the module as public, like this.
```rust
pub use data::*;
pub use executor::*;
pub use parse_sql::*;
pub use plan::*;
pub use result::*;
pub use store::*;
pub use translate::*;

pub use glue::Glue;
pub use storages::*;
```
This can confuse users, because it may present multiple paths to import.
For example, the `Value` enum can be imported with those paths.
```rust
use gluesql::Value;
use gluesql::data::value::Value;
``` 
## Organizing with prelude pattern
We can solve this problem by implementing prelude pattern, which is to re-export only the commonly used userside APIs in `pub mod prelude{ ... }`. Rust's `std` library also uses prelude pattern for commonly used APIs like `Option<T>` which should be in `std::option::Option`, and `Result<T>` which should be in `std::result::Result<T>`, and much more!
## Which are the "userside APIs"?
That's a very subjective question. Since this is the first implementation, I had to use my "guess"(+ discussions with other contributors) about which APIs will be used frequently. Currently, I added the APIs used in our GlueSQL `example`.
```rust
pub mod prelude {
    #[cfg(feature = "sled-storage")]
    pub use crate::storages::SledStorage;

    #[cfg(feature = "memory-storage")]
    pub use crate::storages::MemoryStorage;

    pub use crate::{
        data::value::Value,
        executor::{execute, Payload},
        glue::Glue,
        parse_sql::parse,
        result::{Error, Result},
        translate::translate,
    };
}
``` 
## Extra works for `tests`
Since Rust's file system treats integration tests in `tests` directory as a different crate, we also need to export the APIs which are used in tests. This part should be edited when we adapt new APIs for tests later.

## This PR is **just a start** of the prelude patch
There will be more adds and gones in this project, and the changes of usage patterns. 
Hence there's no concrete answer for which to put inside the prelude module. It will also change.

# Reference Materials
`std::prelude` : https://doc.rust-lang.org/beta/std/prelude/index.html 
`prelude` example in `diesel-rs` crate : https://github.com/diesel-rs/diesel/blob/8a3c4391476d92f25a8b48b6c18e4d7197a4a8ef/diesel/src/lib.rs#L377


src/lib.rs
```diff
@@ -86,13 +86,46 @@ pub mod store;
...
-pub use storages::*;
+    #[cfg(feature = "memory-storage")]
+    pub use crate::storages::MemoryStorage;
+
+    pub use crate::{
+        data::value::Value,
+        executor::{execute, Payload},
+        glue::Glue,
+        parse_sql::parse,
+        translate::translate,
```
panarch, line 101: `parse` -> `translate` -> `plan` -> `execute`
This path above is generally expected routine if user does not use `Glue`.
Let's add `plan::plan` to prelude :)

src/lib.rs
```diff
@@ -86,13 +86,46 @@ pub mod store;
...
+            IntervalError, Literal, LiteralError, RowError,
+        },
+        executor::*,
+        parse_sql::{parse, parse_expr},
+        plan::plan,
+        result::*,
+        store::{GStore, GStoreMut, StoreMut},
+        translate::{translate, translate_expr, TranslateError},
+    };
+}
```
panarch, line 131: Exposing `test` module may sound unclear because we already have `tests` module for providing integration test cases.
What we basically need as an external custom storage maker is the way to use our internal module.
Then, it would be just enough to provide our modules as public.

```rust
pub mod ast;      
pub mod data;     
pub mod executor; 
pub mod parse_sql;
pub mod plan;     
pub mod result;   
pub mod storages; 
pub mod store;    
pub mod tests;    
pub mod translate;
pub mod utils;    
```

Let's provide a single way to access specific module or function.
(Only the exception is `prelude`)

To remove this `test` module, all the test case module imports need to be updated.
However, I think that is necessary task to do.
How do you think?

panarch (changes requested):

src/lib.rs
MRGRAVITY817: That's reasonable. I also thought the role of my test mod was kinda ambiguous. Thanks for the  suggestion!

panarch (approved):
Great job, thanks!!
All looks good, merging! 👍 👍 👍 👍 👍

MRGRAVITY817:
Thanks!

PR #396, opened 2021-10-22
385 factorial operator

devil-k:
supported factorial unary operator

### DIFF
--- a/src/data/value/mod.rs
+++ b/src/data/value/mod.rs
@@ -319,7 +319,10 @@ impl Value {
         use Value::*;
 
         let factorial_function = |a: i64| -> Result<i64> {
-            (1..(a+1)).into_iter().try_fold(1i64, |mul, x| mul.checked_mul(x)).ok_or(ValueError::FactorialOverflow.into())
+            (1..(a + 1))
+                .into_iter()
+                .try_fold(1i64, |mul, x| mul.checked_mul(x))
+                .ok_or_else(|| ValueError::FactorialOverflow.into())
         };
 
         match self {
--- a/src/data/value/mod.rs
+++ b/src/data/value/mod.rs
@@ -319,18 +319,7 @@ impl Value {
         use Value::*;
 
         let factorial_function = |a: i64| -> Result<i64> {
-            let mut result: i64 = 1;
-
-            for x in 1..(a + 1) {
-                match result.checked_mul(x) {
-                    Some(x) => result = x,
-                    None => {
-                        return Err(ValueError::UnaryFactorialOverflow.into());
-                    }
-                }
-            }
-
-            Ok(result)
+            (1..(a+1)).into_iter().try_fold(1i64, |mul, x| mul.checked_mul(x)).ok_or(ValueError::UnaryFactorialOverflow.into())
         };
 
         match self {
--- a/src/data/value/mod.rs
+++ b/src/data/value/mod.rs
@@ -319,18 +319,18 @@ impl Value {
         use Value::*;
 
         let factorial_function = |a: i64| -> Result<i64> {
-            let mut result:i64 = 1;
-            
-            for x in 1..(a+1) {
+            let mut result: i64 = 1;
+
+            for x in 1..(a + 1) {
                 match result.checked_mul(x) {
                     Some(x) => result = x,
                     None => {
                         return Err(ValueError::UnaryFactorialOverflow.into());
-                    },
+                    }
                 }
             }
 
-            return Ok(result);
+            Ok(result)
         };
 
         match self {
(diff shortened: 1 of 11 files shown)


src/data/value/mod.rs
```diff
@@ -315,6 +315,33 @@ impl Value {
...
+                    match factorial_function(*a) {
+                        Ok(x) => Ok(I64(x)),
+                        Err(x) => Err(x),
+                    }
+                }
+            }
+            F64(_) => Err(ValueError::FactorialOnNonInteger.into()),
+            Null => Ok(Null),
+            _ => Err(ValueError::FactorialOnNonNumeric.into()),
+        }
```
panarch, line 342: Tiny simplification, Rust `match` is quite powerful.
```rust
match self {                                                     
    I64(a) if *a >= 0 => factorial_function(*a).map(I64),        
    I64(_) => Err(ValueError::FactorialOnNegativeNumeric.into()),
    F64(_) => Err(ValueError::FactorialOnNonInteger.into()),     
    Null => Ok(Null),                                            
    _ => Err(ValueError::FactorialOnNonNumeric.into()),          
}                                                                
```

src/executor/evaluate/evaluated.rs
```diff
@@ -138,6 +138,13 @@ impl<'a> Evaluated<'a> {
         }
     }
 
+    pub fn unary_factorial(&self) -> Result<Evaluated<'a>> {
+        match self {
+            Evaluated::Literal(v) => v.unary_factorial().map(Evaluated::Literal),
+            Evaluated::Value(v) => v.unary_factorial().map(Evaluated::from),
+        }
```
panarch, line 145: In the `unary_factorial` impl in `data/literal.rs`, it converts `Literal::Number` only into `i64` which can be directly matched to `Value::I64`.
Value or literal flows from less informative type to more informative type, and `Value` has more info than `Literal`.
Current conversion from `BigDecimal` to `i64` really makes sense. However, we don't need to convert back to `BigDecimal` after the calculation has done.
Then, we might not need `Literal::unary_factorial`.
Rather than `Literal::unary_factorial`, we can convert `Literal` into `Value` here.

```rust
match self {                                                                      
    Evaluated::Literal(v) => Value::try_from(v).and_then(|v| v.unary_factorial()),
    Evaluated::Value(v) => v.unary_factorial(),                                   
}.map(Evaluated::from)
```

src/tests/error.rs
```diff
@@ -9,10 +9,6 @@ test_case!(error, async move {
             TranslateError::UnsupportedStatement("TRUNCATE TABLE TableA".to_owned()).into(),
             "TRUNCATE TABLE TableA;",
         ),
-        (
-            TranslateError::UnsupportedUnaryOperator("!".to_owned()).into(),
-            "SELECT 2! FROM TableA;",
-        ),
```
panarch, line 15: Now, it looks that we support all unary operator which `sqlparser-rs` supports.
Let's rename the error.
```rust
#[error("unreachable unary operator: {0}")]
UnreachableUnaryOperator(String),          
```

I saw the parser code, because of the `PostgreSqlDialect`, `Unreachable-` prefix in our code looks inevitable.
image
https://github.com/sqlparser-rs/sqlparser-rs/blob/9c87d3b195ecc05f92927197e29f5285c3aa5e9b/src/parser.rs#L416-L434

panarch (changes requested):
Codes looks neat and integration tests what you wrote is really awesome, which you carefully distinguish different error types to have different errors and added all the test cases.

I left some minor change requests, thanks!

src/data/value/mod.rs
devil-k: @panarch 

thanks.
aligned line..

src/executor/evaluate/evaluated.rs
devil-k: @panarch 

thanks..
it is very beautiful..

modified code.
and, delete literal factorial code

src/tests/unary_operator.rs
```diff
@@ -0,0 +1,107 @@
...
+        (
+            r#"INSERT INTO Test VALUES (10, 10.5, "hello", -5, 1000)"#,
+            Ok(Payload::Insert(1)),
+        ),
+        (
+            "SELECT -v1 as v1, -v2 as v2, v3, -v4 as v4 FROM Test",
+            Ok(select_with_null!(
+                v1  |           v2  |                   v3  |                   v4;
+                I64(-10)      F64(-10.5)            Str("hello".to_owned())                   I64(5)
+            )),
```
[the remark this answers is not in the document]
devil-k: @panarch

thanks.
aligned line..

src/data/value/mod.rs
devil-k: @panarch

change code..

panarch (approved):
Beautiful.. All looks great, thanks a lot! merging!

panarch:
resolve #396

PR #407, opened 2021-11-16
Add int8 data type

### LINKED ISSUE #390: Add `INT(8)` data type
### Please check the following

- [ ] This data type is 1byte
- [ ] Parsing in `sqlparser-rs`: <https://github.com/sqlparser-rs/sqlparser-rs/blob/main/src/parser.rs#L2045>
    - Ex result) `sqlparser::DataType::Int(Option<u64>)`

### In the future
numerics can be implemented efficiently by grouping them into modules once more in `Value`.

ever0de:
resolved #390 

## Detail

- Implement Int8 Data Type

## Agenda 

- The items below are suggestions, The checked list has been applied to the relevant PR.
- Unchecked lists require communication.

1. [ ] Add test code for comparison  between `Literal` and `Value`
2. [ ] Make numeric related operations `***_for_numeric` function.
3. [ ] Rename Int to Int64


ever0de:
TODO: Group integer type

ever0de:
Rebase upstream/main

src/data/value/mod.rs
```diff
@@ -155,21 +184,20 @@ impl Value {
         use Value::*;
 
         match (self, other) {
-            (I64(a), I64(b)) => Ok(I64(a + b)),
-            (F64(a), F64(b)) => Ok(F64(a + b)),
-            (I64(a), F64(b)) | (F64(b), I64(a)) => Ok(F64(*a as f64 + b)),
+            (I8(a), _) => TryBinaryOperator::try_add(a, other),
+            (I64(a), _) => TryBinaryOperator::try_add(a, other),
+            (F64(a), _) => TryBinaryOperator::try_add(a, other),
```
panarch, line 189: This looks enough to use `try_add` only.
```rust
(I8(a), b) => a.try_add(b), 
(I64(a), b) => a.try_add(b),
(F64(a), b) => a.try_add(b),
```

src/data/value/mod.rs
```diff
@@ -81,9 +98,19 @@ impl PartialOrd<Value> for Value {
 }
 
 impl Value {
+    pub fn is_zero(&self) -> bool {
```
panarch, line 101: Nice, thanks 👍

src/data/value/mod.rs
```diff
@@ -36,14 +38,30 @@ pub enum Value {
...
+    fn try_divide(&self, rhs: &Self::Rhs) -> Result<Value>;
+    fn try_modulo(&self, rhs: &Self::Rhs) -> Result<Value>;
+}
+
+trait BinaryOperator {
+    type Rhs;
+
+    fn eq(&self, rhs: &Self::Rhs) -> bool;
+    fn partial_cmp(&self, rhs: &Self::Rhs) -> Option<Ordering>;
+}
```
panarch, line 56: 1. This would be enough to be located in `value/binary_op.rs`
Then, in here we can use these from `binary_op` module.

2. `BinaryOperator` looks unnecessary
There already exists `PartialEq` and `PartialOrd` which are provided by Rust itself, we can simply use those rather than declaring custom `BinaryOperator` trait.

src/tests/data_type/numeric.rs
```diff
@@ -0,0 +1,144 @@
...
+        Ok(select!(
+            modulo
+            I8;
+            0;
+            0;
+            0;
+            0
+        )),
+        "SELECT field_one % field_two AS modulo FROM Item;"
+    );
```
panarch, line 141: `select!` macro does not recognize line-break so,
```rust
test!(                                                 
    Ok(select!(modulo I8; 0; 0; 0; 0)),                
    "SELECT field_one % field_two AS modulo FROM Item;"
);                                                     
```
It would be good to use single line `select!` for simple cases.

src/tests/mod.rs
```diff
@@ -99,6 +99,7 @@ macro_rules! generate_store_tests {
         glue!(ordering, ordering::ordering);
         glue!(order_by, order_by::order_by);
         glue!(sql_types, data_type::sql_types::sql_types);
+        glue!(numeric, data_type::numeric::numeric);
```
panarch, line 102: Current `numeric` test looks that it only covers `INT(8)` data type, how about using `int8`?

panarch (changes requested):
Trait approach looks really, really beautiful.
There's a few comments I left, thanks a lot!

src/data/value/mod.rs
ever0de: 2 will probably have a problem with the core implementation.
When implemented for i8 ...

image

panarch: Those are supposed to work, hmm.. I don't get it why the conflict occurs but I've tested the codes below and there was no issue.

```rust
impl PartialEq<Value> for i8 {                                
    fn eq(&self, other: &Value) -> bool {                     
        match other {                                         
            I8(other) => self.eq(other),                      
            I64(other) => (*self as i64).eq(other),           
            F64(other) => (*self as f64).eq(other),           
            _ => false,                                       
        }                                                     
    }                                                         
}                                                             
                                                              
impl PartialOrd<Value> for i8 {                               
    fn partial_cmp(&self, other: &Value) -> Option<Ordering> {
        match other {                                         
            I8(other) => self.partial_cmp(other),             
            I64(other) => (*self as i64).partial_cmp(other),  
            F64(other) => (*self as f64).partial_cmp(other),  
            _ => None,                                        
        }                                                     
    }                                                         
}                                                             
```

ever0de: It seems to be because it is defined in Associate type. I edited

panarch (approved):
Beautiful! Thanks a lot!

PR #433, opened 2021-12-11
Support arithmetic operations for decimal and null

### LINKED ISSUE #428: Support arithmetic operations for Decimal and Null.
It would be good to move Decimal to binary_op and do it together.

ever0de:
fixed #428


panarch (approved):
Thanks a lot!

Issue #418, opened 2021-12-01
Support `SELECT * FROM (VALUES (1)) t1 (INTEGER);` like queries


panarch:
```sql
SELECT * FROM (VALUES (1)) t1 (INTEGER);
```

This query is supported by `sqlparser-rs` but not us.
GlueSQL returns `Translate(UnsupportedQueryTableFactor("(VALUES (1)) AS t1 (INTEGER)"))` error.

panarch:
blocked by #417

Issue #417, opened 2021-12-01
Support `SetExpr::Values` in `select` module


panarch:
Current `select` module does not support `SetExpr::Values`
ref.
https://github.com/gluesql/gluesql/blob/874710024e5a3d6df7925a834ac80927433f7445/src/executor/select/mod.rs#L161-L172

However, it is already being implemented and used in handling `INSERT` query.
https://github.com/gluesql/gluesql/blob/874710024e5a3d6df7925a834ac80927433f7445/src/executor/execute.rs#L171-L176

Let's merge this `SetExpr::Values` handler into `select` module.

devgony:
Let me try this issue!

PR #435, opened 2021-12-11
Move `order_by` from query to select

devgony:
To implement #417 , #418 
We decided to move `order_by` from `Query` to `Select` first. 

At sqlparser::ast::query, `order_by` is on `Query`,
but from now, our translated gluesql::ast::query will consider `order_by` as on `Select`.

### DIFF
--- a/src/plan.rs
+++ b/src/plan.rs
@@ -7,6 +7,7 @@ use {
         data::{get_name, Schema, SchemaIndex, SchemaIndexOrd},
         result::Result,
         store::Store,
+        utils::Vector,
     },
     async_recursion::async_recursion,
     std::fmt::Debug,
@@ -121,7 +122,7 @@ async fn plan_query<T: Debug>(storage: &dyn Store<T>, query: Query) -> Result<Qu
                 selection,
                 group_by,
                 having,
-                order_by,
+                order_by: Vector::from(order_by).pop().0.into(),
             };
 
             Ok(Query {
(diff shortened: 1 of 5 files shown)


src/plan.rs
```diff
@@ -83,13 +80,13 @@ async fn plan_query<T: Debug>(storage: &dyn Store<T>, query: Query) -> Result<Qu
...
                 body: SetExpr::Select(select),
-                order_by,
                 limit,
                 offset,
             });
         }
     };
 
+    let Select { order_by, .. } = select.as_ref();
     let index = order_by.last().and_then(|value_expr| {
```
panarch, line 90: you can simply use `select.order_by` without destructuring
```rust
let index = select.order_by.last().and_then(|value_expr| {
 ..
```

src/plan.rs
```diff
@@ -124,11 +122,11 @@ async fn plan_query<T: Debug>(storage: &dyn Store<T>, query: Query) -> Result<Qu
                 selection,
                 group_by,
                 having,
+                order_by,
             };
 
             Ok(Query {
                 body: SetExpr::Select(Box::new(select)),
-                order_by: Vector::from(order_by).pop().0.into(),
```
panarch, line 131: Here is the key of `index_order_by` failure.
If the planner succeeded to find the index in `ORDER BY`, then it needs to remove the last order_by item which is consumed by the index.

devgony: Thanks.. I realized we should never remove previous logic while refactoring...

panarch (approved):
Looks all good!! thanks a lot, merging 👍

PR #438, opened 2021-12-13
Add `Metadata` optional store trait,

panarch:
`Metadata` provides two methods which are currently matching `SHOW {VARIABLE};` queries.
- `schema_names`: `SHOW TABLES;`
- `version`: `SHOW VERSION;`

Add `Metadata` store traits and implement it for both `MemoryStorage` and `SledStorage`.

### DIFF
--- a/src/executor/execute.rs
+++ b/src/executor/execute.rs
@@ -24,6 +24,9 @@ use super::alter::alter_table;
 #[cfg(feature = "index")]
 use super::alter::{create_index, drop_index};
 
+#[cfg(feature = "metadata")]
+use crate::{ast::Variable, result::TrySelf};
+
 #[derive(ThisError, Serialize, Debug, PartialEq)]
 pub enum ExecuteError {
     #[error("table not found: {0}")]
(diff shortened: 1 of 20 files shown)


src/executor/execute.rs
```diff
@@ -57,6 +60,16 @@ pub enum Payload {
     Commit,
     #[cfg(feature = "transaction")]
     Rollback,
+
+    #[cfg(feature = "metadata")]
+    ShowVariable(PayloadVariable),
+}
```
ever0de, line 66: It has nothing to do with metadata, but how about attaching indexes to each other?

panarch: I think... I don't get the point. Could you provide more details?

ever0de: This is the sorting story inside `ExecuteError`.
It's related to formatting, so I don't think you need to think too much about it.

Group them in the form below.
```rust
     #[cfg(feature = "index")]
     CreateIndex,
     #[cfg(feature = "index")]
     DropIndex,

     #[cfg(feature = "transaction")]
     StartTransaction,
     #[cfg(feature = "transaction")]
     Commit,
     #[cfg(feature = "transaction")]
     Rollback,

     #[cfg(feature = "metadata")]
     ShowVariable(PayloadVariable),
}
```

tests/sled_transaction.rs
```diff
@@ -729,3 +732,54 @@ async fn sled_transaction_timeout_index() {
         Ok(select!(id I64; 1))
     );
 }
+
+#[cfg(feature = "metadata")]
+#[test]
+fn sled_transaction_metadata() {
+    macro_rules! test_tables {
+        ($glue: ident []) => {
```
ever0de, line 740: Q) [] What is syntax?
I wanted to study, but I haven't found the document yet.
Can you leave a reference?

ever0de (approved):
LGTM 👍 💯 🥇 🚀 👍 💯 🥇 🚀 👍 💯 🥇 🚀

Issue #46, opened 2020-08-27
CLI interface


MichaelMcDonnell:
Is a command line interface planned? I liked playing around with the web interface but there doesn't seem to be something equivalent for the terminal?

panarch:
I have a plan but I may use a separate repository, but it might take some time to work on cli task.
For now, you can expect same SQL support between Rust crate and web version.
Though they use different storages, they all pass same integration test cases so you can think as same at least about functionality.

willy610:
hej folks.
Just made some kind of console application built on ``gluesql  = {version="0.3"}`` or later of cource. Not on the project source. ok.

It's not present at GitHub yet. Trademarks, maintain, could be separate repository as a sister to gluesql.
It's called abc_sql at the moment to avoid any interferences with gluesql

It has the following poor help
```
here > ./target/debug/abc_sql --help
OK 
abc-sql   --newdb filename
        | --insql file --olddb filename [--printsqlstm]
        | --incsv file --olddb filename --table tablename 
            [ --append ]
            #NOT_YET ( 
              (--headers "col1,col2" [ --ignorefirtslineheaders ] )
              | --headersfirstlineincsv 
            ) 
            #NOT_YET [--colsep ','] [ --linsep ' ']
        | --olddb filename #interactive usage !
        | --outcsv filename --olddb filename --table tablename
        | --outsql filename --olddb filename --table tablename
            [--droptable] [--createtable] 
here > 
```

Create db, import of sql scripts, export of sql, csv import/export and interactive support

Interactive

```
abc-sql  > select * from S                                
         > left join P on S.CITY = P.CITY;
"select * from S left join P on S.CITY = P.CITY ;"


+-----+-------+--------+--------+------+-------+-------+--------+--------+
| SNO | SNAME | STATUS |  CITY  | PNO  | PNAME | COLOR | WEIGHT |  CITY  |
+-----+-------+--------+--------+------+-------+-------+--------+--------+
| S1  | Smith |     20 | London | P1   | Nut   | Red   |     12 | London |
| S1  | Smith |     20 | London | P4   | Screw | Red   |     14 | London |
| S1  | Smith |     20 | London | P6   | Cog   | Red   |     19 | London |
| S2  | Jones |     10 | Paris  | P2   | Bolt  | Green |     17 | Paris  |
| S2  | Jones |     10 | Paris  | P5   | Cam   | Blue  |     12 | Paris  |
[...]

PR #422, opened 2021-12-05
Refactor Monorepo

ever0de:
## File Structure

```
- book
- gluesql [packaging crate] - all features
    - dependencies
         [core, storages/*, examples, test-suite]
- core
    - dependencies
        [utils]
- storages/*
    - dependencies
        [core, utils]
    - <storage>
        - tests -> integration tests
        - bench
        - dependencies
            [test-suite]
- cli
    - dependencies
        [core, storages/*]
- utils
- test-suite * -> deploy other crate
```

## TODO

- [x] Remove features in the `utils` crate
- [x] add packagin crate

## Check List

- [ ] 

## Reference

- [serde](https://github.com/serde-rs/serde)

TBD

PR #440, opened 2021-12-25
Implement CLI application

panarch:
resolve https://github.com/gluesql/gluesql/issues/46

- [x] Setup CLI program base
- [x] Argument parsing using `clap`
- [x] Support both `MemoryStorage` and `SledStorage`
- [x] Replace `println!` to `writeln!` (for stdout tests)

```
// show help
cargo run -- --help

// memory-storage
cargo run

// sled-storage
cargo run -- --path [PATH]
```

After https://github.com/gluesql/gluesql/pull/422 done, this `cli/` will also be converted to workspace (currently it is an independent project inside of gluesql repo)


cli/src/print.rs
```diff
@@ -0,0 +1,244 @@
...
+            writeln!(
+                self.output,
+                "{} row{} {}\n",
+                n,
+                if n > 1 { "s" } else { "" },
+                msg
+            )
+        };
+
+        match payload {
```
24seconds, line 27: [nit]
Is it possible to print something when the sql is `CREATE TABLE ~~`?

cli/src/print.rs
```diff
@@ -0,0 +1,244 @@
...
+    table
+        .load_preset(UTF8_BORDERS_ONLY)
+        .apply_modifier(UTF8_ROUND_CORNERS)
+        .set_header(header);
+
+    table
+}
+
+#[cfg(test)]
+mod tests {
```
24seconds, line 87: nice!

cli/src/cli.rs
```diff
@@ -0,0 +1,100 @@
...
+                    println!("bye\n");
+                    break;
+                }
+                Err(e) => {
+                    println!("[unknown error] {:?}", e);
+                    break;
+                }
+            };
+
+            rl.add_history_entry(line.as_str());
```
24seconds, line 70: very useful

cli/src/cli.rs
```diff
@@ -0,0 +1,100 @@
...
+            glue,
+            print,
+            _marker: PhantomData,
+        }
+    }
+
+    pub fn run(&mut self) -> Result<()> {
+        macro_rules! println {
+            ($($p:tt),*) => ( writeln!(&mut self.print.output, $($p),*)?; )
+        }
```
24seconds, line 46: [Question]
Is there a reason to redefine println macro ?

24seconds (approved):
LGTM

cli/src/cli.rs
ever0de: Perhaps this is to reuse `writeln!(&mut ...)`.

panarch: The only reason is simplifying `&mut self.print.output`.
Currently, `cli.rs` does not have unit test but this can be used for it.

cli/src/print.rs
panarch: Sure, do you have any suggestion on it?

cli/src/print.rs
```diff
@@ -0,0 +1,244 @@
+use {
+    comfy_table::{modifiers::UTF8_ROUND_CORNERS, presets::UTF8_BORDERS_ONLY, Row, Table},
+    gluesql::prelude::{Payload, PayloadVariable},
+    std::io::{Result, Write},
+};
+
+pub struct Print<W> {
```
ever0de, line 7: I think I missed the Write type.

cli/src/cli.rs
24seconds: Aha!

cli/src/cli.rs
```diff
@@ -0,0 +1,100 @@
...
+                }
+                Err(e) => {
+                    println!("[unknown error] {:?}", e);
+                    break;
+                }
+            };
+
+            rl.add_history_entry(line.as_str());
+
+            let command = match Command::parse(line.as_str()) {
```
ever0de, line 72: nit picking
You can replace `as_str` with `&`.
String also implements AsRef<str>, and if you borrow it, there is no ownership problem.

cli/src/cli.rs
```diff
@@ -0,0 +1,100 @@
...
+                    continue;
+                }
+                Command::Quit => {
+                    println!("bye\n");
+                    break;
+                }
+                Command::Execute(sql) => match self.glue.execute(sql.as_str()) {
+                    Ok(payload) => self.print.payload(payload)?,
+                    Err(e) => {
+                        println!("[error] {}\n", e);
```
ever0de, line 93: nit picking
In main, we use `eprintln` to print the error. How about using it here?

cli/src/main.rs
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    fn run<T: Debug, U: GStore<T> + GStoreMut<T>>(storage: U) {
+        let output = std::io::stdout();
+        let mut cli = Cli::new(storage, output);
+
+        if let Err(e) = cli.run() {
+            eprintln!("{}", e);
+        }
+    }
```
ever0de, line 44: How about a function outside the main function?

ever0de (approved):

PR #441, opened 2021-12-31
Move `join_columns` argument from `Join::apply()` to `Join::new()`

MRGRAVITY817:
# [WIP] Move `join_columns` argument
Currently `join_columns` argument is passed through `apply()` method of `Join`.
It's better to move into constructor.


panarch (approved):
Thanks a lot!

PR #442, opened 2021-12-31
Refactor monorepo

ever0de:
ref: #422 

## File Structure

```
- book
- gluesql [packaging crate] - all features
    - dependencies
         [core, storages/*, examples, test-suite]
- core
    - dependencies
        [utils]
- storages/*
    - dependencies
        [core, utils]
    - <storage>
        - tests -> integration tests
        - bench
        - dependencies
            [test-suite]
- cli
    - dependencies
        [core, storages/*]
- utils
- test-suite * -> deploy other crate
```

## Reference

- [serde](https://github.com/serde-rs/serde)

PR #446, opened 2022-01-01
Convert `gluesql` workspace to root lib-package

ever0de:
It is an element that I did not think of when implementing #442.
ref: <https://github.com/lapce/lapce>

In the cargo manifest, workspace and package can exist at the same time.
Accordingly, `gluesql`, the workspace for packaging, was moved to the root.

This seems to depend on personal taste.
Even if there is a current `gluesql` workspace, it is unlikely that it will make much of a difference in terms of ux for our users.

Please close or merge according to your comments.


panarch (approved):
Awesome! All looks great, thanks a lot 👍 👍 👍 👍 👍

PR #450, opened 2022-01-07
Remove sorter cfg feature, support all ORDER BY by default

### LINKED ISSUE #420: Remove `sorter` optional feature, enable `ORDER BY` for non-indexed expr by default
`sorter` is an optional feature which enables `ORDER BY` query for non-indexed expr.
This option is for storage makers to have precise memory usage control, because `ORDER BY` for non-indexed expr forces loading all rows to memory.
However, current executor has another examples (e.g. `GROUP BY`, ..) which load target rows to the memory, but there's no feature to prevent this.
For the consistency, it would be simply better to remove `sorter` feature.

panarch:
resolve #420


ever0de (commented):
LGTM

PR #451, opened 2022-01-07
Add [[bin]] to Cargo.toml, update cli and storage workspace versions,

panarch:
When user runs CLI, proper version number must be returned.
Currently, `gluesql --version` returns 0.0.0, so update cli/Cargo.toml.
To show proper versions when users run `SHOW VERSION;` in the CLI program,
update storages/ Cargo.toml versions, too.

For now, I applied the same version number with root Cargo.toml (0.9.0)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,10 @@ keywords = [
 	"webassembly",
 ]
 
+[[bin]]
+name = "gluesql"
+path = "cli"
+
 [package.metadata.docs.rs]
 all-features = true
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,10 +16,6 @@ keywords = [
 	"webassembly",
 ]
 
-[[bin]]
-name = "gluesql"
-path = "cli"
-
 [package.metadata.docs.rs]
 all-features = true
 
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "gluesql-cli"
-version = "0.0.0"
+version = "0.9.0"
 authors = ["Taehoon Moon <taehoon.moon@outlook.com>"]
 edition = "2021"
 
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -4,6 +4,10 @@ version = "0.9.0"
 authors = ["Taehoon Moon <taehoon.moon@outlook.com>"]
 edition = "2021"
 
+[[bin]]
+name = "gluesql"
+path = "src/main.rs"
+
 [dependencies]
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -16,6 +16,10 @@ keywords = [
 	"webassembly",
 ]
 
+[[bin]]
+name = "gluesql"
+path = "cli"
```
ever0de, line 21: `cli` is a directory.
First of all, let's upload only the workspace version in this PR.
Even if you specify `cli/src/main.rs`, you will have to import clap (+ other crates) because the dependency is `gluesql`.

How about making the cli into a library later and going in the direction of holding main in gluesql?

panarch: > How about making the cli into a library later and going in the direction of holding main in gluesql?

This makes sense, but we still need to update `memory-storage` and `sled-storage` Cargo.toml versions.

> cli is a directory.
First of all, let's upload only the workspace version in this PR.
Even if you specify cli/src/main.rs, you will have to import clap (+ other crates) because the dependency is gluesql.

I think I don't get the point, can you explain more details about this?

cli/Cargo.toml
```diff
@@ -4,6 +4,10 @@ version = "0.9.0"
 authors = ["Taehoon Moon <taehoon.moon@outlook.com>"]
 edition = "2021"
 
+[[bin]]
+name = "gluesql"
+path = "src/main.rs"
+
```
panarch, line 10: @ever0de Oh..! Proper place where `[[bin]]` needs to be located was `cli/Cargo.toml`! not `Cargo.toml` in root.

Cargo.toml
ever0de: 1. I agree with the version update of storages workspaces.
2. bin/path is where you enter the path. The reason why it is not currently being built is that the directory is entered in the bin/path, not the main file (in terms of execution context).

I think it would be nice to separate the 2 from another PR.

panarch: About "1.", there was an another thing that I misunderstood.
Currently, default implementation of `version()` in `Metadata` is in `core` workspace, so `core/Cargo.toml` version was supposed to be the same as the root, not storages.

panarch:
Oh..
https://github.com/gluesql/gluesql/blob/d7f8c81b6c575b8bcd39cafcbb5709e55c8d931e/test-suite/src/metadata.rs#L18-L22

Currently,
`run!("SHOW VERSION;")` depends on `core` but `env!("CARGO_PKG_VERSION")` brings the version from storage's Cargo.toml version.

ever0de (approved):
👍 👍 👍 👍

PR #453, opened 2022-01-07
Apply Serde to MemoryStorage instance and make it public,

### LINKED ISSUE #452: Is it possible to construct an in-memory database from bytes?
Apologize if this is answered in the documentation somewhere, but I was unable to find it.

I have a static set of data, that I would like to be able to index and query efficiently from short-lived processes. (Rust compiled to wasm at Fastly's compute@edge network)

I would like to be able to create my database once, and embed it into my application binary, perhaps by writing it to a file and using `include_bytes!` or similar.
I'd then like to map these bytes into gluesql backed by a `memory` storage.

Is there any way of accomplishing this?

panarch:
Currently, there is no official Store trait related to import/export data, yet.
Serde can provide custom import/export functionality to MemoryStorage.

@GeeWee
ref #452

### DIFF
--- a/storages/memory-storage/Cargo.toml
+++ b/storages/memory-storage/Cargo.toml
@@ -12,6 +12,7 @@ gluesql-core = { path = "../../core", features = [
 ] }
 
 async-trait = "0.1"
+serde = { version = "1", features = ["derive"] }
 
 [dev-dependencies]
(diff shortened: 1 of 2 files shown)


ever0de (approved):
LGTM!

PR #466, opened 2022-01-09
Support `limit`, `offset` clause in `Insert into Table Select ~`

devgony:
It is possible to use `limit`, `offset` clause in `Insert into Table Select ~` sentence.
So, remove redundant query closure.
```rs
let query = || Query {
                            body: SetExpr::Select(select_query.clone()),
                            limit: None,
                            offset: None,
                        };
```


panarch (approved):
Awesome! Thanks a lot 👍

PR #469, opened 2022-01-11
Fix src/lib.rs doc comment,

panarch:
Rename gluesql_sled_storage to sled-storage.
Fix it to work with or without sled-storage feature

`cargo test -p gluesql --no-default-features` now works.


ever0de (approved):
👍  Thanks a lot

PR #470, opened 2022-01-11
Bump all versions to 0.10.2

panarch:
blocked by
ref. https://github.com/gluesql/gluesql/pull/469


ever0de (approved):
gooooooood!!

Glad the bug was fixed quickly.

PR #471, opened 2022-01-12
Update evaluate_stateless to support all existing SQL functions,

panarch:
Like `evaluate`, now `evaluate_stateless` also provides all SQL functions supported by GlueSQL

- [x] Update `evaluate_stateless` to support all GlueSQL SQL functions
- [x] Add tests

### DIFF
--- a/core/src/executor/evaluate/mod.rs
+++ b/core/src/executor/evaluate/mod.rs
@@ -1,14 +1,15 @@
 mod error;
 mod evaluated;
 mod expr;
+mod function;
 mod stateless;
 
 use {
     super::{context::FilterContext, select::select},
     crate::{
-        ast::{Aggregate, Expr, Function, TrimWhereField},
+        ast::{Aggregate, Expr, Function},
         data::Value,
-        result::{Error, Result},
+        result::Result,
         store::GStore,
     },
     async_recursion::async_recursion,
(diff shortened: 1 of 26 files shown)


core/src/executor/evaluate/mod.rs
```diff
@@ -204,407 +199,156 @@ async fn evaluate_function<'a, T: Debug>(
...
-
-    macro_rules! eval_to_integer {
-        ($v : expr) => {
-            match eval_to_integer($v).await? {
-                Nullable::Value(i) => i,
-                Nullable::Null => return Ok(Evaluated::from(Value::Null)),
-            }
-        };
-    }
+    let name = || func.to_string();
```
ever0de, line 211: Since there is no place to take the Name after the match expression, there seems to be no ownership problem.
How about converting from closure type to String?

```suggestion
    let name = func.to_string();
```

panarch: This is because.. there are some functions which does not require `name`.

e.g.
```rust
Function::GenerateUuid() => Ok(f::generate_uuid()),
Function::Now() => Ok(Value::Timestamp(Utc::now().naive_utc())),
```

ever0de (approved):
Looks good 👍

PR #472, opened 2022-01-23
Update Join::apply to take the whole table(rows) as a stream iterator

MRGRAVITY817:
# Update `Join::apply` to take the whole table(rows) as a `stream` iterator

`Join::apply` used to take a `init_context`, which is a starting point of a incoming rows.
Instead, our future join operation should compare table by table, so I changed like the following.


panarch:
ref #63

panarch (approved):
Thanks a lot 👍

PR #473, opened 2022-01-23
Implement bump version script

### LINKED ISSUE #464: Make bump verison script
To publish, we have a dependency.

We want the deployment to be like this:
1. utils
2. core
3. test-suite
4. storages/*

ever0de:
resolved #464 

This is only available to administrators (those with distribution rights on crates.io).

### How to run?

shalljs works like a shell script.
We just need to run it with +x permissions and simply ./bump_script.js .
```shell
chmod +x ./bump_script.js # If applied, you don't need it.
yarn
./bump_scripts.js
```


panarch:
Thanks a lot :)
However, I checked [cargo-release](https://github.com/crate-ci/cargo-release) and it looks that `cargo-release` already provides more functionalities. So.. it would be better to use `cargo-release`

ever0de:
> Thanks a lot :) However, I checked [cargo-release](https://github.com/crate-ci/cargo-release) and it looks that `cargo-release` already provides more functionalities. So.. it would be better to use `cargo-release`

oh that looks interesting
If you have a good tool, you can use it.

PR #476, opened 2022-02-02
Decimal precision and scale

boomkim:
Implementing Decimal precision and scale. 
I have implemented create, and insert logic with deciaml precision and scale. 
However, while I was trying to implement `insert into t1 select ~~~` I thought we should discuss it before I go any further. 

below is what I implemented in this draft. 

###  We can create table with `decimal(precision,scale)` type.
```sql
CREATE TABLE DECIMAL_PRECISION_SCALE (d1 DECIMAL(5,2))
```

### Also we can insert value into Decimal type. It validates the precision, and rounds conform to declared scale. 
```sql 
INSERT INTO DECIMAL_PRECISION_SCALE (d1) VALUES (123456); -- fail, exceeds precision. 
INSERT INTO DECIMAL_PRECISION_SCALE (d1) VALUES (123.456); -- success with rounded value: 123.46
```

### Since `eq` for Decimal type is not implemented, for now, it only supports create, insert, and select. 
```sql
UPDATE DECIMAL_PRECISION_SCALE SET d1=234.567 WHERE d1=123.46 -- fail, we cannot compare decimal values and literal (eq is not implemented) 
```

---

And below is what I have in concerned to go further. 

### We need to implement validate and round decimal value to conform to the declared schema. 

For example, `insert into t1 select decimal_value from t2` is available but validation and round is not implemented yet.(which means not perfect decimal type). For now, it only validate and round for literals. 

In this situation, I thought about how to work only when inserting and updating values from literal without precision verification and scale of decimal.

One way to think of it is to add precision validation to the row validation process, and scale round process to round in try_into_value. The advantage of this method is. The process of converting from literal to value is kept as before, and validation and round are applied only to value, thereby simplifying the decimal operation process. The downside. Because logic for validation and round is added to the process of creating a row itself, the whole row operation may be slightly heavier.

If logic can be added to the row side, we will modify the code in the direction of applying precision and scale to decimal values that have already been converted from literal to value. If there are other better ways, that's even better.

### DIFF
--- a/core/src/ast/data_type.rs
+++ b/core/src/ast/data_type.rs
@@ -14,5 +14,5 @@ pub enum DataType {
     Uuid,
     Map,
     List,
-    Decimal,
+    Decimal(Option<u64>, Option<u64>),
 }
--- a/core/src/data/value/decimal.rs
+++ b/core/src/data/value/decimal.rs
@@ -2,50 +2,40 @@ use {
     super::{Value, ValueError},
     crate::result::Result,
     bigdecimal::BigDecimal,
-    bigdecimal::FromPrimitive,
     bigdecimal::ToPrimitive,
     rust_decimal::Decimal,
-    std::borrow::Cow,
 };
 
-
-impl Value {
-    pub fn parse_decimal(precision: &Option<u64>, scale: &Option<u64>, value: &Cow<'_, BigDecimal>) -> Result<Value> {
-        
-        println!("precision: {:?}, scale: {:?}, value: {:?}", precision, scale, value);
-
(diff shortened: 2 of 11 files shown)


panarch:
Awesome!
It looks that now you may need to add `Decimal` wrapper module to `data/` which wraps `rust_decimal::Decimal`.
Then, you can implement
* binary operations between other primitive types in `data/decimal` module.
* binary operations between `Value::Decimal` and other `Value` in `data/value/binary_op/decimal.rs`.
Then you can clearly define binary-op and validation by keeping module dependency acyclic.

PR #484, opened 2022-02-14
Support `limit`, `offset` clause in Insert into values ~

devgony:
### Now It is possible to execute below queries.
```sql
INSERT INTO Test VALUES (1), (2), (3), (4) LIMIT 1; 
-- 1 rows inserted.
INSERT INTO Test VALUES (1), (2), (3), (4) OFFSET 1;
-- 3 rows inserted.
INSERT INTO Test VALUES (1), (2), (3), (4) LIMIT 3 OFFSET 2;
-- 2 rows inserted.
```

### DIFF
--- a/test-suite/src/limit.rs
+++ b/test-suite/src/limit.rs
@@ -53,28 +53,59 @@ test_case!(limit, async move {
             ),
         ),
         (
-            "INSERT INTO Test SELECT * FROM Test OFFSET 1;",
+            "CREATE TABLE InsertTest (
+                case_no INTEGER,
+                id INTEGER
+            )",
+            Payload::Create,
+        ),
+        (
+            "INSERT INTO InsertTest SELECT 1, id FROM Test OFFSET 1;",
             Payload::Insert(7),
         ),
         (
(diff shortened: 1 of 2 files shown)


test-suite/src/limit.rs
```diff
@@ -64,6 +64,18 @@ test_case!(limit, async move {
...
+            Payload::Insert(1),
+        ),
+        (
+            "INSERT INTO Test VALUES (1), (2), (3), (4) OFFSET 1;",
+            Payload::Insert(3),
+        ),
+        (
+            "INSERT INTO Test VALUES (1), (2), (3), (4) LIMIT 3 OFFSET 2;",
+            Payload::Insert(2),
+        ),
```
panarch, line 78: Can we have a `SELECT` test case to verify rows are properly added?
```rust
(
    "SELECT id FROM Test;",
    ...
```
kind of above would be enough.

panarch: Everything else all looks great :) Thanks a lot

panarch (approved):
👍 👍 👍

PR #486, opened 2022-02-20
Remove `Debug` macro from the generic requirements for key

ever0de:
We don't need to require the `Debug` macro for the Generic type `T` we use as `Key`. 
because the `Debug` macro is not used internally.


ever0de:
Rebase upstream main...

panarch (approved):
Awesome, thanks so much

PR #487, opened 2022-02-20
Bump rust version 1.58

ever0de:
# Announce

<https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html>

# Detail

There is a problem with the current CI. <https://github.com/gluesql/gluesql/runs/5264834300?check_suite_focus=true>
I inferred a version problem.


panarch (approved):
Thanks a lot!

PR #488, opened 2022-02-20
Remove unused generic in `IndexMut`

ever0de:
`IndexMut` trait does not require the `T` generic for the Key value.


panarch (approved):
Good catch! Thanks a lot, merging 👍

PR #495, opened 2022-03-12
Simplify lifetime uses in data/literal binary operators

panarch:
Replace `'b` to `'static`, make it to use more narrow lifetimes.

e.g.
```rust
// from
pub fn add<'b>(&self, other: &Literal<'a>) -> Result<Literal<'b>> {
 ..
}

// to
pub fn add(&self, other: &Literal<'a>) -> Result<Literal<'static>> {
 ..
}
```

### DIFF
--- a/core/src/data/literal.rs
+++ b/core/src/data/literal.rs
@@ -133,7 +133,7 @@ impl<'a> Literal<'a> {
         }
     }
 
-    pub fn add<'b>(&self, other: &Literal<'a>) -> Result<Literal<'b>> {
+    pub fn add(&self, other: &Literal<'a>) -> Result<Literal<'static>> {
         match (self, other) {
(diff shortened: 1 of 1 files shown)


ever0de (approved):
Looks good.

It seems that you can set the lifetime based on `self`. How about this one?

panarch:
> seems that you can set the lifetime based on `self`. How about this one?

Oh, I think.. I don't get it. Can you provide more details about it?

ever0de:
> > seems that you can set the lifetime based on `self`. How about this one?
> 
> Oh, I think.. I don't get it. Can you provide more details about it?

I thought the lifetime of Literal didn't have to be longer than self.

But it doesn't actually work.

```rust
pub fn add(&self, other: &Literal) -> Result<Literal> {
 ..
}
```

Issue #481, opened 2022-02-09
Add a functionality to read-and-execute SQL file in GlueSQL CLI


MRGRAVITY817:
## Support SQL file read-and-execute
It would be nice to read `*.sql` files and execute queries inside it.
Below I suggest some usages.
### Using with command line argument
```bash
$ gluesql --execute filename.sql
``` 
### Using inside the GlueSQL CLI
```bash
$ gluesql
[memory-storage] initialized

gluesql> .execute filename.sql
Executed queries in filename.sql

gluesql> .tables
╭────────╮
│ tables │
╞════════╡
│ hello  │
╰────────╯
```

earney:
I believe this enhancement has been implemented and is currently in production.  Can this issue be closed?

ever0de:
> I believe this enhancement has been implemented and is currently in production. Can this issue be closed?

Thanks for the report!

PR #496, opened 2022-03-14
Add support SQL file read-and-execute

pythonbrad:
#### Add a functionality to read-and-execute SQL file in GlueSQL CLI
* Using with command line argument
* Using inside the GlueSQL CLI
### Demo
a

### DIFF
--- a/cli/src/cli.rs
+++ b/cli/src/cli.rs
@@ -8,6 +8,9 @@ use {
     std::io::{Result, Write},
 };
 
+use std::fs::File;
+use std::io::Read;
+
 pub struct Cli<T, U, W>
 where
     U: GStore<T> + GStoreMut<T>,
@@ -82,6 +85,30 @@ where
                         println!("[error] {}\n", e);
                     }
                 },
+                Command::ExecuteFromFile(filename) => {
+                    if let Err(e) = self.load(&filename) {
+                        println!("[error] {}\n", e);
+                    }
+                }
+            }
+        }
+
+        Ok(())
+    }
+
+    pub fn load(&mut self, filename: &str) -> Result<()> {
+        let mut sqls = String::new();
+        // We open the file
+        File::open(filename)?.read_to_string(&mut sqls)?;
+        // We get all the sql query
+        for sql in sqls.split(";\n") {
+            if sql.trim().len() > 0 {
+                match self.glue.execute(&(sql.to_owned() + ";")) {
+                    Ok(payload) => self.print.payload(payload)?,
+                    Err(e) => {
+                        println!("[error] {}\n", e);
+                    }
+                }
             }
         }
 
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -18,6 +18,10 @@ struct Args {
     /// sled-storage path to load
     #[clap(short, long, parse(from_os_str))]
     path: Option<PathBuf>,
+
+    /// SQL file to execute
+    #[clap(short, long, parse(from_os_str))]
+    execute: Option<PathBuf>,
 }
 
 pub fn run() {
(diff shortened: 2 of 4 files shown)


cli/src/cli.rs
```diff
@@ -8,6 +8,9 @@ use {
     std::io::{Result, Write},
 };
 
+use std::fs::File;
+use std::io::Read;
```
panarch, line 12: To avoid code style conflict between contributors, we gather all `use` into a single line.
Then, `cargo fmt` will do everything for us.

e.g.
```rust
use {                                                          
    crate::{command::Command, helper::CliHelper, print::Print},
    gluesql_core::{                                            
        prelude::Glue,                                         
        store::{GStore, GStoreMut},                            
    },                                                         
    rustyline::{error::ReadlineError, Editor},                 
    std::{                                                     
        fs::File,                                              
        io::{Read, Result, Write},                             
    },                                                         
};
```

cli/src/cli.rs
```diff
@@ -82,6 +85,30 @@ where
...
+        }
+
+        Ok(())
+    }
+
+    pub fn load(&mut self, filename: &str) -> Result<()> {
+        let mut sqls = String::new();
+        // We open the file
+        File::open(filename)?.read_to_string(&mut sqls)?;
+        // We get all the sql query
```
panarch, line 103: Your codes are good & obvious enough, so it works itself as a document. These two supplementary comments look unnecessary.

cli/src/cli.rs
```diff
@@ -82,6 +85,30 @@ where
...
+        for sql in sqls.split(";\n") {
+            if sql.trim().len() > 0 {
```
panarch, line 105: `filter` can reduce the indent.

```rust
for sql in sqls.split(";\n").filter(|sql| sql.trim().len() > 0) {
    ...
}
```

cli/src/cli.rs
```diff
@@ -82,6 +85,30 @@ where
...
+                match self.glue.execute(&(sql.to_owned() + ";")) {
```
panarch, line 106: In the cli app, we use `;` to recognize the command.
However, for `glue.execute`, `;` at the end of line is optional.

```rust
self.glue.execute(sql)
```
would be enough

cli/src/cli.rs
```diff
@@ -82,6 +85,30 @@ where
...
+                    Ok(payload) => self.print.payload(payload)?,
+                    Err(e) => {
+                        println!("[error] {}\n", e);
```
panarch, line 109: It looks that we need `break;` in here.

panarch (approved):
Awesome!
Thanks a lot 👍 👍 👍 

ref. https://github.com/gluesql/gluesql/issues/481

cli/src/lib.rs
```diff
@@ -27,16 +31,25 @@ pub fn run() {
...
     }
 
-    fn run<T, U: GStore<T> + GStoreMut<T>>(storage: U) {
+    fn run<T, U: GStore<T> + GStoreMut<T>>(storage: U, input: Option<PathBuf>) {
         let output = std::io::stdout();
         let mut cli = Cli::new(storage, output);
 
+        if let Some(path) = input {
+            if let Err(e) = cli.load(path.as_path().to_str().expect("wrong path")) {
+                println!("[error] {}\n", e);
```
24seconds, line 49: [nit]
Right below code prints meesage using eprintln! macro. So how about using eprintln! here also?

24seconds:
Looks good to me

cli/src/cli.rs
```diff
@@ -82,6 +85,27 @@ where
...
+                }
+            }
+        }
+
+        Ok(())
+    }
+
+    pub fn load(&mut self, filename: &str) -> Result<()> {
+        let mut sqls = String::new();
+        File::open(filename)?.read_to_string(&mut sqls)?;
```
ever0de, line 101: <https://doc.rust-lang.org/stable/std/fs/struct.File.html#method.open>

[nit-picking]
How about using `AsRef<Path>` for extensibility?

ever0de (approved):
Good!

cli/src/lib.rs
ever0de: It's not std's println! that's used here.

I thought about this too, but noticed that there is a separate macro declaration above.

Why not change the name in another PR? @panarch

panarch: Interesting.. I think that we might not need to use `stderr` pipe at all in here.
In `lib.rs`,
https://github.com/gluesql/gluesql/blob/69211ebe96158e0ecf1d8fc373dbf0596899ad49/cli/src/lib.rs#L40-L42

This is the only case we use `eprintln!`. All else, we use `stdout`.
Considering this is CLI for interacting with GlueSQL, it would be better simply not to use `stderr`.
Then, we can think this as a different task and do it in the different PR.

How do you think? @24seconds @ever0de

ever0de: Looks good

cli/src/cli.rs
panarch: @ever0de 
Wow, you mean that replace `filename` type to `AsRef<Path>`?
```rust
pub fn load<P: AsRef<Path>>(&mut self, filename: P) -> Result<()> { 
    ..
```
Nice comment, thanks 👍

cli/src/lib.rs
24seconds: looks good

cli/src/cli.rs
pythonbrad: Ok!

PR #500, opened 2022-03-20
Implement shared memory storage

### LINKED ISSUE #493: support multi-threaded memory-sharing features
Can the MemoryStorage support multi-threaded memory-sharing features? Like supporting MVCC

ever0de:
resolved #493 

It is implemented by wrapping the items field of the existing `MemoryStorage` with `Arc<RwLock>`.


storages/multi-threaded-memory-storage/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "gluesql-multi-threaded-memory-storage"
+version = "0.10.2"
+edition = "2021"
+authors = ["Taehoon Moon <taehoon.moon@outlook.com>"]
```
panarch, line 5: You're the main author of this :)

```toml
authors = [
  "Jiseok CHOI <jiseok.dev@gmail.com>",
  "Taehoon Moon <taehoon.moon@outlook.com>",
]
```

storages/multi-threaded-memory-storage/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "gluesql-multi-threaded-memory-storage"
```
panarch, line 2: How about shorter name?
e.g. `gluesql-shared-memory-storage`
This storage itself is single threaded but it can be accessed by multiple different threads.
The term shared memory implies that memory data can be accessed my multiple process, so it would be good description for this.

storages/multi-threaded-memory-storage/Cargo.toml
ever0de: I'll add it! thank you

storages/multi-threaded-memory-storage/Cargo.toml
ever0de: I agree with the use of short names.

Can you explain a bit more what you mean by single-threaded storage?

ever0de:
Adding a test for concurrent support will open the PR.

storages/multi-threaded-memory-storage/Cargo.toml
panarch: Good point, "single-threaded storage" keyword may be not good to describe it.
The key difference between this and `MemoryStorage` is whether the storage can be accessed concurrently or not by default.
I thought `MultiThreaded-` meaning as a kind of that the storage itself spawns multiple threads to handle data faster.

ever0de: > Good point, "single-threaded storage" keyword may be not good to describe it. The key difference between this and `MemoryStorage` is whether the storage can be accessed concurrently or not by default. I thought `MultiThreaded-` meaning as a kind of that the storage itself spawns multiple threads to handle data faster.

Oh, I understand!
thank you

ever0de:
Memo <https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use>

ever0de:
The discussion proceeded in this way.

Let's start with a simple SharedMemoryStorage operation. (apart from performance improvements)
This storage has MemoryStorage and is wrapped in Mutex.

When the user accesses, the MemoryStorage will be cloned (Arc) and used.

ever0de:
depend on #597

storages/shared-memory-storage/tests/concurrent_access.rs
```diff
@@ -0,0 +1,47 @@
...
+
+    let payloads = glue.execute("SELECT * FROM Thread").unwrap();
+    assert_eq!(payloads.len(), 1);
+
+    let payload = &payloads[0];
+    let rows = match &payload {
+        Payload::Select { rows, .. } => rows.iter().flatten().collect::<Vec<_>>(),
+        _ => unreachable!(),
+    };
+    assert_eq!(rows.len(), 2);
```
panarch, line 46: How about simplifying by exposing the entire result struct?
Payload result is simple in here so it would be good to..
```rust
let actual = glue.execute("SELECT * FROM Thread").unwrap();
let expected = vec![Payload::Select {
    labels: vec!["id".to_owned()],
    rows: vec![vec![Value::I64(1)], vec![Value::I64(2)]],
}];
assert_eq!(actual, expected);
```
This might look more readable.

ever0de: fixed it! thanks

panarch (approved):
Awesome!!
Looks all great 👍 👍 👍

PR #501, opened 2022-03-24
Implement TryFrom between Value and serde_json::Value

panarch:
Implement two conversions
* `TryFrom<Value> for serde_json::Value`
* `TryFrom<serde_json::Value> for Value`

Note.
Some `Value` types converge into the same `serde_json::Value` type (`serde_json::Value::String`)
Loss of information occurs when conversion from `Value` to `serde_json::Value`.

e.g.
```
Value::Date => serde_json::Value::String
Value::Timestamp => serde_json::Value::String
Value::Time => serde_json::Value::String
Value::Interval => serde_json::Value::String
```

Therefore,
```rust
use serde_json::Value as JsonValue;

let original_value: Value = Value::Time(NaiveTime.from(14, 0, 0)); // "14:00:00"
let converted_value: Value = Value::try_from(JsonValue::try_from(original_value).unwrap()).unwrap();

assert_eq!(converted_value, Value::String("14:00:00".to_owned());
// converted_value has `String` type, not `Time`.
```


ever0de (approved):
look good

LGTM

PR #502, opened 2022-03-26
Bump up toolchain ver. from 1.58 -> 1.59

MRGRAVITY817:
# Bump up toolchain ver. from 1.58 -> 1.59

The latest version of the Rust is 1.59, and some of dependencies use the latest one.
It's encouraged to use latest one as well.

### Better monad pattern 
As for the newer version, `clippy` encourages to rewrite monad pattern from `map().flatten()` into `and_then()`.

### Remove unnecessarily creating ownership
For functions that require `AsRef<str>` as an argument, there's no need to convert `ref/slice` string to `String`.


panarch (approved):
Thanks a lot 👍

PR #505, opened 2022-03-28
Fix value/ big_endian module to handle negative integers correctly,

### LINKED ISSUE #503: Indexing/serialising negative numbers
Per @KyGost & @panarch discord conversation, starting:
https://discord.com/channels/780298017940176946/780298017940176949/957870515706535997
ending:
https://discord.com/channels/780298017940176946/780298017940176949/957890136442560552

`i64::to_be_bytes()` doesn't convert negatives such that they will be sorted as less than positives.

Per @panarch simple and effective solution is to prepend a positive/negative byte.

Example solution:
https://github.com/KyGost/multisql/blob/c7c9d8480ddfdd59cca67d397ddd87da0a924a78/src/data/value/big_endian.rs#L18

panarch:
For i8 and i64 types, add sign byte to the prefix.
Add big_endian negative integer comparison test cases.

fix #503

### DIFF
--- a/core/src/data/value/big_endian.rs
+++ b/core/src/data/value/big_endian.rs
@@ -18,16 +18,24 @@ impl Value {
                     vec![VALUE, 0]
                 }
             }
-            Value::I8(v) => [VALUE]
-                .iter()
-                .chain(v.to_be_bytes().iter())
-                .copied()
-                .collect::<Vec<_>>(),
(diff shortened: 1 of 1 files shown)


ever0de (approved):
Oh... now I know the sign wasn't being checked.
thanks to the informant

LGTM

PR #506, opened 2022-03-29
Correct the plural of index from `indexes` to `indices`

MRGRAVITY817:
# Indexes -> Indices
Just a tiny spelling correction.
The plural of the `index` should be `indices`.

## Reference
https://www.merriam-webster.com/dictionary/indices


panarch:
https://www.merriam-webster.com/dictionary/indexes

Both `indexes` & `indices` look acceptable, then I think `indexes` is better for variable names because `indexes` has more same form with singular word `index`.

MRGRAVITY817:
That sounds reasonable! I'll close this.

PR #507, opened 2022-04-01
Minor improvements

ever0de:
- It seems that receiving a variable that implements `AsRef<str>` would have a better experience for the user than receiving only the `&str` type
- I create a dialect every time I parse a sql statement, but it doesn't seem to be necessary.
    - It would be nice to make this for the user to choose later.

### DIFF
--- a/core/src/glue.rs
+++ b/core/src/glue.rs
@@ -27,7 +27,7 @@ impl<T, U: GStore<T> + GStoreMut<T>> Glue<T, U> {
         }
     }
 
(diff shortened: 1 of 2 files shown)


MRGRAVITY817:
> * It seems that receiving a variable that implements `AsRef<str>` would have a better experience for the user than receiving only the `&str` type

I agree with this. It adds better flexibility, which is recommended in Rust API Guidelines.

panarch (approved):
I am also for this, thanks a lot!
Both looks nice, merging 👍

Issue #475, opened 2022-01-26
Support DESCRIBE TABLES or INFORMATION_SCHEMA or something else to get tables schema


boomkim:
Currently, we cannot check table's schema. 
Let's have getter or print_schema() to get table's schema.

earney:
Just my opinion but shouldn't an information_schema be part of a db management system instead of a library such as gluesql?  I know the boundary is a gray area, but I'm wondering what others think?

panarch:
@earney Thanks for the opinion

> But yes, something like a describe table, would be a good addition.

I think that this kind of feature is necessary to have because gluesql covers sql parser layer so without this, users cannot support `DESCRIBE TABLE` like queries. It is not easy to say what would be the enough returning information about this, but at least we need to provide something.

Somewhat this kind of approach may work (extending metadata)
e.g.
```rust
pub trait Metadata<T> where T: Display + Debug {
  ..
  fn describe_schema(table_name) -> Result<T> { // or struct containing T
    ..
  }
}
```

earney:
Show columns from table has been implemented.  Can we close this issue or is there still something this issue addresses that is not fixed yet?

ever0de:
resolved #517 

> Show columns from table has been implemented. Can we close this issue or is there still something this issue addresses that is not fixed yet?

Thanks @earney 

If you register this function at the beginning of PR conversation, it will be closed automatically. 
Can you check it out?
ex) resolved #517 

- <https://stackoverflow.com/questions/12235620/automatically-closing-issue-from-pull-request-in-github>
- <https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword>

PR #509, opened 2022-04-02
Show columns implementation (describe table)

earney:
This pull request implements show columns so that we can see how a table was defined.
for example:
create table mytable(id int, name text);
show columns from mytable;
will show the following:
Field    Type
id         Int
name   text


earney:
This resolves the enhancement suggested in issue #475.

ever0de:
Rust's cargo function has a command called fmt .
See: <https://github.com/gluesql/gluesql/runs/5803579821?check_suite_focus=true>

core/src/data/interval/primitive.rs
```diff
@@ -199,4 +199,15 @@ mod tests {
...
+    #[test]
+    fn arithmetic_microsecond() {
+        use Interval::*;
+
+        assert_eq!(Microsecond(2) * 3_i8, Microsecond(6));
+        assert_eq!(2_i8 * Microsecond(3), Microsecond(6));
+
+        assert_eq!(Microsecond(2) * 3_f64, Microsecond(6));
+        assert_eq!(2_f64 * Microsecond(3), Microsecond(6));
+    }
```
panarch, line 212: This does not look related to this issue. Code itself looks nice, thanks :)
You can make the separate PR for this, then I'll approve right after.

Issue #283, opened 2021-08-10
Support Function :: ABS, SIGN


Goinging98:
# Support Fuction :: `ABS`, `SIGN`

## ABS
`ABS` is to calculate the absolute value of a number.
Return type will be same as input.
```
-- Reference from Postgresql 
select ABS(-15.5);    -> 15.5
select ABS(50 - 100); -> 50 
```

## SIGN
`SIGN` is for sign of the argument (-1, 0, +1)
Return type will be same as input.
```
-- Reference from Postgresql 
select sign(-8.4);  -> -1 
select sign(0);     -> 0 
select sign(8.4);   -> 1
```

## Reference from ..
1. https://www.postgresql.org/docs/9.1/functions-math.html
2. https://www.postgresqltutorial.com/postgresql-abs/

earney:
pull request #510 implements this issue.  I believe it can be closed.

Goinging98:
Thank you!

PR #510, opened 2022-04-02
add ABS and SIGN functions

earney:
I noticed there is another pull request for adding ABS and SIGN, and thought I would try an implementation myself.
This also resolves issue #283

### DIFF
--- a/core/src/ast/function.rs
+++ b/core/src/ast/function.rs
@@ -7,7 +7,7 @@ use {
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, Display)]
 pub enum Function {
     #[strum(to_string = "ABS")]
-    ABS(Expr),
+    Abs(Expr),
     #[strum(to_string = "LOWER")]
     Lower(Expr),
     #[strum(to_string = "UPPER")]
--- a/core/src/ast/function.rs
+++ b/core/src/ast/function.rs
@@ -6,6 +6,8 @@ use {
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, Display)]
 pub enum Function {
+    #[strum(to_string = "ABS")]
+    ABS(Expr),
     #[strum(to_string = "LOWER")]
     Lower(Expr),
     #[strum(to_string = "UPPER")]
@@ -88,6 +90,8 @@ pub enum Function {
     Reverse(Expr),
     #[strum(to_string = "REPEAT")]
     Repeat { expr: Expr, num: Expr },
+    #[strum(to_string = "SIGN")]
+    Sign(Expr),
     #[strum(to_string = "SUBSTR")]
     Substr {
         expr: Expr,
--- a/core/src/executor/evaluate/function.rs
+++ b/core/src/executor/evaluate/function.rs
@@ -232,7 +232,9 @@ pub fn abs(name: String, n: Evaluated<'_>) -> Result<Value> {
 
 pub fn sign(name: String, n: Evaluated<'_>) -> Result<Value> {
     let x = eval_to_float!(name, n);
-    if x == 0.0 {return Ok(Value::I8(0))}
+    if x == 0.0 {
+        return Ok(Value::I8(0));
+    }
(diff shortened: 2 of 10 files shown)


ever0de (approved):
LGTM
Let's wait for the maintainer's review.

ever0de (commented):
Can you add test cases related to ABS and SIGN to the `test-suite` crates?
Check this: <https://github.com/gluesql/gluesql/tree/main/test-suite/src/function>

core/src/ast/function.rs
```diff
@@ -6,6 +6,8 @@ use {
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, Display)]
 pub enum Function {
+    #[strum(to_string = "ABS")]
+    ABS(Expr),
```
panarch, line 10: Let's simply follow the Rust's naming patterns.
`Abs` would be good to use for this.

test-suite/src/function/abs.rs
```diff
@@ -0,0 +1,95 @@
...
+            Ok(Payload::Create),
+        ),
+        (
+            r#"INSERT INTO SingleItem VALUES (0)"#,
+            Ok(Payload::Insert(1)),
+        ),
+        (
+            "SELECT abs(1) AS abs1, 
+                    abs(-1) AS abs2, 
+                    abs(+1) AS abs3
```
panarch, line 19: sql is case-insensitive, so there's no problem to execute this statement.
However we used uppercase letters for all other keyword `SELECT`, `AS` and `FROM` only except for `abs` which is also the same keyword like others.
It looks better to use uppercase letters for all keyword for the consistency.

panarch (changes requested):
Thanks a lot for the contribution, welcome 👍 👍 👍  

There are.. a few small requests mainly about tests.
0. Two inline comments
1. Let's also have integration tests for `SIGN` function.
2. To make newly added integration tests to work, you also need to edit these two files
[test-suite/src/function/mod.rs](https://github.com/gluesql/gluesql/blob/main/test-suite/src/function/mod.rs)
[test-suite/src/lib.rs](https://github.com/gluesql/gluesql/blob/main/test-suite/src/lib.rs)
Then, `cargo test abs` or `cargo test sign` will show that new tests work.

core/src/ast/function.rs
earney: Thanks..  will make necessary modifications.

earney:
I have pushed the changes.  Please let me know if I need to make additional modifications.
I did find one issue.
when someone executes a query such as:  select -0 from mytable;  The -0 automatically gets modified to 0.  I'm not sure if this is part of the sql standard but this does affect a few of my test cases, which I had to give the responses it was looking for instead of the correct response.

earney:
> 

These suggestions make sense to me.  I will look into implementing them.  Thanks for the feedback!

earney:
This pull request should be ready to merge with main.  If you find other issues (or if I missed an issue), please let me know.

ever0de (approved):
Looks good thanks!

core/src/executor/evaluate/function.rs
```diff
@@ -215,6 +215,25 @@ pub fn substr(
 
 // --- float ---
 
+pub fn abs(name: String, n: Evaluated<'_>) -> Result<Value> {
+    match n.try_into()? {
+        Value::I8(v) => Ok(Value::I8(v.abs())),
+        Value::I64(v) => Ok(Value::I64(v.abs())),
+        Value::Decimal(v) => Ok(Value::Decimal(v.abs())),
+        Value::F64(v) => Ok(Value::F64(v.abs())),
+        Value::Null => Ok(Value::Null),
```
panarch, line 224: ❤️

panarch (approved):
Looks all great!! Thanks a lot 👍 👍 👍

Until the first PR merged, to run GitHub Actions, the maintainer needs to approve to run so it takes some time to check whether your commit pass the tests or not.
From now on, your first PR has merged :) so, Action will be triggered just after you push the new code to PRs.

Issue #511, opened 2022-04-03
Benchmark Suite


KyGost:
Per discord messages, starting [here](https://discord.com/channels/780298017940176946/780298017940176949/960070448685674537), it'd be awesome to have benchmark suites for both GlueSQL & MultiSQL for comparison.

Could be useful for highlighting areas where we differ and as such can improve.

Requesting assignment to myself; I'll look into this either tonight or at some point.

KyGost:
KyGost/multisql#45

PR #513, opened 2022-04-04
Benchmarks

KyGost:
See: #511
Resolves: #511

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ sled-storage = { package = "gluesql_sled_storage", path = "./storages/sled-stora
 
 [dev-dependencies]
 futures = "0.3"
+criterion = {version = "0.3.5", features = ["html_reports"] }
 
 [features]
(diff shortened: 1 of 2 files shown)


KyGost:
Works but probably not set up in the manor desired.

Feel free to commit changes to this PR

panarch:
Awesome :) Thanks

There is an one thing which needs to be considered.
The purpose of `test-suite` is providing test cases to other storage implementations.
`test-suite` itself does not execute any thing. It is simply bunch of test cases.
Benchmarks are also the good example which can be covered by `test-suite`.
However, it is not supposed to be executed in here.

We can define benchmark test codes in `test-suite`, but should not be run inside `test-suite`.
If we want to test `SledStorage`, then we need to add benchs to `storages/sled-storage` and import `test-suite` bench test cases.
`memory-storage` & `sled-storage` import `test-suite` but the opposite should not happen.

PR #516, opened 2022-04-06
add abs and sign functions

earney:
This is a new PR for adding Abs and Sign functions.

### DIFF
--- a/core/src/executor/evaluate/function.rs
+++ b/core/src/executor/evaluate/function.rs
@@ -221,12 +221,8 @@ pub fn abs(name: String, n: Evaluated<'_>) -> Result<Value> {
         Value::I64(v) => Ok(Value::I64(v.abs())),
         Value::Decimal(v) => Ok(Value::Decimal(v.abs())),
         Value::F64(v) => Ok(Value::F64(v.abs())),
-        Value::Null => {
-            return Ok(Value::Null);
-        }
-        _ => {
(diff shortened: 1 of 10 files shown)


core/src/executor/evaluate/function.rs
```diff
@@ -215,6 +215,29 @@ pub fn substr(
...
+        Value::I64(v) => Ok(Value::I64(v.abs())),
+        Value::Decimal(v) => Ok(Value::Decimal(v.abs())),
+        Value::F64(v) => Ok(Value::F64(v.abs())),
+        Value::Null => {
+            return Ok(Value::Null);
+        }
+        _ => {
+            return Err(EvaluateError::FunctionRequiresFloatValue(name).into());
+        }
+    }
```
ever0de, line 230:
```suggestion
        Value::Null => Ok(Value::Null),
        _ => Err(EvaluateError::FunctionRequiresFloatValue(name).into())
    }
```
The rest of the value passed in the `match` is used as the return value, so I think you can delete `return` too

panarch:
I can squash and merge to bundle the commits in the PR so you don't need to hesitate adding new commits as many as you want.

panarch:
You can run these commands below before the push. It would be useful.
`cargo fmt`, `cargo clippy --all-targets` and `cargo test` in the root of the repo.

earney:
Sorry about that..  I think I have it fixed now.

PR #517, opened 2022-04-08
Show columns take2

earney:
This pull request is similar to the show columns  (describe table) except I remove the unrelated tests from this pull request.


panarch:
Thanks a lot :)
Awesome! Feature really looks nice.
There are some review comments I need to do, but sorry for the delay.
It's getting late because of the lack of my time for now.
I'll be back at most in a week and leave comments.

For now...
could you add some tests for this?
Two points would be good to consider to add tests.
* `gluesql/cli/src/print.rs` module - unit tests
* `gluesql/test-suite/src/metadata.rs` module - integration tests

earney:
I'm kind of new to rust, and I'm not sure what needs to go on line 20 of metadata.rs.  Can someone give me some help on this?
Thanks!

panarch:
> I'm kind of new to rust, and I'm not sure what needs to go on line 20 of metadata.rs. Can someone give me some help on this? Thanks!

Nice, I think that was really good comments, now I also got the point. Thanks :)
Here, for `SHOW COLUMNS --`, this looks something that is related to `metadata` but it is not.

For example, `SHOW TABLES` is related to `metadata` because using existing store traits methods there's no way to return table list of the storage. Therefore, we added a new method for fetching schema names, `Metadata -> schema_names`

However, in the case of `SHOW COLUMNS --`, as you already implemented well without using any `metadata` related methods, this only requires `fetch_schema` method, nothing else required.
This means that we don't need to handle `Statement::ShowColumns { .. }` as an optional feature because `Store` trait is mandatory trait to implement for using GlueSQL.

This looks that it would be good to put `Statement::ShowColumns` code in `src/executor/execute.rs` to the below of `Statement::Query` (above the `Statement::ShowVariable`)
image

With this updates and having tests for `SHOW COLUMNS - `, we can go to the next step.

earney:
> > I'm kind of new to rust, and I'm not sure what needs to go on line 20 of metadata.rs. Can someone give me some help on this? Thanks!
> 
> Nice, I think that was really good comments, now I also got the point. Thanks :)
> Here, for `SHOW COLUMNS --`, this looks something that is related to `metadata` but it is not.
> 
> For example, `SHOW TABLES` is related to `metadata` because using existing store traits methods there's no way to return table list of the storage. Therefore, we added a new method for fetching schema names, `Metadata -> schema_names`
> 
> However, in the case of `SHOW COLUMNS --`, as you already implemented well without using any `metadata` related methods, this only requires `fetch_schema` method, nothing else required.
> This means that we don't need to handle `Statement::ShowColumns { .. }` as an optional feature because `Store` trait is mandatory trait to implement for using GlueSQL.
> 
> This looks that it would be good to put `Statement::ShowColumns` code in `src/executor/execute.rs` to the below of `Statement::Query` (above the `Statement::ShowVariable`)
> image
> 
> With this updates and having tests for `SHOW COLUMNS - `, we can go to the next step.
> 

Ok.  I have made the changes and added some tests.

earney:
implements issue #475

ever0de:
tip)
<https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue>
You can close the issue automatically using the link above.

ever0de (approved):
Looks good

core/src/executor/execute.rs
```diff
@@ -309,7 +312,27 @@ pub async fn execute<T, U: GStore<T> + GStoreMut<T>>(
...
+            });
+
+            let mut output: Vec<Vec<String>> = vec![vec![]];
+
+            for key in keys {
+                let mut row = Vec::new();
+                row.push(key.name);
+                row.push(format!("{:?}", key.data_type));
+                output.push(row);
+            }
```
panarch, line 333: The above codes can be simplified like this without using `mut`.

```rust
let output = keys
    .into_iter()
    .map(|key| vec![key.name, format!("{:?}", key.data_type)])
    .collect();
```

cli/src/print.rs
```diff
@@ -242,5 +251,39 @@ mod tests {
...
+│ name      Text │
+│ isabear   Bool │
+╰────────────────╯",
+            Payload::ShowColumns {
+                rows: vec![
+                    vec!["id".to_string(), "Int".to_string(),],
+                    vec!["name".to_string(), "Text".to_string(),],
+                    vec!["isabear".to_string(), "Bool".to_string(),],
+                ],
+            }
```
panarch, line 286: These two test cases look redundant.
Lack of tests are the issue, but we also don't need to add redundant ones.
Let's use the last one which has more info.

core/src/executor/execute.rs
```diff
@@ -309,7 +312,27 @@ pub async fn execute<T, U: GStore<T> + GStoreMut<T>>(
 
             Ok((storage, Payload::Select { labels, rows }))
         }
+        Statement::ShowColumns { table_name } => {
+            let keys = try_block!(storage, {
+                let table_name = get_name(table_name)?;
+                let Schema { column_defs, .. } = storage
+                    .fetch_schema(table_name)
+                    .await?
+                    .ok_or_else(|| ExecuteError::TableNotFound(table_name.to_owned()))?;
```
panarch, line 321: The current integration test `show columns` does not cover this error.
Let's add one for this :)

core/src/executor/execute.rs
```diff
@@ -36,6 +36,9 @@ pub enum ExecuteError {
 
 #[derive(Serialize, Debug, PartialEq)]
 pub enum Payload {
+    ShowColumns {
+        rows: Vec<Vec<String>>,
+    },
```
panarch, line 41: `Vec<Vec<String>>` representation covers more than what we actually want to describe.
Rather than using superset, let's use identical data structure for this.
```rust
// use ast::DataType

ShowColumns {
    columns: Vec<(String, DataType)>,
}
```

Then, `ast::DataType` is explicitly exposed to the user.
Let's add `ast::DataType` to `prelude`.
We don't need to use `prelude` in `core`, but it would be good to use in `test-suite` which you already used well.

In `/core/src/lib.rs`,
image

panarch (changes requested):
Thanks 👍 👍 
Sorry for the late review
Now, after resolving a few ones and let's ship this to the main!

earney:
I'm getting the following error, and I'm stuck.  Anyone got some suggestions on how to fix it?  I've pushed my code to the repo.

``
cargo test --workspace
   Compiling gluesql-core v0.10.2 (/home/billy/projects/gluesql/core)
   Compiling gluesql-test-suite v0.10.2 (/home/billy/projects/gluesql/test-suite)
   Compiling gluesql_sled_storage v0.10.2 (/home/billy/projects/gluesql/storages/sled-storage)
   Compiling gluesql_memory_storage v0.10.2 (/home/billy/projects/gluesql/storages/memory-storage)
   Compiling gluesql-cli v0.10.2 (/home/billy/projects/gluesql/cli)
warning: unused import: `gluesql_core::ast::DataType`
 --> cli/src/print.rs:3:5
  |
3 |     gluesql_core::ast::DataType,
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: `gluesql-cli` (lib) generated 1 warning
error[E0433]: failed to resolve: use of undeclared type `DataType`
   --> cli/src/print.rs:266:40
    |
266 |                     ("id".to_string(), DataType::Int),
    |                                        ^^^^^^^^ use of undeclared type `DataType`

error[E0433]: failed to resolve: use of undeclared type `DataType`
   --> cli/src/print.rs:267:42
    |
267 |                     ("name".to_string(), DataType::Text),
    |                                          ^^^^^^^^ use of undeclared type `DataType`

error[E0433]: failed to resolve: use of undeclared type `DataType`
   --> cli/src/print.rs:268:45
    |
268 |                     ("isabear".to_string(), DataType::Boolean),
    |                                             ^^^^^^^^ use of undeclared type `DataType`

For more information about this error, try `rustc --explain E0433`.
``

cli/src/print.rs
```diff
@@ -1,5 +1,6 @@
 use {
     comfy_table::{modifiers::UTF8_ROUND_CORNERS, presets::UTF8_BORDERS_ONLY, Row, Table},
+    gluesql_core::ast::DataType,
```
panarch, line 3: unit test module has independent scope, you need to move this import to 
image

test-suite/src/showcolumns.rs
```diff
@@ -0,0 +1,35 @@
...
+        Ok(select!(
+            Field               | Type
+            Str                 | Str;
+            "id".to_owned()       "Int".to_owned();
+            "rate".to_owned()     "Float".to_owned();
+            "flag".to_owned()     "Boolean".to_owned();
+            "text".to_owned()     "Text".to_owned();
+            "null_value".to_owned()   "Text".to_owned()
+        )),
+        r#"Show columns from mytable"#
```
panarch, line 28: `select!` macro works with `Payload::Select` but `SHOW COLUMNS` returns `Payload::ShowColumns`.
We can simply set expected result without using macro.

test-suite/src/lib.rs
```diff
@@ -19,6 +19,7 @@ pub mod nested_select;
 pub mod nullable;
 pub mod order_by;
 pub mod ordering;
+pub mod showcolumns;
```
panarch, line 22: You also need to specify the test like this below, then gluesql integration test will recognize it.
image

core/src/executor/execute.rs
```diff
@@ -36,6 +36,9 @@ pub enum ExecuteError {
 
 #[derive(Serialize, Debug, PartialEq)]
 pub enum Payload {
+    ShowColumns {
+        rows: Vec<(String, DataType)>,
```
panarch, line 40: The command is `SHOW COLUMNS`, then how about using `columns` rather than `rows`?
Or, this might be enough to declare with `ShowColumns(Vec<(String, DataType)>)` because it has an single value and this looks obvious that the value inside `ShowColumns` has `columns`.

cli/src/print.rs
earney: Thanks.. I knew it was something like what you suggested.  Nothing I tried seemed to work.   Now i"m having an issue with the test-suite-src/showcolumns.rs tests.  I've tried creating my own showcolumns macro, but what is occuring in the (copied from select macro) is a bit beyond my skills at this time.  Can you take a look?
Thanks!

core/src/executor/execute.rs
earney: +1

test-suite/src/showcolumns.rs
```diff
@@ -1,29 +1,77 @@
...
+            "dec".to_owned()      DataType::Decimal.to_string();
+            "flag".to_owned()     DataType::Boolean.to_string();
+            "text".to_owned()     DataType::Text.to_string();
+            "DOB".to_owned()      DataType::Date.to_string();
+            "Tm".to_owned()       DataType::Time.to_string();
+            "ival".to_owned()     DataType::Interval.to_string();
+            "tstamp".to_owned()   DataType::Timestamp.to_string();
+            "uid".to_owned()      DataType::Uuid.to_string();
+            "hash".to_owned()     DataType::Map.to_string();
+            "glist".to_owned()    DataType::List.to_string()
```
panarch, line 74: Oh, I think adding a new macro only for `ShowColumns` is too much.
`select!` simply returns `Payload::Select` and it is for others to read easier than pure `Vec<Vec<Value>>` representation in Rust.
However, in the case of `showcolumns!`, It is simply a type of `Vec<(String, DataType)>` so it would be readable enough to be only with Rust codes without macros.

panarch: You also don't need to convert `String` and `DataType` to `Value`.
Both `String` and `DataType` has `PartialEq` implementation. We can directly compare with each other
https://github.com/gluesql/gluesql/blob/1213d8288a427745862c0b1c40f83fa057033fd3/core/src/ast/data_type.rs#L3-L5

test-suite/src/lib.rs
earney: Thanks for point it out.  I think I got it resolved.

test-suite/src/showcolumns.rs
earney: I believe this is resolved as well.

test-suite/src/showcolumns.rs
earney: Thanks for the info..  It helped quite a bit..

PR #519, opened 2022-04-09
Fix CLI --execute {file}.sql option to support CRLF line break,

panarch:
Replace ";\n" to ";" in --execute line splitting.

SQL files which are generated in Windows OS usually use CRLF symbol for line-break.
Current CLI `--execute` option was handling only the first statement and ignores the remainders.
Fix splitting symbol not to contain `\n` to resolve this issue.


ever0de (approved):
👍

PR #521, opened 2022-04-13
Concat function implementation

ghost:
I have a basic concat function working (with exactly 3 arguments).  I am struggling to get this working with a variable number of arguments.  Any suggestions of how best to implement this?


earney:
please make this a draft PR.   I don't see the option on my end to make this a draft.
Thanks!

earney:
Okay, guys.  This turned out to be easier to implement then I thought.  Can you take a look at this and see if there are any adjustments that need to be made?

Thanks!

core/src/executor/evaluate/function.rs
```diff
@@ -50,6 +50,22 @@ macro_rules! eval_to_float {
 
 // --- text ---
 
+pub fn concat(name: String, exprs: Vec<Evaluated<'_>>) -> Result<Value> {
+    let mut s = String::from("");
+
+    for expr in exprs {
+        match expr.try_into()? {
+            Value::Str(value) => s.push_str(&value),
+            Value::Null => {}
```
ever0de, line 59: In MySQL, if a NULL value exists, the result string is NULL.
I think this is more natural than being treated as a blank space.

What do you think?

core/src/executor/evaluate/stateless.rs
```diff
@@ -126,6 +126,14 @@ fn evaluate_function<'a>(
...
     match func {
         // --- text ---
+        Function::Concat(exprs) => {
+            let mut vec: Vec<Evaluated<'_>> = vec![];
+            for expr in exprs {
+                vec.push(eval(expr)?);
+            }
+
+            f::concat(name(), vec) // _s).await?),
+        }
```
ever0de, line 136:
```suggestion
            let exprs = exprs.iter().map(eval).collect::<Result<_>>()?;

            f::concat(name(), exprs)
```

How about code like this?

core/src/translate/function.rs
```diff
@@ -142,6 +142,14 @@ pub fn translate_function(sql_function: &SqlFunction) -> Result<Expr> {
         .collect::<Result<Vec<_>>>()?;
 
     match name.as_str() {
+        "CONCAT" => {
+            check_len_range(name, args.len(), 2, 255)?;
```
ever0de, line 146: It would be good if the `CONCAT` function works if the length is 1 or more.
Is there a reason there is a 255 count limit?

ever0de: e.g)
```sql
mysql> select concat("Adsf");
+----------------+
| concat("Adsf") |
+----------------+
| Adsf           |
+----------------+
1 row in set (0.00 sec)
```

ever0de (commented):
generally good

I would like to talk about the following.
1. When a NULL value is entered
2. Length limit (2, 255)

core/src/translate/function.rs
earney: I went with a limit of 2 to 255 because of the major sql databases (postgres or maybe it was MSSQL) had those limit  We can definitely change those limits.   We can go with 1 for the lower limit and no upper limit

core/src/executor/evaluate/stateless.rs
earney: Awesome!  I was hoping to convert it so that it was more rust like.   I still don't understand it, but I guess in time I will. ;)

core/src/executor/evaluate/function.rs
earney: While implementing the function I went with the postgres definition for concat (which ignores NULL):
https://www.postgresql.org/docs/14/functions-string.html
But you are correct for mysql the whole statement becomes NULL if a single arg is null. 
https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_concat  

I coded it this way be accident, but now I am thinking that since || returns NULL if an argument is NULL, maybe concat should do the opposite (ignores NULL) so that users have an option for either result. What do you think?

core/src/translate/function.rs
ever0de: I have a personal question. (since the database I mainly use is MySQL)
PostgreSQL, why limit the length in MSSQL?
I wonder what the definition for it is

core/src/executor/evaluate/function.rs
ever0de: Obviously... it would be nice to give the user an option.
How about providing it in a new PR (issue)?

core/src/translate/function.rs
```diff
@@ -73,6 +73,20 @@ fn check_len_range(
...
 
+fn check_len_min(name: String, found: usize, expected_minimum: usize) -> Result<()> {
+    if found >= expected_minimum {
+        Ok(())
+    } else {
+        Err(TranslateError::FunctionArgsLengthNotWithinRange {
+            name,
+            expected_minimum,
+            expected_maximum,
+            found,
```
ever0de, line 84: How about creating a new error type?
'expected_minimum'

core/src/translate/function.rs
earney: I've used mostly MySQL and Server Server (MSSQL).   I've played a bit with postgres just for fun.  
Here's info on SQL Server's implementation of concat: https://docs.microsoft.com/en-us/sql/t-sql/functions/concat-transact-sql?view=sql-server-ver15

They seem to have gone the same route as postgres and just ignore NULL values by converting them to the empty space.

core/src/executor/evaluate/function.rs
earney: I really could care less which way we go with Concat.   If you want I can have it mimic mysql.  By the way, for most functions should we try to emulate mysql's behavior?

ever0de: no.
I think it's a process of finding the best one.

core/src/translate/function.rs
earney: I believe the number of args issue has been resolved.  If not, let me know.

core/src/executor/evaluate/function.rs
earney: Went ahead and changed concat so that it returns Null when concat contains a Null arg.   I will submit an issue so that we can discuss if we want this functionality or if we want to change it so that concat ignores null args

core/src/translate/function.rs
```diff
@@ -142,6 +155,14 @@ pub fn translate_function(sql_function: &SqlFunction) -> Result<Expr> {
         .collect::<Result<Vec<_>>>()?;
 
     match name.as_str() {
+        "CONCAT" => {
+            check_len_min(name, args.len(), 1)?;
+            let mut exprs: Vec<Expr> = vec![];
+            for arg in args {
+                exprs.push(translate_expr(arg)?);
+            }
```
ever0de, line 163:
```suggestion
            let exprs = args
                .into_iter()
                .map(translate_expr)
                .collect::<Result<Vec<_>>>()?;
```

ever0de (approved):
Thank you for reflecting my opinion.
Other than the following code improvements, I like it.

In case of disagreement with each other, we discuss it with the maintainer.

Can we get some answers to @panarch comments?
1. If NULL is entered
2. Limit the number of arguments

core/src/plan/expr/function.rs
```diff
@@ -114,6 +115,7 @@ impl Function {
                 start: expr2,
                 count: Some(expr3),
             } => Exprs::Tripple([expr, expr2, expr3].into_iter()),
+            Self::Concat(exprs) => Exprs::VariableArgs(exprs.iter()),
```
panarch, line 118: Nice decision to add `VariableArgs` to here :)
Let's also have a unit test for this.
You can add it to the below of this file.

core/src/executor/evaluate/function.rs
```diff
@@ -50,6 +50,22 @@ macro_rules! eval_to_float {
...
+            Value::Str(value) => s.push_str(&value),
+            Value::Null => return Ok(Value::Null),
+            _ => {
+                return Err(EvaluateError::FunctionRequiresStringValue(name).into());
+            }
+        }
+    }
+
+    Ok(Value::Str(s))
+}
```
panarch, line 67: `Value` itself has `concat` method using `||` binary operator.
https://github.com/gluesql/gluesql/blob/1213d8288a427745862c0b1c40f83fa057033fd3/core/src/data/value/mod.rs#L165-L170

ref. [test-suite/ concat.rs](https://github.com/gluesql/gluesql/blob/main/test-suite/src/concat.rs)

If you don't have intend `||` and `CONCAT` to have different behaviors, then how about to use `Value::concat`?

earney:
> Can we get some answers to @panarch comments?
> 
>     1. If NULL is entered
> 
>     2. Limit the number of arguments

1 Right now if NULL is entered  => concat("abc", NULL) => NULL,    null is returned (just like ||)
2. Right now concat requires at least 1 arg, and there is no maximum number of arguments.

core/src/executor/evaluate/function.rs
earney: concat has a variable number of arguments, (not just  2).

core/src/translate/function.rs
earney: 1 Right now if NULL is entered => concat("abc", NULL) => NULL, null is returned (just like ||)
2. Right now concat requires at least 1 arg, and there is no maximum number of arguments.

core/src/executor/evaluate/function.rs
earney: I think the concat function should have a different behavior (ie, ignore nulls), but it was suggested that I create another issue so we can discuss this.

earney:
Unless I'm overlooking something, I believe this is ready to be merged with main.

core/src/executor/evaluate/mod.rs
```diff
@@ -211,6 +211,14 @@ async fn evaluate_function<'a, T>(
 
     match func {
         // --- text ---
+        Function::Concat(exprs) => {
+            let mut vec: Vec<Evaluated<'_>> = vec![];
+            for expr in exprs {
+                vec.push(eval(expr).await?);
+            }
```
panarch, line 218: In `core` workspace, except for very few cases, `mut` keyword is not allowed.
Let's get familiar with FP styles :)

```rust
```

core/src/executor/evaluate/mod.rs
```diff
@@ -211,6 +211,14 @@ async fn evaluate_function<'a, T>(
...
+
+            f::concat(name(), vec) // _s).await?),
```
panarch, line 220: Unnecessary comment is in here, it would be good to remove it.

earney: +1

core/src/executor/evaluate/mod.rs
earney: This is the first time I've seen stream::iter.. What is the difference between stream::iter(exprs) and expers.iter()?

ever0de: You can think of [Stream](https://docs.rs/futures/latest/futures/prelude/trait.Stream.html) as an asynchronous version of Iterator.

The name is ambiguous, so this time I have an RFC that I want to change to something like AsyncIterator.

test-suite/src/concat.rs
```diff
@@ -90,4 +90,30 @@ test_case!(concat, async move {
...
+        r#"select concat("ab", "cd", "ef") as myconcat from Concat;"#
+    );
+
+    test!(
+        Ok(select_with_null!(
+           myconcat;
+           Null
+        )),
+        r#"select concat("ab", "cd", NULL, "ef") as myconcat from Concat;"#
+    );
```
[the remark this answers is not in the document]
earney: I added two error cases based on your suggestion.  One checks for when we give non string arguments, and the other checks when we have zero arguments.

core/src/executor/evaluate/function.rs
```diff
@@ -50,6 +50,33 @@ macro_rules! eval_to_float {
...
+            //Value::Interval(value) => s.push_str(&value.to_string()),
+            Value::Uuid(value) => s.push_str(&value.to_string()),
+            //Value::Map(value) => s.push_str(value.to_string()),
+            //Value::List(value) => s.push_str(value.to_string()),
+            Value::Null => {} // ignore null values
+            _ => {
+                return Err(EvaluateError::FunctionRequiresStringValue(name).into());
+            }
+        }
+    }
```
panarch, line 75: We can still use `Value::concat`.
`Value::concat` handles the operation between other values and `Null` to make `Null`.
However, simply we can filtering out `Null` and use `Value::concat`

earney: ok.   I changed the code to use Value::concat.

panarch (changes requested):

panarch (approved):
Beautiful 👍 👍 👍 
Test cases cover all the codes and everything looks good.
Thanks a lot!

PR #523, opened 2022-04-15
Support Inline view (FROM clause subquery)

### LINKED ISSUE #522: Support `Inline view` (FROM clause subquery)
### Currently GlueSQL supports subqueries at
- SELECT clause (Scala)
- WHERE clause (Subquery)
### But not yet
- FROM clause (Inline view)
### Examples
```SQL
CREATE TABLE Test (id INTEGER, name TEXT);
INSERT INTO Test VALUES (1, 'GLUE'), (2, 'SQL');
```
```SQL
SELECT * FROM (SELECT COUNT(*) cnt FROM Test);
```
should return
```SQL
cnt
---
  2
```
It looks like similar with issue: #152 and is related with #417.
### Todo
- [x] Add initial test cases at `test-suite/src/inline_view.rs`
- [x] Add `TableFactor::Derived {subquery, alias}` at `core/src/ast/query.rs`
- [x] Fix UnsupportedQueryTableFactor -> Return TableFactor::Derived at `core/src/translate/query.rs`
- [ ] Should we separate TableFactor to TableFactorEvaluate and TableFactorTranslate?
- [ ] Sth to do in plan?
- [ ] Impl if relation == Derived, select(subquery) in `select_with_label` at `core/src/executor/select/mod.rs`

devgony:
### test-suite/src/inline_view.rs
- [x]  Add test cases
### core/src/ast/query.rs
- [x]  Implement TableFactor.{get_name, get_alias}
### core/src/data/table.rs
- [x]  Derived throws TableError::Unreachable
    Can’t we get just TableFactor::Table?
### core/src/executor/alter/table.rs
- [x]  Derived throws TableError::Unreachable
    create_table should not be called with Devied
### core/src/executor/fetch.rs
- [x]  Implement Rows::{Table, Derived}, fetch_relation
    - [x]  Handle Derived case with calling fetch_relation at core/src/executor/join.rs
    - [x]  modify return type of select, select_with_lables from `impl TryStream<Ok = Row, Error = Error> + 'a` to `impl TryStream<Ok = Row, Error = Error, Item = Result<Row>> + 'a` at core/src/executor/select/mod.rs
### core/src/executor/select/mod.rs
- [x]  Get (rows, columns) depending on each Table and Derived at top of select_with_labels
- [x]  Get join_columns depending on each Table and Derived
    To get join_columns, I used get_labels but it needs join_columns.
    Should join_columns be Option?
### core/src/plan/evaluable.rs
- [x]  Derived returns [alias.name](http://alias.name/)
### core/src/plan/index.rs
- [x]  Derived returns Unreachable
    Derived does not have index yet
### core/src/plan/join.rs
- [x]  Derived returns next
### core/src/plan/schema.rs
- [x]  Derived scans subquery and return
### core/src/translate/query.rs
- [x]  If Derived does not have alias, throw TranslateError::LackOfAlias


core/src/executor/select/mod.rs
```diff
@@ -166,93 +171,141 @@ pub async fn select_with_labels<'a, T>(
...
 
-            async move { blend.apply(aggregated, context).await }
-        });
-    let rows = limit.apply(rows);
+                    async move { blend.apply(aggregated, context).await }
+                });
+            let rows = limit.apply(rows);
 
-    Ok((labels, rows))
+            Ok((labels, rows))
```
devgony, line 262: cannot return value referencing temporary value
returns a value referencing data owned by the current functionrustc[E0515](https://doc.rust-lang.org/error-index.html#E0515)
mod.rs(267, 28): temporary value created here

core/src/executor/select/mod.rs
```diff
@@ -151,6 +156,7 @@ pub async fn select_with_labels<'a, T>(
     filter_context: Option<Rc<FilterContext<'a>>>,
     with_labels: bool,
 ) -> Result<(Vec<String>, impl TryStream<Ok = Row, Error = Error> + 'a)> {
+    // Should we add Future here?
```
devgony, line 159: recursion in an `async fn` requires boxing
a recursive `async fn` must be rewritten to return a boxed `dyn Future`
consider using the `async_recursion` crate: https://crates.io/crates/async_recursionrustc[E0733](https://doc.rust-lang.org/error-index.html#E0733)

devgony: solved with deriving async_recursion

core/src/executor/join.rs
```diff
@@ -245,17 +255,42 @@ impl<'a> JoinExecutor<'a> {
...
+                    .await?
+                    .map(|item| item.map(|(_, row)| row));
+
+                // Box::new(stream::iter(rows))
+                Rows::Table(stream::iter(rows))
+                /*
+                .map_ok(|(key, row)| row)
+                .map(stream::iter) // .map(|r| r.unwrap().1),
+                */
+            }
```
panarch, line 284: This error was because `select` return result type does not have an associated type of `Item`.
image

First, specifying `Item` to `select` and `select_with_labels` are required.
e.g.
```rust
pub async fn select<'a>(                                                       
    storage: &'a dyn GStore,                                                   
    query: &'a Query,                                                          
    filter_context: Option<Rc<FilterContext<'a>>>,                             
) -> Result<impl TryStream<Ok = Row, Error = Error, Item = Result<Row>> + 'a> {
```

Then,
```rust
pub async fn test<'a>(                                                         
    storage: &'a dyn GStore,                                                   
    table_factor: &'a TableFactor,                                             
    filter_context: Option<Rc<FilterContext<'a>>>,                             
) -> Result<impl TryStream<Ok = Row, Error = Error, Item = Result<Row>> + 'a> {
    #[derive(futures_enum::Stream)]                                            
    enum Rows<I1, I2> {                                                        
        Derived(I1),                                                           
        Table(I2),                                                             
    }                                                                          
                                                                               
    match table_factor {                                                       
        TableFactor::Derived { subquery, .. } => {                             
            let filter_context = filter_context.as_ref().map(Rc::clone);       
            let rows = select(storage, subquery, filter_context).await?;       
                                                                               
            Ok(Rows::Derived(rows))                                            
        }                                                                      
        TableFactor::Table { name, .. } => {                                   
            let rows = storage                                                 
                .scan_data(crate::data::get_name(name)?)                       
                .await?                                                        
                .map(|item| item.map(|(_, row)| row));                         
            let rows = stream::iter(rows);                                     
                                                                               
            Ok(Rows::Table(rows))                                              
        }                                                                      
    }                                                                          
}                                                                              
```
These kind of above approach will work.

core/src/executor/select/mod.rs
```diff
@@ -166,40 +171,80 @@ pub async fn select_with_labels<'a>(
...
+                    let Select {
+                        from: TableWithJoins { relation, .. },
+                        ..
+                    } = statement.as_ref();
+
+                    let Select { projection, .. } = statement.as_ref();
+                    let inner_table_name = relation.get_name()?;
+                    let columns = fetch_columns(storage, inner_table_name).await?;
+                    let join_columns = &[(&"null".to_string(), vec![])]; // todo: join_columns should be Option?
+                    let columns = get_labels(projection, inner_table_name, &columns, join_columns)?;
```
devgony, line 229: To get join_columns, I used get_labels but it needs join_columns.
Should join_columns be Option?

test-suite/src/inline_view.rs
```diff
@@ -0,0 +1,253 @@
+use {
```
panarch, line 1: Test cases looks beautiful, easily readable, very informative and well aligned 👍

core/src/ast/query.rs
```diff
@@ -58,6 +59,24 @@ pub enum TableFactor {
...
+            Self::Table { .. } => Ok(Table::new(self)?.get_name()),
+            Self::Derived { alias, .. } => Ok(&alias.name),
+        }
+    }
+    pub(crate) fn get_alias(&self) -> Result<&String> {
+        match self {
+            Self::Table { .. } => Ok(Table::new(self)?.get_alias()),
+            Self::Derived { alias, .. } => Ok(&alias.name),
+        }
+    }
```
panarch, line 79: Here is in the `ast` module, and it is supposed to describe the shape of query, so.. it would be not proper to have logic implementation code in here.

core/src/data/table.rs
```diff
@@ -30,6 +31,7 @@ impl<'a> Table<'a> {
 
                 Ok(Self { name, alias, index })
             }
+            TableFactor::Derived { .. } => Err(Error::Table(TableError::Unreachable)),
```
panarch, line 34: Interesting...
`data::Table` is a kind of... temporal data struct which is only for handling `TableFactor::Table`.
Now `TableFactor` enum has more than a single types.

`data::Table` also was being used in `join` and `select.
Now, we may not need `data::Table` anymore, then the code can be even quite more simple I think..

core/src/executor/select/mod.rs
panarch: I think.. as you've successfully add a new fetch method of `fetch_relation` to fetch rows using `TableFactor` param.

Here this also looks possible to find the common pattern in this module.
In this code, we may need to add `fetch_columns` for `TableFactor`.

There exists two cases which we handle `TableFactor` enum directly in this file `select/mod.rs`.
fetching columns using TableFactor might be a key clue.. I think.

I will also look and dig more on this and share if I find something.

core/src/executor/select/mod.rs
```diff
@@ -74,24 +76,29 @@ async fn fetch_blended<'a>(
...
     Ok(stream::iter(rows))
 }
 
-fn get_labels<'a>(
+pub fn get_labels<'a>(
     projection: &[SelectItem],
     table_alias: &str,
     columns: &'a [String],
-    join_columns: &'a [(&String, Vec<String>)],
+    join_columns: Option<&'a [(&String, Vec<String>)]>,
```
devgony, line 88: From now, we should be able to get labels before join so that join_columns become Option

core/src/data/relation.rs
```diff
@@ -51,6 +55,5 @@ impl<'a> Table<'a> {
```
panarch, line 55: I think we can reduce or remove `Relation` struct dependency.

For a small single step,
we can consider to add `get_alias` which directly takes `TableFactor`.
Then, no more making `Relation` to get_alias
```rust
pub fn get_alias(table_factor: &TableFactor) -> Result<&String> {
    match table_factor {
        TableFactor::Table { 
            name, alias: None, ..
        } => get_name(name),
        TableFactor::Table {
            alias: Some(TableAlias { name, .. }),
            ..
        }
        | TableFactor::Derived {
            alias: TableAlias { name, .. },
            ..
        } => Ok(name),
    }   
}   
```

core/src/executor/select/mod.rs
```diff
@@ -166,40 +183,52 @@ pub async fn select_with_labels<'a>(
...
+                Ok(BlendContext::new(
+                    &alias.name,
+                    Rc::clone(&labels_rc),
+                    Some(row),
+                    None,
+                ))
+            });
+            (Rows::Derived(stream::iter(rows)), Rc::clone(&labels))
+        }
+    };
```
panarch, line 211: I mainly focused on these code above.
It requires fetching `rows` and `columns` at the same time but it looks... possible to use shared function in `fetch`.

This kind of shape may work. I renamed `fetch_relation` to `fetch_relation_rows`.
```rust
let TableWithJoins { relation, joins } = &table_with_joins;
let columns = fetch_relation_columns(storage, relation).await?;
let columns = Rc::from(columns);
let rows = {
    let columns = Rc::clone(&columns);
        
    fetch_relation_rows(storage, relation, &None)
        .await?
        .map(move |row| {
            let row = Some(row?);
            let columns = Rc::clone(&columns);
            let alias = get_alias(relation)?;
            Ok(BlendContext::new(alias, columns, row, None))
        })
};  
```

core/src/executor/fetch.rs
```diff
@@ -60,3 +62,73 @@ pub async fn fetch<'a>(
...
+                .scan_data(crate::data::get_name(name)?)
+                .await?
+                .map_ok(|(_, row)| row);
+            let rows = stream::iter(rows);
+
+            Ok(Rows::Table(rows))
+        }
+    }
+}
+
```
panarch, line 94: I am still doing test on it.

Maybe this kind of code may work.
We can consider to add `fetch_relation_columns` to `executor/fetch.rs`
```rust
pub async fn fetch_relation_columns(
    storage: &dyn GStore,
    table_factor: &TableFactor,
) -> Result<Vec<String>> {
    match table_factor {
        TableFactor::Table { name, .. } => {
            let table_name = get_name(name)?;

            fetch_columns(storage, table_name).await
        }
        /*
        TableFactor::Derived { query: SetExpr::Select , .. } => {
            panic!();
            // we can fill this
        }
        */
        _ => {
            panic!();
        }
    }
}

```

panarch (commented):
I'd really respect your passion & patience what you've showed in this PR.
It now looks that we are almost closed to the real final goal.

I left some comments, and I am still looking on the codes. so I will add comments if I find something more. Thanks a lot!!

core/src/data/relation.rs
devgony: I removed Relation and implemented get_name, get_alias, get_index at fetch.rs
Also realized that there were two `get_name` so that i renamed `get_name(&ObjectName)` to `fetch_name(&Object_name)`

core/src/executor/select/mod.rs
```diff
@@ -166,40 +186,50 @@ pub async fn select_with_labels<'a>(
...
-
-                let columns = fetch_columns(storage, table_name).await?;
+    let (rows, columns) = match relation {
+        TableFactor::Table { .. } => {
+            let columns = fetch_columns(storage, get_name(relation)?).await?;
+            let columns = Rc::from(columns);
+            (
+                Rows::Table(fetch_blended(storage, relation, Rc::clone(&columns)).await?),
+                columns,
+            )
```
panarch, line 196: nit cleanup suggestion
```rust
let table_name = get_name(relation)?;
let columns = fetch_columns(storage, table_name).await.map(Rc::from)?;
let rows = fetch_blended(storage, relation, Rc::clone(&columns)).await?;

(Rows::Table(rows), columns)
```

core/src/executor/select/mod.rs
```diff
@@ -166,40 +186,50 @@ pub async fn select_with_labels<'a>(
...
+            let labels = Rc::clone(&labels_rc);
+            let rows = inline_view.into_iter().map(move |row| {
+                Ok(BlendContext::new(
+                    &alias.name,
+                    Rc::clone(&labels_rc),
+                    Some(row),
+                    None,
+                ))
+            });
+            (Rows::Derived(stream::iter(rows)), Rc::clone(&labels))
```
panarch, line 211: Here we don't need to collect stream into `Vec`.
```rust
let inline_view = inline_view.try_collect::<Vec<_>>().await?;
```

Rather than that, this can be handled using chaining without evaluation.

```rust
let (labels, rows) = select_with_labels(storage, subquery, None, true).await?;
let labels = Rc::from(labels.clone());
let rows = {
    let labels = Rc::clone(&labels);

    rows.and_then(move |row| {
        let labels = Rc::clone(&labels);
        let alias = &alias.name;

        async move { Ok(BlendContext::new(alias, labels, Some(row), None)) }
    })
};

(Rows::Derived(rows), labels)
```

core/src/executor/fetch.rs
```diff
@@ -60,3 +71,110 @@ pub async fn fetch<'a>(
...
+pub async fn fetch_join_columns<'a>(
```
panarch, line 104: This looks that `fetch_join_columns` is currently only being used by `select/mod.rs`.
Then it would be better for `select/mod.rs` to have `fetch_join_columns` as private function.

core/src/executor/select/mod.rs
devgony: I tried what you recommended but it was hard to reuse `fetch_relation_columns`.
Below test was failed 😢
```sql
SELECT *                                                                                                                                                                                                                                                                                   FROM (
    SELECT *
    FROM (
        SELECT COUNT(*) AS cnt FROM InnerTable
    ) AS InlineView
) AS InlineView2"
```
If there are more than two nested inline view, it looks like needs some temporary rows to select like,
```rs
let inline_view = inline_view.try_collect::<Vec<_>>().await?;
```

devgony: I just solved that with recursive labels though there is another bug in index test yet.

core/src/executor/fetch.rs
devgony: After implementing `fetch_relation_columns`, I realized that `fetch_relation_columns` and `fetch_join_columns` can refer each other to remove duplicated lines.
Now `fetch_join_columns` is used in both `select/mod.rs` and `fetch.rs` 😭

panarch: Oh, I missed that point. Then current is right place to be 👍

core/src/executor/select/mod.rs
devgony: After implementing what you recommanded `fetch_relation_rows`, columns and rows are splited.
Which one is better between above and `fetch_relation_rows` way?

devgony:
This PR does not conflict with #613 , Please keep reviewing 😃

core/src/plan/index.rs
```diff
@@ -69,10 +70,7 @@ fn plan_query(schema_map: &HashMap<String, Schema>, query: Query) -> Result<Quer
     };
 
     let TableWithJoins { relation, .. } = &select.from;
-    let table_name = match relation {
-        TableFactor::Table { name, .. } => name,
-    };
-    let table_name = get_name(table_name)?;
+    let table_name = get_name(relation)?;
```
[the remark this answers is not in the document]
devgony: 1. => Eary returning way looks so nice. Thank you!
2. => Done
3. => I moved including get_index which made a bit verbose changes. if it was not proper move, let me know!

panarch (approved):
Incredible work you've done. I'd really respect this.
The previous `AST` and execution layer which described `TableFactor` were a bit messy and it was because I didn't know well at the point when I firstly wrote the codes about this.
Not only adding the new feature, but also you even cleaned up the AST & execution layer codes.
Thanks really, really a lot. Now everything looks awesome!

Issue #430, opened 2021-12-11
Support CAST from other data types to decimal


ever0de:
## SQL Example

```sql
CAST("10.2" AS DECIMAL);
CAST(10.2 AS DECIMAL);
...
```

## Solution

- `data/value/into.rs`

Implement `TryInto<Decimal>` for  Value, &Value

- `data/value/mod.rs/cast`

Add try_into from other data types to decimal in cast function

earney:
I believe this issue has been resolved...  see PR #529

panarch:
Thanks 👍 closing this

PR #527, opened 2022-04-17
Cast to decimal (resolves issue #430)

earney:
This issue resolves issue #430, which requests that the cast function handle values being casted to decimals.

### DIFF
--- a/core/src/ast/function.rs
+++ b/core/src/ast/function.rs
@@ -6,6 +6,8 @@ use {
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, Display)]
 pub enum Function {
+    #[strum(to_string = "ABS")]
+    Abs(Expr),
     #[strum(to_string = "LOWER")]
     Lower(Expr),
     #[strum(to_string = "UPPER")]
@@ -88,6 +90,8 @@ pub enum Function {
     Reverse(Expr),
     #[strum(to_string = "REPEAT")]
     Repeat { expr: Expr, num: Expr },
+    #[strum(to_string = "SIGN")]
+    Sign(Expr),
     #[strum(to_string = "SUBSTR")]
     Substr {
         expr: Expr,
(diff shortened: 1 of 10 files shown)


core/src/data/value/into.rs
```diff
@@ -13,7 +13,7 @@ use {
 };
 
 impl From<&Value> for String {
-    fn from(v: &Value) -> Self {
+    fn from(v: &Value) -> String {
```
ever0de, line 16: It's not open, but I'll leave a comment. :)

<https://doc.rust-lang.org/stable/std/convert/trait.From.html>

`From` is the implementation of the above std Trait. `Self` and `String` are the same here, but it seems good to leave it semantically converted to `Self`.

example) <https://doc.rust-lang.org/rust-by-example/trait.html>

Examples from other famous libraries)
- <https://github.com/serde-rs/serde/blob/ce0844b9ecc32377b5e4545d759d385a8c46bc6a/test_suite/tests/test_remote.rs#L160-L182>

core/src/data/value/into.rs
```diff
@@ -218,6 +218,46 @@ impl TryInto<f64> for Value {
...
+                    Decimal::ZERO
+                }
+            }
+            Value::I8(value) => Decimal::from_i8(*value).ok_or(ValueError::ImpossibleCast)?,
+            Value::I64(value) => Decimal::from_i64(*value).ok_or(ValueError::ImpossibleCast)?,
+            Value::F64(value) => Decimal::from_f64(*value).ok_or(ValueError::ImpossibleCast)?,
+            Value::Str(value) => match Decimal::from_str(value) {
+                Ok(v) => return Ok(v),
+                Err(_v) => return Err(ValueError::ImpossibleCast.into()),
+            },
```
ever0de, line 239:
```suggestion
            Value::Str(value) => {
                Decimal::from_str(value).map_err(|_| ValueError::ImpossibleCast)?
            }
```

How about this?

ever0de (commented):
👍 I left a few comments before opening. thank you :)

PR #529, opened 2022-04-19
implement cast to decimal

earney:
continuation of PR #527 .   I tried to do a git rebase, and things got messy so I started over, and github closed the old PR, so I just created a new one.

### DIFF
--- a/core/src/data/value/error.rs
+++ b/core/src/data/value/error.rs
@@ -110,8 +110,6 @@ pub enum ValueError {
     #[error("unreachable literal cast from number to float: {0}")]
     UnreachableLiteralCastFromNumberToFloat(String),
 
-    #[error("unreachable literal cast from number to decimal: {0}")]
-    UnreachableLiteralCastFromNumberToDecimal(String),
 
     #[error("unimplemented literal cast: {literal} as {data_type:?}")]
     UnimplementedLiteralCast {
--- a/core/src/data/value/error.rs
+++ b/core/src/data/value/error.rs
@@ -110,7 +110,6 @@ pub enum ValueError {
     #[error("unreachable literal cast from number to float: {0}")]
     UnreachableLiteralCastFromNumberToFloat(String),
 
-
     #[error("unimplemented literal cast: {literal} as {data_type:?}")]
     UnimplementedLiteralCast {
         data_type: DataType,
--- a/core/src/data/value/error.rs
+++ b/core/src/data/value/error.rs
@@ -96,6 +96,9 @@ pub enum ValueError {
     #[error("literal cast failed from text to float: {0}")]
     LiteralCastFromTextToFloatFailed(String),
 
+    #[error("literal cast failed from text to decimal: {0}")]
+    LiteralCastFromTextToDecimalFailed(String),
+
     #[error("literal cast failed to boolean: {0}")]
     LiteralCastToBooleanFailed(String),
 
@@ -113,6 +116,9 @@ pub enum ValueError {
 
     #[error("unreachable literal cast from number to float: {0}")]
     UnreachableLiteralCastFromNumberToFloat(String),
+
+    #[error("unreachable literal cast from number to decimal: {0}")]
+    UnreachableLiteralCastFromNumberToDecimal(String),
 
     #[error("unimplemented literal cast: {literal} as {data_type:?}")]
     UnimplementedLiteralCast {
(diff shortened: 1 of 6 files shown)


ever0de (commented):
Looks good. thank you.

Below is my personal coding style. Could you possibly consider it?

1. The alias of `Dec` is good because it is short, but it seems to get through the name of `Decimal`. How about using `Decimal`?
2. When casting to `Decimal` in the test code, `decimal` and `Decimal` are mixed. How about unifying this? `Decimal`

This is a story about style, PR itself approves.

test-suite/src/function/cast.rs
```diff
@@ -74,6 +83,105 @@ test_case!(cast_literal, async move {
...
+            Err(ValueError::LiteralCastFromTextToDecimalFailed("foo".to_owned()).into()),
+        ),
+        (
+            r#"SELECT CAST(NULL AS Decimal) AS cast FROM Item"#,
+            Ok(select_with_null!(cast; Null)),
+        ),
+        (
+            r#"SELECT CAST(mytext AS Decimal) AS cast FROM test"#,
+            Err(ValueError::ImpossibleCast.into()),
+            //Err(ValueError::LiteralCastFromTextToDecimalFailed("foo".to_owned()).into()),
```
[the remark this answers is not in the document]
earney: Sure, I will make the change from Dec to Decimal.  I chose Dec because I thought there was a conflict with Value::Decimal, but I will double check.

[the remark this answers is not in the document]
earney: Using Decimal instead of Dec seems to work ok, so I made the change.  Let me know if anything else needs to be changed.

I believe the code is now up to date with your suggestions.  If not, let me know

ever0de (approved):
LGTM

core/src/data/value/error.rs
```diff
@@ -107,6 +110,9 @@ pub enum ValueError {
     #[error("unreachable literal cast from number to float: {0}")]
     UnreachableLiteralCastFromNumberToFloat(String),
 
+    #[error("unreachable literal cast from number to decimal: {0}")]
+    UnreachableLiteralCastFromNumberToDecimal(String),
```
ever0de, line 114: There was something we missed.
When converting BigDecimal -> Decimal we are using `FromStr` function.
The following implementation: <https://github.com/paupino/rust-decimal/blob/2b582d625592f1170fc3cc7ee03231bcdce2dd70/src/decimal.rs#L1859-L1866>

The error you're getting here is not unreachable code.
Possible error: <https://github.com/paupino/rust-decimal/blob/2b582d625592f1170fc3cc7ee03231bcdce2dd70/src/error.rs#L6-L14>

Perhaps this should be implemented as an error about conversion...
cc. @panarch

earney: Where are we converting from BigDecimal to Decimal?  Or are you referring to the from* function in general?

core/src/data/value/literal.rs
```diff
@@ -204,6 +204,23 @@ impl Value {
...
+                .parse::<Decimal>()
+                .map(Value::Decimal)
+                .map_err(|_| ValueError::LiteralCastFromTextToDecimalFailed(v.to_string()).into()),
+            (DataType::Decimal, Literal::Number(v)) => v
+                .to_string()
+                .parse::<Decimal>()
+                .map(Value::Decimal)
+                .map_err(|_| {
+                    ValueError::UnreachableLiteralCastFromNumberToDecimal(v.to_string()).into()
+                }),
```
ever0de, line 217: Here we convert from `BigDecimal` to `String` and from `String` to `Decimal`.

core/src/data/value/literal.rs
```diff
@@ -204,6 +204,21 @@ impl Value {
...
             }
+            (DataType::Decimal, Literal::Text(v)) => v
+                .parse::<Decimal>()
+                .map(Value::Decimal)
+                .map_err(|_| ValueError::LiteralCastFromTextToDecimalFailed(v.to_string()).into()),
+            (DataType::Decimal, Literal::Number(v)) => v
+                .to_string()
+                .parse::<Decimal>()
+                .map(Value::Decimal)
+                .map_err(|_| ValueError::FailedToParseDecimal(v.to_string()).into()),
```
ever0de, line 215: Let's unify the message `LiteralCastFromTextToDecimalFailed'.
Most of the errors I get in the `try_cast_from_literal` function seem to be `LiteralCastTo...`.

```suggestion
                .map_err(|_| ValueError::LiteralCastFromTextToDecimalFailed(v.to_string()).into()),
```

ever0de (approved):
👍  😃

panarch (approved):
Awesome, thanks a lot!
Thanks for the careful review @ever0de 

All looks great, merging 👍

Issue #427, opened 2021-12-11
Supports arithmetic operation between Decimal and primitive types


ever0de:
ex)
- Decimal + i64 = Decimal
- Decimal + f64 = Decimal 
...

earney:
I believe this has been resolvedi in PR #530, which has been merged.  Can we close this issue?

ever0de:
> I believe this has been resolved in PR #530, which has been merged. Can we close this issue?

Thanks

PR #530, opened 2022-04-20
Decimal math

earney:
resolves issue #427 by adding basic decimal math

### DIFF
--- a/core/src/data/value/binary_op/decimal.rs
+++ b/core/src/data/value/binary_op/decimal.rs
@@ -15,10 +15,9 @@ impl PartialEq<Value> for Decimal {
         match other {
             I8(other) => *self == Decimal::from(*other),
             I64(other) => *self == Decimal::from(*other),
-            F64(other) => match Decimal::from_f64_retain(*other) {
-                Some(x) => *self == x,
-                _ => false,
-            },
+            F64(other) => Decimal::from_f64_retain(*other)
+                .map(|x| *self == x)
+                .unwrap_or(false),
             Decimal(other) => *self == *other,
             _ => false,
         }
@@ -51,7 +50,7 @@ impl TryBinaryOperator for Decimal {
             I64(rhs) => Ok(Decimal(lhs + Decimal::from(rhs))),
             F64(rhs) => match Decimal::from_f64_retain(rhs) {
                 Some(x) => Ok(Decimal(lhs + x)),
-                _ => Err(Error::F64ToDecimalConversionError(rhs.to_string())),
+                _ => Err(ValueError::F64ToDecimalConversionError(rhs)),
             },
             Decimal(rhs) => Ok(Decimal(lhs + rhs)),
             Null => Ok(Null),
@@ -72,7 +71,7 @@ impl TryBinaryOperator for Decimal {
             I64(rhs) => Ok(Decimal(lhs - Decimal::from(rhs))),
             F64(rhs) => match Decimal::from_f64_retain(rhs) {
                 Some(x) => Ok(Decimal(lhs - x)),
-                _ => Err(Error::F64ToDecimalConversionError(rhs.to_string())),
+                _ => Err(ValueError::F64ToDecimalConversionError(rhs)),
             },
             Decimal(rhs) => Ok(Decimal(lhs - rhs)),
             Null => Ok(Null),
@@ -93,7 +92,7 @@ impl TryBinaryOperator for Decimal {
             I64(rhs) => Ok(Decimal(lhs * Decimal::from(rhs))),
             F64(rhs) => match Decimal::from_f64_retain(rhs) {
                 Some(x) => Ok(Decimal(lhs * x)),
-                _ => Err(Error::F64ToDecimalConversionError(rhs.to_string())),
+                _ => Err(ValueError::F64ToDecimalConversionError(rhs)),
             },
             Decimal(rhs) => Ok(Decimal(lhs * rhs)),
             Null => Ok(Null),
@@ -114,7 +113,7 @@ impl TryBinaryOperator for Decimal {
             I64(rhs) => Ok(Decimal(lhs / Decimal::from(rhs))),
             F64(rhs) => match Decimal::from_f64_retain(rhs) {
                 Some(x) => Ok(Decimal(lhs / x)),
-                _ => Err(Error::F64ToDecimalConversionError(rhs.to_string())),
+                _ => Err(ValueError::F64ToDecimalConversionError(rhs)),
             },
 
             Decimal(rhs) => Ok(Decimal(lhs / rhs)),
@@ -145,11 +144,11 @@ impl TryBinaryOperator for Decimal {
                     Some(y) => Ok(Decimal(y)),
                     None => Err(Error::OverflowError("%".to_string())),
                 },
-                _ => Err(Error::F64ToDecimalConversionError(rhs.to_string())),
+                _ => Err(ValueError::F64ToDecimalConversionError(rhs)),
             },
             Decimal(rhs) => match lhs.checked_rem(rhs) {
                 Some(x) => Ok(Decimal(x)),
-                _ => Err(Error::F64ToDecimalConversionError(rhs.to_string())),
+                _ => Err(Error::OverflowError("%".to_string())),
             },
 
             Null => Ok(Null),
@@ -201,20 +200,13 @@ mod tests {
 
     #[test]
     fn try_add() {
-        let dec_epislon: Decimal = Decimal::from_f64_retain(f64::EPSILON).unwrap();
         let base = Decimal::ONE;
 
+        assert!(matches!(base.try_add(&I8(1)), Ok(Decimal(x)) if x == Decimal::TWO ));
+        assert!(matches!(base.try_add(&I64(1)), Ok(Decimal(x)) if x == Decimal::TWO));
+        assert!(matches!(base.try_add(&F64(1.0)), Ok(Decimal(x)) if x == Decimal::TWO));
         assert!(
-            matches!(base.try_add(&I8(1)), Ok(Decimal(x)) if (x - Decimal::TWO).abs() < dec_epislon )
-        );
-        assert!(
-            matches!(base.try_add(&I64(1)), Ok(Decimal(x)) if (x - Decimal::TWO).abs() < dec_epislon )
-        );
-        assert!(
-            matches!(base.try_add(&F64(1.0)), Ok(Decimal(x)) if (x - Decimal::TWO).abs() < dec_epislon )
-        );
-        assert!(
-            matches!(base.try_add(&Decimal(Decimal::ONE)), Ok(Decimal(x)) if (x - Decimal::TWO).abs() < dec_epislon )
+            matches!(base.try_add(&Decimal(Decimal::ONE)), Ok(Decimal(x)) if x == Decimal::TWO)
         );
 
         assert_eq!(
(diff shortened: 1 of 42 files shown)


earney:
Maybe merge this with pr 476?

core/src/data/value/binary_op/decimal.rs
```diff
@@ -0,0 +1,257 @@
...
+    std::cmp::Ordering,
+    Value::*,
+};
+
+impl PartialEq<Value> for Dec {
+    fn eq(&self, other: &Value) -> bool {
+        match other {
+            I8(other) => *self == Dec::from(*other),
+            I64(other) => *self == Dec::from(*other),
+            F64(other) => *self == Dec::from_f64_retain(*other).unwrap(),
```
ever0de, line 14: This feature looks great. Get good knowledge thanks! 
<https://docs.rs/rust_decimal/latest/rust_decimal/prelude/struct.Decimal.html#method.from_f64_retain>

But in the PartialEq implementation, it doesn't seem like it should cause a failure.
If none is displayed, it should be treated as false. What do you think?

core/src/data/value/binary_op/decimal.rs
```diff
@@ -0,0 +1,257 @@
...
+        }
+    }
+}
+
+impl PartialOrd<Value> for Dec {
+    fn partial_cmp(&self, other: &Value) -> Option<Ordering> {
+        match *other {
+            I8(rhs) => self.partial_cmp(&(Dec::from(rhs))),
+            I64(rhs) => self.partial_cmp(&(Dec::from(rhs))),
+            F64(rhs) => self.partial_cmp(&(Dec::from_f64_retain(rhs).unwrap())),
```
ever0de, line 26: `PartialOrd` is thinking of returning an Option, so it seems better to return None if None is found.

core/src/data/value/binary_op/decimal.rs
```diff
@@ -0,0 +1,257 @@
...
+
+    fn try_multiply(&self, rhs: &Self::Rhs) -> Result<Value> {
+        let lhs = *self;
+
+        match *rhs {
+            I8(rhs) => Ok(Decimal(lhs * Dec::from(rhs))),
+            I64(rhs) => Ok(Decimal(lhs * Dec::from(rhs))),
+            F64(rhs) => Ok(Decimal(lhs * Dec::from_f64_retain(rhs).unwrap())),
+            Decimal(rhs) => Ok(Decimal(lhs * rhs)),
+            //Interval(rhs) => Ok(Interval(lhs * rhs)),
```
ever0de, line 70: Remove comments?

```suggestion
```

core/src/data/value/binary_op/decimal.rs
```diff
@@ -0,0 +1,257 @@
...
+            Null => Ok(Null),
+            _ => Err(ValueError::DivideOnNonNumeric(Decimal(lhs), rhs.clone()).into()),
+        }
+    }
+
+    fn try_modulo(&self, rhs: &Self::Rhs) -> Result<Value> {
+        let lhs = *self;
+
+        match *rhs {
+            I8(rhs) => Ok(Decimal(lhs.checked_rem(Dec::from(rhs)).unwrap())),
```
ever0de, line 93: This feature looks good too. thank you
Can you return an error instead of `unwrap` when an overflow occurs?
- ex) Add `OverflowError { operator: char }` to the Error enum

core/src/data/value/binary_op/f64.rs
```diff
@@ -1,16 +1,32 @@
 use {
     super::TryBinaryOperator,
     crate::{data::ValueError, prelude::Value, result::Result},
+    rust_decimal::prelude::Decimal as Dec,
     std::cmp::Ordering,
     Value::*,
 };
 
+impl PartialEq<Value> for f64 {
```
ever0de, line 9: There was something we were missing! thanks for the good catch

ever0de (commented):
❤️  👍

core/src/data/value/binary_op/decimal.rs
earney: good idea..  I've implemented it in the latest push.

core/src/data/value/binary_op/decimal.rs
earney: good catch..

core/src/data/value/binary_op/decimal.rs
earney: I've added an Overflow error.   I believe this describes the checked_rem failures.
what do you think would be a good error for when from_f64_retain fails?  F64ToDecimalConversionError?

core/src/data/value/binary_op/f64.rs
earney: +1  I wondered why it was in i64, and realized it must have been a mistake.. ;)

core/src/data/value/binary_op/decimal.rs
earney: +1

core/src/data/value/binary_op/decimal.rs
ever0de: As explained in the following link, 
<https://docs.rs/rust_decimal/latest/rust_decimal/prelude/struct.Decimal.html#method.checked_rem>
it seems that an overflow occurred when trying Decimal % Decimal.

What about messages like `ModuloOverflowError` rather than Conversion?

We don't actually do any conversions in the `check_ram` method.

earney: I'm referring to the from_f64_retain function, not checked_rem.    the docs doesn't say anything about overflow or conversion errors for from_f64_retain.

ever0de: Hmm...? You obviously use `checked_rem` in your code snippet.
Conversion related errors are correct by using `from_f64_retain` elsewhere.
But isn't `checked_rem` the one you're currently using?
image

ever0de: > what do you think would be a good error for when from_f64_retain fails? F64ToDecimalConversionError?

Ah, sorry. I just checked the code snippet and answered it.

In the case of `from_f64_retain` , it would be nice to have a Conversion error message.

earney: F64toDecimalConversionError has been added.

core/src/data/value/binary_op/decimal.rs
```diff
@@ -0,0 +1,346 @@
...
+};
+
+impl PartialEq<Value> for Decimal {
+    fn eq(&self, other: &Value) -> bool {
+        match other {
+            I8(other) => *self == Decimal::from(*other),
+            I64(other) => *self == Decimal::from(*other),
+            F64(other) => match Decimal::from_f64_retain(*other) {
+                Some(x) => *self == x,
+                _ => false,
```
ever0de, line 20:
```suggestion
            F64(other) => Decimal::from_f64_retain(*other)
                .map(|x| *self == x)
                .unwrap_or(false),
```

How about this code?

core/src/data/value/binary_op/decimal.rs
```diff
@@ -0,0 +1,346 @@
...
+                operator: NumericBinaryOperator::Add,
+                rhs: Bool(true).clone(),
+            }
+            .into()),
+        );
+    }
+
+    #[test]
+    fn try_subtract() {
+        let dec_epislon: Decimal = Decimal::from_f64_retain(f64::EPSILON).unwrap();
```
ever0de, line 233: In a Decimal type, wouldn't it be okay not to compare with epsilon values?

ever0de: Why was the F64 comparison done using epsilon?
Answer) <https://rust-lang.github.io/rust-clippy/master/#float_cmp>

core/src/result.rs
```diff
@@ -31,6 +31,12 @@ pub enum Error {
     #[error("parsing failed: {0}")]
     Parser(String),
 
+    #[error("OverflowError: {0}")]
+    OverflowError(String),
```
ever0de, line 35: I think it can be improved by using Operator Enum, Lhs, and Rhs.

core/src/result.rs
```diff
@@ -31,6 +31,12 @@ pub enum Error {
...
+
+    #[error("failed to convert Float to Decimal: {0}")]
+    F64ToDecimalConversionError(String),
```
ever0de, line 38:
```suggestion
    F64ToDecimalConversionError(Decimal),
```

How about dealing with String as Decimal?

ever0de (commented):
I think it would be good to improve the `match` syntax for other options as well.

core/src/result.rs
earney: Shouldn't it be F64ToDecimalConversionError(F64) instead of F64ToDecimalConversionError(Decimal)?

ever0de: Oh right..! thank you 
I edited

core/src/result.rs
earney: Ok.. will do.

ever0de: thanks for your hard work...

core/src/data/value/binary_op/decimal.rs
earney: Looks good to me!  I'll also try to do something similar with other calls to Decimal::from_f64_retain...

core/src/result.rs
earney: When I move the error message to the ValueError enum, I get the following error:
 _ => Err(ValueError::F64ToDecimalConversionError(lhs)),
    |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected enum `result::Error`, found enum `ValueError` 

I have gotten this before, and I cannot figure out why Err doesn't want to accept a ValueError sometimes.

ever0de: What the function returns is the top-level `Error` enum.
But where we declared the error is in the `ValueError` enum.
If so, here's a very nice way to convert from ValueError to top-level Error.
That's using the `Into` trait.

Since we already use the Error macro + `#from` function of `thiserror`, it is already implemented.

try it like this
`Err(ValueError::F64ToDecimalConversionError(lhs).into())`

earney: Thanks..  I was wondering why into was used in the error messages..

core/src/data/value/binary_op/decimal.rs
```diff
@@ -0,0 +1,321 @@
...
+}
+
+impl PartialOrd<Value> for Decimal {
+    fn partial_cmp(&self, other: &Value) -> Option<Ordering> {
+        match *other {
+            I8(rhs) => self.partial_cmp(&(Decimal::from(rhs))),
+            I64(rhs) => self.partial_cmp(&(Decimal::from(rhs))),
+            F64(rhs) => Decimal::from_f64_retain(rhs)
+                .map(|x| self.partial_cmp(&x))
+                .unwrap_or(None),
```
ever0de, line 34: nit) Good method: [`ok`](https://doc.rust-lang.org/stable/std/result/enum.Result.html#method.ok)

core/src/result.rs
```diff
@@ -31,6 +31,9 @@ pub enum Error {
     #[error("parsing failed: {0}")]
     Parser(String),
 
+    #[error("OverflowError: {0}")]
+    OverflowError(String),
```
ever0de, line 35: Now that we have a BinaryOperator <https://github.com/gluesql/gluesql/commit/53579312d1950e181ac81de231a2ef79668c0fbb> let's improve it with this.

core/src/data/value/binary_op/decimal.rs
```diff
@@ -0,0 +1,321 @@
...
+        );
+    }
+
+    #[test]
+    fn try_subtract() {
+        let base = Decimal::ONE;
+
+        assert!(
+            matches!(base.try_subtract(&Decimal(Decimal::ONE)), Ok(Decimal(x)) if x == Decimal::ZERO)
+        );
```
ever0de, line 222: Even so, the test will pass!

But to give us enough information when debugging, how about using `assert_eq` to know the lhs and rhs values?

core/src/data/value/binary_op/decimal.rs
```diff
@@ -0,0 +1,321 @@
...
+        assert!(
+            matches!(base.try_modulo(&Decimal(Decimal::ONE)), Ok(Decimal(x)) if x == Decimal::ZERO)
+        );
+
+        assert_eq!(
+            base.try_modulo(&Bool(true)),
+            Err(ValueError::NonNumericMathOperation {
+                lhs: Decimal(base),
+                operator: NumericBinaryOperator::Modulo,
+                rhs: Bool(true).clone(),
```
ever0de, line 316:
```suggestion
                rhs: Bool(true),
```

ever0de: Usually `clone` is used to clone a variable, but it doesn't seem necessary here.

Another story) I wonder why this is not caught in clippy error.

core/src/data/value/error.rs
```diff
@@ -23,6 +23,9 @@ pub enum ValueError {
     #[error("failed to parse number")]
     FailedToParseNumber,
 
+    #[error("failed to convert Float to Decimal: {0}")]
+    F64ToDecimalConversionError(f64),
```
ever0de, line 27: <https://github.com/gluesql/gluesql/blob/c49073a76e7a775a8a64752f72d8315c876de68f/core/src/data/value/error.rs#L63-L64>

1. Use `Float` instead of `F64`.
Float type is exposed to the user, but F64 is an internal type, so it is recommended to use `Float` for this error value exposed to the user.
2. Let's clear the `Error` suffix.
Declaring a variant in the ValueError enum is duplicated.
How about something like `FloatToDecimalConversionFailure` instead?

ever0de (approved):
In the logic part, it looks good, so I'll leave the approval first.

Most of the stories in the comments are trivial.

core/src/data/value/binary_op/decimal.rs
earney: +1

core/src/data/value/error.rs
earney: Good suggestions!

core/src/data/value/binary_op/decimal.rs
earney: will do!

core/src/data/value/binary_op/decimal.rs
earney: ??

panarch (approved):
Awesome!! This is also really, really meaningful contribution. Thanks a lot!
Also thanks for the careful review @ever0de 

Merging 👍 👍 👍

Issue #424, opened 2021-12-08
Add arithmetic operation enum for Error types


ever0de:
```rust
pub enum Arithmetic {
    Plus,
    Minus,
    // ...
}
```

We currently define the error types as follows:

```rust
    #[error("add on non-numeric values: {0:?} + {1:?}")]
    AddOnNonNumeric(Value, Value),

    #[error("subtract on non-numeric values: {0:?} - {1:?}")]
    SubtractOnNonNumeric(Value, Value),

    #[error("multiply on non-numeric values: {0:?} * {1:?}")]
    MultiplyOnNonNumeric(Value, Value),

    #[error("divide on non-numeric values: {0:?} / {1:?}")]
    DivideOnNonNumeric(Value, Value),
```

However, if you have an arithmetic operation as an operator, it will be resolved with one error type.

```rust 
    #[error("non-numeric values: {lhs:?} {operator:?} {rhs:?}")]
    NonNumeric {
        lhs: Value,
        rhs: Value,
        operator:  ArithmeticOperatortion
    }
```

ever0de:
ref: <https://github.com/gluesql/gluesql/pull/407/files#diff-36c814425eac10888a46489d522e6ca31f8b426e4890bf563e03291420e8b7c7R135-R141>

earney:
I'll take this issue. It should be pretty easy to make these changes.

PR #534, opened 2022-04-22
resolve issue #424

earney:
This PR removes the ArithmeticNonNumeric enum error codes for a single NonNumericMathOperation error code with a lhs, operator and rhs args.   See issue #424 for more info.

### DIFF
--- a/core/src/data/value/error.rs
+++ b/core/src/data/value/error.rs
@@ -2,6 +2,7 @@ use {
     crate::{ast::DataType, ast::DateTimeField, data::Value},
     serde::Serialize,
     std::fmt::Debug,
+    strum_macros::Display,
     thiserror::Error,
 };
 
@@ -37,11 +38,11 @@ pub enum ValueError {
     #[error("failed to parse Decimal: {0}")]
     FailedToParseDecimal(String),
 
-    #[error("non-numeric values {lhs:?} {operator:?} {rhs:?}")]
+    #[error("non-numeric values {lhs:?} {operator} {rhs:?}")]
     NonNumericMathOperation {
         lhs: Value,
         rhs: Value,
-        operator: String,
+        operator: NumericBinaryOperator,
     },
 
     #[error("the divisor should not be zero")]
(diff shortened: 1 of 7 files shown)


core/src/data/value/error.rs
```diff
@@ -37,24 +37,16 @@ pub enum ValueError {
...
-    MultiplyOnNonNumeric(Value, Value),
-
-    #[error("divide on non-numeric values: {0:?} / {1:?}")]
-    DivideOnNonNumeric(Value, Value),
+    #[error("non-numeric values {lhs:?} {operator:?} {rhs:?}")]
+    NonNumericMathOperation {
+        lhs: Value,
+        rhs: Value,
+        operator: String,
+    },
```
panarch, line 45: Nice changes, thanks a lot 👍 

`operator: String` covers more than its actual range.
This `NonNumericMathOperation` is only for five numeric binary operations.
However, `String` is superset, does not have the identical with the previous state.

This was why @ever0de suggested to use enum for this.
```rust
pub enum Arithmetic {
    Plus,
    Minus,
    // ...
}
```

Let's have `NumericBinaryOperator` enum in `data/value/error.rs`
e.g.
```rust
// we can use strum_macros::Display in here
#[derive(Debug, PartialEq, Eq, Serialize, Display)]
pub enum NumericBinaryOperator {                   
    #[strum(to_string = "+")]                      
    Add,                                           
    #[strum(to_string = "-")]                      
    Subtract,                                      
    #[strum(to_string = "*")]                      
    Multiply,                                      
    #[strum(to_string = "/")]                      
    Divide,                                        
    #[strum(to_string = "%")]                      
    Modulo,                                        
}                                                  
```

Then,
```rust
#[error("non-numeric values {lhs:?} {operator} {rhs:?}")]
NonNumericMathOperation {                                
    lhs: Value,                                          
    rhs: Value,                                          
    operator: NumericBinaryOperator,                     
},                                                       
```
operator implements `Display` so you can replace `{operator:?}` to `{operator}`.

panarch (changes requested):

core/src/data/value/error.rs
earney: enum value BinaryOperationOverflow  also have an operator with values of string.  Should I go ahead and change this to use NumericBinaryOperator?

ever0de: you are right I have an enum error using exactly char type.

earney: Ok.  I will also make that change was well to remove char/string from enum errors.

earney: Done!!

ever0de (approved):
Looks good 👍 

Now users don't have to deal with different strings and chars in terms of operators, so there is no unreachable code.

panarch (approved):
Awesome, thanks a lot 👍

PR #537, opened 2022-04-25
Scalar subquery should throw Error when more than one row returned

### LINKED ISSUE #528: Scalar subquery should throw Error when more than one row returned
### Test
```
gluesql> CREATE TABLE A (NO INT, NAME TEXT);
gluesql> CREATE TABLE B (NO INT, NAME TEXT);
gluesql> INSERT INTO A VALUES (1, 'first');
1 row inserted

gluesql> INSERT INTO B VALUES
 (1, 'first')
,(2, 'second')
,(1, 'third');

3 rows inserted

gluesql> SELECT (SELECT NAME FROM B WHERE A.NO = B.NO) from A;
```
### Expected
```
ERROR:  more than one row returned by a subquery used as an expression
```
### Actual
```
╭────────────────────────────────────────╮
│ (SELECT NAME FROM B WHERE A.NO = B.NO) │
╞════════════════════════════════════════╡
│ first                                  │
╰────────────────────────────────────────╯
```

devgony:
To resolve #528 `Scalar subquery should throw Error when more than one row returned`,

## Implements
```rs
Expr::Subquery(query) => {
            let mut stream = select(storage, query, context.as_ref().map(Rc::clone))
                .await?
                .map_ok(|row| row.take_first_value().map(Evaluated::from));
            let first = stream
                .next()
                .await
                .unwrap_or_else(|| Err(EvaluateError::NestedSelectRowNotFound.into()))?;
            let second = stream.next().await;
            match second {
                Some(_) => Err(EvaluateError::MoreThanOneRowReturned.into()),
                None => first,
            }
        }
```
## Test Added
```rs
(
  EvaluateError::MoreThanOneRowReturned.into(),
  "select (select id from TableA) as id from TableA", // when TableA has 2 rows
)
```

### DIFF
--- a/core/src/executor/evaluate/mod.rs
+++ b/core/src/executor/evaluate/mod.rs
@@ -68,12 +68,20 @@ pub async fn evaluate<'a, T>(
             }
             .map(Evaluated::from)
         }
-        Expr::Subquery(query) => select(storage, query, context.as_ref().map(Rc::clone))
-            .await?
-            .map_ok(|row| row.take_first_value().map(Evaluated::from))
-            .next()
-            .await
-            .unwrap_or_else(|| Err(EvaluateError::NestedSelectRowNotFound.into()))?,
+        Expr::Subquery(query) => {
+            let mut stream = select(storage, query, context.as_ref().map(Rc::clone))
+                .await?
+                .map_ok(|row| row.take_first_value().map(Evaluated::from));
+            let first = stream
+                .next()
+                .await
+                .unwrap_or_else(|| Err(EvaluateError::NestedSelectRowNotFound.into()))?;
+            let second = stream.next().await;
+            match second {
+                Some(_) => Err(EvaluateError::MoreThanOneRowReturned.into()),
+                None => first,
+            }
+        }
         Expr::BinaryOp { op, left, right } => {
             let left = eval(left).await?;
             let right = eval(right).await?;
--- a/core/src/executor/evaluate/mod.rs
+++ b/core/src/executor/evaluate/mod.rs
@@ -79,7 +79,7 @@ pub async fn evaluate<'a, T>(
                 0 => Err(EvaluateError::NestedSelectRowNotFound.into()),
                 1 => evaluations
                     .into_iter()
-                    .nth(0)
+                    .next()
                     .unwrap_or_else(|| Err(EvaluateError::NestedSelectRowNotFound.into())),
                 _ => Err(EvaluateError::MoreThanOneRowReturned.into()),
             }
--- a/core/src/executor/evaluate/mod.rs
+++ b/core/src/executor/evaluate/mod.rs
@@ -75,14 +75,15 @@ pub async fn evaluate<'a, T>(
                 .take(2)
                 .try_collect::<Vec<_>>()
                 .await?;
-            match evaluations.len() {
-                0 => Err(EvaluateError::NestedSelectRowNotFound.into()),
-                1 => evaluations
-                    .into_iter()
-                    .next()
-                    .unwrap_or_else(|| Err(EvaluateError::NestedSelectRowNotFound.into())),
-                _ => Err(EvaluateError::MoreThanOneRowReturned.into()),
+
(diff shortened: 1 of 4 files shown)


core/src/executor/evaluate/mod.rs
```diff
@@ -68,12 +68,20 @@ pub async fn evaluate<'a, T>(
...
+                .map_ok(|row| row.take_first_value().map(Evaluated::from));
+            let first = stream
+                .next()
+                .await
+                .unwrap_or_else(|| Err(EvaluateError::NestedSelectRowNotFound.into()))?;
+            let second = stream.next().await;
+            match second {
+                Some(_) => Err(EvaluateError::MoreThanOneRowReturned.into()),
+                None => first,
+            }
```
panarch, line 83: Thanks!

In here, `take(2)` would be useful in this case to simplify the codes.
```rust
select(storage, query, context.as_ref().map(Rc::clone))
    .await?
    .map_ok(|row| row.take_first_value().map(Evaluated::from))
    .take(2)
    .await
    .try_collect::<Vec<..
```

core/src/executor/evaluate/mod.rs
```diff
@@ -68,12 +68,22 @@ pub async fn evaluate<'a, T>(
...
+                .map_ok(|row| row.take_first_value().map(Evaluated::from))
+                .take(2)
+                .try_collect::<Vec<_>>()
+                .await?;
+            match evaluations.len() {
+                0 => Err(EvaluateError::NestedSelectRowNotFound.into()),
+                1 => evaluations
+                    .into_iter()
+                    .nth(0)
+                    .unwrap_or_else(|| Err(EvaluateError::NestedSelectRowNotFound.into())),
```
devgony, line 83: Thanks for suggetion of `take(2)` way.
But I'm not sure whether it's fine to use `NestedSelectRowNotFound` error again here🥲

core/src/executor/evaluate/mod.rs
```diff
@@ -79,7 +79,7 @@ pub async fn evaluate<'a, T>(
                 0 => Err(EvaluateError::NestedSelectRowNotFound.into()),
                 1 => evaluations
                     .into_iter()
-                    .nth(0)
+                    .next()
                     .unwrap_or_else(|| Err(EvaluateError::NestedSelectRowNotFound.into())),
```
devgony, line 83: Thanks for suggestion of take(2) way.
I tried take(2) and matching len().
But I'm not sure whether it's fine to use NestedSelectRowNotFound error again here🥲

panarch: Good progress!
In this case, using `if` looks better than `match`

```rust
if evaluations.len() > 1 {
    return Err(EvaluateError::MoreThanOneRowReturned.into());
}

evaluations
    .into_iter()
    .next()
    .unwrap_or_else(|| Err(EvaluateError::NestedSelectRowNotFound.into()))
```

test-suite/src/ordering.rs
```diff
@@ -35,7 +35,7 @@ test_case!(ordering, async move {
         (0, "SELECT * FROM Operator WHERE 3 > 3;"),
         (
             5,
-            "SELECT * FROM Operator o1 WHERE 3 > (SELECT id FROM Operator WHERE o1.id < 100);",
+            "SELECT * FROM Operator o1 WHERE 3 > (SELECT min(id) FROM Operator WHERE o1.id < 100);",
```
panarch, line 38: nit.
All other SQL keywords are using uppercase letters but `min` does not though `min` is also the SQL keyword.
It would be good to use all uppercase for consistency.

panarch:
btw. if you write "resolve" keyword in the main PR content,
e.g.
resolve #528

Then, it will automatically close the issue when the PR is merged.

devgony:
> btw. if you write "resolve" keyword in the main PR content, e.g. resolve #528
> 
> Then, it will automatically close the issue when the PR is merged.

🤭 I didn't know that.
Thank you for the cool Tip!
I just edited to `resolve`!

Issue #525, opened 2022-04-16
Bump `sqlparser-rs` version to 0.16.0


ever0de:
It's an internally known issue, but I'm posting it for public notice.

The change in sqlparser is bigger than you think, so it should be taken as a conversion(enhancement) point.

earney:
Do you think it would be better to upgrade one version of sqlparser-rs  from v13 to v14 then from v14 to v15 and then v15 to v16 or just do a jump from v13 to v16?

I did try to modify the Cargo.toml from v13 to v16, and yes, there are a bunch of changes.

ever0de:
> Do you think it would be better to upgrade one version of sqlparser-rs from v13 to v14 then from v14 to v15 and then v15 to v16 or just do a jump from v13 to v16?
> 
> I did try to modify the Cargo.toml from v13 to v16, and yes, there are a bunch of changes.

you're right.
It would be rather easy to go up step by step.
It would be good to think of this as a goal.

earney:
I believe this issue has been resolved.

panarch:
@earney Thanks!

PR #541, opened 2022-04-29
Migrate sqlparser-rs version from v0.13 to v0.14,

panarch:
In v0.14, function arg takes FunctionArgExpr, not Expr.
Remove `Expr::Wildcard` and `Expr::QualifiedWildcard`.
Add `CountArgExpr` enum, only `COUNT` aggregate function takes wildcard expression.

PR #542, opened 2022-04-30
Migrate code from sqlparser v14 to v16 

earney:
update gluesql so that it works with v15 and v16 of sqlparser-rs.

### DIFF
--- a/core/src/ast/ast_literal.rs
+++ b/core/src/ast/ast_literal.rs
@@ -24,6 +24,22 @@ pub enum DateTimeField {
     Hour,
     Minute,
     Second,
+    Century,
+    Decade,
+    Week,
+    Dow, //day of week
+    Doy, //day of year
+    Epoch,
+    Isodow,
+    Isoyear,
+    Julian,
+    Microseconds,
+    Millenium,
+    Milliseconds,
+    Quarter,
+    Timezone,
+    TimezoneHour,
+    TimezoneMinute,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
--- a/core/src/ast/ast_literal.rs
+++ b/core/src/ast/ast_literal.rs
@@ -24,22 +24,6 @@ pub enum DateTimeField {
     Hour,
     Minute,
     Second,
-    Century,
-    Decade,
-    Week,
-    Dow, //day of week
-    Doy, //day of year
-    Epoch,
-    Isodow,
-    Isoyear,
-    Julian,
-    Microseconds,
-    Millenium,
-    Milliseconds,
-    Quarter,
-    Timezone,
-    TimezoneHour,
-    TimezoneMinute,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
--- a/core/src/data/interval/mod.rs
+++ b/core/src/data/interval/mod.rs
@@ -287,7 +287,7 @@ impl Interval {
                 format!("{:?}", to),
             )
             .into()),
-            (_, _) => Err(IntervalError::UnsupportedExtractPart.into()),
+            (None, _) => Err(IntervalError::Unreachable.into()),
         }
     }
 }
--- a/core/src/data/interval/mod.rs
+++ b/core/src/data/interval/mod.rs
@@ -287,7 +287,7 @@ impl Interval {
                 format!("{:?}", to),
             )
             .into()),
-            (None, _) => Err(IntervalError::Unreachable.into()),
+            (_, _) => Err(IntervalError::UnsupportedExtractPart.into()),
         }
     }
 }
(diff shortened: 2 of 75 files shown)


earney:
resolves issue #525

core/src/ast/ast_literal.rs
```diff
@@ -24,6 +24,22 @@ pub enum DateTimeField {
...
+    Isodow,
+    Isoyear,
+    Julian,
+    Microseconds,
+    Millenium,
+    Milliseconds,
+    Quarter,
+    Timezone,
+    TimezoneHour,
+    TimezoneMinute,
```
panarch, line 42: These are new enum values which GlueSQL does not support yet.

fyi.
`translate` module has two main roles. Second role would be good to consider in this case.

1. Generating executable AST which GlueSQL `executor` can handle
Not like other many SQL databases, GlueSQL's query planner layer is not mandatory.
Storage makers can choose whether to use query planner or not, or using partial planners can be also the choice.
`translate` module converts `sqlparser-rs`'s AST to `gluesql`'s internal AST which also includes a kind of default execution plan.
So, in the `plan` module, it takes AST and returns the same AST (same type, but internal info can be different)

2. Filtering out superset AST values which `sqlparser-rs` supports but `gluesql` does not.
We use `sqlparser-rs` to parse SQL text so we don't need to implement SQL parser by ourselves.
This saves tons of our time, but there's something to consider. `sqlparser-rs` aims to support multiple SQL syntaxes as many as possible. This means that the coverage `sqlparser-rs` has is wider than what GlueSQL supports.
Therefore, it would be good to remove all features & values which GlueSQL does not support in the `translate` module.
Then, theoretically `executor` layer can have zero `Unreachable-` errors. (currently lots of them has removed but still some of them remains)

---
```
    Century,
    Decade,
    Week,
    Dow, //day of week
    Doy, //day of year
    Epoch,
    Isodow,
    Isoyear,
    Julian,
    Microseconds,
```
We can think of adding these values to our AST as a new feature to implement.
Then, it is not necessary to include these in the code before we actually support them.

---
e.g.
https://github.com/gluesql/gluesql/pull/541

Currently, GlueSQL supports `COUNT(*)` wildcard expression but `COUNT(table.*)` is not supported.
Because of this, though `sqlparser-rs` has `FunctionArgExpr::QualifiedWildcard(_)` but GlueSQL AST `CountArgExpr` does not.
```
pub enum CountArgExpr {
    Expr(Expr),
    Wildcard,
}
```
When we support `COUNT(table.*)`, then that would be the proper time to add `QualifiedWildcard` to `CountArgExpr`.

earney: So translate_datetime_field function needs to return a Result<DataTimeField> so that an error can be produced if non supported sql is produced by the user?

earney: Okay.. I think I got it now..  Can you take another look?

Thanks!

core/src/data/interval/mod.rs
```diff
@@ -287,7 +287,7 @@ impl Interval {
                 format!("{:?}", to),
             )
             .into()),
-            (None, _) => Err(IntervalError::Unreachable.into()),
+            (_, _) => Err(IntervalError::UnsupportedExtractPart.into()),
```
panarch, line 290: Is this code reachable? Then, we need to have a test case for generting this `IntervalError::UnsupportedExtractPart`

core/src/translate/ast_literal.rs
```diff
@@ -33,15 +37,20 @@ pub fn translate_ast_literal(sql_value: &SqlValue) -> Result<AstLiteral> {
     })
 }
 
-pub fn translate_datetime_field(sql_datetime_field: &SqlDateTimeField) -> DateTimeField {
-    match sql_datetime_field {
+pub fn translate_datetime_field(sql_datetime_field: &SqlDateTimeField) -> Result<DateTimeField> {
```
panarch, line 40: 👍

panarch (changes requested):

core/src/translate/ast_literal.rs
```diff
@@ -23,8 +23,12 @@ pub fn translate_ast_literal(sql_value: &SqlValue) -> Result<AstLiteral> {
...
         } => AstLiteral::Interval {
             value: value.to_owned(),
-            leading_field: leading_field.as_ref().map(translate_datetime_field),
-            last_field: last_field.as_ref().map(translate_datetime_field),
+            leading_field: leading_field
+                .as_ref()
+                .map(|x| translate_datetime_field(x).unwrap()),
+            last_field: last_field
+                .as_ref()
+                .map(|x| translate_datetime_field(x).unwrap()),
```
panarch, line 31:
```rust
leading_field: leading_field      
    .as_ref()                     
    .map(translate_datetime_field)
    .transpose()?,                
last_field: last_field            
    .as_ref()                     
    .map(translate_datetime_field)
    .transpose()?,                
```

I checked in the last review but forgot to mention it.
This is the last one, everything else looks great. Thanks! :)

ever0de (approved):
Looks good 👍

panarch (approved):
Awesome, thanks a lot! merging 👍

Issue #545, opened 2022-05-02
Adding unsigned ints and signed ints together (ie overflow issues)


earney:
### Discussed in https://github.com/gluesql/gluesql/discussions/544

<div type='discussions-op-text'>

<sup>Originally posted by **earney** May  2, 2022</sup>
sqlparser-rs version 16, has the capability to have unsigned integers.   I'm wondering how we want to address arithmetic operations between signed and unsigned numbers?   Or maybe a better question is how do we address out of range and overflow issues in gluesql?

Here is how mysql addresses these issues: [out of range/overflow](https://dev.mysql.com/doc/refman/8.0/en/out-of-range-and-overflow.html)

Addressing my first question..
How should we handle the following:

let x:i8 = 127
let y:u8 = 128

x can contain any value between -127 and 127
y can contain any value between 0 and 255.
The result could contain any value between -127 and 382 which is outside the range for an i8 and an u8.  Should we promote the result to an i16 (or i32, or i64),?   If the result is being put into a table with a given datatype (maybe an i8), then let it complain if the result is out of range?   What do you think?

It seems that currently in gluesql, when we do something like " select 2+125 from mytable", the 2 and 125 are treated as ints (i64), and if try to insert them into a column that is an i8, it gives the following error:

insert into myi8 select 2 + 125 from myi8;
[error] incompatible data type, data type: Int8, value: I64(
    127,
)

For most things, this seems to work just fine. We can use the cast function to convert the result to an i8.  I guess we could convert the i8 and u8 (from my example above) to i64? and then do the math?

for math between i64 and u64, we could convert both operands to an i128?

I guess for u128 or i128 and anything else just promote to i128 and hope for the best, and produce an error if an overflow occurs?

Billy</div>

earney:
I decided to convert this discussion into an issue because I ran into the following which I think should raise an error (overflow error):

```
gluesql> create table myi8 (id int(8));
gluesql> insert into myi8 values (1);
1 row inserted

gluesql> insert into myi8 select cast(255 as int(8)) from myi8;
1 row inserted

gluesql> select * from myi8;
╭─────╮
│ id  │
╞═════╡
│ 1   │
│ 127 │
╰─────╯

gluesql>

```
Note that the value 127 is inserted into the table instead of 255.  127 is the maximum allowed value for an int(8) (i8).   Should an error be printed to the screen when we do the insert statement above?

earney:
So, just to make sure I understand what you are proposing...
[...]

PR #546, opened 2022-05-02
cast 255 as int8, should return an error, but returns 127 instead (resolves issue #545)

earney:
Here's the error I found (this should probably produce an error, not insert 127 into an int(8) field).

```
gluesql> create table myi8 (id int(8));
gluesql> insert into myi8 values (1);
1 row inserted

gluesql> select cast(255 as int(8)) from myi8;
╭─────────────────────╮
│ CAST(255 AS INT(8)) │
╞═════════════════════╡
│ 127                 │
╰─────────────────────╯
gluesql>
```

This PR (resolves issue #545) implements the following result:

```
luesql> select cast(255 as int(8)) from dual;
[error] literal cast failed to Int(8): 255
```

### DIFF
--- a/core/src/data/value/error.rs
+++ b/core/src/data/value/error.rs
@@ -98,6 +98,9 @@ pub enum ValueError {
     #[error("literal cast failed to date: {0}")]
     LiteralCastToDateFailed(String),
 
+    #[error("literal cast failed to Int(8): {0}")]
+    LiteralCastToInt8Failed(String),
+
     #[error("literal cast failed to time: {0}")]
     LiteralCastToTimeFailed(String),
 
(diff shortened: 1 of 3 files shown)


earney:
If the code looks good, I can look into converting the other (*, Literal::Number) cases so that they rely on BigDecimal to do the heavy lifting to convert from bigdecimal to i8, i64, f64, etc.

earney:
resolves issue #545

ever0de:
> resolves issue #545

This phrase should be at the top of your PR conversation. 👍 

e.g) <https://github.com/gluesql/gluesql/pull/512>
image

earney:
This PR is still officially a draft, but I would like to receive comments about it.  I believe we can use the same basic code for other number types (int, float, decimal), and let bigdecimal do the heavy lifting by using bigdecimals to_i8, to_i64, to_f64 functions to convert the bigdecimal to the appropriate data type.   I am also looking at implementing solutions for i32, i128, and the unsigned versions (u8, u32, u64, and u128), in which the cast function would need code for each data type.

PR #547, opened 2022-05-06
Implement SledStorage export & import features

### LINKED ISSUE #533: "fetch failed - expired transaction has used (txid)" error after sled import
I want to migrate data from an existing gluesql database using sled inner function `export` and `import`, but after imported even a query operation will throw "fetch failed - expired transaction has used (txid)" error.

The following is a simple example inspired by unit test sled_transaction_gc in "sled_transaction.rs":
```rust
    let path = &format!("{}/transaction_gc", PATH_PREFIX);
    fs::remove_dir_all(path).unwrap_or(());

    let storage1 = SledStorage::new(path).unwrap();
    let mut glue1 = Glue::new(storage1);

    // add some tables and data...

    let storage = glue1.storage.unwrap();
    let exports = storage.tree.export();

    let path3 = &format!("{}/transaction_gc3", PATH_PREFIX);
    fs::remove_dir_all(path3).unwrap_or(());

    let storage3 = SledStorage::new(path).unwrap();
    storage3.tree.import(exports);
    let mut glue3 = Glue::new(storage3);

    test!(glue3 "SELECT * FROM NewGarlic", Err(Error::StorageMsg("fetch failed - expired transaction has used (txid)".to_owned())));
```

The `txid` after imported does not sync as desired. So is there anything else I should do to let pass, or there is another way to migrate database?

panarch:
Native export & import methods supported by Sled can cause problem if these are used for GlueSQL migration.
SledStorage wraps Sled export & import methods and provide its own ones.

@raindust
I also added a test case which you mentioned in the issue, thanks :)
Now you will be able to use `export` and `import` provided by `SledStorage` itself.
I'll soon to release v0.11 and this fix will be included.
ref. `storages/sled-storage/tests/export_and_import.rs`

fix #533

### DIFF
--- a/storages/sled-storage/src/error.rs
+++ b/storages/sled-storage/src/error.rs
@@ -23,6 +23,8 @@ pub enum StorageError {
     Str(#[from] str::Utf8Error),
     #[error(transparent)]
     SystemTime(#[from] time::SystemTimeError),
+    #[error(transparent)]
+    TryFromSlice(#[from] std::array::TryFromSliceError),
 }
 
 impl From<StorageError> for Error {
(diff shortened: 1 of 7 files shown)


ever0de (approved):
LGTM

Issue #482, opened 2022-02-12
Support `INT(128) unsigned` data type


ever0de:

ever0de:
ref #1019

Issue #384, opened 2021-10-13
Support `INT(32)` data type


heka1024:
Like `MySQL` and `PostgreSQL`, many sql engines support a wide range of integers. For example, `MySQL` has a numeric type `TINYINT` which range is -128 to 127. Like this, it would be nice if we support such a wide range of integers.

kevingzhang:
If I need to use UINT(128), I know GlueSQL currently doesn't support 128bit int. Is there any workaround for this?
thank you in advance

ever0de:
> If I need to use UINT(128), I know GlueSQL currently doesn't support 128bit int. Is there any workaround for this? thank you in advance

<https://github.com/gluesql/gluesql/issues/482>

It will be added when the next task is completed.
thank you

panarch:
Thanks to @earney 
resolved in #608

PR #548, opened 2022-05-06
Add int32, int128, uint8, uint8, uint32, uint64 and uint128

earney:
This PR implements basic support for int32, int128, and unsigned versions of int8, uint32, uint64 and uint128.  It implements improvements made by issue #545, #482 and #384.

I implemented addition, subtraction, multiplication, division and modulo by implementing the following rules:
if a and b are signed integers then the result datatype is the largest datatype of a and b.
if a and b are unsigned integers then the result datatype is also the largest datatype of a * b.
If one operand is an unsigned integer and the other is a signed integer, I convert the unsigned integer to a signed integer like so:
u8 => i32
u32 => i64
u64 => i128
u128 => i128  (we use extra check and through error if the u128 cannot be converted to a i128.
and then I use the 1st rule above to figure out the result datatype.

This is a large PR, so there will probably be errors and issues that need to be addressed.
I also pulled the PR that implements sqlparser-rs v14 to v16 so this implementation could be tested, etc.
Comments and Suggestions welcomed!


earney:
I know there is still a lot of clean up ,etc that needs to happen, but I think I'm ready for some comments and suggestions on how to improve the code.

earney:
This is failing on one of the int32 and int128 test-suites, I added.  Any idea why it is having problems with comparing an Int128 and an Int32 vs zero (0)?

```
[SQL] SELECT field_one FROM Item WHERE field_one > 0
thread 'int32' panicked at 'assertion failed: `(left == right)`
  left: `2`,
 right: `0`: 
[err: number of rows]
expected: [[I32(1)], [I32(3)]]
   found: []', test-suite/src/tester.rs:52:5

```

ever0de:
Sorry for the late reply.
Could you possibly break the PR into each type(ex. int32 PR, int128 PR, ...)?
thank you.

earney:
Sure..  I will partiition by the different data types.
Thanks for the feedback..

Issue #520, opened 2022-04-12
implement show indexes from table


ghost:
implement show index from table syntax

earney:
> Is it correct to support the syntax?
> 
>     * `SHOW INDEXES FROM table_name;`
> 
> 
> assigned to you. thanks for the report

Yes, you are correct.  I'll try to fix the title.

earney:
If I'm viewing sqlparser correctly, it doesn't support show indexes syntax, so I put in an issues with the sqlparser team requesting that enhancement.

earney:
In sqlparser-rs version 0.16, this syntax is supported:

ql: "show indexes from mytable"
AST: [ShowVariable { variable: [Ident { value: "indexes", quote_style: None }, Ident { value: "from", quote_style: None }, Ident { value: "mytable", quote_style: None }] }]

earney:
PR #554  adds a third column to the output (displays expression used to define the index).  Once that PR is merged, this issue can be closed.

panarch:
resolved in #551

PR #551, opened 2022-05-13
Show indexes from table... (resolves issue #520)

earney:
allows users to issue:
show indexes from table 
and it shows the indexes created for a table.

resolves issue #520


panarch:
If you plan to return the result of `SHOW INDEXES {table_name};` as … including not only a tuple of (index_name, ordering) but also (index_name, ordering, expr), then it would be too big one for now.

`SchemaIndexOrd` is simple. but the issue is `Expr`.
GlueSQL does not support display method for `Expr` to string and I am also not sure that translated AST needs to preserve original raw SQL data. In some cases, gluesql::ast can be generated by other APIs, SQL is not the only way to generate gluesql::ast.

How about simplifying the functionality?
Rather than returning all index info, it would be enough to have only (index_name, ordering) info.

earney:
When displaying the output for show indexes, it shows BinaryOp.  (see below in the image). What is the best way to get this to show id + num?

image

panarch:
> When displaying the output for show indexes, it shows BinaryOp. (see below in the image). What is the best way to get this to show id + num?

Default behavior of `Display` macro simply returns the name of enum type (e.g. `BinaryOp`)
If we want to see something like `id + num`, then this means that we need to implement `Display` trait for `Expr` all by ourselves.
That requires huge resource and it will add cost to maintain codes, too.
So... there is no simple way to show `id + num` for now.

earney:
Just returning the index name and Order is fine with me.  I might create an issue on decoding the Exprs, so that we can eventually have something like id + num instead of what is produced now.  I will work on making the changes to this PR.

earney:
removing the 3rd column is complete.

core/src/executor/execute.rs
```diff
@@ -64,6 +64,7 @@ pub enum Payload {
 
     #[cfg(feature = "metadata")]
     ShowVariable(PayloadVariable),
+    ShowIndexes(Vec<SchemaIndex>),
```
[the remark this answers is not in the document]
earney: ok.  I just made the change..

core/src/data/schema.rs
```diff
@@ -2,16 +2,17 @@ use {
...
+    strum_macros::Display,
 };
 
-#[derive(Clone, Copy, Debug, Serialize, Deserialize)]
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Display)]
 pub enum SchemaIndexOrd {
     Asc,
     Desc,
     Both,
 }
```
panarch, line 13: When we use string representation for SQL keyword, we use uppercased letters.
e.g.
image

Could we have same patterns?
```rust
#[strum(to_string = "ASC")]
Asc,
```

core/src/translate/mod.rs
```diff
@@ -146,6 +146,25 @@ pub fn translate(sql_statement: &SqlStatement) -> Result<Statement> {
                 "VERSION" => Ok(Statement::ShowVariable(Variable::Version)),
                 v => Err(TranslateError::UnsupportedShowVariableKeyword(v.to_string()).into()),
             },
+            (3, Some(keyword)) => match keyword.value.to_uppercase().as_str() {
+                //   AST: [ShowVariable { variable: [Ident { value: "indexes", quote_style: None },
+                //     Ident { value: "from", quote_style: None }, Ident { value: "mytable", quote_style: None }] }]
```
panarch, line 151: For the PRs which are ready for review, it is recommended to have less comments (except for doc comment).
This looks better not to have these.

cli/src/print.rs
```diff
@@ -47,6 +47,13 @@ impl<W: Write> Print<W> {
 
                 writeln!(self.output, "{}\n", table)?;
             }
+            Payload::ShowIndexes(indexes) => {
+                let mut table = get_table(vec!["Index Name", "Order"]);
+                for index in indexes {
+                    table.add_row([index.0, index.1.to_string()]);
+                }
+                writeln!(self.output, "{}\n", table)?;
+            }
```
panarch, line 56: Could you add a test case for `SHOW INDEXES --` CLI result?

cli/src/print.rs
```diff
@@ -47,6 +47,13 @@ impl<W: Write> Print<W> {
 
                 writeln!(self.output, "{}\n", table)?;
             }
+            Payload::ShowIndexes(indexes) => {
+                let mut table = get_table(vec!["Index Name", "Order"]);
+                for index in indexes {
```
panarch, line 52: This looks good place to use destruct pattern 
```rust
for (index_name, order) in indexes {
```

core/src/ast/expr.rs
```diff
@@ -4,9 +4,10 @@ use {
         UnaryOperator,
     },
     serde::{Deserialize, Serialize},
+    strum_macros::Display,
 };
 
-#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, Display)]
```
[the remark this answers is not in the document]
earney: good catch.. thank you.

core/src/translate/mod.rs
earney: ok.. done!

cli/src/print.rs
earney: +1.. done!

core/src/ast/mod.rs
```diff
@@ -99,6 +99,7 @@ pub enum Statement {
     /// SHOW VARIABLE
     #[cfg(feature = "metadata")]
     ShowVariable(Variable),
+    ShowIndexes(ObjectName),
```
panarch, line 102: Sorry for the late review.
I was thinking about to provide you a good one shot review, but it becomes too delayed.

Here, `SHOW INDEXES - ` only works for storages which has `Index` support.
```rust
#[cfg(feature = "index")]
CreateIndex,             
#[cfg(feature = "index")]
DropIndex,               
#[cfg(feature = "index")]
ShowIndexes(ObjectName), 
```

It would be good to handle `Payload::ShowIndexes(_)` as optional feature.

panarch (approved):
Awesome!
Looks great, now we support fetching index info feature!
Thanks a lot 👍

PR #553, opened 2022-05-15
Unify GroupKey, HashKey(join) and UniqueKey into a single data::Key s…

panarch:
…truct

There were three different structs for describing converted value which implements `Hash`.
* GroupKey - used in `GROUP BY` aggregation
* HashKey - used for handling `JOIN`
* UniqueKey - used for `UNIQUE` constraint

They had fundamentally same role but there were being maintained separately.
In this PR, I added `data::Key` struct and all existing `GroupKey`, `HashKey` and `UniqueKey` are replaced by the `Key`.

`data::Key` can be also used for many other roles if we need a kind of `Value` but with `Hash` impl.

fyi. @ever0de 
Using `data::Key`, for the next work, I'll remove generic uses in store traits (for key) and replace it to use `data::Key` instead user customizable generics `T`.
All of these are also a kind of prior tasks for supporting `PRIMARY KEY`.


core/src/executor/join.rs
```diff
@@ -273,10 +270,9 @@ impl<'a> JoinExecutor<'a> {
                     .await?
                     .try_into()?;
 
-                    let hash_key = match hash_key {
-                        Some(hash_key) => hash_key,
-                        None => return Ok(None),
-                    };
+                    if matches!(hash_key, Key::None) {
+                        return Ok(None);
+                    }
```
ever0de, line 275: Rather than using this as `Option<Key>`,
Is there any reason to add a variant named None to the key enum?

panarch: Good question, I also actually tried `Option<Key>` first :)
Then, I changed it to just use `Key` with `Key::None`.
Though existing three cases handle the `None` case of the key slightly differently, those were the same that they all use `None`.
This meant that for all three cases I needed to specify `Option<Key>` and there were no use cases of using `Key` without `Option`, so I changed the codes to include `None` in the `Key`.

`Key::None` is actually closer to `Null`. However in this case `Key::None == Key::None` but `Value::Null != Value::Null`, so I thought that using `Null` keyword is not a good idea.

PR #554, opened 2022-05-16
Implement `ToSql` trait which displays AST as SQL text format

earney:
Implementing this so that we can add a definition/description column for the show indexes command.

For example, if we have the following expression;

```
Expr::BinaryOp {
                left: Box::new(Expr::Identifier("id".to_string())),
                op: BinaryOperator::Plus,
                right: Box::new(Expr::Identifier("num".to_string()))
 }
```
this would be converted to:
id + num


earney:
I am thinking about merging this PR with PR #551 since the decode function will primarily be used with the show indexes from table functionality.  Should I keep it as a separate PR, or merge with #551?

earney:
Will continue this PR once PR #551 is merged.  Changed status to draft until then..

core/src/ast/expr_decoder.rs
```diff
@@ -0,0 +1,323 @@
...
+            Aggregate::Avg(e) => format!("Avg({:})", decode(e)),
+        },
+        Expr::Function(f) => {
+            format!("{:}(todo:args)", f)
+        }
+        // todo's...  these require enum query..
+        //Expr::InSubquery {expr, subquery, negated} => format!("InSubquery({:}, subquery:{:}, negated: {:})", decode(*expr),  *subquery, negated),
+        //Expr::Exists(q) => format!("Exists({:})", *query),
+        //Expr::Subquery(q) => format!("Subquery({:})", *query),
+        _ => format!("Unimplemented Decode Expression: {:#?}", expr),
```
panarch, line 116: This looks good start!
We don't need to handle all the display impl in here.
As you did for `Expr::Function`,
```rust
Expr::Function(f) => {
    format!("{:}(todo:args)", f)
}
```

How about wrapping up something like this for now?
```rust
Expr::Exists(q) => "EXISTS(..)",
```

Then, we can keep this SQL display task in following PRs.

earney: sounds good to me.

core/src/ast/expr_decoder.rs
```diff
@@ -0,0 +1,313 @@
...
+            // is there a better way of doing this?  (ie adding '.' between each string?)
+            // tried fold and couldn't figure out how add . between each item
+            let mut str: String = "".to_string();
+            for _s in s {
+                if !str.is_empty() {
+                    str += "."
+                }
+                str += _s;
+            }
+            str
```
panarch, line 20: You can consider this, 
```
Expr::CompoundIdentifier(idents) => idents.join(".")
```

core/src/ast/expr_decoder.rs
```diff
@@ -0,0 +1,313 @@
+use super::{Aggregate, AstLiteral, CountArgExpr, Expr};
+
+pub fn decode(expr: &Expr) -> String {
+    match expr {
+        Expr::Identifier(s) => s.to_string(),
+        Expr::BinaryOp { left, op, right } => {
+            format!("{:} {:} {:}", decode(&*left), op, decode(&*right))
```
panarch, line 7:
```rust
format!("{:} {:} {:}",
```
Here `{:}` looks unnecessary, when you expect to use `Display` trait, then `{}` is enough

core/src/ast/operator.rs
```diff
@@ -1,29 +1,48 @@
...
-#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, Display)]
 pub enum BinaryOperator {
+    #[strum(to_string = "+")]
     Plus,
+    #[strum(to_string = "-")]
     Minus,
+    #[strum(to_string = "*")]
     Multiply,
+    #[strum(to_string = "/")]
```
panarch, line 24: In this case, `strum_macros::Display` conversion is too radical.
How about.. rather than using Display, to impl `ToSql` trait for these (BinaryOperator and UnaryOperator)

panarch (changes requested):
I am also for this decision you add feature for converting from `gluesql::ast` to SQL string.
Thanks a lot for the meaningful changes.

Though `gluesql::ast` is a superset of SQL string so this conversion can lose some information, it would be still very helpful for db users to recognize and figure out what the AST means.

For this, rather than `decode`, how about adding a `ToSql` trait for this?
`decode` sounds like that this function would provide 1:1 conversion from AST to SQL but it cannot. (e.g. some query planned result will be lost)
ref.
image

Then, we can have
```rust
pub trait ToSql {              
    fn to_sql(&self) -> String;
}                              
```
in the root of ast, `ast/mod.rs`
and you can move current codes and unit tests to `expr.rs` (from `expr_decoder.rs`).
I think that it would be good to be located with AST declaration because it can also provide good extra info to users.

---
And.. this PR contains two different works.
Each one is big and meaningful enough to be as an independent change.

1. provide sql representation to gluesql ast
2. add description column to show indexes from table command

Could we proceed the first one first then get moved to the second?

core/src/ast/expr_decoder.rs
ever0de: > format!("{}", decode(&*left), op, decode(&*right))

Note that `{}` takes only one parameter.
In the code you wrote above, there is only one parentheses, but there are 3 parameters.

core/src/ast/operator.rs
```diff
@@ -1,29 +1,48 @@
...
 pub enum UnaryOperator {
+    #[strum(to_string = "+")]
     Plus,
+    #[strum(to_string = "-")]
     Minus,
+    #[strum(to_string = "~")]
     Not,
-    Factorial,
+    // #[strum(to_string="factorial")]
+    Factorial, // why is this here and not in the function enum?
```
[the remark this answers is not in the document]
earney: ok.  I thought that it was gonna be called like so: factorial(3)

core/src/ast/expr.rs
```diff
@@ -61,3 +61,303 @@ pub enum Expr {
...
+                        Expr::Literal(AstLiteral::Number(BigDecimal::from_str("2").unwrap())),
+                        Expr::Literal(AstLiteral::QuotedString("b".to_string()))
+                    )
+                ],
+                else_result: Some(Box::new(Expr::Literal(AstLiteral::QuotedString(
+                    "c".to_string()
+                ))))
+            }
+            .to_sql()
+        );
```
panarch, line 321: Test code readability is also important which we need to make it.
We could consider using regex to make it easier to read

e.g. you can change these to apply to other tests.
```rust
use regex::Regex;                                                                     
                                                                                      
let re = Regex::new(r"\n\s+").unwrap();                                               
let trim = |s: &str| re.replace_all(s.trim(), "\n").into_owned();                     
                                                                                      
assert_eq!(                                                                           
    trim(                                                                             
        r#"                                                                           
        CASE id                                                                       
          WHEN 1 THEN "a"                                                             
          WHEN 2 THEN "b"                                                             
          ELSE "c"                                                                    
        END                                                                           
        "#                                                                            
    ),                                                                                
    Expr::Case {                                                                      
        operand: Some(Box::new(Expr::Identifier("id".to_string()))),                  
        when_then: vec![                                                              
            (                                                                         
                Expr::Literal(AstLiteral::Number(BigDecimal::from_str("1").unwrap())),
                Expr::Literal(AstLiteral::QuotedString("a".to_string()))              
            ),                                                                        
            (                                                                         
                Expr::Literal(AstLiteral::Number(BigDecimal::from_str("2").unwrap())),
                Expr::Literal(AstLiteral::QuotedString("b".to_string()))              
            )                                                                         
        ],                                                                            
        else_result: Some(Box::new(Expr::Literal(AstLiteral::QuotedString(            
            "c".to_string()                                                           
        ))))                                                                          
    }                                                                                 
    .to_sql()                                                                         
);                                                                                    
```

panarch (changes requested):

core/src/ast/expr.rs
```diff
@@ -61,3 +61,303 @@ pub enum Expr {
...
+                    if !s.is_empty() {
+                        s += ",";
+                    }
+                    s += &item.to_sql();
+                }
+
+                match negated {
+                    true => format!("{} NOT IN ({})", expr.to_sql(), s),
+                    false => format!("{} IN ({})", expr.to_sql(), s),
+                }
```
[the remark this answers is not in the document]
earney: I did not know you could do "{expr} ....   {list}"..  Thanks!

core/src/ast/expr.rs
```diff
@@ -61,3 +61,303 @@ pub enum Expr {
...
+                    &*high.to_sql()
+                ),
+
+                false => format!(
+                    "{} BETWEEN {} AND {}",
+                    &*expr.to_sql(),
+                    &*low.to_sql(),
+                    &*high.to_sql()
+                ),
+            },
```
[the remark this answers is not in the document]
earney: yes. that looks much better.

cli/src/print.rs
```diff
@@ -1,3 +1,4 @@
+use gluesql_core::ast::ToSql;
 use {
```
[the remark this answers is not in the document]
earney: ok.  I will try to only use one use statement from now on.

core/src/ast/expr.rs
earney: good idea.

core/src/ast/expr.rs
```diff
@@ -61,3 +61,294 @@ pub enum Expr {
...
+                };
+                for (_when, _then) in when_then {
+                    str += format!("\nWHEN {} THEN {}", _when.to_sql(), _then.to_sql()).as_str();
+                }
+
+                match else_result {
+                    Some(s) => str += format!("\nELSE {}", s.to_sql()).as_str(),
+                    None => str += "", // no operation?
+                };
+                str + "\nEND"
```
panarch, line 134: It would be good not to use `mut` here
In Rust, it is recommended to use variable shadowing because it can explicitly show the lifetime of each variable.

e.g.
```rust
let operand = match operand {
    Some(operand) => format!("CASE {}", operand.to_sql()),
    None => "CASE".to_owned(),
};  
    
let when_then = when_then
    .iter()
    .map(|(when, then)| format!("WHEN {} THEN {}", when.to_sql(), then.to_sql()))
    .collect::<Vec<_>>()
    .join("\n");

let else_result = match else_result {
    Some(else_result) => format!("ELSE {}", else_result.to_sql()),
    None => String::new(),
};  

[operand, when_then, else_result, "END".to_owned()].join("\n")
```

panarch (changes requested):
Nice!
Here... could you split the PR into two?
Two independent tasks exist in the single PR.

1. provide sql representation to gluesql ast
2. add description column to show indexes from table command

earney:
> Nice!
> Here... could you split the PR into two?
> Two independent tasks exist in the single PR.
> 
> 1. provide sql representation to gluesql ast
> 2. add description column to show indexes from table command

OK.  Will do

cli/src/print.rs
```diff
@@ -1,5 +1,6 @@
...
     comfy_table::{modifiers::UTF8_ROUND_CORNERS, presets::UTF8_BORDERS_ONLY, Row, Table},
+    gluesql_core::ast::ToSql,
     gluesql_core::prelude::{Payload, PayloadVariable},
```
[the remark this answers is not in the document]
earney: ok. thanks I missed that one.

core/src/ast/ast_literal.rs
```diff
@@ -1,6 +1,10 @@
+use crate::ast::ToSql;
+use crate::data::Interval;
+//use crate::data::Literal;
 use {
     bigdecimal::BigDecimal,
     serde::{Deserialize, Serialize},
+    strum_macros::Display,
```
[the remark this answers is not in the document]
earney: Thanks!

core/src/ast/expr.rs
```diff
@@ -61,3 +61,294 @@ pub enum Expr {
...
+            Expr::UnaryOp { op, expr } => format!("{}{}", op.to_sql(), &*expr.to_sql()),
+            Expr::Cast { expr, data_type } => {
+                format!("CAST({} AS {})", &*expr.to_sql(), data_type)
+            }
+            Expr::Extract { field, expr } => {
+                format!("EXTRACT({} FROM \"{}\")", field, &*expr.to_sql())
+            }
+            Expr::Nested(expr) => format!("todo:Nested({})", &*expr.to_sql()),
+            Expr::Literal(s) => s.to_sql(),
+            Expr::TypedString { data_type, value } => format!("{data_type}(\"{value}\")"),
```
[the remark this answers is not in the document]
earney: ok.. thanks!

core/src/ast/expr.rs
earney: Thanks for the code rewrite..  It helps a lot.

core/src/ast/ast_literal.rs
```diff
@@ -17,7 +19,27 @@ pub enum AstLiteral {
...
+            AstLiteral::HexString(hs) => format!("\"{}\"", hs),
+            AstLiteral::Interval {
+                value,
+                leading_field,
+                last_field,
+            } => Interval::try_from_literal(value, leading_field.as_ref(), last_field.as_ref())
+                .unwrap()
+                .into(),
+            AstLiteral::Null => "Null".to_string(),
+        }
```
panarch, line 37: 1. SQL keywords need to be displayed using uppercased letters.
2. No need of parsing interval, we can directly format using `AstLiteral::Interval`

e.g.
```rust
match self {                                                     
    AstLiteral::Boolean(b) => b.to_string().to_uppercase(),      
    AstLiteral::Number(n) => n.to_string(),                      
    AstLiteral::QuotedString(qs) => format!(r#""{qs}""#),        
    AstLiteral::HexString(hs) => format!(r#""{hs}""#),           
    AstLiteral::Interval {                                       
        value,                                                   
        leading_field,                                           
        last_field,                                              
    } => {                                                       
        let value = format!(r#"INTERVAL "{value}""#);            
        let leading = leading_field                              
            .as_ref()                                            
            .map(ToString::to_string)                            
            .unwrap_or("".to_owned());                           
                                                                 
        match last_field {                                       
            Some(last) => format!("{value} {leading} TO {last}"),
            None => format!("{value} {leading}"),                
        }                                                        
    }                                                            
    AstLiteral::Null => "NULL".to_string(),                      
}                                                                
```

Then, it would be good to have unit tests in `ast/ast_literal.rs`.
```rust
#[cfg(test)]                                                                  
mod tests {                                                                   
    use {                                                                     
        crate::ast::{AstLiteral, DateTimeField, ToSql},                       
        bigdecimal::BigDecimal,                                               
    };                                                                        
                                                                              
    #[test]                                                                   
    fn to_sql() {                                                             
        assert_eq!("TRUE", AstLiteral::Boolean(true).to_sql());               
        assert_eq!("123", AstLiteral::Number(BigDecimal::from(123)).to_sql());
        assert_eq!(                                                           
            r#""hello""#,                                                     
            AstLiteral::QuotedString("hello".to_owned()).to_sql()             
        );                                                                    
        assert_eq!(                                                           
            r#"INTERVAL "1-2" YEAR TO MONTH"#,                                
            AstLiteral::Interval {                                            
                value: "1-2".to_owned(),                                      
                leading_field: Some(DateTimeField::Year),                     
                last_field: Some(DateTimeField::Month),                       
            }                                                                 
            .to_sql()                                                         
        );                                                                    
        assert_eq!(                                                           
            r#"INTERVAL "10" HOUR"#,                                          
            AstLiteral::Interval {                                            
                value: "10".to_owned(),                                       
                leading_field: Some(DateTimeField::Hour),                     
                last_field: None,                                             
            }                                                                 
            .to_sql()                                                         
        );                                                                    
        assert_eq!("NULL", AstLiteral::Null.to_sql());                        
    }                                                                         
}                                                                             
```

panarch (changes requested):
Now, I've checked everything.
With these changes, we can merge this meaningful changes to the main.
Thanks really a lot!

core/src/ast/function.rs
```diff
@@ -115,8 +116,29 @@ pub enum Aggregate {
...
 }
 
+impl ToSql for Aggregate {
+    fn to_sql(&self) -> String {
+        match self {
+            Aggregate::Count(cae) => format!("Count({})", cae.to_sql()),
+            Aggregate::Sum(e) => format!("Sum({})", e.to_sql()),
+            Aggregate::Max(e) => format!("Max({})", e.to_sql()),
+            Aggregate::Min(e) => format!("Min({})", e.to_sql()),
+            Aggregate::Avg(e) => format!("Avg({})", e.to_sql()),
```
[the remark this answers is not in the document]
earney: will do. Thanks!

core/src/ast/expr.rs
```diff
@@ -61,3 +61,294 @@ pub enum Expr {
...
+                }
+            }
+            Expr::UnaryOp { op, expr } => format!("{}{}", op.to_sql(), expr.to_sql()),
+            Expr::Cast { expr, data_type } => {
+                format!("CAST({} AS {data_type})", expr.to_sql())
+            }
+            Expr::Extract { field, expr } => {
+                format!(r#"EXTRACT({field} FROM "{}")"#, expr.to_sql())
+            }
+            Expr::Nested(expr) => format!("todo:Nested({})", expr.to_sql()),
```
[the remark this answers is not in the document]
panarch: It looks that this is the only missed one. Now it's really close to the goal!

panarch (approved):
Beautiful.. I'd really appreciate this.
There exists lots of cases which `ToSql` can be used to provide much better info to db users.
e.g.
Better more detailed error message, ..

Previously, it was not possible to provide readable info about AST but now `ToSql` can directly hit the needs!

Thanks a lot again for this meaningful contribution.
All looks great, merging 👍 👍 👍

PR #556, opened 2022-05-16
Simplify repetitive `#[derive(Debug, Clone, PartialEq, ...)]` macros by adding attribute type proc-macro `#[basic_derives]`.

### LINKED ISSUE #555: Simplify repetitive `#[derive(Debug, Clone, PartialEq, ...)]`macros by adding proc-macro subcrate.
# Problem and Suggestion
image
I see repetitive use of `#derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]` across several `struct`s in `gluesql-core`. I came up with an idea to simplify this using proc macro.
```rust
// before
pub enum SetExpr {
    Select(Box<Select>),
    Values(Values),
}

// after
#[basic_derives]
pub enum SetExpr {
    Select(Box<Select>),
    Values(Values),
}
```
For few `struct` that needs extra `Display` trait in derive macro, we can do it like this.
```rust
// before
pub enum Function {
    #[strum(to_string = "ABS")]
    Abs(Expr),
    ...

// after
#[basic_derives(add = ["Display"])]
pub enum Function {
    #[strum(to_string = "ABS")]
    Abs(Expr),
    ...
```
With the same method we can also add `remove` attribute in that proc macro to prune out several traits.
Let me know if this idea is good or bad.

MRGRAVITY817:
# Resolves #555 
This pull request tries to solve the problem referred in #555.

## Expected changes
```rust
// before
use serde::{Serialize, Deserialize};

pub enum SetExpr {
    Select(Box<Select>),
    Values(Values),
}

// after
#[basic_derives]
pub enum SetExpr {
    Select(Box<Select>),
    Values(Values),
}
```

Important thing is that this new `#[basic_derives]` macro also auto-imports `Serialize, Deserialize` from `serde`.

# Work in progress ⏰


panarch:
The one point that I am concerning is about maintaining cost.
There might be no way not to publish `gluesql-core-macros` to crates.io

Is that possible to use subcrate without modifying root workspace?
https://github.com/gluesql/gluesql/blob/main/Cargo.toml

MRGRAVITY817:
> Is that possible to use subcrate without modifying root workspace? 

I am not quite sure .. I think it's possible not to publish on crates.io, but seems like it should be added in root Cargo.toml anyway. Maybe I am wrong, but anyway I can try and close this PR to find better one if this ain't possible :) Wish me luck!

|Should be added on root `Cargo.toml`|But can be hidden or not exported in crates.io|
|----|----|
| image | image|

MRGRAVITY817:
That sounds like a great hint. Thanks for the research!

MRGRAVITY817:
I searched Diesel's subcrate again and actually found that it has been actually published to crates.io. 
image
Seems like there's no fair clue to stop publishing the subcrates ... yet I think it ain't so bad to reveal the internally used macros. What do you think?

MRGRAVITY817:
I agree with that. 
Btw, what about using `[exclude]` section in root Cargo.toml?
image
According to explanation it seems like we can choose some directories to be excluded in publishing.
I haven't yet tried publishing monorepo crates like gluesql before, so I have no idea if this would work. 
Any thoughts about this?

MRGRAVITY817:
I concluded that cost of maintaining extra proc-macro crate isn't enough for it's use, so I am closing this PR.
Hope to seek other use cases for proc macros in near future! 🚀

PR #557, opened 2022-05-16
Update rusttoolchain version to 1.60 and sqlparser-rs to 0.17

earney:
upgrade rust to use version 1.60  (currently uses 1.59).
upgrade sqlparser-rs to 0.17 (currently uses 0.16)


panarch (approved):
Thanks a lot!
In this kind of case, you can split the PR into two.
Though each one has only a single line change, these are different changes so it would be better to be separated

PR #559, opened 2022-05-17
Fix the link in `gluesql-js/README.md` Webpack example

MRGRAVITY817:
The link seemed like pointing the rollup example, so I fixed it.


panarch (approved):
I wasn't aware of this at all!
Thanks a lot 👍

Issue #423, opened 2021-12-05
Update aggregate functions to support `Expr`


zmrdltl:
## Description
The [get_value](https://github.com/gluesql/gluesql/blob/main/core/src/executor/aggregate/state.rs#L241) of the existing `closure of function accumulate` determines the value after parsing the parameter to be entered into the function in the enum `Expr`.  It supports only some of the Expr syntax and returns a value, otherwise it returns an error.

Currently supported parameter
1. `Expr::Identifier`
2. `Expr::CompoundIdentifier`
3. `wildcard for function COUNT`
</br>

Therefore, it is necessary to change the aggregate function to support all expressions as a parameter

panarch:
e.g.
```sql
/* Supported */
SUM(id)
COUNT(*)
MAX(col),
MIN(ta.col2)

/* To be supported */
SUM(id + 1)
MAX(col + col2)
SUM(id * rate)
```

earney:
This issue is also discussed in issue #151

PR #561, opened 2022-05-18
Agg functions supports enum Expr (implements issue #423)

ghost:
This PR implements the ability for aggregate expr expressions such as :
SUM(id + 1)
MAX(col + col2)
SUM(id * rate) 

implements improvements stated in issue #423


panarch:
Awesome!
I saw that you consider how to deal with `Value::Null` in `SUM` aggregation.
That is already defined in the existing tests.

https://github.com/gluesql/gluesql/blob/69c9e8445255e3a24fddb3820a948337122a645f/test-suite/src/aggregate.rs#L48-L53

Someone who wants to ignore nulls then they can use `WHERE age IS NOT NULL`
```sql
SELECT SUM(age), MAX(age), MIN(age) FROM Item WHERE age IS NOT NULL
```

Only except for `COUNT(*)`, there's no exceptional handlers for aggregate functions.

earney:
I will continue this PR once the Show Indexes PR is merged.

core/src/executor/aggregate/state.rs
```diff
@@ -175,39 +177,87 @@ impl<'a> State<'a> {
...
+                    BinaryOperator::Minus => left_value.subtract(right_value),
+                    BinaryOperator::Multiply => left_value.multiply(right_value),
+                    BinaryOperator::Divide => left_value.divide(right_value),
+                    BinaryOperator::Modulo => left_value.modulo(right_value),
+                    _ => Err(AggregateError::UnsupportedAggregateFunction.into()),
+                }
+            }
+            _ => Err(AggregateError::UnsupportedAggregateFunction.into()),
+        }
+    }
```
panarch, line 245: Aggregate evaluation is not supposed to be done in here.
You may need to use
https://github.com/gluesql/gluesql/blob/e4c04c1135680b6199b965d732591029f46d808e/core/src/executor/evaluate/mod.rs#L27-L33

panarch: > so we will need to supply a storage variable,
yes I think so. Corresponding interface change would be required for `aggregate/state.rs`

earney:
Making the requested changes, I believe are a bit beyond by current skill level, so I'm gonna put this on pause.  If someone is interested in implementing this, please feel free.

PR #563, opened 2022-05-19
Add i128 (int128) data type

earney:
This PR adds int128 data type to gluesql.


core/src/data/value/error.rs
```diff
@@ -72,6 +72,16 @@ pub enum ValueError {
...
 
+    #[error("failure to convert {a:#?} to {b:#?} for value:{value:#?}")]
+    ConversionErrorFromDataTypeAToDataTypeB {
+        a: DataType,
+        b: DataType,
+        value: Value,
+    },
+
+    #[error("conflict - unique constraint cannot be used for {0} type")]
+    ConflictDataTypeOnUniqueConstraint(String),
```
panarch, line 83: This is not being used in this PR. you can remove this

core/src/data/value/into.rs
```diff
@@ -484,8 +548,64 @@ mod tests {
...
+        );
+        test!(
+            Value::Map(HashMap::new()),
+            Err(ValueError::ImpossibleCast.into())
+        );
+        test!(
+            Value::List(Vec::new()),
+            Err(ValueError::ImpossibleCast.into())
+        );
+        test!(Value::Null, Err(ValueError::ImpossibleCast.into()));
```
panarch, line 586: Thanks a lot for adding tests!
However, these does not match to the title `INT128` data type except
```rust
test!(Value::I128(122), Ok(122));
```

This improvements are sufficient to exists as an independent PR.
Could you split and make a PR for these?

core/src/data/value/into.rs
```diff
@@ -637,10 +759,4 @@ mod tests {
             Ok(I::Month(274))
         )
     }
-
-    #[test]
-    fn try_into_u128() {
-        let uuid = 195965723427462096757863453463987888808;
-        assert_eq!((&Value::Uuid(uuid)).try_into() as Result<u128>, Ok(uuid))
-    }
```
panarch, line 645: This codes removal looks not intentional.
It could happen when you add support for unsigned 128-bit int type.

core/src/data/value/error.rs
```diff
@@ -95,6 +105,9 @@ pub enum ValueError {
     #[error("literal cast failed to date: {0}")]
     LiteralCastToDateFailed(String),
 
+    #[error("literal cast from {1:#?} to ({0:#?} failed")]
+    LiteralCastToDataTypeFailed(DataType, String),
```
panarch, line 109: 1. You can use
`#[error("literal cast from {1} to ({0} failed")]`
`{:#?}` does not use `Display` so it would return verbose result.

2. Applying this only for INT128 first
The decision to add `LiteralCastToDataTypeFailed` is really nice, thanks.
However, it's perfect to start with applying to int128 but it would be better to make a separate PR for applying this to existing data types (e.g. INT8)

core/src/data/value/into.rs
earney: Yes, I removed this when I removed the u128 parts..  I'll make sure I put it back in.. Thanks for catching this!

core/src/data/value/error.rs
earney: ok.  sounds good.  I will try to make sure we aren't using LIteralCastToDatatypeFailed for anything but Int128 at this time.

core/src/data/value/error.rs
earney: good catch..   I will remove it.

earney:
This PR has problems with the following test:

test-suite/src/data_type/int128.rs: (~ line 55)

 SELECT field_one FROM Item WHERE field_one > 0
thread 'int128' panicked at 'assertion failed: `(left == right)`
  left: `2`,
 right: `0`: 
[err: number of rows]
expected: [[I128(1)], [I128(3)]]
   found: []', test-suite/src/tester.rs:52:5

For some reason the comparison (or equal) is not working with correctly.  What am I overlooking?

core/src/data/interval/mod.rs
```diff
@@ -287,7 +287,7 @@ impl Interval {
                 format!("{:?}", to),
             )
             .into()),
-            (None, _) => Err(IntervalError::Unreachable.into()),
+            (_, _) => Err(IntervalError::UnsupportedExtractPart.into()),
```
panarch, line 290: This is also not related to this issue.
We can remove this `UnsupportedExtractPart` error.

core/src/data/value/literal.rs
```diff
@@ -166,12 +182,12 @@ impl Value {
...
-                .map(|v| Value::I64(v.trunc() as i64))
-                .ok_or_else(|| {
-                    ValueError::UnreachableLiteralCastFromNumberToInteger(v.to_string()).into()
-                }),
+            (DataType::Int, Literal::Number(v)) => match v.to_i64() {
+                Some(x) => Ok(Value::I64(x)),
+                None => {
+                    Err(ValueError::UnreachableLiteralCastFromNumberToInteger(v.to_string()).into())
+                }
+            },
```
panarch, line 190: This change also does not look related to this issue.
We can take this in the different issue or PR.

core/src/data/value/literal.rs
```diff
@@ -372,6 +405,7 @@ mod tests {
             rust_decimal::Decimal,
             std::{borrow::Cow, str::FromStr},
         };
+        //use crate::result::Error::Value;
```
panarch, line 408: Unnecessary comment, let's remove this.

core/src/data/value/literal.rs
```diff
@@ -19,7 +20,13 @@ impl PartialEq<Literal<'_>> for Value {
...
             (Value::Bool(l), Literal::Boolean(r)) => l == r,
             (Value::I8(l), Literal::Number(r)) => r.to_i8().map(|r| *l == r).unwrap_or(false),
-            (Value::I64(l), Literal::Number(r)) => r.to_i64().map(|r| *l == r).unwrap_or(false),
+            (Value::I64(l), Literal::Number(r)) => match r.to_string().contains('.') {
+                true => match l.to_f64() {
+                    Some(x) => r.to_f64().map(|r| x == r).unwrap_or(false),
+                    None => false,
+                },
+                false => r.to_i64().map(|r| *l == r).unwrap_or(false),
+            },
```
panarch, line 29: These changes look also not for this PR.

core/src/data/value/binary_op/i8.rs
```diff
@@ -52,7 +54,29 @@ impl TryBinaryOperator for i8 {
...
+                .checked_add(rhs)
+                .ok_or_else(|| {
+                    ValueError::BinaryOperationOverflow {
+                        lhs: I8(lhs),
+                        rhs: I64(rhs),
+                        operator: NumericBinaryOperator::Add,
+                    }
+                    .into()
+                })
+                .map(I64),
```
panarch, line 67: This PR is adding `INT(128)` type, so codes changes which affect existing behavior needs to be done in the different PR.

panarch (changes requested):
It's so pleasure to see huge improvements.
To proceed the review, could you remove unrelated codes changes in here?
This PR is for adding `INT(128)` type but there looks some portion of codes are not related to this.

core/src/data/value/literal.rs
earney: Yes, I found some error while doing some testing and added this.  I'll look into creating an issue/PR about this.

core/src/data/value/binary_op/i8.rs
earney: Ok.  I will post another PR for changing math to use checked_add etc.  This will allow us to catch overflows and underflows.  Ive added some tests to check for overflows and i will add them to the new PR as well.

earney:
This PR is put on hold until PR #571 is resolved.  I will the use the code base from that PR to add the I128 data type.

core/src/data/value/binary_op/i128.rs
```diff
@@ -0,0 +1,660 @@
...
+                .ok_or_else(|| {
+                    ValueError::BinaryOperationOverflow {
+                        lhs: I128(lhs),
+                        rhs: I128(rhs),
+                        operator: NumericBinaryOperator::Subtract,
+                    }
+                    .into()
+                })
+                .map(I128),
+            F64(rhs) => Ok(F64(lhs as f64 - rhs)), // should this be f128??
```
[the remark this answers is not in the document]
ever0de: Same for other calculations!

core/src/data/value/binary_op/i128.rs
```diff
@@ -0,0 +1,660 @@
...
+        assert_eq!(I128(1), I128(1));
+        assert_eq!(I128(1), F64(1.0));
+
+        assert_eq!(-1i128, I128(-1));
+        assert_eq!(0i128, I128(0));
+        assert_eq!(1i128, I128(1));
+        assert_eq!(type_min, I128(type_min));
+        assert_eq!(type_max, I128(type_max));
+
+        //try_add
```
ever0de, line 313: I like to separate it into a test function.

core/src/data/value/binary_op/i128.rs
```diff
@@ -0,0 +1,660 @@
...
+        let type_min: i128 = i128::MIN;
+
+        assert_eq!(I128(1), I8(1));
+        assert_eq!(I128(1), I64(1));
+        assert_eq!(I128(1), I128(1));
+        assert_eq!(I128(1), F64(1.0));
+
+        assert_eq!(-1i128, I128(-1));
+        assert_eq!(0i128, I128(0));
+        assert_eq!(1i128, I128(1));
```
ever0de, line 309:
```suggestion
        assert_eq!(-1_i128, I128(-1));
        assert_eq!(0_i128, I128(0));
        assert_eq!(1_i128, I128(1));
```
It looks pretty.

core/src/data/value/binary_op/i128.rs
```diff
@@ -0,0 +1,660 @@
...
+            }
+            .into())
+        );
+    }
+
+    #[test]
+    fn try_multiply() {
+        let base = 3_i128;
+
+        // 3 * 2 = 6
```
ever0de, line 567:
```suggestion
```

Remove comments?

core/src/data/value/binary_op/i128.rs
```diff
@@ -0,0 +1,660 @@
...
+    #[test]
+    fn try_divide() {
+        let base = 6_i128;
+
+        // 6/2 = 3
+        assert_eq!(base.try_divide(&I8(2)), Ok(I8(3)));
+        assert_eq!(base.try_divide(&I64(2)), Ok(I64(3)));
+        assert_eq!(base.try_divide(&I128(2)), Ok(I128(3)));
+
+        // 6/-6 = -1
```
ever0de, line 606:
```suggestion
        assert_eq!(base.try_divide(&I8(2)), Ok(I8(3)));
        assert_eq!(base.try_divide(&I64(2)), Ok(I64(3)));
        assert_eq!(base.try_divide(&I128(2)), Ok(I128(3)));


```

Remove comments?

core/src/data/value/binary_op/i128.rs
```diff
@@ -0,0 +1,660 @@
...
+        assert_eq!(base.try_divide(&I8(-6)), Ok(I8(-1)));
+        assert_eq!(base.try_divide(&I64(-6)), Ok(I64(-1)));
+        assert_eq!(base.try_divide(&I128(-6)), Ok(I128(-1)));
+
+        assert!(
+            matches!(base.try_divide(&F64(1.0)), Ok(F64(x)) if (x - 6.0).abs() < f64::EPSILON )
+        );
+
+        let _decimal_result = Decimal::from(base);
```
ever0de, line 615: Variables prefixed with `_` in Rust usually indicate "unused" .
<https://stackoverflow.com/questions/48361537/why-do-underscore-prefixed-variables-exist>

Additionally, if used in the same way as `let _`, it will be dropped on the next line.
playground: <https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=6aac0bfc6812eabdfe2e1f2ca0eb29e6>

core/src/data/value/json.rs
```diff
@@ -123,6 +124,9 @@ mod tests {
             Value::I64(100).try_into(),
             Ok(JsonValue::Number(100.into()))
         );
+        // this doesn't work
+        // assert_eq!(Value::I128(i128::MAX).try_into(), Ok(JsonValue::Number(i128::MAX.into())));
```
ever0de, line 128: <https://docs.rs/serde_json/latest/src/serde_json/number.rs.html#28>

Maybe it's because of the size of the number type?

ever0de (commented):
Sorry for leaving a review late.

core/src/data/value/binary_op/i128.rs
```diff
@@ -0,0 +1,660 @@
...
+    Value::*,
+};
+
+impl PartialEq<Value> for i128 {
+    fn eq(&self, other: &Value) -> bool {
+        match other {
+            I8(other) => self == &(*other as i128),
+            I64(other) => self == &(*other as i128),
+            I128(other) => self == other,
+            F64(other) => (*self as f64) == *other, // is this right?
```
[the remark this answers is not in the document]
earney: I will create another PR for this.

core/src/data/value/binary_op/i128.rs
```diff
@@ -0,0 +1,660 @@
...
+    }
+}
+
+impl PartialOrd<Value> for i128 {
+    fn partial_cmp(&self, other: &Value) -> Option<Ordering> {
+        match other {
+            I8(other) => PartialOrd::partial_cmp(self, &(*other as i128)),
+            I64(other) => PartialOrd::partial_cmp(self, &(*other as i128)),
+            I128(other) => PartialOrd::partial_cmp(self, other),
+            F64(other) => PartialOrd::partial_cmp(&(*self as f64), other),
```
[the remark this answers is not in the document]
earney: ok.

core/src/data/value/binary_op/i128.rs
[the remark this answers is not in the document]
earney: I'll leave it as is for now.

core/src/data/value/json.rs
earney: If it is okay,  I'll leave it in commented out, since it generates an error.  Maybe someone else can figure it out?

core/src/data/value/binary_op/i128.rs
earney: bad habit from my python days...   Thank you..

core/src/data/value/binary_op/i64.rs
```diff
@@ -295,6 +352,8 @@ mod tests {
         let type_max: i64 = i64::MAX;
         let type_min: i64 = i64::MIN;
         let type_maxi64: i64 = type_max;
+        let type_maxi128: i128 = type_max.into();
+        let type_mini128: i128 = type_min.into();
```
ever0de, line 356: Since this is a typecast from `i64::MAX` , it looks different from the intended `max_i128` value.

core/src/data/value/binary_op/i8.rs
```diff
@@ -284,14 +341,15 @@ mod tests {
         let type_min: i8 = i8::MIN;
         let type_maxi64: i64 = type_max.into();
         let type_mini64: i64 = type_min.into();
+        let type_maxi128: i128 = type_max.into();
+        let type_mini128: i128 = type_min.into();
```
ever0de, line 345: maybe this too

core/src/data/value/json.rs
```diff
@@ -38,6 +38,7 @@ impl TryFrom<Value> for JsonValue {
             Value::Bool(v) => Ok(JsonValue::Bool(v)),
             Value::I8(v) => Ok(v.into()),
             Value::I64(v) => Ok(v.into()),
+            Value::I128(v) => Ok(v.to_string().into()),
```
ever0de, line 41: If this happens, why not try to convert the `JsonValue` to a `String`?
Maybe we want a type like `BigInt` or a truncated `Number` type...?
What do @panarch think?

core/src/data/value/literal.rs
```diff
@@ -85,6 +89,7 @@ impl TryFrom<&Literal<'_>> for Value {
                 .map(Value::I64)
                 .or_else(|| v.to_f64().map(Value::F64))
                 .ok_or_else(|| ValueError::FailedToParseNumber.into()),
+
```
ever0de, line 92:
```suggestion
```

remove unnecessary blank

core/src/data/value/literal.rs
```diff
@@ -174,7 +183,6 @@ impl Value {
                 }),
             (DataType::Int, Literal::Boolean(v)) => {
                 let v = if *v { 1 } else { 0 };
-
```
ever0de, line 177: Is there any reason to change these?

core/src/data/value/mod.rs
```diff
@@ -366,19 +388,20 @@ impl Value {
...
+                .try_fold(1i128, |mul, x| mul.checked_mul(x))
                 .ok_or_else(|| ValueError::FactorialOverflow.into())
         };
 
         match self {
-            I8(a) => factorial_function(*a as i64).map(I64),
-            I64(a) => factorial_function(*a).map(I64),
+            I8(a) => factorial_function(*a as i128).map(I128),
+            I64(a) => factorial_function(*a as i128).map(I128),
+            I128(a) => factorial_function(*a as i128).map(I128),
```
ever0de, line 404:
```suggestion
            I128(a) => factorial_function(*a).map(I128),
```

core/src/data/value/binary_op/i64.rs
ever0de: `Into` is a trait for transforms. Now, since we have converted the i64 variable to i128, the value I inferred from the variable name is the same as `i128::MAX` or `MIN`. How about like this?

earney: Do you mean something like maxI64intoi128 as a variable name?

panarch: How about.. not to use variables? explicit use of `i64::MAX` or `(i64::MAX as i128)` would be better in these cases.

```rust
let type_max: i64 = i64::MAX;
// i64:MAX

let type_min: i64 = i64::MIN;
// i64::MIN

let type_maxi64: i64 = type_max;
// i64::MAX

let type_maxi128: i128 = type_max.into();
// (i64::MAX as i128) or i64::MAX.into()

let type_mini128: i128 = type_min.into();
// (i64::MIN as i128) or i64::MIN.into()
```

core/src/data/value/binary_op/i8.rs
panarch: Like the above ones, it would be also just good not to use variables, explicit use of `i8::MIN` would be straightforward enough.

test-suite/src/data_type/int128.rs
```diff
@@ -0,0 +1,110 @@
...
+    run!("INSERT INTO Item VALUES (1, -1), (-2, 2), (3, 3), (-4, -4);");
+
+    let parse_i128 = |text: &str| -> i128 { text.parse().unwrap() };
+
+    let max_str = "170141183460469231731687303715884105728";
+    let min_str = "-170141183460469231731687303715884105729";
+
+    let str = format!("INSERT INTO Item VALUES ({:}, {:})", max_str, max_str);
+
+    test!(Err(ValueError::FailedToParseNumber.into()), &str);
```
panarch, line 22: Though `str` keyword is allowed for variable names, but it overlaps with Rust primitive type `str`.
It would be better to use different names.

In here,
Simply we may not need to use declare variables.
```rust
test!(
    Err(ValueError::FailedToParseNumber.into()),
    &format!("INSERT INTO Item VALUES ({}, {})", i128::MAX, i128::MAX)
);
```
The above would be enough.

However...
It looks that there exists some issues parsing i128 literal
In this test below, it is supposed not to generate error but `FailedToParseNumber` occurs.
```rust
test!(
    Err(ValueError::FailedToParseNumber.into()),
    &format!("INSERT INTO Item VALUES ({}, {})", i128::MAX - 100000, i128::MAX + 100000)
);
```

panarch (changes requested):

core/src/data/value/json.rs
earney: You might want to check what I did with this.  Is this what you are after?

core/src/data/value/json.rs
earney: I get the following when I try the assert_eq you gave:
left: `Ok(Number(1.7014118346046923e38))`,
 right: `Ok(String("170141183460469231731687303715884105727"))`', core\src\data\value\json.rs:135:9

core/src/data/bigdecimal_ext.rs
```diff
@@ -11,13 +12,18 @@ impl BigDecimalExt for BigDecimal {
...
             true => bigdecimal::ToPrimitive::to_i64(self),
             false => None,
         }
     }
+    fn to_i128(&self) -> Option<i128> {
+        match self.is_integer() {
+            true => bigdecimal::ToPrimitive::to_i128(self),
+            false => None,
+        }
+    }
```
panarch, line 26: This comment is only for sharing information, you don't need to do action now at least in this PR.
Current implementation code looks good enough.

fyi.
I tested using this code below and it is expected to pass but it failed with `ValueError::FailedToParseNumber`.
```rust
test!(  
    Ok(Payload::Insert(1)),
    &format!("INSERT INTO Item VALUES ({}, {})", i128::MAX, i128::MAX - 10)
);
```

And the reason was because of `bigdecimal::ToPrimitive::to_i128(self)`
https://docs.rs/bigdecimal/latest/bigdecimal/trait.ToPrimitive.html#method.to_i128
https://docs.rs/num-traits/0.2.15/src/num_traits/cast.rs.html#62-64
```rust
fn to_i128(&self) -> Option<i128> {
    self.to_i64().map(From::from)
}
```
That... is wrong implementation and
> The default implementation converts through to_i64(). Types implementing this trait should override this method if they can represent a greater range.

They also mentioned... quite... not promising.
We can consider to replace `to_i128` to our version (not to use `bigdecimal`'s) later.

test-suite/src/data_type/int128.rs
```diff
@@ -0,0 +1,106 @@
...
+        field_one INT(128),
+        field_two INT(128),
+    );"
+    );
+    run!("INSERT INTO Item VALUES (1, -1), (-2, 2), (3, 3), (-4, -4);");
+
+    let parse_i128 = |text: &str| -> i128 { text.parse().unwrap() };
+
+    let max_str = "170141183460469231731687303715884105728";
+    let min_str = "-170141183460469231731687303715884105729";
```
panarch, line 18: I saw that the comment which describes these values are located in the below.
It would be better for attaching comments to here.
It would be also good to use `const`
```rust
// i128::MAX + 1
const MAX_STR: &str = "170141183460469231731687303715884105728";

// i128::MIN - 1
const MIN_STR: &str = "-170141183460469231731687303715884105729";
```

test-suite/src/data_type/int128.rs
```diff
@@ -0,0 +1,106 @@
...
+        Ok(select!(
+            field_one        | field_two
+            I128               |    I128;
+            1                   parse_i128("-1");
+            parse_i128("-2")        2;
+            3                      3;
+            parse_i128("-4")      parse_i128("-4")
+        )),
+        "SELECT field_one, field_two FROM Item"
+    );
```
panarch, line 48: This is very nit request.
Could you make this looks like table shape?
e.g.
```rust
test!(
    Ok(select!(
        field_one        | field_two
        I128             | I128;
        1                  parse_i128("-1");
        parse_i128("-2")   2;
        3                  3;
        parse_i128("-4")   parse_i128("-4")
    )),
    "SELECT field_one, field_two FROM Item"
);
```

btw. thanks to use `parse_i128`.
`select!` macro needs to be upgrade to handle `-10` like negative numbers but.. current it does not.

panarch (changes requested):
Sorry for the late review, now I'm back.

There are some very nit requests about test codes.
With this change, let's merge this to the main.

Everything else looks great.
I appreciate the careful checks you've made on adding this new data type.
Numeric types especially interact with other types quite a lot. That's really pleasure to see the changes that you've carefully handled all the corresponding operations.
Thanks a lot!

core/src/data/bigdecimal_ext.rs
earney: ok.  Thanks for looking into this, and documenting the issue.

PR #565, opened 2022-05-21
Replace the `to_string` macro with the `serialize_all` macro

ever0de:
If `to_string` is used, a human error may occur because the enum name and the name to be written in to_string exist separately.

How about improving it using `serialize_all`?


panarch (approved):
Thanks a lot!
We can apply this to all `strum_macros` use cases, nice 👍 
`#[strum(serialize_all = "SCREAMING_SNAKE_CASE")]`

PR #567, opened 2022-05-21
Benchmarks part 2

earney:
I've take the previous benchmark PR and updated it to use the current main as well as replace the eval closures with function calls.  I'm not sure I am noticing a huge increase in performance.   Do you guys see a difference?


earney:
Here's what I get when I run perf on the whole benchmark:
image

PR #569, opened 2022-05-22
implements IFNull function

earney:
This PR implements mysql's IfNull function:
https://dev.mysql.com/doc/refman/5.7/en/flow-control-functions.html#function_ifnull

### DIFF
--- a/core/src/ast/function.rs
+++ b/core/src/ast/function.rs
@@ -38,6 +38,8 @@ pub enum Function {
     Ceil(Expr),
     #[strum(to_string = "CONCAT")]
     Concat(Vec<Expr>),
+    #[strum(to_string = "IFNULL")]
+    IfNull { expr: Expr, expr2: Expr },
     #[strum(to_string = "ROUND")]
     Round(Expr),
     #[strum(to_string = "FLOOR")]
--- a/core/src/executor/evaluate/function.rs
+++ b/core/src/executor/evaluate/function.rs
@@ -236,6 +236,28 @@ pub fn abs(name: String, n: Evaluated<'_>) -> Result<Value> {
     }
 }
 
+pub fn ifnull(expr1: Evaluated<'_>, expr2: Evaluated<'_>) -> Result<Value> {
+    // should we pass expr2 as an Expr enum and evaluate it if expr1 is null?
+    // that would be more efficient than evaluating it before hand..
+    // what do you think?
+    match expr1.try_into()? {
+        Value::Bool(b) => Ok(Value::Bool(b)),
+        Value::Date(d) => Ok(Value::Date(d)),
+        Value::I8(n) => Ok(Value::I8(n)),
+        Value::I64(v) => Ok(Value::I64(v.abs())),
+        Value::Decimal(v) => Ok(Value::Decimal(v.abs())),
+        Value::F64(v) => Ok(Value::F64(v.abs())),
+        Value::Interval(i) => Ok(Value::Interval(i)),
+        Value::List(l) => Ok(Value::List(l)),
+        Value::Map(m) => Ok(Value::Map(m)),
+        Value::Str(s) => Ok(Value::Str(s)),
+        Value::Timestamp(t) => Ok(Value::Timestamp(t)),
+        Value::Time(t) => Ok(Value::Time(t)),
+        Value::Uuid(u) => Ok(Value::Uuid(u)),
+        Value::Null => expr2.try_into(),
+    }
+}
+
 pub fn sign(name: String, n: Evaluated<'_>) -> Result<Value> {
     let x = eval_to_float!(name, n);
     if x == 0.0 {
--- a/test-suite/src/aggregate.rs
+++ b/test-suite/src/aggregate.rs
@@ -75,6 +75,10 @@ test_case!(aggregate, async move {
     }
 
     let error_cases = vec![
+        (   // once PR #561 is merged this should not produce an error anymore
+            AggregateError::OnlyIdentifierAllowed.into(),
+           "SELECT SUM(ifnull(age, 0)) from Item;",
+        ),
         (
             AggregateError::UnsupportedCompoundIdentifier(expr!("id.name.ok")).into(),
             "SELECT SUM(id.name.ok) FROM Item;",
--- a/test-suite/src/function/ifnull.rs
+++ b/test-suite/src/function/ifnull.rs
@@ -1,0 +1,53 @@
+use crate::*;
+
+test_case!(ifnull, async move {
+    use chrono::NaiveDate;
+    use gluesql_core::{
+        //executor::EvaluateError,
+        executor::Payload,
+        prelude::Value::*,
+        //translate::TranslateError,
+    };
+    use rust_decimal::Decimal;
+    let test_cases = vec![
+        (
+            "CREATE TABLE SingleItem (id integer null, int8 int(8) null, dec decimal null, dt date null, mystring Text null)",
+            Payload::Create,
+        ),
+        (
+            r#"INSERT INTO SingleItem VALUES (0, 1, 2, "2022-05-23", "this is a string")"#,
+            Payload::Insert(1),
+        ),
+        (
+            r#"INSERT INTO SingleItem VALUES (null, null, null, null, null)"#,
+            Payload::Insert(1),
+        ),
+        (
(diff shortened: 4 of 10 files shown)


core/src/executor/evaluate/function.rs
```diff
@@ -236,6 +236,28 @@ pub fn abs(name: String, n: Evaluated<'_>) -> Result<Value> {
...
+        Value::F64(v) => Ok(Value::F64(v.abs())),
+        Value::Interval(i) => Ok(Value::Interval(i)),
+        Value::List(l) => Ok(Value::List(l)),
+        Value::Map(m) => Ok(Value::Map(m)),
+        Value::Str(s) => Ok(Value::Str(s)),
+        Value::Timestamp(t) => Ok(Value::Timestamp(t)),
+        Value::Time(t) => Ok(Value::Time(t)),
+        Value::Uuid(u) => Ok(Value::Uuid(u)),
+        Value::Null => expr2.try_into(),
+    }
```
panarch, line 258: `Evaluated` has `is_null` method.
We can use `is_null` in here.
https://github.com/gluesql/gluesql/blob/c716a72f31f6bc94d2f019c0c4e2e136c2d3bbdb/core/src/executor/evaluate/evaluated.rs#L225-L230

core/src/ast/function.rs
```diff
@@ -33,6 +33,10 @@ pub enum Function {
     },
     Ceil(Expr),
     Concat(Vec<Expr>),
+    IfNull {
+        expr: Expr,
+        expr2: Expr,
+    },
```
panarch, line 39: How about using `then` than `expr2`?
Then... if `expr` is null, then we use `then`.
```rust
IfNull {
  expr: Expr,
  then: Expr,
}
```

ref.
https://github.com/gluesql/gluesql/blob/c716a72f31f6bc94d2f019c0c4e2e136c2d3bbdb/core/src/ast/expr.rs#L58-L62

earney: in plan/expr/function.rs:  Exprs::Double([expr, expr2].into_iter()), wants expr and expr2.   We could create a separate category that has expr, then, etc.  But it probably isn't worth the effort.  Is there any other way to use then?

core/src/executor/evaluate/function.rs
earney: ok.

core/src/ast/function.rs
earney: ok.  should we delay the eval until we need the value in the variable then or should we go ahead and evaluate then before?

core/src/executor/evaluate/function.rs
```diff
@@ -236,6 +236,20 @@ pub fn abs(name: String, n: Evaluated<'_>) -> Result<Value> {
...
+    // what do you think?
+
+    Ok(match expr1 {
+        Evaluated::Value(v) => match v.is_null() {
+            true => then.try_into()?,
+            false => v.into_owned(),
+        },
+        Evaluated::Literal(l) => l.try_into()?,
+    })
+}
```
panarch, line 251: `Evaluated::Literal(l)` can be also `NULL`.
You can consider to use `expr.is_null()` to handle this.

test-suite/src/aggregate.rs
```diff
@@ -75,6 +75,11 @@ test_case!(aggregate, async move {
     }
 
     let error_cases = vec![
+        (
+            // once PR #561 is merged this should not produce an error anymore
+            AggregateError::OnlyIdentifierAllowed.into(),
+            "SELECT SUM(ifnull(age, 0)) from Item;",
+        ),
```
panarch, line 82: It is expected to merge this PR earlier than #561 
How about adding this test to #561 rather than here?

test-suite/src/function/ifnull.rs
```diff
@@ -0,0 +1,74 @@
...
+                          "01:02:03", "1970-01-01 00:00:00 -00:00")"#,
+            Payload::Insert(1),
+        ),
+        (
+            r#"INSERT INTO SingleItem VALUES (null, null, null, null, null, null, null, null, null)"#,
+            Payload::Insert(1),
+        ),
+        (
+            r#"SELECT ifnull(id, 1) as ID, ifnull(int8, 2) as int8, ifnull(dec, 3) 
+            FROM SingleItem where id is not null"#,
```
panarch, line 26: One also important role of integration tests is providing usage guide to users.
In this kind of SQL examples, both using all-uppercased or all-lowercased for SQL keywords look ok.
However, mixing uppercase and lowercase letters would be confusing for users.
```
SELECT IFNULL(id, 1) AS ID, IFNULL(INT8, 2) AS int8, IFNULL(dec, 3)
FROM SingleItem WHERE id IS NOT NULL
```
or
```
select ifnull(id, 1) as ID, ifnull(int8, 2) as int8, ifnull(dec, 3)
from SingleItem where id is not null
```

core/src/executor/evaluate/function.rs
```diff
@@ -236,6 +236,20 @@ pub fn abs(name: String, n: Evaluated<'_>) -> Result<Value> {
     }
 }
 
+pub fn ifnull(expr1: Evaluated<'_>, then: Evaluated<'_>) -> Result<Value> {
+    // should we pass expr2 as an Expr enum and evaluate it if expr1 is null?
+    // that would be more efficient than evaluating it before hand..
+    // what do you think?
```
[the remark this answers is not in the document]
earney: okay.. should we leave the comment in to remind us in the future, or just remove it for now?

test-suite/src/function/ifnull.rs
```diff
@@ -0,0 +1,74 @@
...
+            select!("ID" | "int8" | "ifnull(dec, 3)"; I64 | I8 | Decimal; 0 1 Decimal::from(2i8)),
+        ),
+        (
+            // notice that this example returns I64 I64 I64, where the previous example returned I64 I8 Decimal.
+            // is this behavior desirable?  see https://dev.mysql.com/doc/refman/5.7/en/flow-control-functions.html#function_ifnull
+            r#"SELECT ifnull(id, 1) as ID, ifnull(int8, 2) as int8, ifnull(dec, 3)
```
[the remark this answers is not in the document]
earney: OK..  sounds good to me,   since IFNULL can return some other data type such as IFNULL(myint, "my string")  then it makes sense to leave it this way..

test-suite/src/function/ifnull.rs
earney: yes, I have a bad habit of typing it in lower case..

core/src/executor/evaluate/function.rs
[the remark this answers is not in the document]
panarch: Except for doc comments which are for db users, I'm for having less comments as possible.
Code itself needs to be readable without depending on extra comments.
Maintaining comments are fully done by human, not compiler so it can be easily outdated.

I think we can deal with this issue in simple.
Both two options which we discussed are acceptable.
If this issues come again later, then we can think about this again.
If it doesn't, then we don't need to put effort on this.

core/src/executor/evaluate/function.rs
```diff
@@ -236,6 +236,30 @@ pub fn abs(name: String, n: Evaluated<'_>) -> Result<Value> {
     }
 }
 
+pub fn ifnull(expr1: Evaluated<'_>, then: Evaluated<'_>) -> Result<Value> {
```
panarch, line 239: Now there's no `expr2`, it would be good to use `expr` rather than `expr1`

test-suite/src/aggregate.rs
panarch: Ok, let's make this discussion simple.
Adding this test to this is acceptable, but the comment does not look necessary.
> // once PR #561 is merged this should not produce an error anymore

Let's leave this test case but removing the comment.
In #561 you will not be able to miss this comment because the integration test will fail if this PR is merged earlier than #561

core/src/executor/evaluate/function.rs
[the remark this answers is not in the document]
earney: okay..   sounds good to me.
Sometimes I add comments (that are temporary), just to get advice or info from reviewers and I plan on deleting them before the code is merged into main.

test-suite/src/aggregate.rs
earney: sounds good to me! ;)

PR #571, opened 2022-05-24
use checked_x functions for i8, i64 and decimal data types so that we don't panic during an overflow

earney:
We use the following functions to add , subtract, multiply, divide and modulo:
checked_add
checked_sub
checked_mul
checked_div
checked_rem
for i8, i64, and decimal data types. 
I could not find something similar for F64, so I let F64 math operations as is.


core/src/data/value/error.rs
```diff
@@ -72,6 +72,13 @@ pub enum ValueError {
     #[error("unreachable failure on parsing number")]
     UnreachableNumberParsing,
 
+    #[error("failure to convert {a:#?} to {b:#?} for value:{value:#?}")]
+    ConversionErrorFromDataTypeAToDataTypeB {
+        a: DataType,
+        b: DataType,
+        value: Value,
+    },
```
panarch, line 80: This looks for the other task. Current PR does not use `ConversionErrorFromDataTypeAToDataTypeB`

core/src/data/value/error.rs
```diff
@@ -95,6 +102,9 @@ pub enum ValueError {
     #[error("literal cast failed to date: {0}")]
     LiteralCastToDateFailed(String),
 
+    #[error("literal cast from {1} to {0} failed")]
+    LiteralCastToDataTypeFailed(DataType, String),
```
panarch, line 106: This, too.
No use case in this PR. `LiteralCastToDataTypeFailed`

core/src/data/value/mod.rs
```diff
@@ -902,4 +962,27 @@ mod tests {
             .into()),
         );
     }
+
+    #[test]
+    fn test_unary_minus() {
```
panarch, line 967: unit tests already include the meaning of `test`. so, how about to use simply.. `unary_minus()` for this test name?

core/src/data/value/mod.rs
```diff
@@ -902,4 +962,27 @@ mod tests {
...
+        );
+
+        assert_eq!(
+            Str("abc".to_string()).unary_minus(),
+            Err(ValueError::UnaryMinusOnNonNumeric.into())
+        );
+    }
+
+    #[test]
+    fn test_factorial() {
```
panarch, line 984: > unit tests already include the meaning of test. so, how about to use simply.. unary_minus() for this test name?

This, too.
It would be enough with
```rust
#[test]
fn factorial() {
 . .
```

panarch (changes requested):
It's really such a pleasure to see this improvements.
I thought several times about this.. it would be nice to do this someday but didn't do in real action. Now, huge panicable codes have removed in the right way! You've made this.
Thanks a lot 👍 👍 👍 

I would like to request some very tiny change requests, left a few comments.

core/src/data/value/error.rs
earney: Thanks for catching this..

core/src/data/value/error.rs
earney: Same here..

PR #572, opened 2022-05-24
test (and fix) Divide by zero

earney:
It appears we do not have a divide by zero test, so I added one and found a bug so here is the PR to fix it.

### DIFF
--- a/core/src/data/value/mod.rs
+++ b/core/src/data/value/mod.rs
@@ -290,7 +290,7 @@ impl Value {
     pub fn divide(&self, other: &Value) -> Result<Value> {
         use Value::*;
 
-        if self.is_zero() {
+        if other.is_zero() {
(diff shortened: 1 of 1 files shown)


core/src/data/value/mod.rs
```diff
@@ -432,6 +432,7 @@ mod tests {
     use {
         super::{Interval, Value::*},
         crate::data::value::uuid::parse_uuid,
+        crate::data::ValueError,
```
panarch, line 435: This could be merge by
```rust
crate::data::{
    value::uuid::parse_uuid,
    ValueError,
};
```

This is a very nit comment and also it is in the unit tests, so I will approve and merge this.
You can consider this for the next time.

panarch (approved):
That was a huge miss, thanks really a lot!
Adding supplementary unit tests look also perfect.

PR #573, opened 2022-05-25
Fix primitive numeric types cast not to panic

earney:
Modify TryInto functions so that tests like the ones below do not panic but throw an ImpossibleCast Error.

```
test!(Value::I64(128), Err(ValueError::ImpossibleCast.into()));
test!(Value::F64(128.0), Err(ValueError::ImpossibleCast.into()));
```


earney:
I realize that an any i8 will cast to an i64 , but figured it was best to use to_i64(), instead of as i64.  the compiler should be smart enough to realize this and can optimize the to_i64 away if it needs to.

core/src/data/value/into.rs
```diff
@@ -534,6 +545,7 @@ mod tests {
         test!(Value::Bool(true), Ok(1.0));
         test!(Value::Bool(false), Ok(0.0));
         test!(Value::I8(122), Ok(122.0));
+        test!(Value::I64(122), Ok(122.0));
```
ever0de, line 548:
```rust
        test!(Value::I64(122), Ok(122.0));
        test!(Value::I64(1234567890), Ok(1234567890.0));
```

If it is the same case, it seems that there is no need to add it.

core/src/data/value/into.rs
```diff
@@ -417,6 +417,10 @@ mod tests {
             Err(ValueError::ImpossibleCast.into())
         );
         test!(Value::Null, Err(ValueError::ImpossibleCast.into()));
+
+        //impossible casts to bool
```
ever0de, line 421: <https://doc.rust-lang.org/book/ch03-04-comments.html>
Similarly, in `GlueSQL` it is common to use a leading space like `// TEMP`.

ever0de, line 421: <https://doc.rust-lang.org/book/ch03-04-comments.html>
Similarly, in `GlueSQL` it is common to use a leading space like `// TEMP`.

core/src/data/value/into.rs
```diff
@@ -436,6 +440,7 @@ mod tests {
         test!(Value::I8(122), Ok(122));
         test!(Value::I64(122), Ok(122));
         test!(Value::F64(122.0), Ok(122));
+        test!(Value::F64(122.1), Ok(122));
```
ever0de, line 443:
```rust
        test!(Value::F64(122.0), Ok(122));
        test!(Value::F64(122.1), Ok(122));
```

If it is the same case, it seems that there is no need to add it.

ever0de, line 443:
```rust
        test!(Value::F64(122.0), Ok(122));
        test!(Value::F64(122.1), Ok(122));
```

If it is the same case, it seems that there is no need to add it.

core/src/data/value/into.rs
```diff
@@ -484,8 +493,10 @@ mod tests {
         test!(Value::Bool(true), Ok(1));
         test!(Value::Bool(false), Ok(0));
         test!(Value::I8(122), Ok(122));
+        test!(Value::I64(122), Ok(122));
         test!(Value::I64(1234567890), Ok(1234567890));
         test!(Value::F64(1234567890.0), Ok(1234567890));
+        test!(Value::F64(1234567890.1), Ok(1234567890));
```
ever0de, line 499:
```rust
        test!(Value::F64(1234567890.0), Ok(1234567890));
        test!(Value::F64(1234567890.1), Ok(1234567890));
```

If it is the same case, it seems that there is no need to add it.

ever0de, line 499:
```rust
        test!(Value::F64(1234567890.0), Ok(1234567890));
        test!(Value::F64(1234567890.1), Ok(1234567890));
```

If it is the same case, it seems that there is no need to add it.

ever0de (commented):
Can you consider deleting duplicate test cases? 🙏 
Thanks 👍

core/src/data/value/into.rs
earney: I did not know that. I will try to keep that in mind for future cases.

core/src/data/value/into.rs
earney: I will change to test from x.1 to x.9.  that example will check for round vs truncate.

ever0de (approved):
LGTM

core/src/data/value/into.rs
```diff
@@ -152,9 +152,9 @@ impl TryInto<i64> for &Value {
                     0
                 }
             }
-            Value::I8(value) => *value as i64,
+            Value::I8(value) => value.to_i64().ok_or(ValueError::ImpossibleCast)?,
```
panarch, line 155: This looks unnecessary (cast from smaller type to bigger)

core/src/data/value/into.rs
```diff
@@ -191,8 +191,8 @@ impl TryInto<f64> for &Value {
                     0.0
                 }
             }
-            Value::I8(value) => *value as f64,
-            Value::I64(value) => (*value as f64).trunc(),
+            Value::I8(value) => value.to_f64().ok_or(ValueError::ImpossibleCast)?,
```
panarch, line 194: This, too.

panarch (changes requested):
Thanks a lot!
There are very tiny comments about cast from i8 to i64

panarch (approved):
All looks great, thanks a lot!

PR #575, opened 2022-05-26
have datatype enum use SCREAMING_SNAKE_CASE

earney:
Use snake case for data type enum:

Boolean => BOOLEAN
    Int8,  => INT8
    Int,  => INT,
etc.


panarch (approved):
Awesome, thanks a lot!

PR #576, opened 2022-05-26
Fix modulo binary op to handle zero divisor correctly

earney:
Should check that "other" is zero not the base number.  Also give a test to check for modulo zero.


ever0de (approved):
Thanks for fixing the bug! 👍

panarch (approved):
Thanks a lot!! merging 👍

PR #577, opened 2022-05-27
Replace MemoryStorage Key to core::data::Key,

panarch:
Replace MemoryStorage::Key to use core::data::Key.
Add PartialOrd impl to data::Key.
Update MemoryStorage to use IndexMap which does not require Ord impl and preserves insertion order.


ever0de (approved):
Looks good to me

PR #578, opened 2022-05-28
Fix rust clippy gh action to check optional features

panarch:
Clippy GitHub action was not working because all the features were activated by default because storages always use all the core features.
Fix rust.yml to run clippy in the core workspace for testing optional features.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -16,22 +16,21 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - run: rustup component add clippy rustfmt
-      - name: Clippy
-        run: cargo clippy --no-default-features -- -D warnings
-      - name: Clippy (alter-table)
-        run: cargo clippy --no-default-features --features alter-table -- -D warnings
-      - name: Clippy (index)
-        run: cargo clippy --no-default-features --features index -- -D warnings
-      - name: Clippy (alter-table & index)
-        run: cargo clippy --no-default-features --features "alter-table index" -- -D warnings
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -16,22 +16,21 @@ jobs:
...
-        run: cargo clippy --no-default-features --features "index transaction" -- -D warnings
-      - name: Clippy (alter-table & index & transaction)
-        run: cargo clippy --no-default-features --features "alter-table index transaction" -- -D warnings
-      - name: Clippy (all)
-        run: cargo clippy --all-features --all-targets -- -D warnings
+      - name: Clippy (root)
+        run: cargo clippy -- -D warnings
+      - name: Clippy (core)
+        run: |
+          cd core
```
ever0de, line 23: <https://doc.rust-lang.org/cargo/reference/workspaces.html#package-selection>

It seems fine to run the core-related command through package selection.

panarch: Oh, is there a command for running clippy by only picking specific package?
It looks that clippy does not have
e.g. `cargo clippy -h`
image

panarch: I thought something like
```
cargo clippy --workspace -p gluesql-core --no-default-features
```
but clippy runs for all the workspaces.

ever0de (approved):
ah..! I didn't know that clippy didn't have a feature.
<https://github.com/rust-lang/rust-clippy/issues/2518>

Current state looks good

PR #580, opened 2022-05-29
Fix `sled_multi_threaded` example code with threads of `INSERT` and `SELECT`,

ever0de:
### Detail

1. You can check whether the example code can be executed in github actions.
2. Fix the broken sled example code after multiple query string is supported.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -39,3 +39,11 @@ jobs:
         run: cargo test --no-default-features --verbose
       - name: Run tests with all features
         run: cargo test --all-features --all-targets --verbose
+      - name: Run hello_world example code
+        run: cargo run --package gluesql --example hello_world
+      - name: Run api_usage example code
+        run: cargo run --package gluesql --example api_usage
+      - name: Run sled_multi_threaded example code
+        run: cargo run --package gluesql --example sled_multi_threaded
+      - name: Run using_config example code
+        run: cargo run --package gluesql --example using_config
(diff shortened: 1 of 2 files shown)


ever0de:
<https://github.com/gluesql/gluesql/runs/6641599693?check_suite_focus=true>
```shell
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: StorageMsg("database is locked")', examples/sled_multi_threaded.rs:38:33
thread 'main' panicked at 'Something went wrong in the world thread: Any { .. }', examples/sled_multi_threaded.rs:43:14
```

Did you not support multithreading in sled storage? Or maybe the way you use it has changed?

ever0de:
As a result of checking, this code seems to be a code that has not been running before.
The `INSERT` statement does not work at the same time, and the ` txid != lock_txid` condition is applied, so an error message related to `database is locked` appears.

Modify the example code with threads of `INSERT` and `SELECT`.

.github/workflows/rust.yml
```diff
@@ -39,3 +39,11 @@ jobs:
...
       - name: Run tests with all features
         run: cargo test --all-features --all-targets --verbose
+      - name: Run hello_world example code
+        run: cargo run --package gluesql --example hello_world
+      - name: Run api_usage example code
+        run: cargo run --package gluesql --example api_usage
+      - name: Run sled_multi_threaded example code
+        run: cargo run --package gluesql --example sled_multi_threaded
+      - name: Run using_config example code
+        run: cargo run --package gluesql --example using_config
```
panarch, line 49: 👍

panarch (approved):
Awesome! root examples are now back to be maintained by our rust github action.
Thanks a lot 👍

PR #583, opened 2022-05-29
Ambiguous column selection should be detected while joining

### LINKED ISSUE #581: Doesn't catch out ambiguous column selection error in joined tables
## Should omit error for selecting ambiguous column
Currently gluesql doesn't omit error for selecting ambiguous column in joined tables.
image
...while it should be like below which is `psql`'s output
image

MRGRAVITY817:
# Resolves #581 
Currently whey we query projections from joined table, gluesql is unaware of ambiguous column names.
See the details from #581 .
### Added new `validate` mod under `plan` module in `gluesql-core`.
Right now the `validate` mod only holds validation logic only for this issue, but it should expand in future.
### Added `ColumnReferenceAmbiguous` error
I've place it in `SelectError`, but I am not quite sure if this is a right place to put in. Need discussion.


panarch:
It looks more likely to add a new feature - validation functionality to plan module
enhancement label would be good for this.

core/src/executor/select/error.rs
```diff
@@ -8,6 +8,9 @@ pub enum SelectError {
     #[error("table alias for blend not found: {0}")]
     BlendTableAliasNotFound(String),
 
+    #[error("column reference {0} is ambiguous, please specify the table name")]
```
MRGRAVITY817, line 11: @panarch @ever0de 
I've added a new error for ambiguous column ref in `SelectError` part because it seems like to happen specifically when using `SELECT` and `JOIN`.
But I am not so sure about this. Should I move this to `TranslateError` or not?

panarch: This looks obvious. We now have a new `validate` module in `plan`.
This means that the location we need to add an error is inside of `plan`.
How about adding `PlanError` to `plan/error.rs`'?

MRGRAVITY817: Sounds great!

core/src/plan/mod.rs
```diff
@@ -1,20 +1,26 @@
...
 
 #[cfg(test)]
 mod mock;
 
 use crate::{ast::Statement, result::Result, store::Store};
 
-pub use {index::plan as plan_index, join::plan as plan_join, schema::fetch_schema_map};
+use self::validate::validate;
+
+pub use {error::*, index::plan as plan_index, join::plan as plan_join, schema::fetch_schema_map};
```
ever0de, line 17:
```suggestion
pub use {error::*, index::plan as plan_index, join::plan as plan_join, schema::fetch_schema_map, self::validate::validate};
```

core/src/plan/validate.rs
```diff
@@ -0,0 +1,48 @@
...
+                            expr: Expr::Identifier(ident),
+                            ..
+                        } = select_item
+                        {
+                            let tables_with_given_col = schema_map
+                                .iter()
+                                .filter_map(|(_, schema)| {
+                                    schema.column_defs.iter().find(|col| &col.name == ident)
+                                })
+                                .collect_vec();
```
ever0de, line 31: <https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.count>

Not try using the iterator's count without having to make it a `Vec`?

MRGRAVITY817: Aha you're right

core/src/plan/mod.rs
MRGRAVITY817: Oops.. sorry!

ever0de (approved):
Looks good to me

PR #584, opened 2022-05-29
Use const `Dialect` in parse_query

ever0de:


panarch (approved):
Thanks for catching this!

PR #586, opened 2022-05-29
Add BYTEA data type support

panarch:
BYTEA type value has format of Vec<u8> which is for storing binary data.
Currently, hexadecimal string values are only accepted for inserting BYTEA values using SQL.
Rather than BLOB (binary large object), BYTEA name has used.
BLOB implies "large" but BYTEA is just a kind of byte array and it fits better for current use.

ref. https://www.postgresql.org/docs/current/datatype-binary.html

### DIFF
--- a/core/src/data/literal.rs
+++ b/core/src/data/literal.rs
@@ -1,6 +1,5 @@
 use {
-    super::StringExt,
-    super::Value,
+    super::{StringExt, Value},
     crate::{
         ast::{AstLiteral, DateTimeField},
         data::BigDecimalExt,
@@ -30,6 +29,9 @@ pub enum LiteralError {
     #[error("unreachable literal unary operation")]
     UnreachableUnaryOperation,
 
+    #[error("failed to decode hex string: {0}")]
+    FailedToDecodeHexString(String),
+
     #[error("operator doesn't exist: {0:?} LIKE {1:?}")]
     LikeOnNonString(String, String),
 
@@ -42,6 +44,7 @@ pub enum Literal<'a> {
     Boolean(bool),
     Number(Cow<'a, BigDecimal>),
     Text(Cow<'a, String>),
+    Bytea(Vec<u8>),
     Interval(super::Interval),
     Null,
 }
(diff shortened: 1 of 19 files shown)


core/src/data/literal.rs
```diff
@@ -261,13 +269,58 @@ impl<'a> Literal<'a> {
...
+        use {
+            super::{Literal, LiteralError},
+            crate::{
+                ast::{AstLiteral, DateTimeField},
+                data::Interval as I,
+                result::Result,
+            },
+        };
+
+        fn test(ast_literal: AstLiteral, literal: Result<Literal>) {
```
ever0de, line 289: Is there any reason not to use closures?

core/src/data/value/big_endian.rs
```diff
@@ -46,6 +46,7 @@ impl Value {
                 .chain(v.as_bytes().iter())
                 .copied()
                 .collect::<Vec<_>>(),
+            Value::Bytea(v) => v.to_vec(),
```
ever0de, line 49: If you need to clone ownership from `&Vec<_>` to `Vec<_>`,  `clone` looks fine.

ever0de (approved):
All comments are nitpicking, so I approve

core/src/data/value/big_endian.rs
panarch: I also get used to use `clone` but I just saw that compiler recommends to use `to_vec()` in this kind of cases (but it seems that though I use `clone`, clippy accepts, too)
Hmm, but It would be good to use `to_vec()` in here.

`to_vec` - Copies self into a new Vec.
description fits well to here

https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.to_vec

core/src/data/literal.rs
panarch: closure has two roles in Rust.
- captures outer environment
- allows omitting types

Here, compiler requires the function to specify types explicitly, so one of the pros was not for this.
And here, `test` does not capture any outer variables, so I thought it would be better to simply use normal fn, not closure.

PR #588, opened 2022-05-31
Replace SledStorage generic key T from IVec to data::Key

panarch:
SledStorage generic key `T` from `IVec` to `data::Key`.

Now.. both MemoryStorage and SledStorage use same `data::Key`.
Everything is ready to remove `T` from Store traits!


ever0de (approved):
LGTM

PR #589, opened 2022-05-31
Remove generic T from Store, StoreMut and Index traits,

panarch:
Replace generic T to data::Key for all store traits: Store, StoreMut and Index

Now it's all ready to add primary key support!

### DIFF
--- a/tests/glue.rs
+++ b/tests/glue.rs
@@ -5,7 +5,7 @@ use gluesql_core::{
     store::{GStore, GStoreMut},
 };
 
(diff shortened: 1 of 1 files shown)


core/src/executor/validate.rs
```diff
@@ -73,8 +73,8 @@ impl UniqueConstraint {
     }
 }
 
-pub async fn validate_unique<T>(
-    storage: &impl Store<T>,
+pub async fn validate_unique(
+    storage: &impl Store,
```
ever0de, line 77: Not related to this PR, but I used `&impl` instead of `&dyn` here.

ever0de (approved):
Looks good

PR #594, opened 2022-06-04
Simplify github rust action - Run tests & Run examples,

panarch:
Merge two test running actions into an one "Run tests".
Merge all running example actions into an one "Run examples"


ever0de (approved):
Looks good!

PR #595, opened 2022-06-04
add Rand function

earney:
This PR adds the Rand function so that random values between 0 and 1 can be generated.
there are a few issues that need to be addressed with this PR before it is merged.
The main issue is that a global variable named something like rng needs to be created.  Currently each time the rand function is called I recreate this variable, but for random function to work properly, the rng variable needs to be more on a global scale, than the current local scope.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -47,14 +47,12 @@ jobs:
           yarn build
           cd ../../
       - name: Run tests
-        run: cargo test --no-default-features --verbose
-      - name: Run tests with all features
-        run: cargo test --all-features --all-targets --verbose
-      - name: Run hello_world example code
-        run: cargo run --package gluesql --example hello_world
-      - name: Run api_usage example code
-        run: cargo run --package gluesql --example api_usage
-      - name: Run sled_multi_threaded example code
-        run: cargo run --package gluesql --example sled_multi_threaded
-      - name: Run using_config example code
-        run: cargo run --package gluesql --example using_config
+        run: |
+          cargo test --no-default-features --verbose
+          cargo test --all-features --all-targets --verbose
+      - name: Run examples
+        run: |
+          cargo run --package gluesql --example hello_world
+          cargo run --package gluesql --example api_usage
+          cargo run --package gluesql --example sled_multi_threaded
+          cargo run --package gluesql --example using_config


core/src/executor/evaluate/function.rs
```diff
@@ -274,6 +276,20 @@ pub fn floor(name: String, n: Evaluated<'_>) -> Result<Value> {
...
+
+    match n {
+        Some(x) => match x.try_into()? {
+            Value::I8(v) => rng = StdRng::seed_from_u64(v as u64),
+            Value::I64(v) => rng = StdRng::seed_from_u64(v as u64),
+            _ => {}
+        },
+        None => {}
+    }
+    Ok(Value::F64(rng.gen()))
```
panarch, line 290: This looks ok with seed, but without seed, it needs not to be deterministic.

```rust
let v = match seed {
    Some(seed) => seed.try_into()?,
    None => {
        return Ok(Value::F64(rand::random()));
    }
};  
    
let v = match v {
    Value::I8(v) => StdRng::seed_from_u64(v as u64).gen(),
    Value::I64(v) => StdRng::seed_from_u64(v as u64).gen(),
    _ => rand::random(), // may be we need to return error?
};  

Ok(Value::F64(v))
```

test-suite/src/function/rand.rs
```diff
@@ -0,0 +1,38 @@
+use crate::*;
+
+test_case!(rand, async move {
+    use gluesql_core::prelude::{Payload, Value::*};
+
+    let test_cases = vec![
+        ("CREATE TABLE mytable (id int)", Ok(Payload::Create)),
+        (r#"INSERT INTO mytable VALUES (1)"#, Ok(Payload::Insert(1))),
+        (
+            "SELECT RAND() as r FROM mytable",
```
panarch, line 10: `RAND` function without seed is not supposed to generate re-produceable result, but it looks that `RAND` always generate same values.

Writing tests for random functions, there exists a good reference.
`GENERATE_UUID`
https://github.com/gluesql/gluesql/blob/main/test-suite/src/function/generate_uuid.rs
https://github.com/gluesql/gluesql/blob/1e0f06f43990e26f9a506f5966ff4177b04b8f46/test-suite/src/function/generate_uuid.rs#L30-L34

core/src/executor/evaluate/function.rs
```diff
@@ -1,6 +1,8 @@
 use {
     super::{EvaluateError, Evaluated},
     crate::{ast::TrimWhereField, data::Value, result::Result},
+    rand::rngs::StdRng,
+    rand::{Rng, SeedableRng},
```
panarch, line 5: nit comment
you can merge this
e.g.
```rust
rand::{
    rngs::StdRng,
    Rng, SeedableRng,
}
```

core/src/ast/function.rs
```diff
@@ -89,6 +89,7 @@ pub enum Function {
         expr: Expr,
         chars: Option<Expr>,
     },
+    Rand(Vec<Expr>),
```
panarch, line 92: This looks that `RAND` function takes 0 or 1 argument but current ast accepts variable number of `Expr`.

It would be enough with `Rand(Option<Expr>)`

panarch (changes requested):

core/src/executor/evaluate/function.rs
earney: One issue, I am having..

error[E0308]: mismatched types
  --> core/src/plan/expr/function.rs:39:26
   |
17 |         match self {
   |               ---- this expression has type `&ast::function::Function`
18 |             Self::Now() | Function::Pi() | Function::GenerateUuid() => Exprs::Empty(empty()),
19 |             Self::Lower(expr)
   |                         ---- first introduced with type `&ast::expr::Expr` here
...
39 |             | Self::Rand(expr)
   |                          ^^^^ expected enum `ast::expr::Expr`, found enum `std::option::Option`
   |
Do we need to create another function similar to Exprs::Single([expr].into_iter()),, that accepts Optional arguments or is there a way to make it work with the current Exprs::Single?

core/src/ast/function.rs
earney: Do we need to create a OptionalExpr function or is there a way to use Exprs::Single([expr].into_iter()) ?
Here isthe error I'm getting:

`
 | Self::Rand(expr)
| ^^^^expected enum `ast::expr::Expr`, found enum `std::option::Option`
`

core/src/translate/function.rs
```diff
@@ -339,6 +339,14 @@ pub fn translate_function(sql_function: &SqlFunction) -> Result<Expr> {
                 divisor,
             })))
         }
+        "RAND" => {
+            check_len_range(name, args.len(), 0, 1)?;
+            let expr = match args.len() {
+                1 => Some(translate_expr(args[0])?),
+                0 => None,
+            };
```
ever0de, line 347: The return type of `len()` is usize.
Using the match syntax will result in unreachable code.
How about writing code like this:

```rust
let expr = args.get(0).map(|expr| translate_expr(expr)).transpose()?;
```

core/src/executor/evaluate/function.rs
```diff
@@ -274,6 +278,23 @@ pub fn floor(name: String, n: Evaluated<'_>) -> Result<Value> {
...
+        Some(seed) => seed.try_into()?,
+        None => {
+            return Ok(Value::F64(rand::random()));
+        }
+    };  
+        
+    let v = match v {
+        Value::I8(v) => StdRng::seed_from_u64(v as u64).gen(),
+        Value::I64(v) => StdRng::seed_from_u64(v as u64).gen(),
+        _ => rand::random(), // may be we need to return error?
```
ever0de, line 292: If you branch here, it seems to be an error because the value of 'seed' is of a type other than integer + signed integer.

core/src/executor/evaluate/stateless.rs
```diff
@@ -165,6 +164,13 @@ fn evaluate_function<'a>(
 
             f::rtrim(name(), expr, chars)
         }
+        Function::Rand(expr) => {
+            let expr = match expr {
+                Some(e) => Some(eval(e)?),
+                _ => None,
+            };
+            f::rand(expr)
```
ever0de, line 172: [nit-picking]
refactor issue

```suggestion
            f::rand(expr.as_ref().map(|expr| eval(expr)).transpose()?)
```

core/src/ast/function.rs
earney: I'm still getting this error.   Any way of fixing it?
 --> core\src\plan\expr\function.rs:39:26
   |
17 |         match self {
   |               ---- this expression has type `&ast::function::Function`
18 |             Self::Now() | Function::Pi() | Function::GenerateUuid() => Exprs::Empty(empty()),
19 |             Self::Lower(expr)
   |                         ---- first introduced with type `&ast::expr::Expr` here
...
39 |             | Self::Rand(expr)
   |                          ^^^^ expected enum `ast::expr::Expr`, found enum `std::option::Option`
   |

PR #598, opened 2022-06-05
Support variance aggregate function & update aggregate test

zmrdltl:
1. Add variance aggregation function
2. modified `avg` aggregate test
The test that compared the existing average value as an integer is modified to a test that compares the decimal result.


core/src/executor/aggregate/state.rs
```diff
@@ -59,7 +75,7 @@ impl<'a> AggrValue {
                     Ok(None)
                 }
             }
-            Self::Sum(value) => Ok(Some(Self::Sum(value.add(new_value).unwrap()))),
+            Self::Sum(value) => Ok(Some(Self::Sum(value.add(new_value)?))),
```
panarch, line 78: I wasn't aware of this, thanks!

panarch (approved):
Awesome!!
New aggregate function to GlueSQL 🥇 
Thanks a lot, merging!

Issue #596, opened 2022-06-05
Support composite type using `CREATE TYPE`


MRGRAVITY817:
## Support composite type using `CREATE TYPE`
Creating a column that takes a custom composite type can be useful. It's like saving a tuple-typed data.
## Example
```sql
-- Defining composite type
CREATE TYPE inventory_item AS (
    name            text,
    supplier_id     integer,
    price           numeric
);

-- Using composite type `inventory_item`
CREATE TABLE on_hand (
    item      inventory_item,
    count     integer
);
INSERT INTO on_hand VALUES (ROW('fuzzy dice', 42, 1.99), 1000);
```
## Reference
https://www.postgresql.org/docs/9.6/rowtypes.html

PR #599, opened 2022-06-05
Introducing `CompositeType`

MRGRAVITY817:
## Prepares entry layer for #596
Introduces new `CompositeType`, which allows you to compose various types into one.
This PR only implements the basic ingredients before the final goal of using `CREATE TYPE`.
### Expected usage when fully implemented
```sql
INSERT INTO Foo
  (id, name.id, name.order)
VALUES
  (1, 2, 3);

--- or

INSERT INTO Foo
  (id, name)
VALUES
  (1, { id: 2, order: 3 });
```
## Work In Progress 🚧

### DIFF
--- a/core/src/ast/data_type.rs
+++ b/core/src/ast/data_type.rs
@@ -17,4 +17,5 @@ pub enum DataType {
     Map,
     List,
     Decimal,
+    CustomType,
 }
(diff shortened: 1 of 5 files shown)


MRGRAVITY817:
Seems like sqlparser does not have SqlStatement::CreateType yet, so I cannot implement the translation part.
I should make a PR in sqlparser first if there's no related one with this.

I'll be back after a while! 👍

### On-going issue & PR from `sqlparser`
- Issue: https://github.com/sqlparser-rs/sqlparser-rs/issues/535
- PR: In progress 🚧

PR #607, opened 2022-06-14
Support `Deserialize` for `Payload`

### LINKED ISSUE #601: Can we make `Payload` derive from `Deserialize` trait?
Current `Payload` struct defines like the following: 

```rust
#[derive(Serialize, Debug, PartialEq)]
pub enum Payload {
  //...
}
```

We can see that it derives the `Serialize` trait from serde, so we can serialize to a buffer. But if we transfer the buffer to another place, we can't deserialize it into the `Payload` struct (by which we can analyze data defined in `Payload`, such as query select rows in `Payload::Select`).

ever0de:
resolved #601

### DIFF
--- a/core/src/executor/execute.rs
+++ b/core/src/executor/execute.rs
@@ -14,7 +14,7 @@ use {
         store::{GStore, GStoreMut},
     },
     futures::stream::{self, TryStreamExt},
-    serde::Serialize,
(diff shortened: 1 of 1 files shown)


panarch (approved):
Looks nice, thanks 👍

PR #611, opened 2022-06-20
Select cast i32 and i64 fix.

earney:
select cast ( i as int(32)) from ...    will output i32::MAX when I is greater than I32::MAX, and will put i32::MIN in for i values less than i32::MIN.  When I is not in i32 range, it should output an error.

This fix does cause one test to fail:
 SELECT CAST(1.1 AS INTEGER) AS cast FROM Item"#,
  should succeed by but fails with this PR.  Do we want this to get converted to an integer value (ie, truncated?)  I believe mysql does truncate the float to an integer.

### DIFF
--- a/test-suite/src/data_type/int32.rs
+++ b/test-suite/src/data_type/int32.rs
@@ -1,0 +1,114 @@
+use {
+    crate::*,
+    gluesql_core::{data::ValueError, prelude::Payload, prelude::Value::*},
+};
+
+test_case!(int32, async move {
+    run!(
+        "CREATE TABLE Item (
+        field_one INT(32),
+        field_two INT(32),
+    );"
+    );
+    run!("INSERT INTO Item VALUES (1, -1), (-2, 2), (3, 3), (-4, -4);");
+
+    let parse_i32 = |text: &str| -> i32 { text.parse().unwrap() };
+
+    let max: i64 = i32::MAX as i64 + 1_i64;
+    let min: i64 = i32::MIN as i64 - 1_i64;
+
+    test!(
+        Err(ValueError::FailedToParseNumber.into()),
+        &format!("INSERT INTO Item VALUES ({:?}, {:?})", max, min)
+    );
+
+    // cast i32::MAX+1
+    // this should produce an error! will create a different PR / issue for fixing this.
+    //test!(
+    //    Err(ValueError::LiteralCastToDataTypeFailed(DataType::Int32, max.to_string()).into()),
+    //    &format!("select cast({} as INT(32)) from Item", max)
+    //);
+
+    // cast i32::MIN-1
+    // this also should produce an error, will create a PR for it as well.
+    //test!(
+    //    Err(ValueError::LiteralCastToDataTypeFailed(DataType::Int32, min.to_string()).into()),
+    //    &format!("select cast({} as INT(32)) from Item", min)
+    //);
+
+    // lets try some valid SQL
+    test!(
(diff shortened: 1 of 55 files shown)


test-suite/src/data_type/int32.rs
```diff
@@ -14,27 +14,42 @@ test_case!(int32, async move {
...
-        &format!("INSERT INTO Item VALUES ({:?}, {:?})", max, min)
+        &format!(
+            "INSERT INTO Item VALUES ({:?}, {:?})",
+            i32::MAX as i64 + 1_i64,
+            i64 = i32::MIN as i64 - 1_i64
+        )
     );
 
     // cast i32::MAX+1
     // this should produce an error! will create a different PR / issue for fixing this.
```
panarch, line 27: Now you can remove these comments.

test-suite/src/data_type/int32.rs
```diff
@@ -14,27 +14,42 @@ test_case!(int32, async move {
...
+        )
+        .into()),
+        &format!(
+            "select cast({} as INT(32)) from Item",
+            i32::MAX as i64 + 1_i64
+        )
+    );
 
     // cast i32::MIN-1
     // this also should produce an error, will create a PR for it as well.
```
panarch, line 41: These, too.

test-suite/src/data_type/int64.rs
```diff
@@ -0,0 +1,123 @@
+use {
+    crate::*,
+    gluesql_core::{data::ValueError, prelude::DataType, prelude::Payload, prelude::Value::*},
```
panarch, line 3: nit clean up
you can merge prelude
```rust
gluesql_core::{
    data::ValueError,
    prelude::{DataType, Payload, Value::*},
},      
```

test-suite/src/function/cast.rs
```diff
@@ -51,10 +51,12 @@ test_case!(cast_literal, async move {
...
-        (
-            r#"SELECT CAST(1.1 AS INTEGER) AS cast FROM Item"#,
-            Ok(select!(cast I64; 1)),
-        ),
+
+        //this now fails..
+        //(
+        //    r#"SELECT CAST(1.1 AS INTEGER) AS cast FROM Item"#,
+        //    Ok(select!(cast I64; 1)),
+        //),
```
panarch, line 59: This looks reasonable decision that you've made.
Then, we don't need to keep this test case anymore.
Let's remove these code block :) thanks.

panarch (changes requested):
Nice improvement, thanks a lot!
With nit change requests about the comments, let's merge this to the main.

test-suite/src/function/cast.rs
```diff
@@ -51,9 +54,10 @@ test_case!(cast_literal, async move {
             r#"SELECT CAST("foo" AS INTEGER) AS cast FROM Item"#,
             Err(ValueError::LiteralCastFromTextToIntegerFailed("foo".to_owned()).into()),
         ),
+
         (
             r#"SELECT CAST(1.1 AS INTEGER) AS cast FROM Item"#,
-            Ok(select!(cast I64; 1)),
+            Err(ValueError::LiteralCastToDataTypeFailed(DataType::Int, "1.1".to_string()).into()),
```
panarch, line 60: Wow, nice, this is even much better than removing the codes which I mentioned.
Thanks!

panarch (approved):
Looks all good,
thanks a lot!

PR #613, opened 2022-06-29
Implement query builder base (AST builder)

panarch:
Let me introduce GlueSQL's new query builder!
Now we can build GlueSQL AST by method chaining.
GlueSQL's query builder is actually an AST builder. Rather than generating SQL text from APIs, it generates GlueSQL AST directly.

e.g.
```rust
let actual = table("Bar")
    .select()
    .filter(col("id").is_null())
    .group_by("id, (a + name)")
    .build();
let expected = "
    SELECT * FROM Bar
    WHERE id IS NULL
    GROUP BY id, (a + name)
";  
```

```rust
let actual = table("Bar")
    .select()
    .group_by("city")
    .project("city, COUNT(name) as num")
    .build();
let expected = "
    SELECT
      city, COUNT(name) as num
    FROM Bar
    GROUP BY city
";  
```

```rust
let actual = table("Person")
    .delete()
    .filter(col("name").is_null())
    .build();
let expected = "DELETE FROM Person WHERE name IS NULL";
```

### AST builder accepts various expressions.
e.g.
```rust
col("id").add(10)
// id + 10

text("name").concat(col("col2"))
// "name" || col2
```

Someone who gets used to use raw SQL, then it just works.
```rust
let query1 = table("Foo")
    .select()
    .filter(col("id").add(num(10)))
    .build();

let query2 = table("Foo")
    .select()
    .filter(col("id").add(10))
    .build();

let query3 = table("Foo")
    .select()
    .filter("id + 10")
    .build();
```

All `query1`, `query2` and `query3` are identical.


core/src/ast_builder/select/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+    pub having: Option<Expr>,
+    pub limit: Option<Expr>,
+    pub offset: Option<Expr>,
+}
+
+trait Prebuild {
+    fn prebuild(self) -> Result<NodeData>;
+}
+
+fn build_stmt(select_data: NodeData) -> Statement {
```
24seconds, line 35: [Question]
Is there a reason to use just function without struct impl? Maybe we could implement this function under the struct `NodeData`.

core/src/ast_builder/expr/is_null.rs
```diff
@@ -0,0 +1,27 @@
+use super::ExprNode;
+
+impl ExprNode {
+    pub fn is_null(self) -> Self {
+        Self::IsNull(Box::new(self))
+    }
+
+    pub fn is_not_null(self) -> Self {
```
24seconds, line 8: [comment]
Though the file name is `is_null`, it contains `is_not_null`. But I understand that making a new file for `is_not_null` seems too much.

core/src/ast_builder/expr/function.rs
```diff
@@ -0,0 +1,53 @@
...
+                .try_into()
+                .map(Function::Abs)
+                .map(Box::new)
+                .map(Expr::Function),
+        }
+    }
+}
+
+impl ExprNode {
+    pub fn abs(self) -> ExprNode {
```
24seconds, line 29: [comment]
Nice. There could  be many functions in the near future!

core/src/ast_builder/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+    expr::{col, expr, nested, num, text, ExprNode},
+    expr_list::ExprList,
+    select::{
+        GroupByNode, HavingNode, LimitNode, LimitOffsetNode, OffsetLimitNode, OffsetNode,
+        ProjectNode, SelectNode,
+    },
+    select_item_list::SelectItemList,
+};
+
+pub fn table(table_name: &str) -> TableNode {
```
24seconds, line 22: [Question]
Is there a reason for `Table` residing in the mod.rs?  I expected separated file for example `table.rs` when I reading the code.

24seconds (approved):
Though I'm not familiar with glueSQL codebase but It looks good to me.
I enjoyed reading this pr. :+1:

core/src/ast_builder/expr/is_null.rs
ever0de: How about omitting the is_ suffix and putting null related stuff in the null name?

core/src/ast_builder/select/mod.rs
panarch: That is really a good point. Thanks a lot! I've just applied this

core/src/ast_builder/select/root.rs
```diff
@@ -0,0 +1,110 @@
...
+        ast_builder::{table, test},
+    };
+
+    #[test]
+    fn select() {
+        let actual = table("App").select().build();
+        let expected = "SELECT * FROM App";
+        test(actual, expected);
+
+        let actual = table("Bar").select().filter("id IS NULL").build();
```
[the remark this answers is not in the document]
panarch: Awesome, this is what we will make it.

1.
```rust
let actual = table("Bar").select().filter("id IS NULL").filter("foo = 1").build();
```

2.
```rust
let actual = table("Bar").select().filter("id IS NULL").filter("foo = 1").build();
```

3.
```rust
let actual = table("Bar").select().filter(expr("id IS NULL").and(expr("foo = 1"))).build();
// or
let actual = table("Bar").select().filter(col("id").is_null().and(col("foo").eq(1))).build();
```

We will support all above codes in near future!

core/src/ast_builder/expr/is_null.rs
panarch: In other cases (`ast/ Expr`)
image
`negated` field is used for merging two different exprs into a single one.

However, `IsNull` and `IsNotNull` are kind of exceptions. so.. I think it would be acceptable to gather `is_null` and `is_not_null` into a single file.

core/src/ast_builder/mod.rs
panarch: Thanks again 👍 this is also really a good point that I missed.
I splitted the codes and add some comments, too.
image

PR #616, opened 2022-07-08
Add floor func in ast_builder

### LINKED ISSUE #615: Implement FLOOR function
# Background
This is well described in #614. 

# Task
Support [`FLOOR` function](https://www.w3schools.com/sql/trysqlserver.asp?filename=trysql_func_sqlserver_floor).

To do this, test case below should pass.

```rust
#[cfg(test)]
mod tests {
    use crate::ast_builder::{abs, col, expr, floor, test_expr};

    #[test]
    fn function() {
        // FLOOR
        let actual = floor(col("num"));
        let expected = "FLOOR(num)";
        test_expr(actual, expected);

        let actual = expr("base - 10").floor();
        let expected = "FLOOR(base - 10)";
        test_expr(actual, expected);
    }
}
```

yujong-lee:
Resolves #615

### DIFF
--- a/core/src/ast_builder/expr/function.rs
+++ b/core/src/ast_builder/expr/function.rs
@@ -9,6 +9,7 @@ use {
 #[derive(Clone)]
 pub enum FunctionNode {
     Abs(ExprNode),
+    Floor(ExprNode),
 }
 
 impl TryFrom<FunctionNode> for Expr {
@@ -21,6 +22,11 @@ impl TryFrom<FunctionNode> for Expr {
                 .map(Function::Abs)
                 .map(Box::new)
                 .map(Expr::Function),
+            FunctionNode::Floor(expr_node) => expr_node
+                .try_into()
+                .map(Function::Floor)
+                .map(Box::new)
+                .map(Expr::Function),
         }
     }
 }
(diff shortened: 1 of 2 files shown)


core/src/ast_builder/expr/function.rs
```diff
@@ -49,5 +62,14 @@ mod tests {
...
+
+        // FLOOR
+        let actual = floor(col("num"));
+        let expected = "FLOOR(num)";
+        test_expr(actual, expected);
+
+        let actual = expr("base - 10").floor();
+        let expected = "FLOOR(base - 10)";
+        test_expr(actual, expected);
     }
```
bugoverdose, line 74: Instead of explaining that certain tests are about `abs()`, `floor()`, etc using comments(`// ABS`, `// FLOOR`), 
how about we separate them into different test functions?

I can't select the lines from 56~64 at github, so it's kind of hard to explain, but it's kind of like this.

```rust
#[cfg(test)]
mod tests {
    use crate::ast_builder::{abs, floor, col, expr, test_expr};

    #[test]
    fn function_abs() {
        let actual = abs(col("num"));
        let expected = "ABS(num)";
        test_expr(actual, expected);

        let actual = expr("base - 10").abs();
        let expected = "ABS(base - 10)";
        test_expr(actual, expected);
    }

    #[test]
    fn function_floor() {
        let actual = floor(col("num"));
        let expected = "FLOOR(num)";
        test_expr(actual, expected);

        let actual = expr("base - 10").floor();
        let expected = "FLOOR(base - 10)";
        test_expr(actual, expected);
    }
}
```

bugoverdose (commented):
Hello, I really think that you did a great job in just a few hours.

I'm not a reviewer, but I have a simple question/suggestion regarding how we are going to organize test codes.

core/src/ast_builder/expr/function.rs
yujong-lee: @bugoverdose  Yes. It totally makes sense. Thank you for pointing it out!

ever0de (approved):
Looks good to me 👍

panarch (approved):
All looks great!
Splitting unit tests look also better, thanks a lot 👍

PR #620, opened 2022-07-09
feat(ast_build): implement ifnull function

### LINKED ISSUE #618: [AST Builder] Implement `ifnull` function
## Backgroud

The following issue is a follow-up issue of #614.
The point is to make it easy to use SQL function `IFNULL` via method chaining.

## Task

The following issue is to implement the `ifnull` function in `ast_builder` module to support the `IfNull` function in the query builder.

Following tests should pass in order to complete the following task.

```rust
#[test]
fn ifnull_function() {
    let actual = col("updated_at").ifnull(col("created_at"));
    let expected = "IFNULL(updated_at, created_at)";
    test_expr(actual, expected);

    let actual = text("HELLO").ifnull(text("WORLD"));
    let expected = "IFNULL('HELLO', 'WORLD')";
    test_expr(actual, expected);
}
```

bugoverdose:
Hello, I implemented the `ifnull` function at `ast_build`.

I was going to name it `if_null` because it's snake_case, 
but I found the already implemented `ifnull` function at `executor/evaluate/function.rs`, so I erased the `_` part.

Please leave an RC if there's any problem, suggestion or conflict at main branch.

Thank you.

Closes #618

### DIFF
--- a/core/src/ast_builder/expr/function.rs
+++ b/core/src/ast_builder/expr/function.rs
@@ -78,6 +78,7 @@ mod tests {
         test_expr(actual, expected);
     }
 
+    #[test]
     fn function_floor() {
         let actual = floor(col("num"));
         let expected = "FLOOR(num)";
@@ -88,6 +89,7 @@ mod tests {
         test_expr(actual, expected);
     }
 
+    #[test]
     fn function_ifnull() {
         let actual = col("updated_at").ifnull(col("created_at"));
         let expected = "IFNULL(updated_at, created_at)";
@@ -95,5 +97,6 @@ mod tests {
 
         let actual = text("HELLO").ifnull(text("WORLD"));
         let expected = "IFNULL('HELLO', 'WORLD')";
+        test_expr(actual, expected);
     }
 }
--- a/core/src/ast_builder/expr/function.rs
+++ b/core/src/ast_builder/expr/function.rs
@@ -65,7 +65,7 @@ pub fn floor<T: Into<ExprNode>>(expr: T) -> ExprNode {
 
 #[cfg(test)]
 mod tests {
-    use crate::ast_builder::{abs, col, expr, floor, test_expr, text};
+    use crate::ast_builder::{abs, col, expr, floor, ifnull, test_expr, text};
 
     #[test]
     fn function_abs() {
@@ -80,12 +80,12 @@ mod tests {
 
     #[test]
     fn function_ifnull() {
-        let actual = col("updated_at").ifnull(col("created_at"));
-        let expected = "IFNULL(updated_at, created_at)";
+        let actual = ifnull(text("HELLO"), text("WORLD"));
+        let expected = "IFNULL('HELLO', 'WORLD')";
         test_expr(actual, expected);
 
-        let actual = text("HELLO").ifnull(text("WORLD"));
-        let expected = "IFNULL('HELLO', 'WORLD')";
+        let actual = col("updated_at").ifnull(col("created_at"));
+        let expected = "IFNULL(updated_at, created_at)";
         test_expr(actual, expected);
     }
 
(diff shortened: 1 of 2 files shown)


core/src/ast_builder/expr/function.rs
```diff
@@ -21,6 +22,13 @@ impl TryFrom<FunctionNode> for Expr {
                 .map(Function::Abs)
                 .map(Box::new)
                 .map(Expr::Function),
+            FunctionNode::IfNull(expr_node, then_node) => expr_node
+                .try_into()
+                .and_then(|expr| then_node
+                    .try_into()
+                    .map(|then| Function::IfNull { expr, then })
+                    .map(Box::new)
+                    .map(Expr::Function))
```
ever0de, line 31: It doesn't seem like rustfmt is applied.
Would you like to run `cargo fmt` ?

bugoverdose: @ever0de Yes, I totally forgot to do so.

I ran the command and pushed the commit. 

Thank you for the fast feedback.

ever0de (approved):
Looks good! thanks

core/src/ast_builder/mod.rs
```diff
@@ -22,7 +22,7 @@ pub use expr::{col, expr, nested, num, text, ExprNode};
 /// Available aggregate or normal SQL functions
 pub use expr::{
     aggregate::{max, sum, AggregateNode},
-    function::{abs, floor, FunctionNode},
+    function::{abs, floor, ifnull, FunctionNode},
 };
```
bugoverdose, line 26: This part gets ordered alphabetically when I run `cargo fmt` though.

bugoverdose (commented):
I fetched the upstream main branch, merged it, and resolved the conflict.

core/src/ast_builder/expr/function.rs
```diff
@@ -44,16 +55,20 @@ pub fn abs<T: Into<ExprNode>>(expr: T) -> ExprNode {
...
 }
 
 #[cfg(test)]
 mod tests {
-    use crate::ast_builder::{abs, col, expr, floor, test_expr};
+    use crate::ast_builder::{abs, col, expr, floor, test_expr, text};
 
     #[test]
-    fn function_abs() {
+    fn abs_test() {
```
ever0de, line 71: Since we used the `#[test]` macro, it turns out that this is a test function.
It seems good to omit the test prefix.

example)
https://github.com/gluesql/gluesql/blob/main/core/src/ast_builder/select/having.rs#L73-L74
https://github.com/gluesql/gluesql/blob/main/core/src/plan/evaluable.rs#L175-L176

bugoverdose: I agree. I really wanted to just write it like `abs`, `ifnull`, `floor`.

But I got an `error[E0255]: the name abs is defined multiple times` message when I do so (because of line 68).

I'll rollback the changes to `function_abs` format for now.

ever0de (approved):
LGTM

core/src/ast_builder/expr/function.rs
```diff
@@ -63,6 +78,17 @@ mod tests {
...
 
+    #[test]
+    fn function_ifnull() {
+        let actual = col("updated_at").ifnull(col("created_at"));
+        let expected = "IFNULL(updated_at, created_at)";
+        test_expr(actual, expected);
+
+        let actual = text("HELLO").ifnull(text("WORLD"));
+        let expected = "IFNULL('HELLO', 'WORLD')";
+        test_expr(actual, expected);
```
panarch, line 89: Current two unit tests are identical.
Both tests use
```rust
pub fn ifnull(self, another: ExprNode) -> ExprNode {
   ..
}
```

Let's replace one of those to use
```rust
pub fn ifnull<T: Into<ExprNode>, V: Into<ExprNode>>(expr: T, then: V) -> ExprNode {
   ..
}
```

bugoverdose: You are right. I'll fix one of them.

panarch (approved):
Looks all good, thanks a lot 👍

PR #621, opened 2022-07-09
feat(ast_build): implement upper function

### LINKED ISSUE #617: [AST Builder] Implement Upper function
# Backgroud

The following issue is a follow-up issue of #614.

# Task

The following issue is to implement the `Upper` function in `ast_builder` module to support the `Upper` function in the query builder.
The following test should pass in order to complete the following task.

```rust
#[test]
fn function() {
    // Upper
    let actual = upper(text("ABC"));
    let expected = "UPPER('ABC')";
    test_expr(actual, expected);

    let actual = expr("HoHo").upper();
    let expected = "UPPER(HoHo)";
    test_expr(actual, expected);
}
```

ChobobDev:
Hello, I implemented the `upper` function at `ast_build`.

I have ran `cargo fmt` to format the files changes.

Please leave an RC if there's any problem, suggestion or conflict at main branch.

resolves #617 

Thank you

### DIFF
--- a/core/src/ast_builder/expr/function.rs
+++ b/core/src/ast_builder/expr/function.rs
@@ -13,6 +13,7 @@ pub enum FunctionNode {
     Floor(ExprNode),
     Left(ExprNode, ExprNode),
     Right(ExprNode, ExprNode),
+    Upper(ExprNode),
 }
 
 impl TryFrom<FunctionNode> for Expr {
@@ -51,6 +52,11 @@ impl TryFrom<FunctionNode> for Expr {
                     .map(Box::new)
                     .map(Expr::Function)
             }),
+            FunctionNode::Upper(expr_node)=>expr_node
+                .try_into()
+                .map(Function::Upper)
+                .map(Box::new)
+                .map(Expr::Function),
         }
     }
 }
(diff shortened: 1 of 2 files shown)


ever0de:
ref: #617 
<https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue>
[Recommended] If you use this, the issue will be automatically resolved when the PR is merged.

ChobobDev:
Thank you for reminding me to link the issue when opening PR.

I think we can solve this by making a PR template as @bugoverdose suggested to make sure this does not happen again.

core/src/ast_builder/expr/function.rs
```diff
@@ -110,6 +125,17 @@ mod tests {
         test_expr(actual, expected);
     }
 
+    #[test]
+    fn function_upper() {
+        // Upper
```
panarch, line 130: Now it looks that this comment `// Upper` is unnecessary.
Everything else looks nice 👍 
Let's merge with this simple nit change.

ChobobDev: I totally agree with you, 
The comments are unnecessary at this point. 
I will fix them.

panarch (approved):
Looks all great, thanks a lot!
Merging 👍

PR #622, opened 2022-07-09
Remove unnecessary `Result`

ever0de:
Ummm... no Context as to why this was, but it doesn't seem necessary?


panarch (approved):
That's really a good catch, looks nice 👍
Thanks a lot!

PR #629, opened 2022-07-10
[DRAFT] Add CSV storage

### LINKED ISSUE #532: [CSVStorage] Implement storage that supports CSV
You have a separate schema file, which will probably be a file with the SQL extension.
ex) `{table_name}.schema.sql`
We read it in and create a table.

Then, when scan data, schema validation is performed. (Wouldn't it be enough to verify only on the first run?)

MRGRAVITY817:
## Resolves #532 
This new storage implementation will have the following functionalities:
- [ ] Read CSV file from the path like reading a table.
- [ ] Check if the CSV file has an interpretable schema using `{table_name}.schema.sql` file.
- [ ] Mutate the original CSV file, or make a copy of mutated result.
- [ ] [...will add more]

## WIP 🚧


storages/csv-storage/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "gluesql_csv_storage"
```
ever0de, line 2: It would be better to use -.

```suggestion
name = "gluesql-csv-storage"
```

ever0de (commented):
😀

storages/csv-storage/Cargo.toml
MRGRAVITY817:
```toml
[package]
name = "gluesql_memory_storage"
```
Hmm... but other storage subcrates in this project use snake case.

ever0de: memory storage has been deployed before so it is :cry:

ever0de: ref: <https://github.com/gluesql/gluesql/blob/main/storages/shared-memory-storage/Cargo.toml#L2>

panarch: Oh, you don't need to care about underscore cases. You can use `-` dash.
I published `gluesql_memory_storage` and `gluesql_sled_storage` before we change our project structure into workspace-based.
And.. `crates.io` does not distinguish `_` and `-` so I wasn't able to rename `gluesql_memory_storage` to `gluesql-memory-storage`.


image

MRGRAVITY817:
@panarch 

Initial structure that I've been thinking for `schema.toml` 

```toml
[table.users]
path = "example/data/users.csv"
columns.id = "Int128"
columns.name = "Text"
columns.age = "Uint8"
columns.role = "Text"

[table.orders]
path = "example/data/orders.csv"
columns.id = "Int128"
columns.name = "Text"
columns.orderer_id = "Int128"
columns.restaurant_id = "Int128"
columns.quantity = "Uint8"
columns.price = "Float"
columns.status = "Text"

```

MRGRAVITY817:
Or maybe this one is better, at least more easy to parse using `toml` crate.

```toml
[[tables]]
name = "users"
path = "example/data/users.csv"
columns = [
	{ name = "id",   type = "Int128", options = ["PKey"]    },
	{ name = "name", type = "Text",   options = ["NotNull"] },
	{ name = "age",  type = "Unit8"                         },
	{ name = "role", type = "Text",   options = ["NotNull"] },
]
```

Issue #636, opened 2022-07-11
[AST Builder] Implement Reverse functions


seonghun-dev:
# Description
Implements ast builders for `Reverse` functions.

## Background
Refered https://github.com/gluesql/gluesql/issues/614

## Task
Implement [`Reverse` function](https://www.w3schools.com/sql/trysqlserver.asp?filename=trysql_func_sqlserver_reverse) in ast_builder module to reverses a string
The following tests should be passed:

``` rust
    #[test]
    fn function_reverse() {
        let actual = reverse(text("GlueSQL"));
        let expected = "REVERSE('GlueSQL')";
        test_expr(actual, expected);

        let actual = expr("GlueSQL").reverse();
        let expected = "REVERSE(GlueSQL)";
        test_expr(actual, expected);
    }
```

PR #637, opened 2022-07-11
Implement add reverse func in ast_builder

seonghun-dev:
## Issue
Implement https://github.com/gluesql/gluesql/issues/636

## Description
Implement `reverse` function at ast_builder


panarch (approved):
Looks all nice, thanks a lot 👍

PR #648, opened 2022-07-15
Support `VALUES List` in `select` module

### LINKED ISSUE #417: Support `SetExpr::Values` in `select` module

devgony:
To resolve #417,

## Expected result
```sql
postgres> VALUES (1, 'a'), (2, 'b');
```
```
 column1 | column2
---------+---------
       1 | a
       2 | b
```

## To do
- [x] Implement values list base
- [ ] Validate the number of values
- [ ] Validate each type of values

### DIFF
--- a/core/src/data/row.rs
+++ b/core/src/data/row.rs
@@ -117,4 +117,26 @@ impl Row {
 
         Ok(())
     }
+
+    pub fn to_rows(exprs_list: &Vec<Vec<Expr>>) -> Result<Vec<Result<Self>>> {
+        let rows = exprs_list
+            .iter()
+            .map(|exprs| Self::to_row(exprs))
+            .collect::<Vec<_>>();
+
+        Ok(rows)
+    }
+
+    fn to_row(exprs: &Vec<Expr>) -> Result<Self> {
+        let values = exprs
+            .into_iter()
+            // .map(|expr| evaluate_stateless(None, expr).unwrap().try_into().unwrap())
+            .map(|expr| evaluate_stateless(None, expr))
+            .filter_map(Result::ok)
+            .map(|evaluated| evaluated.try_into())
+            .filter_map(Result::ok)
+            .collect::<Vec<_>>();
+
+        Ok(Self(values))
+    }
 }
--- a/core/src/data/row.rs
+++ b/core/src/data/row.rs
@@ -23,6 +23,12 @@ pub enum RowError {
 
     #[error("conflict! row cannot be empty")]
     ConflictOnEmptyRow,
+
+    #[error("VALUES lists must all be the same length")]
+    NumberOfValuesDifferent,
+
+    #[error("VALUES types {0} and {1} cannot be matched")]
+    ValuesTypeDifferent(String, String),
 }
 
 #[derive(iter_enum::Iterator)]
--- a/core/src/executor/select/mod.rs
+++ b/core/src/executor/select/mod.rs
@@ -132,34 +132,12 @@ pub async fn select_with_labels<'a>(
                 .enumerate()
                 .map(|(i, _)| format!("column{}", i + 1))
                 .collect::<Vec<_>>();
-            // let rows = values_list
-            //     .iter()
-            //     .map(|values| {
-            //         let values = values
-            //             .iter()
-            //             .map(|value| {
-            //                 let value: Value =
-            //                     evaluate_stateless(None, value).unwrap().try_into().unwrap();
-            //                 value
-            //             })
-            //             .collect::<Vec<_>>();
-            //         Ok(Row(values))
-            //     })
-            //     .collect::<Vec<_>>();
-
             let rows = Row::to_rows(values_list)?;
-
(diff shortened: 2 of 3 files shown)


core/src/data/row.rs
```diff
@@ -111,4 +117,39 @@ impl Row {
...
         Ok(())
     }
+
+    pub fn to_rows(exprs_list: &[Vec<Expr>]) -> Result<Vec<Result<Self>>> {
+        let rows = exprs_list.iter().map(Row::try_from).collect::<Vec<_>>();
+
+        Ok(rows)
+    }
+
+    // fn to_row(exprs: &[Expr]) -> Result<Self> {
```
devgony, line 127: Which is better to use between Associated Function `to_row` and below `try_from`?

core/src/executor/select/mod.rs
```diff
@@ -124,8 +125,18 @@ pub async fn select_with_labels<'a>(
...
-        _ => {
-            return Err(SelectError::Unreachable.into());
+        SetExpr::Values(Values(values_list)) => {
+            let limit = Limit::new(query.limit.as_ref(), query.offset.as_ref())?;
+            let labels = values_list
+                .iter()
+                .enumerate()
+                .map(|(i, _)| format!("column{}", i + 1))
+                .collect::<Vec<_>>();
+            let rows = Row::to_rows(values_list)?;
```
panarch, line 135: Instead of adding `Row::to_rows`, how about including
image
TryFrom impl in here?

If we have more than a single case which requires conversion from `Vec<Expr>` to `Row`, then we can consider to make this as a shared function. But currently it is only for evaluating `SetExpr::Values` in `select` module, so it would be  enough to be located only for here.

core/src/executor/select/mod.rs
```diff
@@ -124,8 +124,35 @@ pub async fn select_with_labels<'a>(
...
+                    if exprs.len() != first_len {
+                        return Err(RowError::NumberOfValuesDifferent.into());
+                    }
+                    let values = exprs
+                        .iter()
+                        .map(|expr| evaluate_stateless(None, expr))
+                        .filter_map(Result::ok)
+                        .map(|evaluated| evaluated.try_into())
+                        .filter_map(Result::ok)
+                        .collect::<Vec<_>>();
```
panarch, line 147: If error occurs in `evaluate_stateless` or `evaluated.try_into()`, then the current code will filter out and return `Ok` because of `filter_map`.
Rather than filter_map, it would be good to consider replacing `filter_map` to `and_then` like ones.
Then, by `collect::<Result<Vec<_>>>()` we can propagate `Result` to outside.

core/src/executor/select/mod.rs
```diff
@@ -124,8 +124,33 @@ pub async fn select_with_labels<'a>(
...
-        _ => {
-            return Err(SelectError::Unreachable.into());
+        SetExpr::Values(Values(values_list)) => {
+            let limit = Limit::new(query.limit.as_ref(), query.offset.as_ref())?;
+            let labels = values_list
+                .iter()
+                .enumerate()
+                .map(|(i, _)| format!("column{}", i + 1))
+                .collect::<Vec<_>>();
+            let first_len = values_list[0].len();
```
MRGRAVITY817, line 134: Will there be any possibility when users will accidentally call values list WITHOUT any actual list items?

e.g. 
```sql
# Calling without any list
gluesql> VALUES;
gluesql> ??
```

If this happens, maybe this `values_list[0].len()` will omit error or warning due to indexing empty list.

devgony: Thank you for pointing out !
Fortunatly, those kinda SQLs are invalid so that it gets ParseError before execute the line !

MRGRAVITY817: Thats awesome! Thanks

panarch: Not necessary work for now, but we may need to use a new data structure which can describe non-empty-vector.
maybe you can add `NonEmptyVec` to `/utils`?
Then, `NonEmptyVec` can become a fundamental solution for this kind of situations.

test-suite/src/values.rs
```diff
@@ -0,0 +1,55 @@
...
+            )),
+        ),
+        (
+            "VALUES (1), (2) offset 1",
+            Ok(select!(
+                column1;
+                I64;
+                2
+            )),
+        ),
```
panarch, line 50: 👍

panarch (approved):
Beautiful... All looks great, thanks a lot!!

PR #649, opened 2022-07-16
Add missing test case in `test-suite` workspace

ever0de:
ref: https://coveralls.io/builds/50921796/source?filename=test-suite%2Fsrc%2Ftransaction%2Falter_table.rs#L8

### DIFF
--- a/test-suite/src/lib.rs
+++ b/test-suite/src/lib.rs
@@ -239,6 +239,10 @@ macro_rules! generate_transaction_alter_table_tests {
         }
 
         glue!(
+            alter_table_rename_table,
+            transaction::alter_table_rename_table
+        );
+        glue!(
             transaction_alter_table_rename_column,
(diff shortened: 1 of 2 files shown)


test-suite/src/lib.rs
```diff
@@ -238,6 +238,10 @@ macro_rules! generate_transaction_alter_table_tests {
             };
         }
 
+        glue!(
+            alter_table_rename_table,
```
panarch, line 242: This test is combination of two optional features `alter-table` and `transaction`.
So, it would be better to use `transaction_alter_table_rename_table` as a test name

ever0de: Thanks!

panarch (approved):
Nice catch! Thank a lot! 
wow.. single entire test was not checked even from the beginning...

PR #653, opened 2022-07-16
Enhance SelectNode::filter

woojinnn:
## Task
I modified `ast_builder::select::filter` function.

## Previous Error
When several filter phrases are present together, that cannot be fully considered.
For example,
```rust
let actual = table("Bar")
    .select()
    .filter("id IS NULL")
    .filter("id > 10")
    .build();
```
was built while `is IS NULL` clause being ignored.  

I modified the code so that the following test can pass:
```rust
let actual = table("Bar")
    .select()
    .filter("id IS NULL")
    .filter("id > 10")
    .filter("id < 20")
    .build();
let expected = "SELECT * FROM Bar WHERE id IS NULL AND id > 10 AND id < 20";

test(actual, expected);
```

## FYI
I added `ast_builder::expr::binary_op::{lt, and}` function too.

### DIFF
--- a/core/src/ast_builder/expr/binary_op.rs
+++ b/core/src/ast_builder/expr/binary_op.rs
@@ -45,12 +45,24 @@ impl ExprNode {
         self.binary_op(BinaryOperator::Gt, other)
     }
 
+    pub fn gteq<T: Into<Self>>(self, other: T) -> Self {
+        self.binary_op(BinaryOperator::GtEq, other)
+    }
+
     pub fn lt<T: Into<Self>>(self, other: T) -> Self {
         self.binary_op(BinaryOperator::Lt, other)
     }
 
+    pub fn lteq<T: Into<Self>>(self, other: T) -> Self {
+        self.binary_op(BinaryOperator::LtEq, other)
+    }
+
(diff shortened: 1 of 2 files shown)


core/src/ast_builder/expr/binary_op.rs
```diff
@@ -45,13 +45,25 @@ impl ExprNode {
...
+        self.binary_op(BinaryOperator::LtEq, other)
+    }
+
     pub fn and<T: Into<Self>>(self, other: T) -> Self {
         self.binary_op(BinaryOperator::And, other)
     }
+
+    pub fn or<T: Into<Self>>(self, other: T) -> Self {
+        self.binary_op(BinaryOperator::Or, other)
+    }
```
panarch, line 66: These would be enough to be as separate PR.
Supporting `and` is dependent to `filter` so `and` is supposed to be with `filter` improvement, but `gteq`, `lteq` and `or` are not related to `filter`.
Could you split and make another PR for supporting `gteq`, `lteq` and `or`?
It would be more common to use `gte` & `lte` than `gteq` & `lteq`.

panarch (approved):
Looks all good, nice improvement! Thanks a lot 👍

PR #655, opened 2022-07-16
Add more expression support to project,

panarch:
Add SelectItemNode to support more expr types not only SQL text
- AST itself, SQL text, ExprNode and Vec of each type.

e.g.
```rust
let actual = table("Foo")
    .select()
    .project(vec!["col1", "col2"])
    .project("col3")
    .project(vec!["col4".into(), col("col5")])
    .project(col("col6"))
    .project("col7 as hello")
    .build();
let expected = "
    SELECT
        col1, col2, col3,
        col4, col5, col6,
        col7 as hello
    FROM
        Foo
";
test(actual, expected);
```

### DIFF
--- a/core/src/ast_builder/mod.rs
+++ b/core/src/ast_builder/mod.rs
@@ -2,6 +2,7 @@ mod delete;
 mod expr;
 mod expr_list;
 mod select;
+mod select_item;
 mod select_item_list;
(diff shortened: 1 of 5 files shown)


ever0de (approved):
Good 👍

PR #635, opened 2022-07-11
feat(ast_builder): Add Agggregate functions

zmrdltl:
resolve #634 
## Tasks
~~count~~ => will do later PR
- [x] : avg
- [x] : min
- [x] : variance

PR #656, opened 2022-07-17
[AST Builder] Implement Agggregate function `COUNT`

zmrdltl:
## Task
Implementing aggregate function `COUNT` is a continuation of the previous work. #635 
enum Implemented by adding a new enum called `CountExprNode` to convert between `ExprNode` and `CountArgExpr`

### DIFF
--- a/core/src/ast_builder/expr/aggregate.rs
+++ b/core/src/ast_builder/expr/aggregate.rs
@@ -1,7 +1,7 @@
 use {
     super::ExprNode,
     crate::{
-        ast::{Aggregate, CountArgExpr, Expr, ToSql},
+        ast::{Aggregate, CountArgExpr, Expr},
         parse_sql::parse_expr,
         result::{Error, Result},
         translate::translate_expr,
@@ -10,7 +10,7 @@ use {
 
 #[derive(Clone)]
 pub enum AggregateNode {
-    Count(CountExprNode),
+    Count(CountArgExprNode),
     Sum(ExprNode),
     Min(ExprNode),
     Max(ExprNode),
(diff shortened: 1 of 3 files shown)


core/src/translate/function.rs
```diff
@@ -105,7 +105,7 @@ fn translate_function_one_arg<T: FnOnce(Expr) -> Function>(
         .map(Expr::Function)
 }
 
-fn translate_aggrecate_one_arg<T: FnOnce(Expr) -> Aggregate>(
```
ever0de, line 108: good catch! thanks

core/src/ast_builder/expr/aggregate.rs
```diff
@@ -1,25 +1,76 @@
...
+                } else {
+                    CountExprNode::Text(CountArgExpr::Expr(Expr::Identifier(str)).to_sql())
+                }
+            }
+            _ => CountExprNode::Text(CountArgExpr::Wildcard.to_sql()),
+        }
+    }
+}
+
+impl TryFrom<CountExprNode> for CountArgExpr {
```
panarch, line 47: AST target enum name is `CountArgExpr`, then by only adding postfix `Node`, it would be clear to understand with `CountArgExprNode`.

core/src/ast_builder/expr/aggregate.rs
```diff
@@ -1,25 +1,76 @@
...
     Min(ExprNode),
     Max(ExprNode),
     Avg(ExprNode),
     Variance(ExprNode),
 }
 
+#[derive(Clone)]
+pub enum CountExprNode {
+    Text(String),
+}
```
panarch, line 24: You can consider to have one more enum type
```rust
pub enum CountExprNode {
    Text(String),
    Expr(ExprNode),
}
```
Then, we don't need to use `to_sql` conversion

core/src/ast_builder/expr/aggregate.rs
```diff
@@ -1,25 +1,76 @@
...
+        match expr_node {
+            ExprNode::Identifier(str) => {
+                if str == *"*" {
+                    CountExprNode::Text(CountArgExpr::Wildcard.to_sql())
+                } else {
+                    CountExprNode::Text(CountArgExpr::Expr(Expr::Identifier(str)).to_sql())
+                }
+            }
+            _ => CountExprNode::Text(CountArgExpr::Wildcard.to_sql()),
+        }
```
panarch, line 43: By adding `Expr(ExprNode)` to CountArgExprNode,
this can become quite simple.
```rust
fn from(expr_node: ExprNode) -> Self{
    Self::Expr(expr_node)
}
```

panarch (changes requested):

panarch (approved):
Looks all great, awesome... 👍 
Thanks a lot, merging!!

PR #659, opened 2022-07-20
Enhance `Values List` to validate column type

devgony:
This is subsequent PR of #648 

## Goal
- `Values List` does not have schema so that there is no minimum validation for each column types yet
- If there is no `NULL`, the first row will create column types
- If there is `NULL` at first row, the next row having type will create colum type

## Todo
- [x] get mutable `column_types` base
- [x] get indexes of None
- [x] if new row has type at those indexes, renew `column_types`
- [x] if there is no None in `column_types`, break
- [x] validate and cast type with try_info_value

## In Next PR
- Handle when specific column has NULL at all rows
  - It seems like postgres also converts NULL to TEXT for accomplishing `CreateTableAsValues`
```sql
postgres=# create table tb_values as values (1, null, null), (2, null, 3);
SELECT 2
postgres=# select column_name, data_type from information_schema.columns where table_name = 'tb_values';
 column_name | data_type 
-------------+-----------
 column1     | integer
 column2     | text
 column3     | integer
``` 
-  refactor `values_list into row` with row::new
  - split common codes
  - blend `column_defs` from `column_types`, `labels`

### DIFF
--- a/core/src/executor/select/mod.rs
+++ b/core/src/executor/select/mod.rs
@@ -139,7 +139,7 @@ pub async fn select_with_labels<'a>(
                 .map(|expr| {
                     evaluate_stateless(None, expr)
                         .and_then(|evaluated| evaluated.try_into())
-                        .and_then(|value: Value| Ok(value.get_type()))
+                        .map(|value: Value| value.get_type())
                 })
                 .collect::<Result<Vec<_>>>()?;
 
@@ -154,7 +154,7 @@ pub async fn select_with_labels<'a>(
 
             for exprs in values_list {
                 let null_indexes = get_null_indexes(&column_types);
-                if null_indexes.len() == 0 {
+                if null_indexes.is_empty() {
                     break;
                 };
                 exprs
--- a/core/src/executor/select/mod.rs
+++ b/core/src/executor/select/mod.rs
@@ -183,8 +183,8 @@ pub async fn select_with_labels<'a>(
                     }
                     let values = exprs
                         .iter()
+                        .map(|expr| evaluate_stateless(None, expr))
                         .zip(column_types.iter())
-                        .map(|(expr, column_type)| (evaluate_stateless(None, expr), column_type))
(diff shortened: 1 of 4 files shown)


core/src/executor/select/mod.rs
```diff
@@ -131,6 +131,44 @@ pub async fn select_with_labels<'a>(
...
+                    .try_for_each(|(i, expr)| -> Result<()> {
+                        let value: Value = evaluate_stateless(None, expr)?.try_into()?;
+                        if let Some(data_type) = value.get_type() {
+                            column_types[i] = Some(data_type);
+                        }
+
+                        Ok(())
+                    })?;
+            }
+
```
ever0de, line 171:
```rust
            let rows = values_list.into_iter().scan(
                iter::repeat(None)
                    .take(first_len)
                    .collect::<Vec<Option<DataType>>>(),
                move |column_types, exprs| {
                    if exprs.len() != first_len {
                        return Some(Err(RowError::NumberOfValuesDifferent.into()));
                    }

                    let values = column_types
                        .iter_mut()
                        .zip(exprs.iter())
                        .map(|(column_type, expr)| -> Result<_> {
                            let evaluated = evaluate_stateless(None, expr)?;

                            let value = match column_type {
                                Some(data_type) => evaluated.try_into_value(data_type, true)?,
                                None => {
                                    let value: Value = evaluated.try_into()?;
                                    *column_type = value.get_type();

                                    value
                                }
                            };

                            Ok(value)
                        })
                        .collect::<Result<Vec<_>>>()
                        .map(Row);

                    Some(values)
                },
            );
```

Since it runs on all `values_list` below, I think it can be integrated.
This eliminates pseudo `unwrap` syntax, such as `[index]` access, and eliminates duplicate execution.

How about this code?

ever0de (approved):
Looks good 👍

panarch (approved):
Awesome!! Looks great, thanks really a lot 👍 👍 👍

PR #672, opened 2022-07-23
Implement AST-builder between expression function

### LINKED ISSUE #671: [AST Builder] Implement between expression function
## Description

Implement `between` and `not_between` in ast_builder

## Task

```rust
fn between() {
    let actual = col("num").between(1, 10);
    let expected = "num BETWEEN 1 AND 10";
    test_expr(actual, expected);

    let actual = col("date").between(text("2022-01-01"), text("2023-01-01"));
    let expected = "date BETWEEN '2022-01-01' AND '2023-01-01'";
    test_expr(actual, expected);

    let actual = col("num").not_between(1, 10);
    let expected = "num NOT BETWEEN 1 AND 10";
    test_expr(actual, expected);

    let actual = col("date").not_between(text("2022-01-01"), text("2023-01-01"));
    let expected = "date NOT BETWEEN '2022-01-01' AND '2023-01-01'";
    test_expr(actual, expected);
}
```

CEOJINSUNG:
## Issue
Resolves #671 

## Description
Implement between expression function at ast_builder


core/src/ast_builder/expr/between.rs
```diff
@@ -0,0 +1,36 @@
+use super::ExprNode;
+
+impl ExprNode {
+    pub fn between<T: Into<Self>>(self, negated: bool, low: T, high: T) -> Self {
```
panarch, line 4: How about splitting `between` into two - `between` and `not_between`?
Then we don't need to take `negated` as a param.
Then...
```rust
let actual = col("num").between(1, 10);
let expected = "num BETWEEN 1 AND 10";
test_expr(actual, expected);
```

```rust
let actual = col("num").not_between(1, 10);
let expected = "num NOT BETWEEN 1 AND 10";
test_expr(actual, expected);
```

CEOJINSUNG: There was a parameter called negated , so I thought I should get this as a value as well. As you said, I will modify the function and make a Pull Request again later. Thanks a lot

CEOJINSUNG:
According to the existing requirements, it has been divided into two functions, the 'between' function and the 'not_between' function.

core/src/ast_builder/expr/between.rs
CEOJINSUNG: According to the existing requirements, it has been divided into two functions, the 'between' function and the 'not_between' function.

panarch (approved):
Looks all great, thanks a lot! merging 👍

Issue #652, opened 2022-07-16
[AST Builder] Implement Index functions


seonghun-dev:
# Description
Implements ast builders for [`Index`](https://www.w3schools.com/sql/sql_create_index.asp) functions.

## Background
Refered https://github.com/gluesql/gluesql/issues/614

## Task
In AST Builder module to create, show and delete index easier

- [x] 1. [`create_index`](https://www.w3schools.com/sql/sql_create_index.asp) function
- [x] 2. [`drop_index`](https://www.w3schools.com/sql/sql_create_index.asp) function 

The following tests should be passed:

``` rust
    #[test]
    fn index() {
        let actual = table("Foo").create_index("name","nameIndex").build();
        let expected = "CREATE INDEX nameIndex ON Foo (name)";
        test(actual, expected);

        let actual = table("Foo").drop_index("nameIndex").build();
        let expected = "DROP INDEX Foo.nameIndex";
        test(actual, expected);

    }
```

PR #673, opened 2022-07-23
[AST Builder] Implement Index Expr

seonghun-dev:
## Issue
Resolve https://github.com/gluesql/gluesql/issues/652

## Description
Implement `index` expr at ast_builder

## Task
- [x] 1. [`create_index`](https://www.w3schools.com/sql/sql_create_index.asp) function
- [x] 2. [`drop_index`](https://www.w3schools.com/sql/sql_create_index.asp) function 
- [x] 3. `OrderByExprNode` node

### DIFF
--- a/core/src/ast_builder/index.rs
+++ b/core/src/ast_builder/index.rs
@@ -61,13 +61,11 @@ mod tests {
 
     #[test]
     fn create_index() {
-        let actual = table("Foo").create_index("nameIndex", "name", true).build();
+        let actual = table("Foo").create_index("nameIndex", "name asc").build();
         let expected = "CREATE INDEX nameIndex ON Foo (name Asc)";
         test(actual, expected);
 
-        let actual = table("Foo")
-            .create_index("nameIndex", "name", false)
-            .build();
+        let actual = table("Foo").create_index("nameIndex", "name desc").build();
         let expected = "CREATE INDEX nameIndex ON Foo (name Desc)";
         test(actual, expected);
     }
--- a/core/src/ast_builder/index.rs
+++ b/core/src/ast_builder/index.rs
@@ -1,41 +1,32 @@
 #![cfg(feature = "index")]
 
 use crate::{
-    ast::{Expr, ObjectName, OrderByExpr, Statement},
+    ast::{ObjectName, Statement},
     result::Result,
 };
+
+use super::OrderByExprNode;
 
 #[derive(Clone)]
 pub struct CreateIndexNode {
     name: String,
     table_name: String,
-    column: String,
-    asc: bool,
+    column: OrderByExprNode,
 }
 
 impl CreateIndexNode {
-    pub fn new(table_name: String, name: String) -> Self {
+    pub fn new(table_name: String, name: String, column: OrderByExprNode) -> Self {
         Self {
             table_name,
             name,
-            column: String::new(),
-            asc: true,
+            column,
         }
-    }
-
-    pub fn column(mut self, column: &str, asc: bool) -> Self {
-        self.column = column.to_string();
-        self.asc = asc;
-        self
     }
 
     pub fn build(self) -> Result<Statement> {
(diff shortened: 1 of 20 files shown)


core/src/ast_builder/index.rs
```diff
@@ -0,0 +1,94 @@
...
+    ast::{Expr, ObjectName, OrderByExpr, Statement},
+    result::Result,
+};
+
+#[derive(Clone)]
+pub struct CreateIndexNode {
+    name: String,
+    table_name: String,
+    column: String,
+    asc: bool,
```
panarch, line 13: https://github.com/gluesql/gluesql/blob/0d2ecc5f2b00a006b1e84417de21c3b8388d3859/core/src/ast/mod.rs#L82-L87

type of field `column` in `Statement::CreateIndex` is `OrderByExpr`

https://github.com/gluesql/gluesql/blob/0d2ecc5f2b00a006b1e84417de21c3b8388d3859/core/src/ast/query.rs#L102-L106

Then, here what we need to do is adding a new node - `OrderByExprNode`.
One expected behavior of `OrderByExprNode` is taking `"id + name` or `col1 || col2 DESC` like strings and converts into `ast::OrderByExpr`.

core/src/ast_builder/index.rs
```diff
@@ -0,0 +1,94 @@
...
+
+#[cfg(all(test, feature = "index"))]
+mod tests {
+    use crate::ast_builder::{table, test};
+
+    #[test]
+    fn create_index() {
+        let actual = table("Foo")
+            .create_index("nameIndex")
+            .column("name", true)
```
panarch, line 75: Here it would be better to combine `create_index` and `column` because there is no use case of using `column` several times.
e.g.
In current structure, this code works without error
```rust
let actual = table("Foo")
    .create_index("test")
    .column("id", true)
    .column("name", true)
    .build();
```
`.column("name", true)` simply overwrites `.column("id", true)`

core/src/ast_builder/mod.rs
```diff
@@ -20,6 +22,8 @@ pub use {
 
 /// Available expression builder functions
 pub use expr::{col, expr, nested, num, text, ExprNode};
+#[cfg(feature = "index")]
+pub use {index::CreateIndexNode, index::DropIndexNode};
```
panarch, line 26: Looks really nice that you've handled `#[cfg(feature = "index")]` properly 👍

core/src/ast_builder/order_by_expr.rs
```diff
@@ -0,0 +1,82 @@
...
+        ast::{Expr, OrderByExpr},
+        parse_sql::parse_order_by_expr,
+        result::{Error, Result},
+        translate::translate_order_by_expr,
+    },
+};
+
+#[derive(Clone)]
+pub enum OrderByExprNode {
+    Expr(ExprNode, Option<bool>),
```
panarch, line 13: How about starting with minimal spec?
It would be enough to start only with `OrderByExprNode::Text`.

```rust
Expr(ExprNode, Option<bool>),
```
This is exposed for ast-builder users and it is expected for them to use this.
Currently to use `Expr(ExprNode, Option<bool>)`, user needs to use something like this
```rust
let actual = (col("id"), Some(true)).into()
```
but I'm not sure that `Some(true)` would be simple enough for users to get it.
So.. I would like to suggest that we make this PR only including `OrderByExprNode::Text` and we can discuss `Expr` like improvement design in other issue or PR.

panarch (changes requested):
Awesome 👍 
There are some comments I left, thanks a lot!

fyi.
image
This coverage report would be helpful to check uncovered lines.
https://coveralls.io/builds/51284545/source?filename=core%2Fsrc%2Fast_builder%2Forder_by_expr.rs#L18

panarch (approved):
All looks great! Thanks a lot 👍

PR #675, opened 2022-07-23
[Data] Add function `sqrt` for enum `Value`

zmrdltl:
## Description
Implement to support evaluating sqrt between enum Value

https://doc.rust-lang.org/std/primitive.f64.html
- For real numbers less than or equal to 0, std::f64::sqrt() returns NaN. Therefore, for mistakes, the sqrt part does not process errors.
- It is not an error to pass real and null as arguments to the sqrt function, and non-numeric errors are handled for the rest of the types.
- Meanwhile, real numbers less than 0 are implemented to indicate an error in the result of the sql sqrt function.
  + This can result in having to handle error for NaN.
However, since NaN does not come as a result of expr parsing, it is independent of sqrt operations between enum `Values`.

### DIFF
--- a/core/src/data/value/error.rs
+++ b/core/src/data/value/error.rs
@@ -158,6 +158,12 @@ pub enum ValueError {
         rhs: Value,
         operator: NumericBinaryOperator,
     },
+
+    #[error("non numeric value in sqrt")]
+    SqrtOnNonNumeric,
+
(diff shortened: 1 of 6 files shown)


panarch (approved):
All looks great, thanks a lot! Merging 👍

Issue #679, opened 2022-07-25
[AST Builder] Implement `ln` function 


gimwonbae:
## Description
implement `ln` function in `ast_builder`

## Test to pass

``` rust
    #[test]
    fn function_ln() {
        let actual = ln(num(2));
        let expected = "LN(2)";
        test_expr(actual, expected);

        let actual = num(2).ln();
        let expected = "LN(2)";
        test_expr(actual, expected);
    }
```

PR #680, opened 2022-07-25
Implement `ln` in ast_builder

gimwonbae:
### Description
implement `ln` function in ast_builder

### Issue https://github.com/gluesql/gluesql/issues/679


devgony:
If there is connected issue, you can write like `resolve #issueNo` which will connect and close issue automatically 👍

PR #682, opened 2022-07-27
Convert `TryInto` traits into `TryFrom`

### LINKED ISSUE #619: Change the implementation of `*Into` to `*From`
<https://doc.rust-lang.org/stable/std/convert/trait.Into.html>

One should avoid implementing [Into](https://doc.rust-lang.org/stable/std/convert/trait.Into.html) and implement [From](https://doc.rust-lang.org/stable/std/convert/trait.From.html) instead. Implementing [From](https://doc.rust-lang.org/stable/std/convert/trait.From.html) automatically provides one with an implementation of [Into](https://doc.rust-lang.org/stable/std/convert/trait.Into.html) thanks to the blanket implementation in the standard library.

nyeong:
It resolves #619.

- changes 13 `TryInto` for `&Values` in `data/value/into.rs`.
- implements 8 `TryFrom` for `Values` in `data/value/into.rs` using macro.
- add tests for `TryFrom` in `data/value/into.rs`.
- changes 2 `TryInto` for `Evaluated<'_>` in `executor/evaluate/evaluated.rs`.

### DIFF
--- a/core/src/data/value/convert.rs
+++ b/core/src/data/value/convert.rs
@@ -46,11 +46,11 @@ impl From<Value> for String {
     }
 }
 
-impl TryInto<bool> for &Value {
+impl TryFrom<&Value> for bool {
     type Error = Error;
 
-    fn try_into(self) -> Result<bool> {
-        Ok(match self {
+    fn try_from(v: &Value) -> Result<Self> {
+        Ok(match v {
             Value::Bool(value) => *value,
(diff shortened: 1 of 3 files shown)


MRGRAVITY817 (approved):
LGTM 👍

panarch (approved):
Awesome, so good to see all the changes.
All looks great, thanks a lot! 👍

Issue #609, opened 2022-06-18
Support `STDEV` aggregate function


zmrdltl:
last PR : https://github.com/gluesql/gluesql/pull/598
ref : https://www.ibm.com/docs/ko/db2/11.1?topic=functions-stddev

The standard deviation is the square root of the variance. Because of this, the following happens: What should I do?
![image](https://user-images.githubusercontent.com/22141521/174432087-2ce428a9-e378-4437-8bea-2ebabd08a74a.png)

We need a way to represent an infinite decimal point.

PR #684, opened 2022-07-27
Add `STDEV` aggregate funtion

zmrdltl:
resolved : https://github.com/gluesql/gluesql/issues/609
Added `STDEV` function to ast_builder and executor.

### DIFF
--- a/core/src/executor/aggregate/state.rs
+++ b/core/src/executor/aggregate/state.rs
@@ -139,13 +139,12 @@ impl<'a> AggrValue {
                 sum_square,
                 sum,
                 count,
-            } => {
-                let sum_expr1 = sum_square.multiply(&Value::I64(count))?;
-                let sum_expr2 = sum.multiply(&sum)?;
-                let expr_sub = sum_expr1.subtract(&sum_expr2)?;
-                let cnt_square = Value::F64(count as f64).multiply(&Value::F64(count as f64))?;
-                expr_sub.divide(&cnt_square)?.sqrt()
-            }
+            } => Self::export(Self::Variance {
+                sum_square,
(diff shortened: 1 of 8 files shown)


core/src/executor/aggregate/state.rs
```diff
@@ -116,6 +135,17 @@ impl<'a> AggrValue {
...
+                sum_square,
+                sum,
+                count,
+            } => {
+                let sum_expr1 = sum_square.multiply(&Value::I64(count))?;
+                let sum_expr2 = sum.multiply(&sum)?;
+                let expr_sub = sum_expr1.subtract(&sum_expr2)?;
+                let cnt_square = Value::F64(count as f64).multiply(&Value::F64(count as f64))?;
+                expr_sub.divide(&cnt_square)?.sqrt()
+            }
```
devgony, line 148: Looks like you can try recursive `export` to reuse Variance
```rs
Self::Stdev {                                                                                                                               [0/1387]
                    sum_square,
                    sum,
                    count,
               } => Self::export(Self::Variance {
                   sum_square,
                   sum,
                   count,
               })?
               .sqrt()
```

devgony (approved):
LGTM! but i left single comment to reuse previouse code

core/src/executor/aggregate/state.rs
panarch: @devgony That's a really good point, thanks!
Then, how about adding `variance` local function or closure inside `Self::export` and let both `Self::Variance` and `Self::Stdev` to use it?

panarch (approved):
Awesome... 👍 👍 
Thanks a lot!!

PR #687, opened 2022-07-28
Implement primary key support,

panarch:
- [x] Adding `fetch_data` to `Store` trait
- [x] Expanding `ast::IndexItem`
- [x] Execution layer - select
- [x] Execution layer - insert
- [x] Execution layer - validate
- [x] Query planner
- [x] Integration tests
- [x] Planner unit tests


core/src/plan/join.rs
```diff
@@ -24,34 +24,30 @@ pub fn plan(schema_map: &HashMap<String, Schema>, statement: Statement) -> State
...
-            SetExpr::Values(_) => body,
+            SetExpr::Values(_) => query.body,
         };
 
-        Query {
-            body,
-            limit,
-            offset,
-        }
+        query
```
ever0de, line 42:
```suggestion
    fn query(&self, outer_context: Option<Rc<Context<'a>>>, query: Query) -> Query {
        let body = match query.body {
            SetExpr::Select(select) => {
                let select = self.select(outer_context, *select);

                SetExpr::Select(Box::new(select))
            }
            SetExpr::Values(_) => query.body,
        };

        Query { body, ..query }
    }
```

panarch: Wow, I didn't know this kind of assignment working. thanks a lot!

core/src/plan/primary_key.rs
```diff
@@ -0,0 +1,577 @@
...
+                    },
+                    join_operator: JoinOperator::Inner(JoinConstraint::None),
+                    join_executor: JoinExecutor::NestedLoop,
+                }],
+            },
+            selection: None,
+            group_by: Vec::new(),
+            having: None,
+            order_by: Vec::new(),
+        });
```
ever0de, line 424: It looks that current_context should contain join relation tables, too

ever0de (approved):
This looks... very very... good.
Thanks to you, I learned a lot about PK. thank you

Issue #693, opened 2022-07-30
[AST Builder] Implement drop_table Expression


seonghun-dev:
# Description
Implements ast builders for [`drop_table`](https://www.w3schools.com/sql/sql_drop_table.asp) Expression.

## Background
Refered https://github.com/gluesql/gluesql/issues/614

## Task
In AST Builder module to [`drop_table`](https://www.w3schools.com/sql/sql_drop_table.asp)  Expression to drop table easier

The following tests should be passed:

``` rust
    #[test]
    fn drop_table() {
        let actual = table("Foo").drop().build();
        let expected = "DROP TABLE Foo";
        test(actual, expected);

        let actual = table("Foo").drop_table().build();
        let expected = "DROP TABLE Foo";
        test(actual, expected);

        let actual = table("Foo").drop_exists().build();
        let expected = "DROP TABLE IF EXISTS Foo";
        test(actual, expected);

        let actual = table("Foo").drop_exists_table().build();
        let expected = "DROP TABLE IF EXISTS Foo";
        test(actual, expected);

    }
```

PR #695, opened 2022-07-30
[AST Builder] Implement drop_table Expr

seonghun-dev:
## Issue
Resolve https://github.com/gluesql/gluesql/issues/693

## Description
Implement `drop table` expr at ast_builder

### DIFF
--- a/core/src/ast_builder/drop_table.rs
+++ b/core/src/ast_builder/drop_table.rs
@@ -31,19 +31,11 @@ mod tests {
 
     #[test]
     fn drop_table() {
-        let actual = table("Foo").drop().build();
-        let expected = "DROP TABLE Foo";
-        test(actual, expected);
-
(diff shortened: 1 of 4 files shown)


panarch (commented):
It looks that we can choose only one rather than supporting multiple options - `drop()` | `drop_table()`

How about these names only?
`table("Foo").drop().build()`
`table("Foo").drop_if_exists().build()`

seonghun-dev:
> It looks that we can choose only one rather than supporting multiple options - `drop()` | `drop_table()`
> 
> How about these names only? `table("Foo").drop().build()` `table("Foo").drop_if_exists().build()`

Change to Support one option `drop_table()` | `drop_table_if_exists()`

panarch (approved):
Thanks a lot! Looks all good, merging 👍 👍

PR #647, opened 2022-07-14
Update coverage action to use grcov & Coveralls

panarch:
Two replacements
- tarpaulin to grcov
- Codecov to Coveralls

ref. https://coveralls.io/github/gluesql/gluesql

resolve #582

PR #699, opened 2022-07-31
Run github action job parallel

24seconds:
## Description
Current github action job takes 10~15 minutes to finish. By separating the jobs, we can run them in parallel.
I took a experiment and it seem that it takes 4 ~ 7 minutes.

I separated the jobs as below
- Clippy (root)
- Clippy (core)
- Rustfmt
- Build
- Build (gluesql-js)
- Run tests
- Run examples

github action experimental running example link:
https://github.com/24seconds/gluesql/actions/runs/2768504708

### Benefits
- Running time reduced (10~15 min -> 4 ~7 min)
- If a job failed, we can re-run it separately
Screenshot from 2022-07-31 15-38-13

## Review point
- used [toolchain action](https://github.com/actions-rs/toolchain) for installing `rustfmt` and `clippy`
- dirty commit history?
  - I found that it is quite hard to test workflow - even triggering the job is impossible because workflow is triggered based on main branch `push` or `pull_request`. So I made several commits. If it seems not okay, then please suggest some good ways to look commit history pretty


panarch:
> dirty commit history?

I also have no better answer for this.
What I've done a few weeks before.. :D 
ref. https://github.com/gluesql/gluesql/pull/647

24seconds:
Seems it runs well - https://github.com/gluesql/gluesql/actions/runs/2768537010

panarch:
> Seems it runs well - https://github.com/gluesql/gluesql/actions/runs/2768537010
> 
> `Rust tests` is the bottleneck but I have no idea how to reduce it. Maybe 5minutes is enough for now?

I think that the current duration is good enough because
image
There exists another task which takes longer than any other `Rust / ` jobs.

panarch (approved):
This is also meaningful improvement. Thanks a lot!!

Issue #667, opened 2022-07-23
Add Assignees automatically when someone creates PR


24seconds:
## Description

Currently in the gluesql when the PR is created by someone - let say Bob, then gluesql maintainers manually assign the Bob as `Assignees` section - which is quite cumbersome. So it is good to let bot add assignees. I remember there was a github action which exactly supports this.

garimellasrisudha:
yes i have idea on this we can do this in settings in codeowners part .
in project setting we can add codeowners and in branch setting we tick that codeowners

24seconds:
> yes i have idea on this we can do this in settings in codeowners part . in project setting we can add codeowners and in branch setting we tick that codeowners. As i can't able to see settings near repo because i don't have access

Ah, thank you for the comment. As I understand what you are talking about is [CODEOWNER](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#about-code-owners). When we set the CODEOWNER then the person who has a responsibility about specified path in CODEOWNER file will be added as reviewer automatically.

What I want to do is to add `pr writer` automatically as Assignees - see the image below. So it's quite different I think
![image](https://user-images.githubusercontent.com/34280965/180593577-486cd9a7-31d2-42f1-bbe5-1a007ab1fccd.png)

gimwonbae:
Can I work on this issue? Kindly assign it to me!

PR #701, opened 2022-08-01
Add Assignees automatically when someone creates PR (#667)

gimwonbae:
### Description
Add Assignees automatically when someone creates PR using github action

### Issue
Resolve https://github.com/gluesql/gluesql/issues/667

### DIFF
--- a/.github/workflows/assign-pr.yml
+++ b/.github/workflows/assign-pr.yml
@@ -1,0 +1,18 @@
+name: Assign a PR to the person who created it
+
+on:
+  pull_request_target:
+    branches: [ "main" ]
+
+jobs:
+  assign-pr:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Add assignees to an PR
+        run: |
+          HTTP_RESP_CODE=$(curl -o /dev/null  -w "%{http_code}" -X POST -H "Accept: application/vnd.github+json" -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" https://api.github.com/repos/${{ github.repository }}/issues/${{ github.event.number }}/assignees -d '{"assignees":["${{ github.event.sender.login }}"]}')
+          if [[ $HTTP_RESP_CODE != 201 ]]; then
+            echo "Add assignees to an PR failed ($HTTP_RESP_CODE)"
+            exit 1
+          fi


24seconds:
Let me review this pr

.github/workflows/assign-pr.yml
```diff
@@ -0,0 +1,18 @@
...
+    branches: [ "main" ]
+
+jobs:
+  assign-pr:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Add assignees to an PR
+        run: |
+          HTTP_RESP_CODE=$(curl -o /dev/null  -w "%{http_code}" -X POST -H "Accept: application/vnd.github+json" -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" https://api.github.com/repos/${{ github.repository }}/issues/${{ github.event.number }}/assignees -d '{"assignees":["${{ github.event.sender.login }}"]}')
```
24seconds, line 14: [Suggestion and IMO]
It seems this code is directly calling the github API. For the maintenance and sustainability, how about using a github action which is released in market place?  
For example, [Assign-author](https://github.com/marketplace/actions/assign-author) or [Auto author assign](https://github.com/marketplace/actions/auto-author-assign)? It seems it is easy to use and provides a bit more features.

24seconds (commented):
Thank you for the work. I left a comment. Could you take a look?

.github/workflows/assign-pr.yml
gimwonbae: For complex or multi-step tasks like docker build/push, I agree that you're better off using marketplace actions as you said.

However, I'm not sure what the maintenance advantage is of using an external action for a simple task that can be executed with one simple api call.

Using the Github API is as scalable and easy as using an external action in the marketplace, and it seems more maintainable than having a dependency on an external repo with fewer stars and less reliability.

Please let me know if there are any advantages to using external actions that I've missed 😀.

24seconds: Thank you for the opinion.  

I think using existing solution - github action I mean - has two advantages.  

1. We don't need to know about the details.
By using github action, we don't need to care much about small changes. I believe they - github action owner - will follow up the problems, bugs or some changes on behalf of us. (For example, github API's behavior change) 
What we need to do is just updating the github action version. But this is based on the assumption that the owner would maintain the project consistently. 😅 

2. We don't need to care about shell script proficiency.
Maybe it is my bias but I think nowadays people are not so much familiar with shell script code. So it would be a bit hard to maintain the job because reviewer and pr author should know about it. So.. I think it's good to use github action. Well, in this case because it is so simple this is not a big problem. 😄 

To sum up, we agreed the problem is so simple. But the action was different.
- Because it's so simple it's okay to use github action
- Because it's so simple it's okay to use simple script
Therefore I think it's okay to choose either way. So it's own this repo owner's choice. Thank you for the comment again!

cc: @panarch

gimwonbae: Thanks for your comments!

p.s. REST API has backward compatibility, so even if the Github API version is changed, there seems to be no side effect.

24seconds: @panarch could you check this pr?

ever0de: I'm not a maintainer, but I leave comments.
1. We may need scalability. At this time, I think it can be easily added using the functions of github-action. (For example, designate a reviewer)
2. If a REST call fails, functions such as retry attempts will come more intuitively when using a github action rather than writing a shell script.

PR #702, opened 2022-08-02
Mark flaky test using target_os cfg

24seconds:
### Problem and Suggestion
tl;dr
When we run `cargo test` on macOS, it's highly likely failed. This is because it takes some time to acquire write lock.  Therefore, for the macOS users, it is good to mark flaky tests using `target_os` cfg to not to be run in macOS.

#### explanation
1. In the Gluesql layer, when we try to acquire a lock internal storage - sled storage in this case - tries to acquire a lock. But it 
2. This problem can be reproduced easily. 
```sh
cargo test -p gluesql_sled_storage --test sled_transaction
```
3. The main root cause was `sled-storage` takes a lot of time to acquire the lock. Because lock is RWlock, it waits for readers to release the read lock.
4. Why it takes so long in macOS? -> I guess the lock implementation is different by each os platform. `sled` package uses `parking_lot` package and inside the `parking_lot`, it seems that lock is implemented with futex and waitAddress in Linux and Windows. But macOS uses mutex and condition variable.
5. Why the test failed frequently especially when the tests are run in parallel? -> In side the sled-storage, many threads are created. And additional threads for running tests are created. Therefore I suspect that running test parallel creates so many threads and it leads the thread scheduling taking much time on macOS.

But this explanation could be wrong. Because due to the lack of my debugging skill, I am not sure how exactly works.
The one truth is that the problem is not related to gluesql and the best effort to solve this is to marking them properly

### DIFF
--- a/storages/sled-storage/tests/sled_transaction.rs
+++ b/storages/sled-storage/tests/sled_transaction.rs
@@ -411,6 +411,7 @@ async fn sled_transaction_gc() {
 const TX_TIMEOUT: Option<u128> = Some(200);
 const TX_SLEEP_TICK: Duration = Duration::from_millis(201);
 
+#[cfg(not(any(target_os = "macos", target_os = "ios")))]
 #[tokio::test]
 async fn sled_transaction_timeout_store() {
     let path = &format!("{}/transaction_timeout_store", PATH_PREFIX);
@@ -542,6 +543,7 @@ async fn sled_transaction_timeout_store() {
     );
 }
 
+#[cfg(not(any(target_os = "macos", target_os = "ios")))]
 #[tokio::test]
 async fn sled_transaction_timeout_alter() {
     let path = &format!("{}/transaction_timeout_alter", PATH_PREFIX);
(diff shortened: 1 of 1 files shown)


ever0de (approved):
That's a perfect explanation. However, it is still unfortunate that CI can sometimes fail.
If I add ignore attribute macro and execute each test function in order (separately) in CI, is success guaranteed?

(Accept the current changes.)

24seconds:
> That's a perfect explanation. However, it is still unfortunate that CI can sometimes fail. If I add ignore attribute macro and execute each test function in order (separately) in CI, is success guaranteed?
> 
> (Accept the current changes.)

Thank you for the comment. Yes, as I understand running test sequential would drop the failure change hugely. Next pr, let me add `cfg ignore` and modify the job - let ignored tests be run and the job run flaky tests in sequence

panarch (approved):
Awesome, this can make contributors happy who uses MacOS. Thanks a lot!!

PR #704, opened 2022-08-03
Support CTAV(Create Table As Values)

devgony:
This is subsequent PR of #659 
### Like Below, we will be able to create table with `Values List` !
```sql
postgres=# CREATE TABLE gluesql AS VALUES (1, null, TRUE, null), (2, 'b', FALSE, null);
SELECT 2
postgres=# select table_name, column_name, data_type from information_schema.columns where table_name = 'gluesql';
 table_name | column_name | data_type
------------+-------------+-----------
 gluesql    | column1     | integer
 gluesql    | column2     | text
 gluesql    | column3     | boolean
 gluesql    | column4     | text
```

## Todo
- [x] Create column names
Currently `CTAV` command runs with some bug it doesn't have column names, column types
```sql
gluesql> CREATE TABLE gluesql AS VALUES (1, null, TRUE, null), (2, 'b', FALSE, null);
gluesql> SELECT * FROM gluesql;
╭─────────────────────────╮
│                         │
╞═════════════════════════╡
│ 1   NULL   TRUE    NULL │
│ 2   b      FALSE   NULL │
╰─────────────────────────╯
```
- [x] Create column types
To create column types, we should know explicit type of all columns.
If all values are NULL in a specific column, we need a policy to get explicit type.
Note the last line of postgres example.
They converted NULL as TEXT
```sql
 gluesql    | column4     | text
```
  - [x] Convert NULL as TEXT after evaluating all Values

### DIFF
--- a/core/src/executor/alter/table.rs
+++ b/core/src/executor/alter/table.rs
@@ -1,9 +1,10 @@
-use crate::ast::{ColumnOption, ColumnOptionDef};
-
 use {
     super::{validate, AlterError},
     crate::{
-        ast::{ColumnDef, ObjectName, Query, SetExpr, TableFactor, Values},
+        ast::{
+            ColumnDef, ColumnOption, ColumnOptionDef, ObjectName, Query, SetExpr, TableFactor,
+            Values,
+        },
         data::{get_name, Schema, TableError},
         executor::{evaluate_stateless, select::select},
         prelude::{DataType, Value},
--- a/core/src/executor/alter/table.rs
+++ b/core/src/executor/alter/table.rs
@@ -8,7 +8,7 @@ use {
         data::{get_name, Schema, TableError},
         executor::{evaluate_stateless, select::select},
         prelude::{DataType, Value},
-        result::{Error, MutResult, Result, TrySelf},
+        result::{Error, IntoControlFlow, MutResult, Result, TrySelf},
         store::{GStore, GStoreMut},
     },
     futures::stream::{self, TryStreamExt},
@@ -55,40 +55,29 @@ pub async fn create_table<T: GStore + GStoreMut>(
                     let column_types =
                         values_list
                             .iter()
-                            .try_fold(Ok(init_types), |column_types, exprs| {
-                                let result = column_types.and_then(
-                                    |column_types| -> Result<(Vec<Option<DataType>>, bool)> {
-                                        let column_types = column_types
-                                            .iter()
-                                            .zip(exprs.iter())
-                                            .map(|(column_type, expr)| -> Result<_> {
-                                                let column_type = match column_type {
-                                                    Some(data_type) => Some(data_type.to_owned()),
-                                                    None => evaluate_stateless(None, expr)
-                                                        .and_then(Value::try_from)?
-                                                        .get_type(),
-                                                };
+                            .try_fold(init_types, |column_types, exprs| {
+                                let column_types = column_types
+                                    .iter()
+                                    .zip(exprs.iter())
+                                    .map(|(column_type, expr)| match column_type {
+                                        Some(data_type) => Ok(Some(data_type.to_owned())),
+                                        None => evaluate_stateless(None, expr)
+                                            .and_then(Value::try_from)
+                                            .map(|value| value.get_type()),
+                                    })
+                                    .collect::<Result<Vec<Option<DataType>>>>()
+                                    .into_control_flow()?;
 
-                                                Ok(column_type)
-                                            })
-                                            .collect::<Result<Vec<Option<DataType>>>>()?;
-
-                                        let has_none = column_types.iter().any(Option::is_none);
-
-                                        Ok((column_types, has_none))
-                                    },
-                                );
-
-                                match result {
-                                    Ok((column_types, true)) => Continue(Ok(column_types)),
-                                    Ok((column_types, false)) => Break(Ok(column_types)),
-                                    Err(error) => Break(Err(error)),
+                                match column_types.iter().any(Option::is_none) {
+                                    true => Continue(column_types),
+                                    false => Break(Ok(column_types)),
                                 }
                             });
                     let column_types = match column_types {
-                        Continue(column_types) | Break(column_types) => column_types,
+                        Continue(column_types) => column_types,
+                        Break(column_types) => column_types?,
                     };
-                    let column_defs = column_types?
+                    let column_defs = column_types
                         .iter()
                         .map(|column_type| match column_type {
                             Some(column_type) => column_type.to_owned(),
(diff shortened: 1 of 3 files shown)


core/src/executor/alter/table.rs
```diff
@@ -20,28 +28,91 @@ pub async fn create_table<T: GStore + GStoreMut>(
...
+                                let has_none = column_types
+                                    .as_ref()
+                                    .unwrap()
+                                    .iter()
+                                    .any(|column_type| column_type.is_none());
+
+                                match has_none {
+                                    true => Continue(column_types),
+                                    false => Done(column_types),
+                                }
```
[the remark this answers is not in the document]
ever0de: How to get rid of unwrap.
In case of Error, try of Result does not occur, just wrap it with Done with match statement.

test-suite/src/values.rs
```diff
@@ -87,8 +85,33 @@ test_case!(values, async move {
...
+        &[
+            DataType::Int,
+            DataType::Text,
+            DataType::Boolean,
+            DataType::Int,
+            DataType::Text,
+        ],
+        "SELECT * FROM TableFromValues"
+    );
+});
```
panarch, line 117: It looks that `type_match` is not registered in integration tests, so it is not being tested.

devgony: Thank you for pointing out!
I just fixed that part but realized that it does not match schema but data.
Is there any test to match schema?

core/src/executor/alter/table.rs
```diff
@@ -20,28 +29,93 @@ pub async fn create_table<T: GStore + GStoreMut>(
...
+                                );
+
+                                match result {
+                                    Ok((column_types, true)) => Continue(Ok(column_types)),
+                                    Ok((column_types, false)) => Done(Ok(column_types)),
+                                    Err(error) => Done(Err(error)),
+                                }
+                            },
+                        )
+                        .into_inner();
```
panarch, line 96: Ok, let's reduce indentation depth little bit for making this easier to read.

```rust
let init_types = iter::repeat(None)
    .take(first_len)
    .collect::<Vec<Option<DataType>>>();
let column_types = values_list
    .iter()
    .fold_while(Ok(init_types), |column_types, exprs| {
        let result = column_types.and_then(|column_types| -> Result<_> {
            let column_types = column_types
                .iter()
                .zip(exprs.iter())
                .map(|(column_type, expr)| -> Result<_> {
                    let column_type = match column_type {
                        Some(data_type) => Some(data_type.to_owned()),
                        None => evaluate_stateless(None, expr)
                            .and_then(Value::try_from)?
                            .get_type(),
                    };

                    Ok(column_type)
                })
                .collect::<Result<Vec<Option<DataType>>>>()?;

            let has_none = column_types.iter().any(Option::is_none);

            Ok((column_types, has_none))
        });

        match result {
            Ok((column_types, true)) => Continue(Ok(column_types)),
            Ok((column_types, false)) => Done(Ok(column_types)),
            Err(error) => Done(Err(error)),
        }
    })
    .into_inner();
```

test-suite/src/values.rs
devgony: I substituted with `show columns from~` test !

core/src/executor/alter/table.rs
```diff
@@ -20,28 +29,85 @@ pub async fn create_table<T: GStore + GStoreMut>(
...
+                                },
+                            );
+
+                            match result {
+                                Ok((column_types, true)) => Continue(Ok(column_types)),
+                                Ok((column_types, false)) => Done(Ok(column_types)),
+                                Err(error) => Done(Err(error)),
+                            }
+                        })
+                        .into_inner();
```
panarch, line 88: Rather than using `itertools`, we can use native iterator methods to solve this.
```rust
// by using
use std::ops::ControlFlow::{Break, Continue};

// e.g.
let column_types =
    values_list
        .iter()
        .try_fold(Ok(init_types), |column_types, exprs| {
            let result = column_types.and_then(
                |column_types| -> Result<(Vec<Option<DataType>>, bool)> {
                    let column_types = column_types
                        .iter()
                        .zip(exprs.iter())
                        .map(|(column_type, expr)| -> Result<_> {
                            let column_type = match column_type {
                                Some(data_type) => Some(data_type.to_owned()),
                                None => evaluate_stateless(None, expr)
                                    .and_then(Value::try_from)?
                                    .get_type(),
                            };

                            Ok(column_type)
                        })
                        .collect::<Result<Vec<Option<DataType>>>>()?;

                    let has_none = column_types.iter().any(Option::is_none);

                    Ok((column_types, has_none))
                },
            );

            match result {
                Ok((column_types, true)) => Continue(Ok(column_types)),
                Ok((column_types, false)) => Break(Ok(column_types)),
                Err(error) => Break(Err(error)),
            }
        });

let column_types = match column_types {
    Continue(column_types) | Break(column_types) => column_types,
};
```

core/src/executor/alter/table.rs
```diff
@@ -20,28 +28,87 @@ pub async fn create_table<T: GStore + GStoreMut>(
...
+                                match result {
+                                    Ok((column_types, true)) => Continue(Ok(column_types)),
+                                    Ok((column_types, false)) => Break(Ok(column_types)),
+                                    Err(error) => Break(Err(error)),
+                                }
+                            });
+                    let column_types = match column_types {
+                        Continue(column_types) | Break(column_types) => column_types,
+                    };
+                    let column_defs = column_types?
```
panarch, line 91: If we add something like...
e.g.
```rust
// core/src/result.rs
pub trait IntoControlFlow<T> {
    fn into_control_flow(self) -> ControlFlow<Result<T>, T>;
}

impl<T> IntoControlFlow<T> for Result<T> {
    fn into_control_flow(self) -> ControlFlow<Result<T>, T> {
        match self {
            Ok(v) => ControlFlow::Continue(v),
            e => ControlFlow::Break(e),
        }
    }
}
```

Then, we imports `IntoControlFlow` in `alter/table.rs`
then...
e.g.
```rust
let column_types =
    values_list
        .iter()
        .try_fold(init_types, |column_types, exprs| {
            let column_types = column_types
                .iter()
                .zip(exprs.iter())
                .map(|(column_type, expr)| match column_type {
                    Some(data_type) => Ok(Some(data_type.to_owned())),
                    None => evaluate_stateless(None, expr)
                        .and_then(Value::try_from)
                        .map(|v| v.get_type()),
                })
                .collect::<Result<Vec<Option<DataType>>>>()
                .into_control_flow()?;

            match column_types.iter().any(Option::is_none) {
                true => Continue(column_types),
                false => Break(Ok(column_types)),
            }
        });

let column_types = match column_types {
    Continue(column_types) => column_types,
    Break(column_types) => column_types?,
};

let column_defs = column_types
  ...
```

Thanks to @ever0de

panarch (approved):
This is also the meaningful new function GlueSQL newly has.
Now looks really great, let's ship this to the main. Thanks a lot!!

PR #708, opened 2022-08-05
[AST Builder] implement Case

woojinnn:
## Description
I implemented `Case` statement in AST_Builder

## Test Cases
```rust
    #[test]
    fn test_case() {
        let re = Regex::new(r"\n\s+").unwrap();
        let trim = |s: &str| re.replace_all(s.trim(), "\n").into_owned();

        let actual = col("id").case(vec![(num(1), text("a")), (num(2), text("b"))], None);
        let expected = trim(
            r#"                                                                           
            CASE id
              WHEN 1 THEN "a"
              WHEN 2 THEN "b"
            END
            "#,
        );
        test_expr(actual, expected.as_str());

        let actual = col("id").case(
            vec![(num(1), text("a")), (num(2), text("b"))],
            Some(text("c").into()),
        );
        let expected = trim(
            r#"                                                                           
            CASE id
              WHEN 1 THEN "a"
              WHEN 2 THEN "b"
              ELSE "c"
            END
            "#,
        );
        test_expr(actual, expected.as_str());

        let actual = case(
            None,
            vec![(col("City").is_null(), col("Country"))],
            Some(col("City").into()),
        );
        let expected = trim(
            r#"                                                                           
            CASE
                WHEN City IS NULL THEN Country
                ELSE City
            END
            "#,
        );
        test_expr(actual, expected.as_str());
    }
```

### DIFF
--- a/core/src/ast_builder/expr/case.rs
+++ b/core/src/ast_builder/expr/case.rs
@@ -1,0 +1,81 @@
+use super::ExprNode;
+
+impl ExprNode {
+    pub fn case(
+        self,
+        when_then: Vec<(ExprNode, ExprNode)>,
+        else_result: Option<Box<ExprNode>>,
+    ) -> Self {
+        Self::Case {
+            operand: Some(Box::new(self)),
+            when_then,
+            else_result,
+        }
+    }
+}
+
+pub fn case(
+    operand: Option<Box<ExprNode>>,
+    when_then: Vec<(ExprNode, ExprNode)>,
+    else_result: Option<Box<ExprNode>>,
+) -> ExprNode {
+    ExprNode::Case {
+        operand,
+        when_then,
+        else_result,
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use regex::Regex;
+
+    use crate::ast_builder::{case, col, num, test_expr, text};
+
+    #[test]
+    fn test_case() {
+        let re = Regex::new(r"\n\s+").unwrap();
+        let trim = |s: &str| re.replace_all(s.trim(), "\n").into_owned();
+
+        let actual = col("id").case(vec![(num(1), text("a")), (num(2), text("b"))], None);
+        let expected = trim(
+            r#"                                                                           
+            CASE id
+              WHEN 1 THEN "a"
+              WHEN 2 THEN "b"
+            END
+            "#,
+        );
+        test_expr(actual, expected.as_str());
+
+        let actual = col("id").case(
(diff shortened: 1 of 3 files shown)


core/src/ast_builder/expr/case.rs
```diff
@@ -0,0 +1,81 @@
...
+              WHEN 2 THEN "b"
+            END
+            "#,
+        );
+        test_expr(actual, expected.as_str());
+
+        let actual = col("id").case(
+            vec![(num(1), text("a")), (num(2), text("b"))],
+            Some(text("c").into()),
+        );
```
panarch, line 54: As we build FSM for handling select statement, it would be also cool to consider building small FSM for case expression.

e.g.
```rust
let actual = col("id")
    .case()
    .when_then(num(1), text("a"))
    .when_then(2, "b")
    .else("c")
    .end()

let actual = case("id > 10")
    .when_then(true, "a")
    .when_then(false, "b")
    .end();
```

core/src/ast_builder/expr/case.rs
```diff
@@ -0,0 +1,81 @@
...
+            None,
+            vec![(col("City").is_null(), col("Country"))],
+            Some(col("City").into()),
+        );
+        let expected = trim(
+            r#"                                                                           
+            CASE
+                WHEN City IS NULL THEN Country
+                ELSE City
+            END
```
devgony, line 76: It would be nice if there is a test case for `nested CASE`!
```sql
CASE
    WHEN City IS NULL THEN CASE WHEN Country IS NULL THEN 'weird'
                                ELSE Country
                                END
    ELSE City
END
```

core/src/ast_builder/expr/case.rs
panarch:
```sql
CASE id
    WHEN 1 THEN "a" 
    WHEN 2 THEN "b"
    ELSE "c"
END

CASE id > 10
    WHEN True THEN "a"
    WHEN False THEN "b"
END
```

ding-young:
@woojinnn Are you still working on this issue?

woojinnn:
@ding-young I don't think I'm going to do this for a while. You can take this one if you want.

ding-young:
> @ding-young I don't think I'm going to do this for a while. You can take this one if you want.

I'll take it. Thank you.

ever0de:
Since this PR was created by @woojinn, it should be closed and assigned to something created by @ding-young.
cc. @devgony

PR #716, opened 2022-08-06
Remove unnecessary `Box`

ever0de:
related <https://github.com/sqlparser-rs/sqlparser-rs/pull/556>

> Previously, Query variant in enum Statement seemed to be used recursively, but this is not the case now.


panarch (approved):
Awesome, thanks a lot for the good catch.
Looks great 👍

PR #717, opened 2022-08-06
Update sled-storage::update_data not to return error on not-existing …

panarch:
…key inputs

Update `SledStorage::update_data` behavior

as-is: `update_data` returns error if not-existing key comes as input param.
to-be: if key does not exist in the storage, it works as insertion mode. now the data is inserted rather than error.


ever0de (approved):
LGTM!

Issue #714, opened 2022-08-06
[AST Builder] Implement update expression


seonghun-dev:
# Description
Implements ast builders for [`update`](https://www.w3schools.com/sql/sql_create_index.asp) Expression.

## Background
Refered https://github.com/gluesql/gluesql/issues/614

## Task
In AST Builder module to create, show and delete index easier

- [x] 1. [`Update`](https://www.w3schools.com/sql/sql_update.asp) expression
- [x] 2. [`Assignment`](https://www.w3schools.com/sql/sql_create_index.asp) expression

The following tests should be passed:

``` rust
    #[test]
    fn update() {
        let actual = table("Foo").update().set("id","2").build();
        let expected = "UPDATE Foo SET id = 2";
        test(actual, expected);

       let actual = table("Foo").update().set("id","2").set("name","😝").build();
       let expected = "UPDATE Foo SET id = 2, name='😝'";
       test(actual, expected);

       let actual = table("Foo").update().set("id","2").set("name","😝").filter("Bar =1").build();
       let expected = "UPDATE Foo SET id = 2, name='😝' WHERE Bar = 1";
       test(actual, expected);

       let actual = table("Foo").update().set("id","2").set("head_item","SELECT id FROM head_item WHERE level = 3 LIMIT 1").filter("body_item =1").build();
       let expected = "UPDATE Foo SET id = 2, head_item = (SELECT id FROM head_item WHERE level = 3 LIMIT 1) WHERE body_item = 1";
       test(actual, expected);

    }
```

PR #720, opened 2022-08-06
[AST Builder] Implement update Expression

seonghun-dev:
Resolve https://github.com/gluesql/gluesql/issues/714


core/src/ast_builder/assignment.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    #[test]
+    fn assignment() {
+        let actual = AssignmentNode::Text("foo = 1".into());
+        let expected = "foo = 1";
+        test(actual, expected);
+
+        let actual = AssignmentNode::Text("foo = \"cocoa\"".into());
+        let expected = "foo = \"cocoa\"";
```
panarch, line 52: This would be useful.
e.g.
```rust
let expected = r#"foo = "cocoa""#;
```

core/src/ast_builder/assignment.rs
```diff
@@ -0,0 +1,55 @@
...
+}
+
+impl TryFrom<AssignmentNode> for Assignment {
+    type Error = Error;
+
+    fn try_from(node: AssignmentNode) -> Result<Self> {
+        match node {
+            AssignmentNode::Text(expr) => {
+                let expr = parse_sql_assignment(expr).and_then(|op| translate_assignment(&op))?;
+                Ok(expr)
```
ever0de, line 26: [nit]

```suggestion
                parse_sql_assignment(expr).and_then(|op| translate_assignment(&op))
```

ever0de: This variable is an assignment, not an operator.
edit variable name

`op` to `assignment`

panarch (approved):
Looks all great, now we support another new statement `UPDATE`! Thanks a lot 👍

PR #723, opened 2022-08-06
[AST Builder] Update FunctionNode structure

zmrdltl:
## Description
Fixed structure for uniform cast in `ast_builder`. 
Instead of being converted to `Expr` directly at each node, conversion is performed to the correct enum type in the intermediate stage.


panarch (approved):
Awesome! Thanks a lot for the meaningful improvement changes :)

PR #727, opened 2022-08-07
Replace bool::then to bool::then_some

panarch:
`bool::then_some` is now available in stable rust 1.62.
Replace some `bool::then` codes to `bool::then_some` which does not require closures

ref. https://doc.rust-lang.org/std/primitive.bool.html#method.then_some


24seconds (approved):
Nice

ever0de (approved):
I love it 
Thanks

PR #730, opened 2022-08-08
Support `ORDER BY` clause in `VALUES list`

devgony:
This is subsequent PR of #648 
## Goal
- Support `ORDER BY` clause in `VALUES list`
```sql
gluesql> VALUES (1, 'a'), (2, 'b') ORDER BY column1 DESC;
 column1 | column2
---------+---------
       2 | b
       1 | a
```

## Todo
- [x] move order_by property from `select` to `query`
- [x] implement values-orderby
- [x] refactor implementation as sort_stateless() for future reuse

### DIFF
--- a/core/src/executor/select/mod.rs
+++ b/core/src/executor/select/mod.rs
@@ -160,7 +160,7 @@ fn rows_from(exprs_list: &[Vec<Expr>]) -> (Vec<Result<Row>>, Vec<String>) {
 fn sort_stateless(
     rows: Vec<Result<Row>>,
     labels: &Vec<String>,
-    order_by: &Vec<OrderByExpr>,
+    order_by: &[OrderByExpr],
 ) -> Result<Vec<Result<Row>>> {
     let sorted = rows
         .into_iter()
--- a/core/src/executor/select/mod.rs
+++ b/core/src/executor/select/mod.rs
@@ -1,6 +1,8 @@
 mod blend;
 mod error;
 
+use std::cmp::Ordering;
+
 pub use error::SelectError;
 
 use {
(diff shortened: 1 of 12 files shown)


core/src/executor/select/mod.rs
```diff
@@ -154,6 +156,66 @@ fn into_rows(exprs_list: &[Vec<Expr>]) -> (Vec<Result<Row>>, Vec<String>) {
     (rows, labels)
 }
 
+fn sort_stateless(
```
panarch, line 159: How about adding this to `executor/sort.rs` module.

Then, we can re-use partial sorting function.
e.g.
```rust
fn sort_by(values_a: &[(Value, Option<bool>)], values_b: &[(Value, Option<bool>)]) -> Ordering {
    let pairs = values_a
        .iter()
        .map(|(a, _)| a)
        .zip(values_b.iter())
        .map(|(a, (b, asc))| (a, b, asc.unwrap_or(true)));

    for (value_a, value_b, asc) in pairs {
        let apply_asc = |ord: Ordering| if asc { ord } else { ord.reverse() };

        match (value_a, value_b) {
            (Value::Null, Value::Null) => {}
            (Value::Null, _) => {
                return apply_asc(Ordering::Greater);
            }
            (_, Value::Null) => {
                return apply_asc(Ordering::Less);
            }
            _ => {}
        };

        match value_a.partial_cmp(value_b) {
            Some(ord) if ord != Ordering::Equal => {
                return apply_asc(ord);
            }
            _ => {}
        }
    }

    Ordering::Equal
}
```

Then, in `sort_stateless`
```rust
.sort_by(|(values_a, _), (values_b, _)| sort_by(values_a, values_b))
```
and in 'Sort::apply`
```rust
.sort_by(|(values_a, _, _), (values_b, _, _)| sort_by(values_a, values_b))
```

Then can share the same sorting function.

panarch (approved):
All looks great, and also thanks that make `order_by` to find its own proper place.
Thanks a lot!!

PR #731, opened 2022-08-08
Support `Derived VALUES`

### LINKED ISSUE #418: Support `SELECT * FROM (VALUES (1)) t1 (INTEGER);` like queries

devgony:
To resolve #418 
## Goal
- Basic Derived VALUES
```sql
SELECT * FROM (VALUES (1, 'a'), (2, 'b')) AS Derived
 column1 | column2
---------+---------
       1 | a
       2 | b
```
- Derived VALUES with Column aliases
```sql
SELECT * FROM (VALUES (1, 'a'), (2, 'b')) AS Derived(id, name)
 id | name
----+------
  1 | a
  2 | b
```
## Todo
- [x] implement Drived VALUES base
- [x] implement Drived VALUES with Column aliases
  - No Column aliases
  - Some Column aliases + original column name 
  - Only Column aliases
- [x] Throw error if the number of Column aliases exceeds columns

### DIFF
--- a/core/src/executor/fetch.rs
+++ b/core/src/executor/fetch.rs
@@ -3,6 +3,7 @@ use {
     crate::{
         ast::{
             ColumnDef, Expr, IndexItem, Join, Query, Select, SetExpr, TableFactor, TableWithJoins,
+            Values,
         },
         data::{get_alias, get_index, get_name, Key, Row, TableError, Value},
         executor::select::{get_labels, select},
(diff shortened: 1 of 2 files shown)


core/src/executor/fetch.rs
```diff
@@ -166,33 +169,48 @@ pub async fn fetch_relation_columns(
...
+                    .into_iter()
+                    .map(|i| format!("column{}", i));
+                let labels = match alias_len > width {
+                    true => {
+                        return Err(
+                            FetchError::TooManyColumnAliases(name.into(), width, alias_len).into(),
+                        )
+                    }
+                    false => columns.iter().cloned().chain(labels).collect::<Vec<_>>(),
+                };
```
panarch, line 209: It looks that we can make make code a bit more simple by doing some rearrangement.
```rust
let total_len = values_list[0].len();
let alias_len = columns.len();

if alias_len > total_len {
    return Err(FetchError::TooManyColumnAliases(
        name.into(),
        total_len,
        alias_len,
    )
    .into());
}

let labels = (alias_len + 1..=total_len)
    .into_iter()
    .map(|i| format!("column{}", i));
let labels = columns.iter().cloned().chain(labels).collect::<Vec<_>>();
```

panarch (approved):
Thanks a lot! Awesome new feature now comes again 👍 
All looks great

PR #733, opened 2022-08-08
Support `SELECT * FROM SERIES(N)` and `SELECT 1`(without table)

### LINKED ISSUE #286: Support `SELECT` without table
```sql
SELECT 1;
SELECT 1 + 2;
SELECT "id" || "wow";
```

Currently `SELECT {expr}` queries without referring table are not supported, `translate` module generates `LackOfTable` error.
Let's add this `SELECT {expr};` query support, then we can remove `LackOfTable` error.

devgony:
To resolve #286
### 1. `SERIES(N)` generates sequential numbers from `1 to N` with reserved column `N`
```sql
gluesql> SELECT * FROM SERIES(3);
╭───╮
│ N │
╞═══╡
│ 1 │
│ 2 │
│ 3 │
╰───╯
```
### 2. `SELECT 1` is equivalent to `SELECT 1 FROM SERIES(1)`
```sql
gluesql> SELECT 1, 'a', true, 1 + 2, 'a' || 'b';
╭─────────────────────────────────────╮
│ 1   'a'   true   1 + 2   'a' || 'b' │
╞═════════════════════════════════════╡
│ 1   a     TRUE   3       ab         │
╰─────────────────────────────────────╯
gluesql> SELECT (SELECT 1);
╭────────────╮
│ (SELECT 1) │
╞════════════╡
│ 1          │
╰────────────╯
gluesql> SELECT 1 AS id, (SELECT MAX(num) FROM TestA) AS max;
╭──────────╮
│ id   max │
╞══════════╡
│ 1    9   │
╰──────────╯
```
## Todo
~~- [x] add Dummy(String) to TableFactor~~
~~- I decided to add a new field to TableFactor instead of Option<TableFactor> to minimized effects to previous code~~
~~- [x] use "$Dummy" string as a dummy table name~~
~~- [x] fill missing match arms~~
~~- [x]  cover all test cases~~
- [x] Add TableFactor::Series(N)
- [x] `SELECT 1` => redirect to SERIES(1)
- [x] support case insensitive like `sErIeS(N)`
- [x] prohibit below queries
```sql
SELECT * FROM SERIES(-1) -- WrongSeriesSize
SELECT * FROM SERIES(0) -- LackOfSeriesSize
SELECT * FROM SERIES() -- LackOfSeriesSize
SELECT * FROM SERIES -- TableNotFound
```
- [x] support CTAS `without table`, with `SERIES(N)`
## References
[Oracle Hierarchical Queries](https://docs.oracle.com/cd/B19306_01/server.102/b14200/queries003.htm)
[MySQL Recursive Common Table Expression](https://dev.mysql.com/doc/refman/8.0/en/with.html#common-table-expressions-recursive-examples)
[PostgreSQL Series Generating Functions](https://www.postgresql.org/docs/current/functions-srf.html)
## In Next PR
~~- looks like there will be common code between `CTAS without Table` and `CTAV` #704, so that the test case is left Unreachable for now.~~


ever0de (commented):
Sorry for the late review. Could you please resolve the conflict?
In general, the changes look good, but the `table_name == $Dummy` side could be improved.
ex) The type of table_name is changed from string to enum, as shown below
```rust
enum TableName {
    Table(String),
    Dummy,
}
```

What do you think?

devgony:
> Sorry for the late review. Could you please resolve the conflict? In general, the changes look good, but the `table_name == $Dummy` side could be improved. ex) The type of table_name is changed from string to enum, as shown below
> 
> ```rust
> enum TableName {
>     Table(String),
>     Dummy,
> }
> ```
> 
> What do you think?
@ever0de 
At first I tried to add Dummy to TableFactor in this [commit](https://github.com/gluesql/gluesql/pull/733/commits/6f31bca46a45e3f4dedf476ae01fa2212602c29e#diff-aac67cf4fb487d82438fcfb74d89cca32495c16b10f3ed47af25e779a7255f4fR65) but it looked like too excessive to handle just simple case.
So i decided to use simple String `$Dummy`.
Currently, @panarch suggested to use `Option<TableFactor>` so that i'm considering which is the best among three options.

core/src/translate/query.rs
```diff
@@ -131,18 +145,84 @@ fn translate_table_with_joins(sql_table_with_joins: &SqlTableWithJoins) -> Resul
...
+                    }
+                    .into()),
+                }
+            }
+            None => Err(TranslateError::LackOfSeriesSize.into()),
+            _ => Err(TranslateError::UnsupportedArgs(format!(
+                "{}",
+                function_args
+                    .iter()
+                    .map(|v| v.to_string())
```
devgony, line 186: I'm not sure this is best way or not 😢

core/src/translate/query.rs
```diff
@@ -1,20 +1,26 @@
+use bigdecimal::BigDecimal;
+use sqlparser::ast::UnaryOperator;
```
panarch, line 2: It looks that we can merge this into `use` in the below

core/src/ast/query.rs
```diff
@@ -65,6 +65,11 @@ pub enum TableFactor {
         subquery: Query,
         alias: TableAlias,
     },
+    Series {
+        name: ObjectName,
+        alias: Option<TableAlias>,
+        size: i64,
```
panarch, line 71: How about simply getting `Expr`?

core/src/executor/fetch.rs
```diff
@@ -140,6 +144,11 @@ pub async fn fetch_relation_rows<'a>(
 
             Ok(Rows::Table(stream::iter(rows)))
         }
+        TableFactor::Series { size, .. } => {
+            let rows = (1..=*size).collect::<Vec<_>>();
+            let rows = stream::iter(rows).map(|v| Ok(Row(vec![Value::I64(v)])));
+            Ok(Rows::Series(rows))
```
panarch, line 150: Use of `collect` looks unnecessary.
It would be ok without `collect`
```rust
let rows = (1..=*size).map(|v| Ok(Row(vec![Value::I64(v)])));

Ok(Rows::Series(stream::iter(rows)))
```

test-suite/src/basic.rs
```diff
@@ -64,6 +71,143 @@ CREATE TABLE TestA (
...
         ),
+        (
+            // SELECT without Table
+            Ok(select!(
+                1   | 'a'       | true | "1 + 2" | "'a' || 'b'"
+                I64 | Str       | Bool | I64     | Str;
+                1     "a".into()  true   3         "ab".into()
+            )),
+            "SELECT 1, 'a', true, 1 + 2, 'a' || 'b'",
+        ),
```
panarch, line 82: `Series` is too big and meaningful to exist only as a part of `basic.rs` tests.
Let's make `test-suite/src/series.rs` to contain these. :D

core/src/translate/query.rs
```diff
@@ -131,18 +145,84 @@ fn translate_table_with_joins(sql_table_with_joins: &SqlTableWithJoins) -> Resul
...
+        }
+        SqlTableFactor::Table { name, alias, .. } => {
+            Ok(TableFactor::Table {
+                name: translate_object_name(name),
+                alias: alias
+                    .as_ref()
+                    .map(|SqlTableAlias { name, columns }| TableAlias {
+                        name: name.value.to_owned(),
+                        columns: translate_idents(columns),
+                    }),
```
panarch, line 222: There exists two translate code of `SqlTableAlias` to `TableAlias`.
It looks good to add `translate_table_alias` function to handle this.

panarch (approved):
This is beautiful, I'd really appreciate..
All looks really great, even the coverage is nice.
Thanks a lot, let's ship this to the main 👍 👍 👍

PR #736, opened 2022-08-09
Support `.columns TABLE` command in CLI

### LINKED ISSUE #735: Support `.columns TABLE` command in CLI
## Goal
- add `.columns Table` command in CLI as a alias of `show columns from Table`
```sql
gluesql> CREATE TABLE Foo(id INT, name TEXT);
gluesql> .columns Foo
╭──────────────╮
│ Field   Type │
╞══════════════╡
│ id      Int  │
│ name    Text │
╰──────────────╯
```
## In Next PR
- Support showing column options
Like below, it should show informations for not only columns but also options like PK, Default.. 
```sql
+-------+-------------+------+-----+---------+-------+
| Field | Type        | Null | Key | Default | Extra |
+-------+-------------+------+-----+---------+-------+
| 1     | varchar(19) | NO   |     |         |       |
+-------+-------------+------+-----+---------+-------+
```

devgony:
To resolve #735
## Goal
add .columns Table command in CLI as a alias of show columns from Table
```sql
gluesql> CREATE TABLE Foo(id INT, name TEXT);
gluesql> .columns Foo
╭──────────────╮
│ Field   Type │
╞══════════════╡
│ id      Int  │
│ name    Text │
╰──────────────╯
```
```sql
gluesql> .columns wrong
[error] table not found: wrong
```
```sql
gluesql> .columns
[error] should specify table. eg: .columns TableName
```
## Todo
- [x] create CommandError::LackOfTable
- [x] parse 2nd param
- If 2nd param does not exists => throw LackOfTable
- else..
  - [x] let .columns command executes `show columns from TABLE`

### DIFF
--- a/cli/src/cli.rs
+++ b/cli/src/cli.rs
@@ -1,3 +1,5 @@
+use crate::command::CommandError;
+
 use {
     crate::{command::Command, helper::CliHelper, print::Print},
     gluesql_core::{
@@ -64,7 +66,11 @@ where
 
             let command = match Command::parse(&line) {
                 Ok(command) => command,
-                Err(_) => {
+                Err(CommandError::LackOfTable) => {
+                    println!("\n[error] should specify table. eg: .columns TableName\n");
+                    continue;
+                }
+                Err(CommandError::NotSupported) => {
                     println!("[error] command not supported: {}", line);
                     println!("\n  type .help to list all available commands.\n");
                     continue;
--- a/cli/src/cli.rs
+++ b/cli/src/cli.rs
@@ -1,7 +1,9 @@
-use crate::command::CommandError;
-
 use {
-    crate::{command::Command, helper::CliHelper, print::Print},
+    crate::{
+        command::{Command, CommandError},
+        helper::CliHelper,
+        print::Print,
+    },
     gluesql_core::{
         prelude::Glue,
         store::{GStore, GStoreMut},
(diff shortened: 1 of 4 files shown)


cli/src/cli.rs
```diff
@@ -1,3 +1,5 @@
+use crate::command::CommandError;
```
ever0de, line 1: Combine it with the import statement below

cli/src/command.rs
```diff
@@ -6,8 +8,16 @@ pub enum Command {
     ExecuteFromFile(String),
 }
 
+#[derive(ThisError, Debug, PartialEq)]
+pub enum CommandError {
+    #[error("should specify table")]
+    LackOfTable,
+    #[error("command not supported")]
+    NotSupported,
+}
```
ever0de, line 17: Good 👍

ever0de (approved):
It's trivial, so I'll admit it first.

cli/src/command.rs
devgony: Do you think it is proper even if we don't use the error messges?
I still doubt if there is another way to distinguish those two cases without using Error

ever0de: Since they're in different branches, you might be able to tell them apart, but what's clear is that this looks good.

panarch (approved):
Awesome! Adding `CommandError` looks also really nice.
Thanks a lot 👍 👍

Issue #744, opened 2022-08-10
[AST Builder] Implement Div, Mod Funciton


seonghun-dev:
# Description
Implements ast builders for `Mod`, `Div` Function.
`mod` is keyword of rust, so using `modulo` based on Modulo Operation

## Background
Refered https://github.com/gluesql/gluesql/issues/614

## Task
In AST Builder module to mod function

The following tests should be passed:

``` rust
    #[test]
    fn function_mod(){
        let actual = modulo(num(64), num(8));
        let expected = "mod(64,8)";
        test_expr(actual, expected);
    }

    #[test]
    fn function_div() {
        let actual = div(num(64), num(8));
        let expected = "div(64,8)";
        test_expr(actual, expected);
    }
```

PR #760, opened 2022-08-13
[AST Builder] Implement Div, Mod Function

seonghun-dev:
Resolve : https://github.com/gluesql/gluesql/issues/744


core/src/ast_builder/expr/function.rs
```diff
@@ -379,12 +397,27 @@ pub fn substr<V: Into<ExprNode>>(expr: V, start: V, count: Option<V>) -> ExprNod
     }))
 }
 
+pub fn div<V: Into<ExprNode>>(dividend: V, divisor: V) -> ExprNode {
```
panarch, line 400: I like the decision to have name `modulo` rather than `mod`, how about applying this also for `div`?
As a separate function to be exported, `divide` would be explicit to understand.

panarch (approved):

PR #762, opened 2022-08-14
Update SledStorage scan_(indexed)_data key not to contain table prefix,

panarch:
prev.
Key info fetched from SledStorage scan_data & scan_indexed_data had
format of {table_name} + {data identifier}.
However, all store mut traits which handles data keys provide table_name info
so it is unnecessary for the fetched keys to have {table_name} info.

changes.
Remove {table_name} prefix info from the keys fetched from
Store::scan_data & Index::scan_indexed_data.
Apply key format changes to store mut trait impl codes.

### DIFF
--- a/storages/sled-storage/src/key.rs
+++ b/storages/sled-storage/src/key.rs
@@ -4,6 +4,20 @@ const TEMP_DATA: &str = "temp_data/";
 const TEMP_SCHEMA: &str = "temp_schema/";
 const TEMP_INDEX: &str = "temp_index/";
 
+pub fn data_prefix(table_name: &str) -> String {
+    format!("data/{table_name}/")
+}
+
+pub fn data(table_name: &str, key: Vec<u8>) -> IVec {
+    let key = data_prefix(table_name)
+        .into_bytes()
+        .into_iter()
+        .chain(key.into_iter())
+        .collect::<Vec<_>>();
+
+    IVec::from(key)
+}
+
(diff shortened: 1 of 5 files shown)


storages/sled-storage/src/key.rs
```diff
@@ -4,6 +4,20 @@ const TEMP_DATA: &str = "temp_data/";
...
+pub fn data_prefix(table_name: &str) -> String {
+    format!("data/{table_name}/")
+}
+
+pub fn data(table_name: &str, key: Vec<u8>) -> IVec {
+    let key = data_prefix(table_name)
+        .into_bytes()
+        .into_iter()
+        .chain(key.into_iter())
+        .collect::<Vec<_>>();
```
ever0de, line 16: https://docs.rs/sled/latest/sled/struct.IVec.html#impl-FromIterator%3Cu8%3E

Can you delete `collect` and use from as an iterator?

ever0de (commented):
Other than minor things, it looks fine.

storages/sled-storage/src/key.rs
panarch: Oh, thanks a lot. This looks that we can apply this not only for `key::data`, but also all other functions in `key.rs` module.
I'll make this as a separate PR for this.

ever0de: Ok i will check

ever0de (approved):

PR #765, opened 2022-08-15
Update SledStorage key.rs to use IVec::from_iter, not IVec::from

panarch:


ever0de (approved):
I've seen that the actual operation is not different, but the code seems to have gotten prettier.

PR #770, opened 2022-08-15
Update `Expr::CompoundIdentifier` to have more accurate data format

### LINKED ISSUE #768: Update `Expr::CompoundIdentifier` to have more accurate data format
Currently, the format of `Expr::CompoundIdentifier(Vec<String>)` is a superset of what GlueSQL execution can handle it.
Only single pair of (table_alias, column_name) can exist, so the current shape `Vec<String>` is too broad.

We can consider to change the type of `Expr::CompoundIdentifier` to have more narrow info.
e.g.
```rust
Expr::AliasedIdentifier {
  alias: String,
  ident: String,
}
```

It would be nice that we can dicuss about this. @ever0de @devgony

devgony:
To resolve #768


core/src/plan/evaluable.rs
```diff
@@ -217,7 +213,6 @@ mod tests {
         test!("Bar.rate", true);
         test!("Foo.rate", false);
         test!("Rand.id", false);
-        test!("a.b.c", false);
```
devgony, line 220: Looks like we have to fix unwrap at tester::expr to keep this unit test.

ever0de: It seems right to disappear as a patch in the current PR.

core/src/plan/join.rs
```diff
@@ -163,6 +163,116 @@ impl<'a> JoinPlanner<'a> {
         (context, join)
     }
 
+    fn subquery_expr(&self, outer_context: Option<Rc<Context<'a>>>, expr: Expr) -> Expr {
```
ever0de, line 166: Why is the change here?

core/src/translate/expr.rs
```diff
@@ -21,7 +21,18 @@ pub fn translate_expr(sql_expr: &SqlExpr) -> Result<Expr> {
...
+                .map(|ident| ident.value.to_owned())
+                .collect::<Vec<_>>();
+
+            match &values[..] {
+                [alias, ident] => Ok(Expr::CompoundIdentifier {
+                    alias: alias.to_owned(),
+                    ident: ident.to_owned(),
+                }),
+                _ => Err(TranslateError::UnsupportedExpr(values.join(".")).into()),
+            }
```
ever0de, line 35: recommended code(my opinion)

```rust
        SqlExpr::CompoundIdentifier(idents) => (idents.len() == 2)
            .then(|| Expr::CompoundIdentifier {
                alias: idents[0].value.clone(),
                ident: idents[1].value.clone(),
            })
            .ok_or_else(|| {
                TranslateError::UnsupportedExpr(translate_idents(idents).join(".")).into()
            }),
```

core/src/plan/expr/mod.rs
```diff
@@ -116,8 +124,10 @@ mod tests {
 
         // PlanExpr::CompoundIdentifier
         let actual = expr("Foo.id");
-        let expected = ["Foo".to_owned(), "id".to_owned()];
-        let expected = PlanExpr::CompoundIdentifier(&expected);
+        let expected = PlanExpr::CompoundIdentifier {
+            alias: &"Foo".to_string(),
+            ident: &"id".to_string(),
```
ever0de, line 129: This is a situation that is caught when the clippy target is tested.
`cargo clippy --release --all-features --all-targets --workspace`

```suggestion
            alias: "Foo",
            ident: "id",
```

Let's change it to

core/src/ast_builder/expr/mod.rs
```diff
@@ -192,12 +197,14 @@ pub fn expr(value: &str) -> ExprNode {
...
-    if idents.len() == 1 {
-        ExprNode::Identifier(value.to_owned())
-    } else {
-        ExprNode::CompoundIdentifier(idents.into_iter().map(ToOwned::to_owned).collect())
+    match &idents[..] {
+        [alias, ident] => ExprNode::CompoundIdentifier {
+            alias: alias.to_owned(),
+            ident: ident.to_owned(),
+        },
+        _ => ExprNode::Identifier(value.to_owned()),
```
ever0de, line 207: I'm sure there's better code... I'll think about it.

ever0de (commented):
👍

core/src/plan/join.rs
devgony: Just realized that there was wrong merging.. Thank you 👍

core/src/ast_builder/expr/mod.rs
devgony: The problem here is that `fn col()` is trying to handle both of `CompoundIdentifier` and `Identifier`.
Do you think it should throw error if `idents.len()` is not 2 or 1 ?

ever0de: Well, I think so.
It seems good to state that we do not support anything other than what we currently support.

panarch: Though it changes some existing behavior but I think that is acceptable and does not have issue for general use cases.

here is just nit codes suggestion
It looks that `map(|v| v.to_string())` is unnecessary.

```rust
    let idents = value.split('.').collect::<Vec<_>>();

    match idents.as_slice() {
        [alias, ident] => ExprNode::CompoundIdentifier {
            alias: alias.to_string(),
            ident: ident.to_string(),
        },
        _ => ExprNode::Identifier(value.to_owned()),
    }
```

test-suite/src/migrate.rs
```diff
@@ -33,7 +33,7 @@ test_case!(migrate, async move {
             "INSERT INTO Test (id, num, name) VALUES (1, 1, a.b);",
         ),
         (
-            EvaluateError::UnsupportedCompoundIdentifier(expr!("Here.User.id")).into(),
```
panarch, line 36: As you removed `AggregateError::UnsupportedCompoundIdentifier`, we now no more need `EvaluateError::UnsupportedCompoundIdentifier`.

PR #774, opened 2022-08-20
Rename StoreMut insert_data & update_data methods,

panarch:
StoreMut::insert_data -> append_data
StoreMut::update_data -> insert_data

Our custom storage makers use Rust lang to build their own storages.
It would be better to follow the method names which are commonly used by standard libraries
e.g.
[Vec::append](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.append)
[Vec::insert](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.insert)

### DIFF
--- a/core/src/executor/execute.rs
+++ b/core/src/executor/execute.rs
@@ -259,8 +259,8 @@ pub async fn execute<T: GStore + GStoreMut>(
             });
 
             match rows {
-                RowsData::Insert(rows) => storage.insert_data(table_name, rows).await,
-                RowsData::Update(rows) => storage.update_data(table_name, rows).await,
+                RowsData::Insert(rows) => storage.append_data(table_name, rows).await,
+                RowsData::Update(rows) => storage.insert_data(table_name, rows).await,
             }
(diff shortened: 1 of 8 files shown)


core/src/executor/execute.rs
```diff
@@ -259,8 +259,8 @@ pub async fn execute<T: GStore + GStoreMut>(
             });
 
             match rows {
-                RowsData::Insert(rows) => storage.insert_data(table_name, rows).await,
-                RowsData::Update(rows) => storage.update_data(table_name, rows).await,
+                RowsData::Insert(rows) => storage.append_data(table_name, rows).await,
+                RowsData::Update(rows) => storage.insert_data(table_name, rows).await,
```
ever0de, line 263: `RowsData` is Update, but the storage method it uses is `insert_data`. How about unifying this?

panarch: Oh!! Thanks a lot, I missed it

ever0de (approved):
Seems a bit more intuitive :+1:

PR #775, opened 2022-08-20
[Test Suite] delete unneccesary keyword `vec!`, `iter()`

### LINKED ISSUE #782: [Test Suite] Refacting tests
## Description
refacting some tests in folder [test_suite](https://github.com/gluesql/gluesql/tree/main/test-suite)
Tests implemented in the existing test suite have become longer as new functions are added. 

To do this, we need to `split the test file by function`.
We also want to remove unnecessary macros and `functions` as new macros are implemented for unity.

The criteria for sorting are as follows.
1. `Minimize` `vec!`, `iter()`
2. `Merge` use crate statements into one
3. `Rearrange` top level test files, and error test cases back to its place
4. `Unification` of the order of test cases by forcing the order as the argument of the `test function` in the order of sql and expected
## Tasks
- [x] #771
- [x] #773 
- [x] #775
- [x] #785
- [x] #788 
- [x] #792
- [x] #794
- [x] #804 <br/>#816 
- [ ] forcing the argument order of test function

zmrdltl:
## Description
resolve #782


panarch (approved):
It's pleasure to see this meaningful improvement, thanks a lot @zmrdltl 
Looks great!

PR #781, opened 2022-08-20
Add in_subquery test case using QueryNode

### LINKED ISSUE #778: [AST Builder] Implement in_subquery Expression using QueryNode Function
## Description

Implement `in_subquery` and `not_in_subquery` statements ast builder

## Task

```rust
#[cfg(test)]
mod test {
    use crate::ast_builder::{col, table, test_expr};

    #[test]
    fn in_subquery() {
        let actual = col("id").in_subquery(table("FOO").select().filter("id IS NULL").limit(10));
        let expected = "id IN (SELECT * FROM FOO WHERE id IS NULL LIMIT 10)";
        test_expr(actual, expected);

        let actual = col("id").not_in_subquery(table("FOO").select().filter("id IS NULL").limit(10));
        let expected = "id NOT IN (SELECT * FROM FOO WHERE id IS NULL LIMIT 10)";
        test_expr(actual, expected);

        let actual = col("id").in_subquery("SELECT id FROM FOO");
        let expected = "id IN (SELECT id FROM FOO)";
        test_expr(actual, expected);

        let actual = col("id").not_in_subquery("SELECT id FROM FOO");
        let expected = "id NOT IN (SELECT id FROM FOO)";
        test_expr(actual, expected);
    }
}
```

CEOJINSUNG:
Resolve #778


panarch (approved):
Thanks a lot 👍 Looks nice

Issue #713, opened 2022-08-06
[AST Builder] Implement create_table statement


ding-young:
## Description
Implement ast_builder `create_table` statement 

## Tasks
- [ ] implement basic create table statement  
- [ ] block invalid column options (ex. NULL NOT NULL)
- [ ] implement create table as statement (with subquery) 

## Issue
referred #614 

## Description
Following tests should be passed. 
```rust
    #[test]
    fn create_table() {
        let actual = table("Foo")
                .create_table()
                .set_col(col("id"), "int", null())
                .set_col(col("num"), "int")
                .set_col(col("name"), "text")
                .build();
        let expected = "CREATE TABLE Foo (id INTEGER NULL, num INTEGER, name TEXT)";

        let actual = table("Foo")
                .create_table_if_not_exists()
                .set_col(col("id"), "uuid", unique())
                .build();
        let expected = "CREATE TABLE IF NOT EXISTS Foo (id UUID UNIQUE)";

        test(actual, expected);
    }
```

ever0de:
It seems to be a problem solved in #784. Can I close it?

pythonbrad:
Hi, this issue has been solved?

ever0de:
> Hi, this issue has been solved?

There doesn't seem to be a subquery implementation yet.

ever0de:
> Hi, this issue has been solved?

There doesn't seem to be a subquery implementation yet.

ding-young:
> Hi, this issue has been solved?

Hi, ast builder for basic create table statement has been implemented, but you can still work on supporting ast builder for create table with subquery

PR #784, opened 2022-08-20
[AST-Builder] Create Table Statement

ding-young:
## Issue
related issue : #713


core/src/ast_builder/create_table.rs
```diff
@@ -0,0 +1,93 @@
...
+                ColumnOption::Unique { is_primary: false },
+            ])
+            .build();
+        let expected = "CREATE TABLE IF NOT EXISTS Foo (id UUID UNIQUE NULL)";
+        test(actual, expected);
+
+        let actual = table("Foo")
+            .create_table_if_not_exists()
+            .set_col("id", "UUID")
+            .option("UNIQUE")
```
panarch, line 80: It would be good enough to start with minimal spec which only supports sql string based column def.

Also thanks for the name suggestion `add_column` @devgony 

```rust
let actual = table("Foo")
    .create_table_if_not_exist()
    .add_column("id UUID UNIQUE")
    .add_column("name TEXT")
    .build();
```

let actual = table("Foo")
    .create_table()

panarch (approved):
New statement for AST builder!
Looks simple and straightforward to use.
Thanks a lot!!
```rust
table("Foo")
    .create_table()
    .add_column("id INTEGER NULL")
    .add_column("num INTEGER")
    .add_column("name TEXT")
    .build()
```

PR #788, opened 2022-08-20
[Test Suite] Merge use crates into one

zmrdltl:
## Description
The scattered use crate syntax has been merged into one.

## e.g
```rust
use crate::*;

test_case!(case, async move {
    use gluesql_core::{
        prelude::{Payload, Value::*},
        translate::TranslateError,
    };
});
```
to
```rust
use {
    crate::*,
    gluesql_core::{
        prelude::{Payload, Value::*},
        translate::TranslateError,
    },
};


test_case!(case, async move {
       ...
});
```

### DIFF
--- a/test-suite/src/alter/alter_table.rs
+++ b/test-suite/src/alter/alter_table.rs
@@ -2,7 +2,10 @@
 
 use {
     crate::*,
-    gluesql_core::{data::Value::*, prelude::Payload, store::AlterTableError},
+    gluesql_core::{
+        ast::*, data::Value::*, executor::AlterError, executor::EvaluateError, prelude::Payload,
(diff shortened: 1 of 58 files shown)


ever0de:
Rather than limiting this to test-suite, how about setting the fmt nightly and running the whole thing?

panarch:
> Rather than limiting this to test-suite, how about setting the fmt nightly and running the whole thing?

I think that might not work for this case.
Because usual cases in `test-suite` is..
They used `use` in different functions.
Mixed of `use` on the top and `use` inside of test case functions

panarch (approved):
Nice improvements, thanks a lot 👍

PR #790, opened 2022-08-20
[AST Builder] Implement EXISTS Expression

### LINKED ISSUE #779: [AST Builder] Implement Exists Expression
## Description

Implement `Exist` Expression

## Task

```rust
#[cfg(test)]
mod test {
    use crate::ast_builder::{col, exists, table, test, test_expr};

    #[test]
    fn exist() {
        let actual = table("FOO")
            .select()
            .filter(exists(
                table("BAR")
                    .select()
                    .filter("id IS NOT NULL")
                    .group_by("name"),
            ))
            .build();
        let expected =
            "SELECT * FROM FOO WHERE EXISTS (SELECT * FROM BAR WHERE id IS NOT NULL GROUP BY name)";
        test(actual, expected);

        let actual = table("FOO")
            .select()
            .filter(exists(table("BAR").select().filter("id IS NOT NULL")).not())
            .build();
        let expected =
            "SELECT * FROM FOO WHERE NOT EXISTS (SELECT * FROM BAR WHERE id IS NOT NULL)";
        test(actual, expected);

        let actual = exists(table("FOO").select().filter(col("id").gt(2)));
        let expected = "EXISTS (SELECT * FROM FOO WHERE id > 2)";
        test_expr(actual, expected);

        let actual = exists(table("FOO").select().filter(col("id").gt(2))).not();
        let expected = "NOT EXISTS (SELECT * FROM FOO WHERE id > 2)";
        test_expr(actual, expected);

        let actual = exists("SELECT * FROM FOO");
        let expected = "EXISTS (SELECT * FROM FOO)";
        test_expr(actual, expected);

        let actual = exists("SELECT * FROM FOO").not();
        let expected = "NOT EXISTS (SELECT * FROM FOO)";
        test_expr(actual, expected);
    }
}
```

CEOJINSUNG:
Resolve #779

### DIFF
--- a/core/src/ast_builder/expr/exists.rs
+++ b/core/src/ast_builder/expr/exists.rs
@@ -41,6 +41,10 @@ mod test {
         let expected = "EXISTS (SELECT * FROM FOO WHERE id > 2)";
         test_expr(actual, expected);
 
+        let actual = exists(table("FOO").select().filter(col("id").gt(2))).not();
+        let expected = "NOT EXISTS (SELECT * FROM FOO WHERE id > 2)";
+        test_expr(actual, expected);
+
         let actual = exists("SELECT * FROM FOO");
         let expected = "EXISTS (SELECT * FROM FOO)";
         test_expr(actual, expected);
--- a/core/src/ast_builder/expr/exists.rs
+++ b/core/src/ast_builder/expr/exists.rs
@@ -31,11 +31,7 @@ mod test {
 
         let actual = table("FOO")
             .select()
-            .filter(exists(
-                table("BAR")
-                    .select()
-                    .filter("id IS NOT NULL")
-            ).not())
+            .filter(exists(table("BAR").select().filter("id IS NOT NULL")).not())
             .build();
         let expected =
             "SELECT * FROM FOO WHERE NOT EXISTS (SELECT * FROM BAR WHERE id IS NOT NULL)";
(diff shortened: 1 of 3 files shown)


CEOJINSUNG:
Currently, the several test cases are missed

col().exists

```
let actual = col("id").exists("SELECT * FROM FOO");
let expected = "id WHERE EXISTS (SELECT * FROM FOO)";
test_expr(actual, expected);

let actual = col("id").exists("SELECT * FROM FOO");
let expected = "id EXISTS (SELECT * FROM FOO)";
test_expr(actual, expected);
```

- These cases make failures

core/src/ast_builder/expr/exists.rs
```diff
@@ -0,0 +1,56 @@
+use {super::ExprNode, crate::ast_builder::QueryNode};
+
+impl ExprNode {
+    pub fn exists<T: Into<QueryNode>>(self, query: T) -> Self {
+        exists(query)
+    }
```
panarch, line 6: It does not use `self` at all.
So...
The proper location of `exists` might be not here.
May be it would be better for select nodes to have this. But it looks too big changes for now.
For this PR, I think it is simply enough without having `ExprNode::exists`

CEOJINSUNG: Thanks good comment a lot I will try to fix it

panarch (approved):
All looks nice and test cases look also really helpful.
Thanks a lot 👍

PR #792, opened 2022-08-21
[Test Suite] move `top level test files` to `new tester folder`

zmrdltl:
## Description
To make it clear that other test files are importing macro.rs and tester.rs, they have been merged into a folder called tester.


panarch (approved):
Very helpful changes for users to understand the shape of test-suite.
Thanks a lot 👍

PR #794, opened 2022-08-21
[Test Suite] rearrange `error test cases` to `correct file`

zmrdltl:
rearrange the `error test cases` to `the implemented test file`. 
Some unclassified `error test cases` have been moved to `blend.rs`

## Unclassified error cases
```rust
(
    EvaluateError::ValueNotFound("noname".to_owned()).into(),
    "SELECT * FROM BlendUser WHERE noname = 1;",
),
(
    EvaluateError::NestedSelectRowNotFound.into(),
    "SELECT * FROM BlendUser WHERE id = (SELECT id FROM BlendUser WHERE id = 9);",
),
(
    FetchError::TableNotFound("Nothing".to_owned()).into(),
    "SELECT * FROM Nothing;",
),
(
    TranslateError::TooManyTables.into(),
    "SELECT * FROM BlendUser, BlendItem",
),
(
    TranslateError::UnsupportedStatement("TRUNCATE TABLE BlendUser".to_owned()).into(),
    "TRUNCATE TABLE BlendUser;",
),
(
    TranslateError::UnsupportedBinaryOperator("^".to_owned()).into(),
    "SELECT 1 ^ 2 FROM BlendUser;",
),
(
    TranslateError::UnsupportedQuerySetExpr(
        "SELECT * FROM BlendItem UNION SELECT * FROM BlendItem".to_owned(),
    )
    .into(),
    "SELECT * FROM BlendItem UNION SELECT * FROM BlendItem;",
),
(
    EvaluateError::MoreThanOneRowReturned.into(),
    "SELECT (SELECT id FROM BlendItem) as id FROM BlendItem",
),
```


test-suite/src/blend.rs
```diff
@@ -159,6 +160,41 @@ test_case!(blend, async move {
...
+            "SELECT * FROM Nothing;",
+        ),
+        (
+            TranslateError::TooManyTables.into(),
+            "SELECT * FROM BlendUser, BlendItem",
+        ),
+        (
+            TranslateError::UnsupportedStatement("TRUNCATE TABLE BlendUser".to_owned()).into(),
+            "TRUNCATE TABLE BlendUser;",
+        ),
```
panarch, line 182: `blend.rs` is for testing.. blend module or in SQL it is commonly known as projection.
These does not look related to blend.
It would be good to find another places.

e.g.
This.. looks good to be located in `join` because it shows that implicit inner join is not supported.
```rust
            TranslateError::TooManyTables.into(),
            "SELECT * FROM BlendUser, BlendItem",
```
This would be good to exist in `nested_select`
```rust
        (
            EvaluateError::NestedSelectRowNotFound.into(),
            "SELECT * FROM BlendUser WHERE id = (SELECT id FROM BlendUser WHERE id = 9);",
        ),
```

test-suite/src/blend.rs
```diff
@@ -159,6 +160,41 @@ test_case!(blend, async move {
...
+            TranslateError::UnsupportedBinaryOperator("^".to_owned()).into(),
+            "SELECT 1 ^ 2 FROM BlendUser;",
+        ),
+        (
+            TranslateError::UnsupportedQuerySetExpr(
+                "SELECT * FROM BlendItem UNION SELECT * FROM BlendItem".to_owned(),
+            )
+            .into(),
+            "SELECT * FROM BlendItem UNION SELECT * FROM BlendItem;",
+        ),
```
panarch, line 193: These would be also.. good to find some other places.
but if it is hard for now, then.. how about using `migrate.rs`?

zmrdltl: good idea. However, there are still test cases with ambiguous clauses. This is the case with `migrate.rs` and `synthesize.rs`. It would be nice to have good classification criteria in the future.

panarch (approved):
Thanks a lot for this meaningful refactoring improvement work.
Looks great, now most of test cases in `error.rs` has found proper place to be located.

PR #795, opened 2022-08-24
Impl `TryFrom<&Value> for usize`

### LINKED ISSUE #789: Impl `TryFrom<&Value> for usize`
Location: [core/src/data/value/convert.rs](https://github.com/gluesql/gluesql/blob/main/core/src/data/value/convert.rs)

`TryFrom<&Value> for usize` would be useful for safe type casting from numeric literal to unsigned integer.

e.g. we can apply this to limit execution.
https://github.com/gluesql/gluesql/blob/06ba99eabcf38662c4fe85946b700a0f22b57b05/core/src/executor/limit.rs#L19-L24
```rust
let eval = |expr| -> Result<usize> {
    let value: Value = evaluate_stateless(None, expr)?.try_into()?;
    let n: i64 = value.try_into()?;


    Ok(n as usize)
};
```

devgony:
To resolve #789

### DIFF
--- a/core/src/data/value/convert.rs
+++ b/core/src/data/value/convert.rs
@@ -323,6 +323,41 @@ impl TryFrom<&Value> for f64 {
     }
 }
 
+impl TryFrom<&Value> for usize {
+    type Error = Error;
+
+    fn try_from(v: &Value) -> Result<usize> {
+        Ok(match v {
+            Value::Bool(value) => {
+                if *value {
+                    1
+                } else {
+                    0
+                }
+            }
+            Value::I8(value) => value.to_usize().ok_or(ValueError::ImpossibleCast)?,
+            Value::I16(value) => value.to_usize().ok_or(ValueError::ImpossibleCast)?,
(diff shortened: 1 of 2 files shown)


devgony:
I tried to adjust `try_into usize` at `fetch.rs` but eventually have to use i64.
I will keep using i64 here for now.
```rs
  152         TableFactor::Series { size, .. } => {                                                         
  153             let value: Value = evaluate_stateless(None, size)?.try_into()?;                           
  154             let size: i64 = value.try_into()?;                                                        
  155             let size = match size {                                                                   
  156                 n if n >= 0 => size,                                                                  
  157                 n => return Err(FetchError::SeriesSizeWrong(n).into()),                               
  158             };                                                                                        
  159             let rows = (1..=size).map(|v| Ok(Row(vec![Value::I64(v)])));                              
  160                                                                                                       
  161             Ok(Rows::Series(stream::iter(rows)))                                                      
  162         } 
```

panarch (approved):
Nice improvement, thanks a lot 👍

PR #797, opened 2022-08-24
Small improvement: add clear guide for web example module

24seconds:
## Description

When I look into the `gluesql-js`, I felt it is quite unclear about how to run the `example/web/module/index.html`.
And the index html under the `gluesql-js/` made me confused.

So I did two things
1. Remove `index.html` under `gluesql-js`
2. Write `how to guide` README.md under `gluesql-js/examples/web/modules/`


panarch (approved):
This can definitely help users to easily try using web browser module to use GlueSQL.
Thanks really a lot 👍 
Also thanks for removing unmaintained index.html file. I wasn't aware of that.

PR #800, opened 2022-08-27
[AST-Builder] change error message in test functions

### LINKED ISSUE #799: [AST-Builder] change error message in test functions
## Background
Currently, when we write invalid sql statements for ast-builder test case, parse error occurs but error message seems not clear. 
ex. 
```rust
fn create_table_without_column() {
        let actual = table("Foo").create_table().build();
        let expected = "CREATE TABLES Foo";  // wrong
        test(actual, expected);
    }
```
```called `Result::unwrap()` on an `Err` value: Parser("ParserError(\n    \"Expected end of statement, found: NOT\",\n)")```

## Task
replace `unwrap()` with `expect(something)` in test functions for ast builder
```rust
// before
let parsed = &parse_expr(expected).unwrap();

// after
let parsed = &parse_expr(expected).expect(expected);
```

ding-young:
## Issue
resolves #799

### DIFF
--- a/core/src/ast_builder/assignment.rs
+++ b/core/src/ast_builder/assignment.rs
@@ -45,7 +45,7 @@ mod tests {
     };
 
     fn test(actual: AssignmentNode, expected: &str) {
-        let parsed = &parse_sql_assignment(expected).unwrap();
+        let parsed = &parse_sql_assignment(expected).expect(expected);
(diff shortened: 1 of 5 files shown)


panarch (approved):
Nice improvement which can help ast builder contributors to recognize errors easier.
Looks all good, thanks a lot!

Issue #766, opened 2022-08-15
[AST Builder] Implement join Expression in Select


seonghun-dev:
# Description
Implements ast builders for `join`, `left join` Expression
Using `join_as` to express table alias
## Background
Refered https://github.com/gluesql/gluesql/issues/614

## Task
In AST Builder module to mod function

The following tests should be passed:
### Inner Join
``` rust
    #[test]
    fn inner_join() {
        let actual = table("Item")
            .select()
            .join_as("Player", "p")
            .on("p.id = Item.player_id")
            .filter("p.id = 1")
            .build();
        let expected = "
        SELECT * FROM Item INNER JOIN Player AS p ON p.id = Item.player_id WHERE p.id = 1;
        ";
        test(actual, expected);

        let actual = table("student")
            .select()
            .join("marks")
            .on("students.id = marks.id")
            .join("attendance")
            .on("marks.id = attendance.id")
            .filter("attendance.attendance >= 75")
            .project(vec![
                "students.id",
                "students.name",
                "marks.rank",
                "attendance.attendance",
            ])
            .build();
        let expected = "
            SELECT s.id, s.name, m.rank, a.attendance
            FROM students AS s
            INNER JOIN marks AS m ON s.id=m.id
            INNER JOIN attendance AS a on m.id=a.id
            WHERE a.attendance>=75;
        ";
        test(actual, expected);

        let acutal = table("Orders")
            .select()
            .join("Customers")
            .project(vec![
                "Orders.OrderID",
                "Customers.CustomerName",
                "Orders.OrderDate",
            ])
            .build();
        let expected = "
            SELECT Orders.OrderID, Customers.CustomerName, Orders.OrderDate 
            FROM Orders INNER JOIN Customers
        ";
    }
```
### Left Outer Join
``` rust

    fn left_join() {
        let actual = table("player")
            .select()
            .projection("player.id", "item.id")
            .left_join("item")
            .on("player.id = item.id")
            .build();
        let expected = "
            SELECT p.id, i.id
            FROM Player p
            LEFT JOIN Item i
            ON p.id = i.player_id
        ";
        test(actual, expected);

        let actual = table("Item")
            .select()
            .left_join("Player")
            .on("Item.player_id = Player.id")
            .left_join_as("Player", "p1")
            .on("p1.id = Item.player_id")
            .left_join_as("Player", "p2")
            .on("p2.id = Item.player_id")
            .left_join_as("Player", "p3")
[...]

PR #803, opened 2022-08-27
[Ast Builder] Implement join expression in select statement

seonghun-dev:
Resolve : https://github.com/gluesql/gluesql/issues/766

## FSM
Project node is last statement, so eliminate in this fsm graph

GlueSQL-1


core/src/ast_builder/select/root.rs
```diff
@@ -1,36 +1,28 @@
...
-        ast::{Expr, ObjectName, SelectItem, Statement, TableFactor, TableWithJoins},
+        ast::{ObjectName, SelectItem, Statement, TableFactor},
         ast_builder::{
-            ExprList, ExprNode, GroupByNode, LimitNode, OffsetNode, ProjectNode, SelectItemList,
+            ExprList, ExprNode, FilterNode, GroupByNode, JoinNode, LimitNode, OffsetNode,
+            ProjectNode, SelectItemList,
         },
         result::Result,
     },
+    std::vec,
```
ever0de, line 11:
```suggestion
```

`std::vec` does not need to be imported.

core/src/ast_builder/select/root.rs
```diff
@@ -62,61 +80,74 @@ impl Prebuild for SelectNode {
...
-            .filter(Expr::BinaryOp {
-                left: Box::new(Expr::Identifier("col1".to_owned())),
-                op: BinaryOperator::Gt,
-                right: Box::new(Expr::Identifier("col2".to_owned())),
-            })
-            .build();
-        let expected = "SELECT * FROM Foo WHERE col1 > col2";
+        // select node to left join node with alias
+        let actual = table("Foo").select().left_join_as("Bar", "b").build();
+        let expected = "SELECT * FROM Foo LEFT JOIN Bar AS b";
```
ever0de, line 150: For test cases, there is a rule that only writes cases that can access `Node` in the current module, but this is not the case here.

However, since developers are expected to see this module first when using the `select` function, it is desirable to have a test case.

Let's decide what to do.

ever0de (approved):
Looks good except for the discussion above.

core/src/ast_builder/select/join.rs
```diff
@@ -0,0 +1,543 @@
...
+        ast_builder::{
+            ExprList, ExprNode, FilterNode, GroupByNode, JoinConstraintNode, LimitNode, OffsetNode,
+            ProjectNode, SelectItemList, SelectNode,
+        },
+        result::Result,
+    },
+};
+
+#[derive(Clone)]
+pub enum JoinType {
```
panarch, line 17: 1. For this kind of small enums, it would be good to dervice `Copy` trait.
Then, we don't need to use `clone()`.

e.g.
```rust
#[derive(Clone, Copy)]
pub enum JoinType { .. }
```

2. Name to `JoinOperatorType`.
There exists two different types which describes join.
One is join operator type which consists of inner or outer joins.
The other is join execution type, currently we support two join execution types - nested loop & hash joins.

So it would be better to use more explicit name for this.
This `JoinType` is for describing join operator types, then how about using `JoinOperatorType`?

core/src/ast_builder/select/join.rs
```diff
@@ -0,0 +1,543 @@
...
+        PrevNode::JoinConstraint(Box::new(node))
+    }
+}
+
+#[derive(Clone)]
+pub struct JoinNode {
+    prev_node: PrevNode,
+    relation: TableFactor,
+    join_operator: JoinOperator,
+    join_operator_type: JoinType,
```
panarch, line 62: This `JoinNode` does not have JoinOperator with `ON` condition.

Only `JoinOperator::Inner(JoinConstraint::None) | JoinOperator::LeftOuter(JoinConstraint::None)` are being used.
Then, we don't need `join_operator: JoinOperator` field in `JoinNode` because that field can be simply generated from `join_operator_type: JoinType`.

panarch (approved):
Awesome!! All looks great 👍 👍 👍 
Now AST builder support all essential SELECT query building functions!
Thanks a lot 👍

PR #805, opened 2022-08-29
Support ORDER BY `ALIAS` and `COLUMN_INDEX`

devgony:
### Support ORDER BY `ALIAS` clause like below
```sql
SELECT column_name AS alias_name FROM Table ORDER BY alias_name DESC
```
- Currently, it throws `[error] value not found: alias_name`
### Original `column_name` is still available at ORDER BY clause though SELECT clause uses `alias_name`
```sql
SELECT column_name AS alias_name FROM Table ORDER BY column_name DESC
```
### Support ORDER BY `COLUMN_INDEX`
 ```sql
SELECT alpha, beta FROM Table ORDER BY 1 DESC
```
- `1` means the column_index which is first column `alpha`

## Todo
-  [x] swap sequence of sort and blend to obtain Row
- [x] concat label_context
- [x] should support `ORDER BY column_name` as before even with alias
- [x] order by column index
- [x] remove clone, unwrap

### CODE REVIEW

File: core/src/executor/sort.rs [hunk 1]
```diff
@@ -41,48 +48,100 @@ impl<'a> Sort<'a> {
...
-            .and_then(move |AggregateContext { aggregated, next }| async move {
-                let blend_context = Rc::clone(&next);
+            .and_then(|(AggregateContext { aggregated, next }, row)| {
+                // let table_alias = next.get_table_alias();
+                let labels = Rc::clone(&labels);
                 let filter_context = Rc::new(FilterContext::concat(
                     self.context.as_ref().map(Rc::clone),
                     Some(Rc::clone(&next)),
                 ));
+                let label_context = BlendContext::new(table_alias, labels, Some(row.clone()), None);
```
devgony, line 74: It looks bad to clone `row`. I'm still struggling to fix it.

core/src/executor/sort.rs
```diff
@@ -41,48 +48,100 @@ impl<'a> Sort<'a> {
...
+                        .await?;
+
+                    Ok((values, row))
+                }
             })
             .try_collect::<Vec<_>>()
             .await
             .map(Vector::from)?
-            .sort_by(|(values_a, _, _), (values_b, _, _)| {
+            .sort_by(|(values_a, ..), (values_b, ..)| {
```
panarch, line 144: Wow, I didn't know that `..` also works in tuples. Looks better!

core/src/executor/sort.rs
panarch: Oh, I think we don't need to make `BlendContext`.
How about making `FilterContext` and chaining it with existing filter_context.
Then we don't need to use `FilterContext::concat`, it can be used just as `next`

core/src/executor/sort.rs
devgony: lol but it's a bit weird that 
```rs
.map(|(.., row)| Ok(row));
```
works but
```rs
.map(|(row, ..)| Ok(row));
```
doesn't 🤣 
https://github.com/gluesql/gluesql/pull/805/files#diff-71b79d5c3dcf3ec4810d6e9149f90c16fff0dae2a728aa95f670d742baafba81R146

panarch: There still exists the thing the Rust compiler can be smarter 😁

core/src/executor/sort.rs
panarch: image
We can simply use `FilterContext::new`

core/src/executor/select/mod.rs
```diff
@@ -298,20 +298,10 @@ pub async fn select_with_labels<'a>(
...
-    // let rows = sort
-    //     .apply(rows)
-    //     .await?
-    //     .and_then(move |(aggregated, context)| {
-    //         let blend = Rc::clone(&blend);
-
-    //         async move { blend.apply(aggregated, context).await }
-    //     });
+    let labels = Rc::new(labels);
+    let rows = sort.apply(rows, labels, get_alias(relation)?).await?;
```
panarch, line 302: Is this compiled?
```rust
    let rows = sort.apply(rows, Rc::clone(&labels), get_alias(relation)?).await?;
```
`Rc::clone` looks needed

devgony: Ahh! I wasn't using Rc at all..
It works now! Thank you so much 🙇🏼

core/src/executor/context/blend_context.rs
```diff
@@ -3,19 +3,25 @@ use {
     std::{fmt::Debug, rc::Rc},
 };
 
+#[derive(Debug)]
+pub enum BlendContextRow {
+    Single(Option<Row>),
+    Shared(Option<Rc<Row>>),
```
panarch, line 9: Currently we don't have ues case of `BlendContextRow::Shared(None)`.
So, `Shared(Rc<Row>)` would be enough for now.

core/src/executor/context/blend_context.rs
```diff
@@ -25,15 +31,21 @@ impl<'a> BlendContext<'a> {
...
                 .iter()
                 .position(|column| column == target)
                 .map(|index| match &self.row {
-                    Some(row) => row.get_value(index),
-                    None => Some(&Value::Null),
+                    BlendContextRow::Single(Some(row)) => row.get_value(index),
+                    BlendContextRow::Shared(Some(row)) => row.get_value(index),
+                    BlendContextRow::Single(None) | BlendContextRow::Shared(None) => {
+                        Some(&Value::Null)
+                    }
```
panarch, line 48: Then, we don't need to consider `BlendContext::Shared(None)` case so we can simplify like..

```rust
BlendContextRow::Shared(row) => row.get_value(index),
BlendContextRow::Single(Some(row)) => row.get_value(index),
BlendContextRow::Single(None) => Some(&Value::Null),
```

core/src/executor/context/blend_context.rs
```diff
@@ -87,8 +105,11 @@ impl<'a> BlendContext<'a> {
 
     pub fn get_all_values(&'a self) -> Vec<Value> {
         let values: Vec<Value> = match &self.row {
-            Some(Row(values)) => values.clone(),
-            None => self.columns.iter().map(|_| Value::Null).collect(),
+            BlendContextRow::Single(Some(row)) => row.0.clone(),
+            BlendContextRow::Shared(Some(row)) => row.as_ref().0.clone(),
+            BlendContextRow::Single(None) | BlendContextRow::Shared(None) => {
+                self.columns.iter().map(|_| Value::Null).collect()
+            }
```
panarch, line 112: This too
```rust
BlendContextRow::Shared(row) => row.as_ref().0.clone(),
BlendContextRow::Single(Some(row)) => row.0.clone(),
BlendContextRow::Single(None) => self.columns.iter().map(|_| Value::Null).collect(),
```

core/src/executor/context/blend_context.rs
```diff
@@ -25,15 +31,19 @@ impl<'a> BlendContext<'a> {
             next,
         }
     }
+    // pub fn get_table_alias(&'a self) -> &'a str {
+    //     self.table_alias
+    // }
```
panarch, line 36: nit. we can delete unused comments :D

core/src/executor/sort.rs
```diff
@@ -1,31 +1,38 @@
...
+        result::{Error, Result},
         store::GStore,
     },
+    bigdecimal::ToPrimitive,
     futures::stream::{self, Stream, StreamExt, TryStreamExt},
-    im_rc::HashMap,
-    std::{cmp::Ordering, pin::Pin, rc::Rc},
+    serde::Serialize,
+    std::fmt::Debug,
+    std::{cmp::Ordering, rc::Rc},
```
panarch, line 17: It looks that we can merge these into a single one.
```rust
std::{
    fmt::Debug,
    cmp::Ordering,
    rc::Rc,
}
```

core/src/executor/sort.rs
```diff
@@ -41,52 +48,116 @@ impl<'a> Sort<'a> {
...
+                        match big_decimal {
+                            Some(n) => {
+                                let index = n.to_usize().ok_or_else(|| {
+                                    crate::result::Error::from(SortError::Unreachable)
+                                })?;
+                                let zero_based = index.checked_sub(1).ok_or_else(|| {
+                                    crate::result::Error::from(SortError::ColumnIndexOutOfRange(
+                                        index,
+                                    ))
+                                })?;
```
panarch, line 97: This would be enough to use
```rust
SortError::ColumnIndexOutOfRange(..).into()
```
rather than
`crate::result::Error::from`

core/src/executor/sort.rs
```diff
@@ -41,52 +48,116 @@ impl<'a> Sort<'a> {
...
+
+                                Ok((SortType::Value(value.clone()), *asc))
+                            }
+                            None => Ok((SortType::Expr(expr), *asc)),
+                        }
+                    })
+                    .collect::<Result<Vec<_>>>();
+                // let table_alias = next.get_table_alias();
```
panarch, line 110: Now it's almost close to the complete. Let's remove some remaining comments :D

test-suite/src/order_by.rs
```diff
@@ -150,4 +150,60 @@ CREATE TABLE Test (
...
+        "SELECT id AS C1, num AS C2 FROM Test ORDER BY id ASC, num DESC",
+        Ok(select!(
+            C1  | C2
+            I64 | I64;
+            1     9;
+            1     2;
+            3     4;
+            4     7
+        ))
+    );
```
panarch, line 176: Thanks to @zmrdltl 
New `test!` macro is now available which can specify test name.
This test already has a good title, so.. how about to start using the new `test!`?

```rust
test! {
    name: "original column_names still work even if aliases were used at SELECT clause",
    sql: "
        SELECT id AS C1, num AS C2
        FROM Test
        ORDER BY id ASC, num DESC
    ",
    expected:
        Ok(select!(
            C1  | C2
            I64 | I64;
            1     9;
            1     2;
            3     4;
            4     7
        ))
};
```

core/src/executor/sort.rs
```diff
@@ -41,52 +47,111 @@ impl<'a> Sort<'a> {
...
     pub async fn apply(
         &self,
-        rows: impl Stream<Item = Result<AggregateContext<'a>>> + 'a,
-    ) -> Result<Pin<Box<dyn Stream<Item = Item<'a>> + 'a>>> {
-        if self.order_by.is_empty() {
-            let rows = rows.map_ok(|aggregate_context| {
-                let AggregateContext { aggregated, next } = aggregate_context;
+        rows: impl Stream<Item = Result<(AggregateContext<'a>, Row)>> + 'a,
+        labels: Rc<Vec<String>>,
+        table_alias: &'a str,
```
devgony, line 52: I'm still not sure if it's redundant to get table_alias here or not 🤔

core/src/executor/sort.rs
```diff
@@ -41,52 +47,111 @@ impl<'a> Sort<'a> {
...
-            .and_then(move |AggregateContext { aggregated, next }| async move {
-                let blend_context = Rc::clone(&next);
+            .and_then(|(AggregateContext { aggregated, next }, row)| {
+                enum SortType<'a> {
+                    Value(Value),
+                    Expr(&'a Expr),
+                }
+
+                let order_by = self
+                    .order_by
```
panarch, line 73: Could we.. have some more line breaks here? It would be better to read if we have some more line breaks.

```rust
        if self.order_by.is_empty() {
            let rows = rows.map_ok(|(_, row)| row);

            return Ok(Rows::NonOrderBy(Box::pin(rows)));
        }

        let rows = rows
            .and_then(|(AggregateContext { aggregated, next }, row)| {
                enum SortType<'a> {
                    Value(Value),
                    Expr(&'a Expr),
                }

                let order_by = self
                    .order_by
                    .iter()
                    .map(|OrderByExpr { expr, asc }| -> Result<_> {
                        let big_decimal = match expr {
                            Expr::Literal(AstLiteral::Number(n)) => Some(n),
                            Expr::UnaryOp {
                                op: UnaryOperator::Plus,
                                expr,
                            } => match expr.as_ref() {
                                Expr::Literal(AstLiteral::Number(n)) => Some(n),
                                _ => None,
                            },
                            _ => None,
                        };

                        match big_decimal {
                            Some(n) => {
                                let index = n
                                    .to_usize()
                                    .ok_or_else(|| -> Error { SortError::Unreachable.into() })?;
                                let zero_based = index.checked_sub(1).ok_or_else(|| -> Error {
                                    SortError::ColumnIndexOutOfRange(index).into()
                                })?;
                                let value = row.get_value(zero_based).ok_or_else(|| -> Error {
                                    SortError::ColumnIndexOutOfRange(index).into()
                                })?;

                                Ok((SortType::Value(value.clone()), *asc))
                            }
                            None => Ok((SortType::Expr(expr), *asc)),
                        }
                    })
                    .collect::<Result<Vec<_>>>();

                let labels = Rc::from(labels.as_slice());
                let filter_context = Rc::new(FilterContext::concat(
                    self.context.as_ref().map(Rc::clone),
                    Some(Rc::clone(&next)),
                ));
                let aggregated = aggregated.map(Rc::new);

                async move {
                    let row = Rc::new(row);
                    let label_context = Rc::from(BlendContext::new(
                        table_alias,
                        labels,
                        Shared(Rc::clone(&row)),
                        None,
                    ));
                    let filter_context = Rc::new(FilterContext::concat(
                        Some(filter_context),
                        Some(Rc::clone(&label_context)),
                    ));

                    let values = stream::iter(order_by?.into_iter())
                        .then(|(sort_type, asc)| {
                            let context = Some(Rc::clone(&filter_context));
                            let aggregated = aggregated.as_ref().map(Rc::clone);

                            async move {
                                let value = match sort_type {
                                    SortType::Value(value) => value,
                                    SortType::Expr(expr) => {
                                        evaluate(self.storage, context, aggregated, expr)
                                            .await?
                                            .try_into()?
                                    }
                                };

                                Ok::<Value, Error>((value, asc))
                            }
                        })
                        .try_collect::<Vec<_>>()
                        .await?;

                    drop(label_context);
                    drop(filter_context);

                    let row = Rc::try_unwrap(row).map_err(|_| SortError::Unreachable)?;

                    Ok((values, row))
                }
            })
```

core/src/executor/select/mod.rs
```diff
@@ -287,15 +287,23 @@ pub async fn select_with_labels<'a>(
...
+        let AggregateContext { aggregated, next } = aggregate_context;
+        let rc_aggregated = aggregated.clone().map(Rc::new);
+
+        async move {
+            let row = blend.apply(rc_aggregated.clone(), Rc::clone(&next)).await?;
 
-            async move { blend.apply(aggregated, context).await }
-        });
+            Ok((AggregateContext { aggregated, next }, row))
+        }
```
panarch, line 299: Currently `sort.rs` apply method takes `AggregateContext` as a param.
image

As we do in `blend.rs` apply method,
image

It looks that if we change the `Sort::apply` the types of input params similar with `Blend::apply`, then we might not need to clone aggregated in here.

core/src/executor/context/blend_context.rs
```diff
@@ -73,8 +81,9 @@ impl<'a> BlendContext<'a> {
     pub fn get_alias_values(&self, alias: &str) -> Option<Vec<Value>> {
         if self.table_alias == alias {
             let values = match &self.row {
-                Some(Row(values)) => values.clone(),
-                None => self.columns.iter().map(|_| Value::Null).collect(),
+                BlendContextRow::Shared(row) => row.as_ref().0.clone(),
```
ever0de, line 84:
```suggestion
                BlendContextRow::Shared(row) => row.0.clone(),
```

I don't think `as_ref` is necessary.

panarch (approved):
Thanks a lot for the meaningful change
All looks awesome!
Breaking change for the next release 👍 👍 👍

PR #812, opened 2022-09-03
[AST] Implement to_sql func for OrderByExpr

### LINKED ISSUE #809: [AST] Implement to_sql func for OrderByExpr
## Description 

Implement to_sql func for OrderByExpr

## Task

```rust
#[test]
fn to_sql_order_by_expr() {
    let actual = "foo ASC".to_string();
    let expected = OrderByExpr {
        expr: Expr::Identifier("foo".to_string()),
        asc: Some(true),
    }
    .to_sql();
    assert_eq!(actual, expected);

    let actual = "foo DESC".to_string();
    let expected = OrderByExpr {
        expr: Expr::Identifier("foo".to_string()),
        asc: Some(false),
    }
    .to_sql();
    assert_eq!(actual, expected);

    let actual = "foo".to_string();
    let expected = OrderByExpr {
        expr: Expr::Identifier("foo".to_string()),
        asc: None,
    }
    .to_sql();
    assert_eq!(actual, expected);
}
```

CEOJINSUNG:
Resolve #809


ever0de (approved):
Looks good :+1:

panarch (changes requested):
Current ToSql impl of `VALUES` is different from what it actually does.
e.g.
```rust
VALUES (1, NULL), (2, 'a'), (3, 4)
```

More information is here in `test-suite/ values.rs`
ref. https://github.com/CEOJINSUNG/gluesql/blob/features/query-to-sql/test-suite/src/values.rs

core/src/ast/mod.rs
```diff
@@ -224,9 +224,10 @@ impl ToSql for Statement {
                 column: _column,
             } => {
                 format!(
-                    "CREATE INDEX {} ON {} (..order_by_expr..)",
+                    "CREATE INDEX {} ON {} {}",
                     name.to_sql(),
-                    table_name.to_sql()
+                    table_name.to_sql(),
+                    _column.to_sql()
```
panarch, line 230: Now the `_column` is being used, so we can remove `_` underscore prefix.
`column.to_sql()` would be enough

panarch (approved):
Awesome! Looks all nice, now `CREATE INDEX` ToSQL impl has completed :D Thanks a lot!

PR #817, opened 2022-09-04
Migrate sqlparser-rs from 0.18 to 0.19,

panarch:
Add negated field to ast::Expr::Exists.
Update Interval to accept numeric value - e.g. INTERVAL 3 DAY
(previously, it only accepted string value - e.g. INTERAVAL "3" DAY)


ever0de (approved):
Looks good :+1:

PR #818, opened 2022-09-05
[Function] Implement `FORMAT` function

### LINKED ISSUE #806: [Function] implement `FORMAT`  function
Support `format` function that formats the `date` , `timestamp` and `datetime` features.

For the reference,
In MySQL its expressed as [DATE_FORMAT](https://www.w3schools.com/sql/trymysql.asp?filename=trysql_func_mysql_date_format).

ChobobDev:
This resolves #806


test-suite/src/function/format.rs
```diff
@@ -0,0 +1,16 @@
...
+
+test_case!(format, async move {
+    use gluesql_core::{
+        executor::{EvaluateError, SelectError},
+        prelude::Value::*,
+    };
+
+    let test_cases = vec![(
+        r#"SELECT FORMAT(DATE "2017-06-15", "%Y-%m")"#,
+        Ok(select!("2017-06")),
```
ever0de, line 11: The column name and type are missing from the `select!` macro.

https://github.com/gluesql/gluesql/blob/7fadbf557505c35569fbcedb2e9c06604ea74ac6/test-suite/src/series.rs#L131-L138

ChobobDev: Thank you for pointing this out, I will apply this to the new test cases

ChobobDev:
I think I am ready for review,
If you think I need more test cases, please give me some recommendations.

Thank you

test-suite/src/function/format.rs
```diff
@@ -0,0 +1,31 @@
...
+            Ok(select!(
+                "TIMESTAMP"
+                Str;
+                "2015-09-05 23".to_owned()
+            )),
+        ),
+        (
+            r#"SELECT FORMAT("2015-09-05 23:56:04", "%Y-%m-%d %H") AS TIMESTAMP"#,
+            Err(EvaluateError::FunctionRequiresFormattableValue("FORMAT".to_string()).into()),
+        ),
```
devgony, line 26: It would be nice if there are plenty of test cases.
eg1) If I don't know how to format minute and second, i could glance test cases
eg2) Someone may want to get `Day of week`
image
(https://livesql.oracle.com/apex/livesql/file/content_GCEY1DN2CN5HZCUQFHVUYQD3G.html)

ChobobDev: As we are using `Chrono` to format the date I read through the document of the `Chrono` and found [this](https://docs.rs/chrono/latest/chrono/format/strftime/index.html). I will try to add the cases with the use of the following document.  But I am still not sure whether I should put all of the possible cases to our unit test.

ever0de: We don't have to consider every case of `strftime`. (This is because `chrono` is doing well)
Instead, I think we should decide how to handle an invalid argument in the format argument.
What does the `format` method of NativeDate and DateTime currently used do when an invalid argument is received?

core/src/executor/evaluate/function.rs
```diff
@@ -410,3 +413,27 @@ pub fn unwrap(name: String, expr: Evaluated<'_>, selector: Evaluated<'_>) -> Res
 pub fn generate_uuid() -> Value {
     Value::Uuid(Uuid::new_v4().as_u128())
 }
+
+pub fn function_format(name: String, expr: Evaluated<'_>, format: Evaluated<'_>) -> Result<Value> {
```
devgony, line 417: Looks like it's okay to use name `format`

test-suite/src/function/format.rs
devgony: I agree it can be redundant to list all cases but i thought each representative of usages can have meaning.
Or at least [format table](https://docs.rs/chrono/latest/chrono/format/strftime/index.html) you found should be included in our doc someday.

ChobobDev: > We don't have to consider every case of `strftime`. (This is because `chrono` is doing well) Instead, I think we should decide how to handle an invalid argument in the format argument. What does the `format` method of NativeDate and DateTime currently used do when an invalid argument is received?

It currently returns the error message that `Chrono` returns.

ChobobDev: when the invalid argument is given, the `Format` returns the wrong argument.
for example if `Y-M` is given to the second argument  which is invalid, it returns 
`Ok(Select { labels: ["TIMESTAMP"], rows: [[Str("Y-M")]] })`

core/src/executor/evaluate/error.rs
```diff
@@ -57,4 +57,10 @@ pub enum EvaluateError {
 
     #[error("subquery returns more than one row")]
     MoreThanOneRowReturned,
+
+    #[error("function requires formattable value: {0}")]
+    FunctionRequiresFormattableValue(String),
+
+    #[error("given specifier is not supported: {0}")]
```
ChobobDev, line 64: I added a new error type and this will be triggered when the invalid specifier is given.

core/src/executor/evaluate/function.rs
```diff
@@ -410,3 +411,38 @@ pub fn unwrap(name: String, expr: Evaluated<'_>, selector: Evaluated<'_>) -> Res
...
+    check_valid_specifier(b, date)
+}
+
+fn format_timestamp(a: NaiveDateTime, b: String) -> Result<Value> {
+    let b = b.as_str();
+    let date = chrono::NaiveDateTime::format(&a, b).to_string();
+    check_valid_specifier(b, date)
+}
+
+fn check_valid_specifier(a: &str, b: String) -> Result<Value> {
```
ChobobDev, line 441: The error handling is done by checking the output from `Chrono` and the specifier that was given as an argument of the function. The `Chrono` returns the specifier given when the `NaiveDate` or NaiveDateTIme` is not formattable.

ChobobDev:
I think I am ready for the second review.
Please give me suggestions on the changes made by the recent commits I pushed,

Thank you

ChobobDev:
As guided, I have removed the error handling part when an invalid specifier is given since we thought this improvement should be done by `Chrono`.

core/src/executor/evaluate/error.rs
```diff
@@ -57,4 +57,7 @@ pub enum EvaluateError {
 
     #[error("subquery returns more than one row")]
     MoreThanOneRowReturned,
+
+    #[error("function does not support following DataType: {0}")]
```
ever0de, line 61: It seems that the information about the `Format` function is missing from the display information.
So it seems to be a very broad error type.

ever0de (approved):
LGTM

panarch (approved):
All looks great! New function has come 👍 
Thanks a lot!

PR #819, opened 2022-09-06
Upgrade sqlparser-rs to v0.22

panarch:
Upgrade sqlparser-rs from v0.19 to v0.22
Migrate translate_trim, add new supported test case.
Move Like, ILike from ast::BinaryOperator to ast::Expr.
Now "User" is a reserved keyword, so rename test cases which used "User" to use "Player".

### DIFF
--- a/core/src/ast/mod.rs
+++ b/core/src/ast/mod.rs
@@ -217,11 +217,15 @@ impl ToSql for Assignment {
 #[cfg(test)]
 mod tests {
     #[cfg(feature = "alter-table")]
-    use {crate::ast::AlterTableOperation, bigdecimal::BigDecimal, std::str::FromStr};
+    use crate::ast::AlterTableOperation;
 
-    use crate::ast::{
-        Assignment, AstLiteral, BinaryOperator, ColumnDef, ColumnOption, ColumnOptionDef, DataType,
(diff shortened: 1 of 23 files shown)


ever0de:
Conflict seems to need to be resolved

core/src/ast/mod.rs
```diff
@@ -167,13 +167,12 @@ impl ToSql for Statement {
...
 #[cfg(test)]
 mod tests {
-    use {
-        crate::ast::{
-            AlterTableOperation, AstLiteral, ColumnDef, ColumnOption, ColumnOptionDef, DataType,
-            Expr, ObjectName, Query, SetExpr, Statement, ToSql, Values,
-        },
-        bigdecimal::BigDecimal,
-        std::str::FromStr,
+    #[cfg(feature = "alter-table")]
```
ever0de, line 170: Oh, that was the missing part(prev review). thank you

core/src/plan/join.rs
```diff
@@ -467,23 +467,23 @@ mod tests {
     #[test]
     fn basic() {
         let storage = run("
-            CREATE TABLE User (
+            CREATE TABLE Player (
```
ever0de, line 470: Now that User has become a reserved word, it has been changed because it cannot be used.

ever0de (approved):
Looks good. There are many changes

PR #831, opened 2022-09-11
Simplify root doc comment code example

panarch:
Hide unnecessary info to users - fn main, cfg info, ...

ref. https://docs.rs/gluesql/latest/gluesql/
prev.
image

now.
image


ever0de (approved):
I didn't know this feature existed. looks really good

PR #832, opened 2022-09-11
Improvement `validate_unique` function when only `PrimaryKey` exists

ever0de:
The existing `validate_unique` function creates a unique constraint by looping through all rows even when only `PrimaryKey` exists.

Use `Store::fetch_data` to improve speed in the above situation.

### DIFF
--- a/core/src/executor/validate.rs
+++ b/core/src/executor/validate.rs
@@ -1,6 +1,6 @@
 use {
     crate::{
-        ast::{ColumnDef, ColumnOption},
+        ast::{ColumnDef, ColumnOption, ColumnOptionDef},
         data::{Key, Row, Value},
         result::Result,
         store::Store,
@@ -17,12 +17,17 @@ pub enum ValidateError {
     #[error("conflict! storage row has no column on index {0}")]
     ConflictOnStorageColumnIndex(usize),
 
-    #[error("duplicate entry '{0:?}' for unique column '{1}'")]
+    #[error("duplicate entry '{}' for unique column '{1}'", String::from(.0))]
     DuplicateEntryOnUniqueField(Value, String),
+
+    #[error("duplicate entry '{0:?}' for primary_key field")]
+    DuplicateEntryOnPrimaryKeyField(Key),
 }
 
 pub enum ColumnValidation {
+    /// `INSERT`
     All(Rc<[ColumnDef]>),
+    /// `UPDATE`
     SpecifiedColumns(Rc<[ColumnDef]>, Vec<String>),
 }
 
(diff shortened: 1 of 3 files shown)


core/src/executor/validate.rs
```diff
@@ -74,39 +79,93 @@ impl UniqueConstraint {
...
+                .count();
+
+            match (primary_key_index, other_unique_column_def_count) {
+                (Some(primary_key_index), 0) => Columns::PrimaryKeyOnly(primary_key_index),
+                _ => Columns::All(fetch_all_unique_columns(column_defs)),
+            }
         }
+        ColumnValidation::SpecifiedColumns(column_defs, specified_columns) => Columns::All(
+            fetch_specified_unique_columns(column_defs, specified_columns),
+        ),
```
ever0de, line 127: `SpecifiedColumns` is for `UPDATE` Statement, but PrimaryKey cannot be updated, so it is not checked.

core/src/executor/validate.rs
```diff
@@ -74,39 +79,93 @@ impl UniqueConstraint {
...
+            for primary_key in row_iter.filter_map(|row| match row.0.get(primary_key_index) {
+                Some(value) => Key::try_from(value).map(Some).transpose(),
+                None => None,
+            }) {
+                let key = primary_key?;
+
+                if storage.fetch_data(table_name, &key).await?.is_some() {
+                    return Err(ValidateError::DuplicateEntryOnPrimaryKeyField(key).into());
+                }
+            }
```
ever0de, line 141: [futures::join](https://docs.rs/futures/latest/futures/macro.join.html) can be used to improve concurrency.

core/src/executor/validate.rs
```diff
@@ -74,39 +79,93 @@ impl UniqueConstraint {
...
-            .try_for_each(|constraint| {
-                let col_idx = constraint.column_index;
-                let val = row
-                    .get_value(col_idx)
-                    .ok_or(ValidateError::ConflictOnStorageColumnIndex(col_idx))?;
+    match columns {
+        Columns::PrimaryKeyOnly(primary_key_index) => {
+            for primary_key in row_iter.filter_map(|row| match row.0.get(primary_key_index) {
+                Some(value) => Key::try_from(value).map(Some).transpose(),
+                None => None,
```
ever0de, line 134: If the entered data is not wrong, it must be inaccessible code?

panarch: Same code but simple simplification
```rust
for primary_key in row_iter.filter_map(|row| row.0.get(index).map(Key::try_from)) {
    let key = primary_key?;

    if storage.fetch_data(table_name, &key).await?.is_some() {
        return Err(ValidateError::DuplicateEntryOnPrimaryKeyField(key).into());
    }
}
```

panarch (approved):
Nice performance improvement! Thanks really a lot 👍 👍

PR #833, opened 2022-09-11
[Function] Implement `TO_DATE` and `TO_TIMESTAMP` function

### LINKED ISSUE #829: [Function] implement `TO_DATE` and `TO_TIMESTAMP` function
Support `to_date` and `to_timestamp` function that converts string into `DATE` and `TIMESTAMP` datatype. 

For your reference,
the following function is expressed as [STR_TO_DATE](https://www.w3schools.com/sql/trymysql.asp?filename=trysql_func_mysql_str_to_date).
Also following is an example converting string to timestamp [STR_TO_DATE ](https://www.w3schools.com/sql/trymysql.asp?filename=trysql_func_mysql_str_to_date4).

Thank you

ChobobDev:
Resolves #829

### DIFF
--- a/core/src/executor/evaluate/error.rs
+++ b/core/src/executor/evaluate/error.rs
@@ -7,6 +7,9 @@ use {
 
 #[derive(Error, Serialize, Debug, PartialEq)]
 pub enum EvaluateError {
+    #[error(transparent)]
+    ChronoFormat(#[from] ChronoFormatError),
+
     #[error("nested select row not found")]
     NestedSelectRowNotFound,
 
@@ -60,4 +63,61 @@ pub enum EvaluateError {
 
     #[error("format function does not support following data_type: {0}")]
     UnsupportedExprForFormatFunction(String),
+}
+
+#[derive(Error, Serialize, Debug, PartialEq)]
+pub enum ChronoFormatError {
+    /// Given field is out of permitted range.
+    #[error("given field is out of permitted range")]
+    OutOfRange,
+
+    /// There is no possible date and time value with given set of fields.
+    ///
+    /// This does not include the out-of-range conditions, which are trivially invalid.
+    /// It includes the case that there are one or more fields that are inconsistent to each other.
+    #[error("the given date and time value is impossible to be formmated")]
+    Impossible,
+
+    /// Given set of fields is not enough to make a requested date and time value.
+    ///
+    /// Note that there *may* be a case that given fields constrain the possible values so much
+    /// that there is a unique possible value. Chrono only tries to be correct for
+    /// most useful sets of fields however, as such constraint solving can be expensive.
+    #[error("given set of field is not enough to be formatted")]
+    NotEnough,
+
+    /// The input string has some invalid character sequence for given formatting items.
+    #[error("given format string has invalid specifier")]
+    Invalid,
+
+    /// The input string has been prematurely ended.
+    #[error("input string has been permaturely ended")]
+    TooShort,
+
+    /// All formatting items have been read but there is a remaining input.
+    #[error("given format string is missing some specifier")]
+    TooLong,
+
+    /// There was an error on the formatting string, or there were non-supported formating items.
+    #[error("given format string includes non-supported formmating item")]
+    BadFormat,
+
+    // TODO: Change this- to `#[non_exhaustive]` (on the enum) when MSRV is increased
+    #[error("unreachable chrono format error")]
+    Unreachable,
+}
+
+impl From<chrono::format::ParseError> for ChronoFormatError {
+    fn from(error: chrono::format::ParseError) -> ChronoFormatError {
+        match error.kind() {
+            chrono::format::ParseErrorKind::OutOfRange => ChronoFormatError::OutOfRange,
+            chrono::format::ParseErrorKind::Impossible => ChronoFormatError::Impossible,
+            chrono::format::ParseErrorKind::NotEnough => ChronoFormatError::NotEnough,
+            chrono::format::ParseErrorKind::Invalid => ChronoFormatError::Invalid,
+            chrono::format::ParseErrorKind::TooShort => ChronoFormatError::TooShort,
+            chrono::format::ParseErrorKind::TooLong => ChronoFormatError::TooLong,
+            chrono::format::ParseErrorKind::BadFormat => ChronoFormatError::BadFormat,
+            chrono::format::ParseErrorKind::__Nonexhaustive => ChronoFormatError::Unreachable,
+        }
+    }
 }
--- a/core/src/executor/evaluate/error.rs
+++ b/core/src/executor/evaluate/error.rs
@@ -1,6 +1,5 @@
 use {
     crate::ast::{Aggregate, Expr},
-    chrono::format::ParseErrorKind::*,
     serde::Serialize,
     std::fmt::Debug,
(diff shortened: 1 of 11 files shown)


core/src/executor/evaluate/function.rs
```diff
@@ -1,3 +1,5 @@
+use super::ChronoFormatError;
```
ever0de, line 1: Combine it with the import statement below

core/src/executor/evaluate/error.rs
```diff
@@ -61,3 +64,60 @@ pub enum EvaluateError {
     #[error("format function does not support following data_type: {0}")]
     UnsupportedExprForFormatFunction(String),
 }
+
+#[derive(Error, Serialize, Debug, PartialEq)]
+pub enum ChronoFormatError {
```
ever0de, line 69:
```rust

impl ChronoFormatError {
    pub fn err_into(error: chrono::format::ParseError) -> crate::result::Error {
        let error: ChronoFormatError = error.into();
        let error: EvaluateError = error.into();
        error.into()
    }
}
```

Let's implement this to improve our two-step `into` operation.

core/src/executor/evaluate/function.rs
```diff
@@ -429,3 +431,37 @@ pub fn format(name: String, expr: Evaluated<'_>, format: Evaluated<'_>) -> Resul
...
+        Value::Str(expr) => {
+            let format = eval_to_str!(name, format);
+            match chrono::NaiveDate::parse_from_str(&expr, &format) {
+                Ok(a) => Ok(Value::Date(a)),
+                Err(error) => {
+                    let error: ChronoFormatError = error.into();
+                    let error: EvaluateError = error.into();
+                    Err(error.into())
+                }
+            }
```
ever0de, line 446:
```suggestion
            chrono::NaiveDate::parse_from_str(&expr, &format)
                .map(Value::Date)
                .map_err(ChronoFormatError::err_into)
```

It gets so simple. How is it?

core/src/executor/evaluate/function.rs
```diff
@@ -429,3 +431,37 @@ pub fn format(name: String, expr: Evaluated<'_>, format: Evaluated<'_>) -> Resul
...
+        Value::Str(expr) => {
+            let format = eval_to_str!(name, format);
+            match chrono::NaiveDateTime::parse_from_str(&expr, &format) {
+                Ok(a) => Ok(Value::Timestamp(a)),
+                Err(error) => {
+                    let error: ChronoFormatError = error.into();
+                    let error: EvaluateError = error.into();
+                    Err(error.into())
+                }
+            }
```
ever0de, line 463:
```suggestion
            chrono::NaiveDateTime::parse_from_str(&expr, &format)
                .map(Value::Timestamp)
                .map_err(ChronoFormatError::err_into)
```

Same above too

core/src/executor/evaluate/error.rs
```diff
@@ -61,3 +64,60 @@ pub enum EvaluateError {
...
+    fn from(error: chrono::format::ParseError) -> ChronoFormatError {
+        match error.kind() {
+            chrono::format::ParseErrorKind::OutOfRange => ChronoFormatError::OutOfRange,
+            chrono::format::ParseErrorKind::Impossible => ChronoFormatError::Impossible,
+            chrono::format::ParseErrorKind::NotEnough => ChronoFormatError::NotEnough,
+            chrono::format::ParseErrorKind::Invalid => ChronoFormatError::Invalid,
+            chrono::format::ParseErrorKind::TooShort => ChronoFormatError::TooShort,
+            chrono::format::ParseErrorKind::TooLong => ChronoFormatError::TooLong,
+            chrono::format::ParseErrorKind::BadFormat => ChronoFormatError::BadFormat,
+            chrono::format::ParseErrorKind::__Nonexhaustive => ChronoFormatError::Unreachable,
```
ever0de, line 120: Repeated `chrono::format::ParseErrorKind` may reduce readability.

```Suggestion
           Use chrono::format::ParseErrorKind::*;

           error.kind() match {
               OutOfRange => ChronoFormatError::OutOfRange,
               Impossible => ChronoFormatError:: Impossible,
               NotEnough => ChronoFormatError::NotEnough,
               invalid => ChronoFormatError::invalid,
               TooShort => ChronoFormatError::TooShort,
               TooLong => ChronoFormatError::TooLong,
               BadFormat => ChronoFormatError::BadFormat,
               __Nonexhaustive => ChronoFormatError::unreachable,
           }
```
How about this?

core/src/executor/evaluate/function.rs
ChobobDev: Oh, thanks for poting this out, I will apply this ASAP

core/src/executor/evaluate/error.rs
ChobobDev: I totally agree with this

core/src/executor/evaluate/function.rs
ChobobDev: It looks way simple than before, I think it also increases readability

core/src/executor/evaluate/error.rs
ever0de: Well... the question may be where is the definition for this.
Even if it exists as a single function, there is no problem. If there is a suitable place, please change it there.

core/src/executor/evaluate/error.rs
```diff
@@ -1,12 +1,16 @@
 use {
     crate::ast::{Aggregate, Expr},
+    chrono::format::ParseErrorKind::*,
```
ever0de, line 3: Importing this across the `evaluate/error` module can be tricky. (e.g. when declaring or using the same name as a variant of `ParseErrorKind`)
This changes it to use only where it is used.

core/src/executor/evaluate/error.rs
```diff
@@ -61,3 +65,68 @@ pub enum EvaluateError {
...
+impl ChronoFormatError {
+    pub fn err_into(error: chrono::format::ParseError) -> crate::result::Error {
+        let error: ChronoFormatError = error.into();
+        let error: EvaluateError = error.into();
+        error.into()
+    }
+}
+
+impl From<chrono::format::ParseError> for ChronoFormatError {
+    fn from(error: chrono::format::ParseError) -> ChronoFormatError {
```
ever0de, line 120:
```suggestion
    fn from(error: chrono::format::ParseError) -> ChronoFormatError {
        use chrono::format::ParseErrorKind::*;
```

core/src/executor/evaluate/error.rs
ChobobDev: Oh, I will move this to where it belongs, thank you for letting me know

ever0de (approved):
Looks good 👍

core/src/executor/evaluate/error.rs
```diff
@@ -7,6 +7,9 @@ use {
 
 #[derive(Error, Serialize, Debug, PartialEq)]
 pub enum EvaluateError {
+    #[error(transparent)]
+    ChronoFormat(#[from] ChronoFormatError),
```
panarch, line 11: That's interesting that we have 2-level depth error conversion.
Looks good :) thanks!

panarch (approved):
We now have support new two helpful which handles dates.

PR #840, opened 2022-09-13
Update core/ chrono dependency version to 0.4.22, specify patch version

panarch:
Currently core compile fails with chrono v0.4.19.
So, update the chrono dep version and specify the patch number.


24seconds (approved):
Thank you for the quick action! LGTM

ever0de (approved):

PR #707, opened 2022-08-04
Draft: Implement ast_builder Select::Order_by

woojinnn:
## Description
Implementation of [#696](https://github.com/gluesql/gluesql/issues/696)

## Review Required:
- [ ] FSM structure
  - Currently, fsm is structure as:
  - {Select, Having, GroupBy} -> {OrderBy} -> {Offset, Limit, Project}

## Remaining Works:  
from Issue 696
- [ ] Support In Sub query Expression
- [ ] Support null first

Issue #696, opened 2022-07-30
[AST Builder] Implement `order_by` expression in ast_builder select


seonghun-dev:
# Description
Implements [`ORDER BY`](https://www.w3schools.com/sql/sql_orderby.asp) Select ast builders for [`order_by`](https://www.w3schools.com/sql/sql_orderby.asp) Expression.
Support vector only

## Background
Refered https://github.com/gluesql/gluesql/issues/614

## Task

- [ ] Support order_by vector
- [ ] Support In Sub query Expression
- [ ] ~~Support null first~~ >> not support in ast

In AST Builder `Select` module to [`order_by`](https://www.w3schools.com/sql/sql_orderby.asp)  Expression to order by

The following tests should be passed:

``` rust
    #[test]
    fn order_by() {
        let actual = table("Bar")
            .select()
            .group_by("city")
            .having("COUNT(name) < 100")
            .limit(3)
            .order_by(vec!["name asc"])
            .build();
        let expected = "
            SELECT * FROM Bar
            GROUP BY city
            HAVING COUNT(name) < 100
            LIMIT 3
            ORDER BY name ASC
        ";
        test(actual, expected);

        let actual = table("Bar").select().order_by(vec!["name asc", "id desc"]);
        let expected = "SELECT * FROM Bar ORDER BY name asc, id desc";
        test(actual, expected);
    }
```

woojinnn:
I'll devote on it!

woojinnn:
I think `order_by` shouldn't be allowed in Sub query Expression, referring this: [Can I use Order By clause in a subquery](https://stackoverflow.com/questions/34378701/can-i-use-order-by-clause-in-a-subquery)  

Personally, I would prefer eliminating `order_by` clauses in subquery in plan/executor layer rather than blocking `order_by` expression in ast_builder level to make the most of the original meaning.

seonghun-dev:
I wrote not clear sentence

- [ ] Support In Sub query Expression


I add Task based on below test
https://github.com/gluesql/gluesql/blob/acb2498734fc1409e174d5b353fbd1c00370633e/test-suite/src/order_by.rs#L111-L146

OrderByExpr support expr, asc. so, we can use below statement
``` rust
let actual = table("Bar").select().order_by(vec!["(SELECT t2.id FROM Test t2 WHERE Test.id = t2.id ORDER BY (Test.id + t2.id) LIMIT 1) asc", "id desc"]);
```
But, I think this statement is not clear, so i added on task

PR #841, opened 2022-09-13
Implement order by ast builder

24seconds:
## Description
Implementation of [#696](https://github.com/gluesql/gluesql/issues/696)
scope: This pr only covers `Support order_by vector str`.

## Review Required:
- [ ] **FSM structure**
  - As woojin mentioned in #707, fsm structure seems to be as:
  - {Select, Having, GroupBy} -> {OrderBy} -> {Offset, Limit, Project}

reference: https://www.postgresql.org/docs/current/queries-order.html

## Remaining Works:  
from issue #696 
- [ ] Support In Sub query Expression
- [ ] Support null first

### DIFF
--- a/core/src/parse_sql.rs
+++ b/core/src/parse_sql.rs
@@ -3,7 +3,7 @@ use {
     sqlparser::{
         ast::{
             Assignment as SqlAssignment, ColumnDef as SqlColumnDef, DataType as SqlDataType,
-            Expr as SqlExpr, Ident as SqlIdent, OrderByExpr, Query as SqlQuery,
+            Expr as SqlExpr, Ident as SqlIdent, OrderByExpr as SqlOrderByExpr, Query as SqlQuery,
             SelectItem as SqlSelectItem, Statement as SqlStatement,
         },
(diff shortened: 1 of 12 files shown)


core/src/parse_sql.rs
```diff
@@ -78,7 +78,7 @@ pub fn parse_interval<Sql: AsRef<str>>(sql_interval: Sql) -> Result<SqlExpr> {
         .map_err(|e| Error::Parser(format!("{:#?}", e)))
 }
 
-pub fn parse_order_by_expr<Sql: AsRef<str>>(sql_order_by_expr: Sql) -> Result<OrderByExpr> {
+pub fn parse_order_by_expr<Sql: AsRef<str>>(sql_order_by_expr: Sql) -> Result<SqlOrderByExpr> {
```
ever0de, line 81: nice catch

ever0de (approved):
Looks good to me

panarch (approved):
Awesome! Looks all nice, thanks a lot 👍 👍

PR #828, opened 2022-09-10
[Data] add data type u8

zmrdltl:
#### About numeric type
- [mysql](https://dev.mysql.com/doc/refman/8.0/en/numeric-types.html)
- [oracle](https://docs.oracle.com/cd/B19306_01/olap.102/b14346/dml_datatypes002.htm)
- [postgresql](https://www.postgresql.org/docs/9.6/datatype-numeric.html)

#### overflow handling
- [mysql](https://dev.mysql.com/doc/refman/8.0/en/out-of-range-and-overflow.html)
- [oracle question](https://asktom.oracle.com/pls/apex/f?p=100:11:0::::P11_QUESTION_ID:9531631800346339791)

## Description
1. When calculating arithmetic operations on u8 and i8, both are converted to i64 and calculated.
2. For the remaining types, cast to the larger type.

Issue #623, opened 2022-07-09
Add Numeric Data Type in `core` workspace 


ever0de:
Issues for tracking purposes

[<https://github.com/gluesql/gluesql/blob/main/core/src/data/value/mod.rs#L24-L43>](https://github.com/gluesql/gluesql/blob/e8323b878851098ad6d2202427ba84b4b7d821ef/core/src/data/value/mod.rs#L24-L43)

### UINT
- [x] `UINT8` #828
- [x] `UINT16` #844 
- [x] `UINT32` <https://github.com/gluesql/gluesql/pull/1019>
- [x] `UINT64` <https://github.com/gluesql/gluesql/pull/1019>
- [x] `UINT128` <https://github.com/gluesql/gluesql/pull/1019>

### INT
- [x] `INT8` <https://github.com/gluesql/gluesql/pull/407>
- [x] `INT16` <https://github.com/gluesql/gluesql/pull/632>
- [x] `INT32` <https://github.com/gluesql/gluesql/pull/608> 
- [x] `INT64` <https://github.com/gluesql/gluesql/commit/0244b769431eae941e669fd8370805507b1aac92>
- [x] `INT128` <https://github.com/gluesql/gluesql/pull/563>

### FLOAT
- [x] `FLOAT32` #1145 
- [x] `FLOAT`, `FLOAT64` <https://github.com/gluesql/gluesql/pull/21> 

### DECIMAL
- [x] `DECIMAL` <https://github.com/gluesql/gluesql/pull/377>

If anything is missing, please reply. I'll add it.

**If there is anyone who can proceed, dig up a new issue
Please leave the 623 issue as a reference.**

ever0de:
reference: <https://dev.mysql.com/doc/refman/8.0/en/integer-types.html>

ever0de:
Adding an integer data type just got easier.
Please refer to
<https://github.com/gluesql/gluesql/pull/915/commits/160d34b8042e7e5247a7009f7d82cb70ed2bcfcf>

Instead, adding a new type should handle some branching in the macro.

PR #844, opened 2022-09-14
[Data] add data type u16

ChobobDev:
# Background
from #623 

ref
[add data type u8](https://github.com/gluesql/gluesql/pull/828)

# Description
- When calculating arithmetic operations on u16 and i16, both are converted to i64 and calculated.
- For the remaining types, cast to the larger type.

# Rules
1. cast `right-hand` element into `left-hand` element type in case of arithmetics between INTEGER types 
2. in case of INTEGER * FLOAT convert to FLOAT
3. in case of INTEGER * DECIMAL convert to DECIMAL
4. in case of FLOAT * DECIMAL  convert to FLOAT


core/src/data/value/binary_op/u16.rs
```diff
@@ -0,0 +1,1186 @@
...
+                .checked_add(rhs as u16)
+                .ok_or_else(|| {
+                    ValueError::BinaryOperationOverflow {
+                        lhs: U16(lhs),
+                        rhs: I64(rhs),
+                        operator: NumericBinaryOperator::Add,
+                    }
+                    .into()
+                })
+                .map(U16),
```
devgony, line 99: Looks like we have to compute with large integer like `i64` and convert to u16 at the end like this.

```rs
I64(rhs) => {
    (lhs as i64)
        .checked_add(rhs)
        .ok_or_else(|| {
            ValueError::BinaryOperationOverflow {
                lhs: U16(lhs),
                rhs: I64(rhs),
                operator: NumericBinaryOperator::Add,
            }
            .into()
        })
        .and_then(|v| {
            v.try_into().map_err(|_| {
                ValueError::BinaryOperationOverflow {
                    lhs: U16(lhs),
                    rhs: I64(rhs),
                    operator: NumericBinaryOperator::Add,
                }
                .into()
            })
        })
        .map(U16)
```

*or we may have to fix UnaryOp logic to work after evaluation

ChobobDev: I think the latter case is better than the former. Since it will make the code much tidier.

devgony: We decided to throw error as soon as rhs fail while casting.
```rs
lhs.checked_add(rhs.try_into().map_err(|_| {
    ValueError::BinaryOperationOverflow {
        lhs: U16(lhs),
        rhs: I64(rhs),
        operator: NumericBinaryOperator::Add,
    }
})?)
.ok_or_else(|| {
    ValueError::BinaryOperationOverflow {
        lhs: U16(lhs),
        rhs: I64(rhs),
        operator: NumericBinaryOperator::Add,
    }
    .into()
})
.map(U16)
```

For now, It is Okay to repeat error code.
Then, refactor with macro in next PR

panarch (approved):
Awesome 👍 👍 
All looks nice!
Thanks for the contribution! New numeric type has come :D

Issue #845, opened 2022-09-15
[Data] update some of `numeric data type`


zmrdltl:
## Description
we use sqlparser-rs becasue of 2 reasons
1. parse sql statement 
2. make AST

parsed sql's ast result looks like this
### e.g1
SQL
```sql
CREATE TABLE Item (
        field_one INT(8),
        field_two INTEGER(8),
    );
```

AST
```json
CreateTable { or_replace: false, temporary: false, external: false, global: None, if_not_exists: false, name: ObjectName([Ident { value: "Item", quote_style: None
        }
    ]), columns: [ColumnDef { name: Ident { value: "field_one", quote_style: None
            }, data_type: Int(Some(8)), collation: None, options: []
        }, ColumnDef { name: Ident { value: "field_two", quote_style: None
            }, data_type: Integer(Some(8)), collation: None, options: []
        }
    ], ...
}
```

### e.g2
SQL
```sql
CREATE TABLE Item (
        field_one INT8,
        field_two INTEGER8,
    );
```
AST
```json

CreateTable { or_replace: false, temporary: false, external: false, global: None, if_not_exists: false, name: ObjectName([Ident { value: "Item", quote_style: None
        }
    ]), columns: [ColumnDef { name: Ident { value: "field_one", quote_style: None
            }, data_type: Custom(ObjectName([Ident { value: "INT8", quote_style: None
                }
            ])), collation: None, options: []
        }, ColumnDef { name: Ident { value: "field_two", quote_style: None
            }, data_type: Custom(ObjectName([Ident { value: "INTEGER8", quote_style: None
                }
            ])), collation: None, options: []
        }
    ], ...
}
```

To declare a column using int8 data in e.g1, it must be enclosed in parentheses. This can be confusing with function types. It is expected to provide a better user experience by creating an ast after parsing it with the custom type provided by sqlparser-rs.

## Tasks
The format of the data to be replaced is as follows.
- [x] i8 #846 
- [x] i16 #847 
- [x] i32 #848 
- [x] i128 #850 
- [x] u8 #851 
- [x] update README #852 

* A simple `INT` with no trailing number, `UNSIGNED` remains converted to 64-bit format.
* The data type int64 was removed #849 
* Since keyword `INT` is a synonym for keyword `INTEGER` in [mysql](https://dev.mysql.com/doc/refman/8.0/en/numeric-types.html), i decided to remain those keywords

panarch:
[...]

PR #846, opened 2022-09-15
[Data] modify int8 to parse sql int(8) format to int8

zmrdltl:
## Background
ref #845


panarch (approved):
Awesome!
Let's get started 👍 👍 👍

PR #847, opened 2022-09-15
[Data] modify int16 to parse sql int(16) format to int16

zmrdltl:
## Background
ref #845


panarch (approved):
Awesome! Thanks a lot 👍

PR #848, opened 2022-09-15
[Data] modify int32 to parse sql int(32) format to int32

zmrdltl:
## Background
ref #845


ever0de (approved):
It's very fast, it looks good

PR #849, opened 2022-09-15
[Data] remove data type int64

zmrdltl:
## Background
ref #845 

## Description
https://github.com/gluesql/gluesql/blob/main/core/src/translate/data_type.rs#L14
Since 64 bit signed integers are already supported as INT and INTEGER, in this case, there is no need for an additional name.


panarch (approved):
I wasn't aware of `INT(64)` support.
Good change, thanks a lot!

PR #862, opened 2022-09-18
Implement function `POSITION`

### LINKED ISSUE #855: Implement `POSITION` expression
## Description
https://www.w3schools.com/sql/func_mysql_position.asp
SQL
```sql
SELECT CAST('CO' IN 'W3Schoolsco') AS MatchPosition;
```
AST
```json
Query(Query { 
    with: None, 
    body: Select(Select { 
        distinct: false, 
        top: None, 
        projection: [
            ExprWithAlias { 
            expr: Position { 
                    expr: Value(SingleQuotedString("CO")), in: Value(SingleQuotedString("W3Schoolsco"))
                }, 
            alias: Ident { 
                    value: "MatchPosition", quote_style: None
                }
            }
        ], 
        into: None, 
        from: [], 
        lateral_views: [], 
        selection: None, 
        group_by: [], 
        cluster_by: [], 
        distribute_by: [], 
        sort_by: [], 
        having: None, 
        qualify: None
    }), 
    order_by: [], 
    limit: None, 
    offset: None, 
    fetch: None, 
    lock: None
})
```

zmrdltl:
## Background
resolve #855

## Tasks
- add structures
  - [x] ast
  - [x] translate
  - [x] executor
  - [x] plan
  - [x] ast_builder

- add tests
  - [x] projection with columns
  - [x] only string
  - [x] errors


core/src/executor/evaluate/function.rs
```diff
@@ -471,3 +471,14 @@ pub fn to_time(name: String, expr: Evaluated<'_>, format: Evaluated<'_>) -> Resu
         _ => Err(EvaluateError::FunctionRequiresStringValue(name).into()),
     }
 }
+
+pub fn position(name: String, expr: Evaluated<'_>, format: Evaluated<'_>) -> Result<Value> {
+    match expr.try_into()? {
+        Value::Str(_) => {
+            let format = eval_to_str!(name, format);
+            let num = format.parse::<i64>().unwrap();
+            Ok(Value::I64(num))
```
ever0de, line 480: See <https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.position>
The index cannot be negative, how about using size (8 bytes)?

zmrdltl: thanks for your careful review

PR #866, opened 2022-09-20
Add cli/ print and command modules to code coverage tests

panarch:
In `cli` workspace, rusty_line dependent modules are not able to be tested.
Except for those, there is no issue for us to run coverage tests.
e.g. `print` and `command` modules

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -43,7 +43,10 @@ jobs:
             --branch \
             --ignore-not-existing \
             --ignore "/*" \
-            --ignore "cli/src/*" \
+            --ignore "cli/src/cli.rs" \
+            --ignore "cli/src/helper.rs" \
+            --ignore "cli/src/lib.rs" \
+            --ignore "cli/src/main.rs" \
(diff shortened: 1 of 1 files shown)


.github/workflows/coverage.yml
```diff
@@ -43,7 +43,10 @@ jobs:
             --branch \
             --ignore-not-existing \
             --ignore "/*" \
-            --ignore "cli/src/*" \
+            --ignore "cli/src/cli.rs" \
+            --ignore "cli/src/helper.rs" \
+            --ignore "cli/src/lib.rs" \
+            --ignore "cli/src/main.rs" \
```
ever0de, line 49:
```
         --ignore <PATH>...
             Ignore files/directories specified as globs
```

It's globs, so it's okay to do it this way?

```
             --ignore "cli/src/{cli,helper,lib,main}.rs" \
```

panarch: That was because of my laziness... thanks a lot 👍

ever0de (approved):
👍

PR #872, opened 2022-09-24
Update test-suite Tester trait to use Glue struct

panarch:
before:
https://github.com/gluesql/gluesql/blob/316fbdb6878bc5051b9cf2c0efa46eeea3e2b20a/test-suite/src/tester/mod.rs#L240-L245

now:
```rust
#[async_trait]
pub trait Tester<T: GStore + GStoreMut> {
    fn new(namespace: &str) -> Self;

    fn get_glue(&mut self) -> &mut Glue<T>;
}
```


ever0de (approved):
Good! 😃

PR #873, opened 2022-09-24
refactor: Replace `ObjectName` with `String`

devgony:
## Goal
Currently table_name, alias and index_name are using `ObjectName(Vec<String>)` but redundant.
Let's replace all `ObjectName` with `String`

## Todo
- [x] replace all `ObjectName` with `String`
- [x] Throw Error if the number of `ObjectName` is more than 2

## In Future
To support multiple schemas or databases in future, it would be inevitable to replace with better struct.

### DIFF
--- a/test-suite/src/basic.rs
+++ b/test-suite/src/basic.rs
@@ -59,7 +59,7 @@ CREATE TABLE TestA (
         ),
         (
             "SELECT id FROM FOO.Test",
-            Err(TranslateError::CompoundObjectNotSupported("FOO.Test".into()).into()),
+            Err(TranslateError::CompoundObjectNotSupported("FOO.Test".to_owned()).into()),
         ),
     ];
 
--- a/test-suite/src/basic.rs
+++ b/test-suite/src/basic.rs
@@ -1,4 +1,7 @@
-use {crate::*, gluesql_core::prelude::Value::*};
+use {
+    crate::*,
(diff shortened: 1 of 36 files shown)


ever0de (commented):
It would look good if you limit the conversion of `&str -> String` to one.
today
1. `into`
2. `to_owned`
3. `to_string`
This is being mixed and changed.

Clippy lint also recommends `to_owned` (default allow, but later changed to deny), so it would be good to unify it later.

ever0de (approved):
Looks good

test-suite/src/basic.rs
```diff
@@ -54,6 +57,10 @@ CREATE TABLE TestA (
             "SELECT id, num FROM Test",
             Ok(select!(id | num; I64 | I64; 2 2; 2 9; 2 4; 2 7)),
         ),
+        (
+            "SELECT id FROM FOO.Test",
+            Err(TranslateError::CompoundObjectNotSupported("FOO.Test".to_owned()).into()),
```
panarch, line 62: 👍

panarch (approved):
Great improvement.... 👍 👍 👍 
At the moment I decided to add `ObjectName`, I wasn't sure whether we need `ObjectName` or not.
However, it's now obvious that it is actually enough to use `String` for table name but it was hard to touch this issue because of so many.. `ObjectName` uses.
Thanks really, really a lot!
Looks all great 👍

PR #875, opened 2022-09-25
Implement str_position function under enum Value

zmrdltl:
## Background
for #855 

### mysql POSITION function return examples
```sql
SELECT POSITION("h" IN "sushi") AS MatchPosition;
```
|MatchPosition|
|----------------|
|4|

```sql
SELECT POSITION("" IN "sushi") AS MatchPosition;
```
|MatchPosition|
|----------------|
|1|

```sql
SELECT POSITION(NULL IN "sushi") AS MatchPosition;
```
|MatchPosition|
|----------------|
|NULL|

```sql
SELECT POSITION(-1 IN "sushi") AS MatchPosition;
```
|MatchPosition|
|----------------|
|0|

```sql
SELECT POSITION(2022-09-25 IN "sushi") AS MatchPosition;
```
|MatchPosition|
|----------------|
|0|

## Description
Added function to support mysql's `POSITION` function.

Since the substring and string to be parsed and received as an argument in the sql statement are value types in glusql, the str_position function has been implemented for the operation.

If `empty_string` found in arguments, I64(0) is returned.
If `substring in string found`, the position of the first occurrence is returned. The minimum value is I64(1).
If `NULL found` in arguments, Null is returned. 
Although mysql returns 0 for other types of arguments, I decided to display an Error for types other than string and null.

## Tasks
- [x] Write implementation
- [x] Add unit, doc Tests
- [x] Add Error

### DIFF
--- a/core/src/data/value/error.rs
+++ b/core/src/data/value/error.rs
@@ -169,7 +169,7 @@ pub enum ValueError {
     SqrtOnNonNumeric(Value),
 
     #[error("non string value found: from_string is {0:?}, sub_string is {1:?}")]
-    StrPositionOnNonString(Value, Value),
+    UnSupportedArgumentInFunctionPosition(Value, Value),
 }
 
 #[derive(Debug, PartialEq, Eq, Serialize, Display)]
--- a/core/src/data/value/error.rs
+++ b/core/src/data/value/error.rs
@@ -168,8 +168,8 @@ pub enum ValueError {
     #[error("non numeric value in sqrt {0:?}")]
     SqrtOnNonNumeric(Value),
 
-    #[error("non string value found: from_string is {0:?}, sub_string is {1:?}")]
-    UnSupportedArgumentInFunctionPosition(Value, Value),
+    #[error("unsupported value by position function: from_str(from_str:?), sub_str(sub_str:?)")]
+    UnSupportedValueByPositionFunction { from_str: Value, sub_str: Value },
 }
 
 #[derive(Debug, PartialEq, Eq, Serialize, Display)]
--- a/core/src/data/value/error.rs
+++ b/core/src/data/value/error.rs
@@ -167,6 +167,9 @@ pub enum ValueError {
 
     #[error("non numeric value in sqrt {0:?}")]
     SqrtOnNonNumeric(Value),
+
+    #[error("non string value found: from_string is {0:?}, sub_string is {1:?}")]
+    StrPositionOnNonString(Value, Value),
 }
 
 #[derive(Debug, PartialEq, Eq, Serialize, Display)]
--- a/core/src/data/value/mod.rs
+++ b/core/src/data/value/mod.rs
@@ -585,10 +585,10 @@ impl Value {
         match (self, other) {
             (Str(from_str), Str(sub_str)) => Ok(I64(str_position(from_str, sub_str) as i64)),
             (Null, _) | (_, Null) => Ok(Null),
-            _ => Err(ValueError::UnSupportedArgumentInFunctionPosition(
-                self.clone(),
-                other.clone(),
-            )
+            _ => Err(ValueError::UnSupportedValueByPositionFunction {
+                from_str: self.clone(),
+                sub_str: other.clone(),
+            }
             .into()),
         }
     }
@@ -1539,7 +1539,11 @@ mod tests {
         assert_eq!(str1.position(&empty_str), Ok(I64(0)));
         assert_eq!(
             str1.position(&I64(1)),
-            Err(ValueError::UnSupportedArgumentInFunctionPosition(str1, I64(1)).into())
+            Err(ValueError::UnSupportedValueByPositionFunction {
+                from_str: str1,
+                sub_str: I64(1)
+            }
+            .into())
         );
     }
 
--- a/core/src/data/value/mod.rs
+++ b/core/src/data/value/mod.rs
@@ -560,9 +560,9 @@ impl Value {
     ///
     /// Returns the position where the first letter of the substring starts if the string contains a substring.
     ///
-    /// Returns `I64(0)` if the string to be found is not found.
-    /// Returns minimum value `I64(1)` when the string is found.
-    /// Returns Null if NULL parameter found.
+    /// Returns [`Value::I64(0)`] if the string to be found is not found.
+    /// Returns minimum value [`Value::I64(1)`] when the string is found.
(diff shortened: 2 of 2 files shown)


core/src/data/value/mod.rs
```diff
@@ -555,6 +555,45 @@ impl Value {
...
+                    return Ok(0);
+                }
+                match from_str.find(sub_str).map(|position| position + 1) {
+                    Some(pos) => Ok(pos),
+                    None => Ok(0),
+                }
+            }
+            (Null, Str(_)) | (Str(_), Null) => Ok(0),
+            _ => Err(ValueError::StrPositionOnNonString(self.clone(), other.clone()).into()),
+        }
```
ever0de, line 595:
```suggestion
        use Value::*;

        match (self, other) {
            (Str(from_str), Str(sub_str)) => {
                if from_str == "" || sub_str == "" {
                    return Ok(0);
                }
                
                Ok(from_str.find(sub_str).map(|position| position + 1).unwrap_or(0))
            }
            (Null, Str(_)) | (Str(_), Null) => Ok(0),
            _ => Err(ValueError::StrPositionOnNonString(self.clone(), other.clone()).into()),
        }
```

ever0de: In mysql, if `Null` is entered, `Null` is output.
Wouldn't it be appropriate to consider this

```sql
mysql> select position(null in null);
+---------------+
| position(null in null) |
+---------------+
| NULL |
+---------------+
1 row in set (0.00 sec)

mysql> select position("" in null);
+-------------------------+
| position("" in null) |
+-------------------------+
| NULL |
+-------------------------+
1 row in set (0.00 sec)

mysql> select position(null in "");
+-------------------------+
| position(null in "") |
+-------------------------+
| NULL |
+-------------------------+
1 row in set (0.00 sec)
```

core/src/data/value/mod.rs
```diff
@@ -555,6 +555,45 @@ impl Value {
...
+    /// assert_eq!(str2.str_position(&str1), Ok(0));
+    /// assert_eq!(Null.str_position(&str2), Ok(0));
+    /// assert_eq!(str1.str_position(&Null), Ok(0));
+    /// assert_eq!(empty_str.str_position(&str2), Ok(0));
+    /// assert_eq!(str1.str_position(&empty_str), Ok(0));
+    /// assert_eq!(
+    ///     str1.str_position(&I8(1)),
+    ///     Err(ValueError::StrPositionOnNonString(str1, I8(1)).into())
+    /// );
+    /// ```
```
ever0de, line 580: <https://doc.rust-lang.org/stable/std/primitive.str.html#method.find>
I rewrote it by referring to the reference. Apply markdown lint rules.

```suggestion
    /// Support only [`Value::Str`] variant
    /// 
    /// Returns the position where the first letter of the substring starts if the string contains a substring.
    /// 
    /// Returns `0` if the string to be found is not found.
    /// Returns minimum value `1` when the string is found.
    /// 
    /// # Examples
    /// 
    /// ```
    /// use gluesql::core::data::{ValueError, value::Value::*};
    ///
    /// let str1 = Str("ramen".to_string());
    /// let str2 = Str("men".to_string());
    /// let empty_str = Str("".to_string());
    /// assert_eq!(str1.str_position(&str2), Ok(3));
    /// assert_eq!(str2.str_position(&str1), Ok(0));
    /// assert_eq!(Null.str_position(&str2), Ok(0));
    /// assert_eq!(str1.str_position(&Null), Ok(0));
    /// assert_eq!(empty_str.str_position(&str2), Ok(0));
    /// assert_eq!(str1.str_position(&empty_str), Ok(0));
    /// assert_eq!(
    ///     str1.str_position(&I8(1)),
    ///     Err(ValueError::StrPositionOnNonString(str1, I8(1)).into())
    /// );
    /// ```
```
image

core/src/data/value/mod.rs
zmrdltl: Current return type is Result<usize>.. then, how about Result<Value>.
If we use that we have to consider general case return type. If found, return Result<I64> will be useful

core/src/data/value/mod.rs
```diff
@@ -555,6 +555,53 @@ impl Value {
...
+    /// Returns minimum value `I64(1)` when the string is found.
+    /// Returns Null if NULL parameter found.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use gluesql_core::{
+    ///     data::ValueError,
+    ///     prelude::Value::*
+    /// };
```
ever0de, line 573: It is very rare for users to use `gluesql_core`, and it would be good to roll back because they will normally use `gluesql`.

core/src/data/value/error.rs
```diff
@@ -167,6 +167,9 @@ pub enum ValueError {
 
     #[error("non numeric value in sqrt {0:?}")]
     SqrtOnNonNumeric(Value),
+
+    #[error("non string value found: from_string is {0:?}, sub_string is {1:?}")]
+    UnSupportedArgumentInFunctionPosition(Value, Value),
```
ever0de, line 172:
```suggestion

    #[error("unsupported value by position function: from_str(from_str:?), sub_str(sub_str:?)")]
    UnSupportedValueByPositionFunction { from_str: Value, sub_str: Value },
```

ever0de: Using a tuple will fade the meaning, so let's write it in a struct form.

core/src/data/value/mod.rs
```diff
@@ -555,6 +555,61 @@ impl Value {
...
+    /// assert_eq!(str1.position(&str2), Ok(I64(3)));
+    /// assert_eq!(str2.position(&str1), Ok(I64(0)));
+    /// assert!(Null.position(&str2).unwrap().is_null());
+    /// assert!(str1.position(&Null).unwrap().is_null());
+    /// assert_eq!(empty_str.position(&str2), Ok(I64(0)));
+    /// assert_eq!(str1.position(&empty_str), Ok(I64(0)));
+    /// assert_eq!(
+    ///     str1.position(&I64(1)),
+    ///     Err(ValueError::UnSupportedArgumentInFunctionPosition(str1, I64(1)).into())
+    /// );
```
ever0de, line 589: At least one representative example

panarch (approved):
Looks great! meaningful beginning of doc comments! Thanks a lot 👍 👍 👍

PR #876, opened 2022-09-25
Remove Result from core/ table::get_alias return type

panarch:
core table::get_alias does not return error so remove `Result` from the return type.
`get_alias` now simply returns `&String`.


ever0de (approved):
Thanks for the improvement!

PR #877, opened 2022-09-25
Remove unused comments in core/ ast/mod.rs

panarch:
I missed this in the previous PR review.

### DIFF
--- a/core/src/ast/mod.rs
+++ b/core/src/ast/mod.rs
@@ -249,21 +249,6 @@ mod tests {
         std::str::FromStr,
     };
 
-    // #[test]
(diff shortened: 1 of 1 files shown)


ever0de (commented):
I thought I'd leave it as a part. Because ObjectName can happen later?

ever0de (approved):
Discord discussion summary

Since sqlparser-rs needs to support a variety of dbs and the goal is to parse it, it receives idents as Vector.
Since we are in the position of execution, it is not necessary right now, but if you need to expand later, it seems better to receive a DB Name separately.

Also, since Rust's compiler support is good, it seems better to delete it than to leave a comment.

PR #878, opened 2022-09-25
refactor: remove redundant TableFactor::Series.name

devgony:
## Goal
The `TableFactor` which does not fetch data from storage with `specific table_name` does not need to own `name`.

Instead, Let's use alias and remove Tablefactor::Series.name. 
(If there is no alias, consider name as a alias)


panarch (approved):
Nice improvement 👍 Thanks a lot!

PR #879, opened 2022-09-28
Implement AST builder execute function

panarch:
Add `execute` support to AST builder which takes `Glue` and executes query directly without using `Glue::execute`

e.g.
```rust
let actual = table("Foo")
    .select()
    .project("id, name")
    .execute(glue)
    .await;
let expected = Ok(select!(
    id  | name
    I64 | Str;
    100   "Pickle".to_owned();
    200   "Lemon".to_owned()
));
assert_eq!(actual, expected, "select");
```


tests/glue.rs
```diff
@@ -63,6 +64,11 @@ async fn basic_async<T: GStore + GStoreMut>(mut glue: Glue<T>) {
         .await,
         Ok(vec![Payload::Create])
     );
+
+    let payload = table("api_test").delete().execute(&mut glue).await.unwrap();
+    println!("{:?}", payload);
+
+    panic!();
```
ever0de, line 71: Remove this 👍

tests/glue.rs
```diff
@@ -1,5 +1,6 @@
 #![cfg(any(feature = "memory-storage", feature = "sled-storage"))]
 use gluesql_core::{
+    ast_builder::*,
```
ever0de, line 3:
```suggestion
```

ever0de: Everything looks good other than that 🚀 🚀

ever0de (approved):

PR #880, opened 2022-09-28
[AST] Support `ArrayIndex`Expr

### LINKED ISSUE #827: Support `ArrayIndex` expression
## Description
Support `ArrayIndex` expression for index based access for `Map`, `List` type values

## Example
```sql
CREATE TABLE Test (
    id INTEGER,
    items LIST
);

INSERT INTO Test VALUES (1, '[{ "foo": 100, "bar": [true, 0, [10.5, false] ] }, 10, 20]');


SELECT id, items[1] as foo FROM Test;  
SELECT items[1]["foo"] as bar FROM Test; 
```

ding-young:
## Issue
resolves #827 

## Detail
ArrayIndex Expr works for accessing value from `List`, `Map` type value. ex) `map["1"]["2"]`
The internal behavior is a bit different from `unwrap` fn. 
1. return `Err` when value is not selectable 
  cf. `unwrap` also works the same at first time but returns `Value::Null` when recursively applied.
2. return `Value::Null` when accessing non-existent key from `List`, `Map` value


core/src/data/value/selector.rs
```diff
@@ -45,4 +45,13 @@ impl Value {
...
     }
+
+    pub fn selector_by_index(&self, selector: Vec<Value>) -> Result<Value> {
+        let keys = selector
+            .iter()
+            .map(String::from)
+            .collect::<Vec<_>>()
+            .join(".");
+        self.selector(&keys)
+    }
```
ding-young, line 56: I just converted `Vec<Value>` into form of **"index1.index2.index3"** instead of modifying `selector` fn (which used for `unwrap` fn). Since this is just a tmp solution, I need suggestions for it.

core/src/executor/evaluate/expr.rs
```diff
@@ -78,3 +78,16 @@ pub fn between<'a>(
 
     Ok(Evaluated::from(Value::Bool(v)))
 }
+
+pub fn array_index<'a>(obj: Evaluated<'a>, indexes: Vec<Evaluated<'a>>) -> Result<Evaluated<'a>> {
+    let value = match obj {
+        Evaluated::Value(value) => value,
+        _ => return Err(EvaluateError::MapOrListTypeRequired.into()),
+        // TODO Q. 의미도 안맞고 실제로 map, list check는 selector에서 ..
```
ding-young, line 86: Actually, this match statement does not check whether the value is either`map` or `list` type.. Better name for this error?

core/src/data/value/selector.rs
```diff
@@ -45,4 +45,26 @@ impl Value {
...
+
+    pub fn selector_by_index(&self, selector: &[Value]) -> Result<Value> {
+        let str_keys = selector.iter().map(String::from).collect::<Vec<_>>();
+
+        let value = str_keys.iter().try_fold(self, |selectable, key| {
+            selectable.get_value_from_compound_type(key)
+        })?;
+
+        Ok(value.clone())
+    }
```
ever0de, line 57:
```suggestion
    pub fn selector_by_index(&self, selector: &[Value]) -> Result<Value> {
        selector
            .iter()
            .map(String::from)
            .try_fold(self, |selectable, key| {
                selectable.get_value_from_compound_type(&key)
            })
            .map(Clone::clone)
    }
```

It looks like you don't need to consume the iterator as a `collect` in the middle.

core/src/data/value/selector.rs
```diff
@@ -45,4 +45,26 @@ impl Value {
...
+        let value = match self {
+            Value::Map(map) => map.get(key),
+            Value::List(list) => key.parse::<usize>().ok().and_then(|i| list.get(i)),
+            _ => return Err(ValueError::SelectorRequiresMapOrListTypes.into()),
+        };
+        match value {
+            None => Ok(&Value::Null),
+            Some(v) => Ok(v),
+        }
+    }
```
ever0de, line 69: minor refactoring

```suggestion
    fn get_value_from_compound_type(&self, key: &str) -> Result<&Value> {
        let value = match self {
            Value::Map(map) => map.get(key),
            Value::List(list) => key.parse::<usize>().ok().and_then(|i| list.get(i)),
            _ => return Err(ValueError::SelectorRequiresMapOrListTypes.into()),
        };

        Ok(value.unwrap_or(&Value::Null))
```

core/src/executor/evaluate/mod.rs
```diff
@@ -228,6 +228,12 @@ pub async fn evaluate<'a>(
                 None => Ok(Evaluated::from(Value::Null)),
             }
         }
+        Expr::ArrayIndex { obj, indexes } => {
+            let obj = eval(obj).await?;
+            let indexes: Vec<_> = indexes.iter().map(eval).collect();
+            let indexes = try_join_all(indexes).await?;
```
ever0de, line 234:
```suggestion
            let indexes = try_join_all(indexes.iter().map(eval)).await?;
```

ever0de (commented):
I think it would be nice to have one test code for the case without alias.

test-suite/src/data_type/map.rs
```diff
@@ -61,6 +61,49 @@ INSERT INTO MapType VALUES
...
+            nested["a"]["blue"] + nested["b"] as sum,
+            nested["c"] AS c
+        FROM MapType2"#,
+        Ok(select_with_null!(
+            id     | fruit        | sum      | c;
+            I64(1)   s("apple")     I64(11)    Null;
+            I64(2)   s("cherry")    I64(22)    Null;
+            I64(3)   s("berry")     I64(33)    Bool(true)
+        ))
+    );
```
zmrdltl, line 105: macro test can define test name what you commented
```rust
test!{
 name: index expr with non-existent key from MapType Value returns Null
 sql: r#"SELECT
            id,
            nested["a"]["red"] AS fruit,
            nested["a"]["blue"] + nested["b"] as sum,
            nested["c"] AS c
        FROM MapType2"#
 expected: Ok(select_with_null!(
            id     | fruit        | sum      | c;
            I64(1)   s("apple")     I64(11)    Null;
            I64(2)   s("cherry")    I64(22)    Null;
            I64(3)   s("berry")     I64(33)    Bool(true)
        ))
}
```

ever0de (approved):
Look good 👍👍 👍

panarch (approved):
New expression! 👍 👍 👍 
Looks all great! Thanks a lot 👍

Issue #798, opened 2022-08-24
[AST] Implement `ascii`, `chr` function in ast


seonghun-dev:
# Description
Implements ast for `ascii`, `chr` Function.

> ascii ( text ) → integer
> 
> Returns the numeric code of the first character of the argument. In UTF8 encoding, returns the Unicode code point of the character. In other multibyte encodings, the argument must be an ASCII character.
> ascii('x') → 120

> chr ( integer ) → text
> 
> Returns the character with the given code. In UTF8 encoding the argument is treated as a Unicode code point. In other multibyte encodings the argument must designate an ASCII character. chr(0) is disallowed because text data types cannot store that character.
> chr(65) → A
> 

## Task
In AST module add `ascii`, `chr`  function

The following tests should be passed:

``` rust
test_case!(ascii, async move {
        (
            "CREATE TABLE SingleItem (id FLOAT DEFAULT 90)",
            Ok(Payload::Create),
        ),
        (
            r#"INSERT INTO SingleItem VALUES (0)"#,
            Ok(Payload::Insert(1)),
        ),
        (
            "SELECT ASCII('x') as to_ascii FROM SingleItem",
            Ok(select!(
                to_ascii
                I64;
                I64::from(120).to_owned()
            )),
        ),
        (
            "SELECT CHR(120) as to_chr FROM SingleItem",
            Ok(select!(
                to_chr
                Str;
                "A".to_owned()
            )),
        ),
}

```
## Reference
https://www.postgresql.org/docs/14/functions-math.html

PR #881, opened 2022-09-28
[Ast] Implement ascii, chr function in ast

seonghun-dev:
Resolve : https://github.com/gluesql/gluesql/issues/798


core/src/executor/evaluate/function.rs
```diff
@@ -223,6 +223,35 @@ pub fn substr(
     Ok(Value::Str(string))
 }
 
+pub fn ascii(name: String, expr: Evaluated<'_>) -> Result<Value> {
+    let string = eval_to_str!(name, expr);
+
+    if string.len() != 1 {
```
ever0de, line 229:
```suggestion
    if !string.is_ascii() {
```

core/src/executor/evaluate/function.rs
```diff
@@ -223,6 +223,35 @@ pub fn substr(
...
+        return Err(EvaluateError::AsciiFunctionRequiresSingleCharacterValue.into());
+    }
+
+    let char = string.chars().next();
+    match char {
+        Some(char) => {
+            let char = char as u8;
```
ever0de, line 236:
```suggestion
```

duplicated type casting

core/src/executor/evaluate/function.rs
```diff
@@ -223,6 +223,35 @@ pub fn substr(
...
+pub fn chr(name: String, expr: Evaluated<'_>) -> Result<Value> {
+    let expr = eval_to_int!(name, expr);
+
+    match expr {
+        0..=255 => {
+            let expr = expr as u8;
+            Ok(Value::Str((expr as char).to_string()))
+        }
+        _ => Err(EvaluateError::ChrFunctionRequiresIntegerValueInRange0To255.into()),
+    }
```
ever0de, line 252: It seems good to convert to `String` here too and check with `is_ascii`. Then it seems that no conversion is necessary for the return type `char` .

core/src/executor/evaluate/function.rs
```diff
@@ -223,6 +223,35 @@ pub fn substr(
...
 }
 
+pub fn ascii(name: String, expr: Evaluated<'_>) -> Result<Value> {
+    let string = eval_to_str!(name, expr);
+
+    if string.len() != 1 {
+        return Err(EvaluateError::AsciiFunctionRequiresSingleCharacterValue.into());
+    }
+
+    let char = string.chars().next();
```
ever0de, line 233: `char` is the name of a primitive type, so we recommend something else.
ex) `character`

core/src/executor/evaluate/function.rs
ever0de: edit) 

```rust
    let string = eval_to_str!(name, expr);
    let mut iter = string.chars();

    match iter.next() {
        Some(text) => {
            if iter.count() > 0 {
                // non char
            }

            if text.is_ascii() {
                // return Ok(Value/...)
            } else {
                // non ascii string
            }

            todo!()
        }
        // empty string?
        None => todo!(),
    }
```

core/src/executor/evaluate/function.rs
ever0de: This is a bad comment because the `CHR` function only accepts u8(numeric).

ever0de (approved):

panarch (approved):
All looks nice! Thanks a lot 👍

PR #884, opened 2022-10-01
Auto Implement `GStore`, `GStoreMut` traits

ever0de:
## Before

```rust
// implement Store, StoreMut, Optional traits ...

// must be written explicitly
impl GStore for Storage {}
impl GStoreMut for Storage {}

let storage = Storage::new();
```

## After

```rust
// implement Store, StoreMut, Optional traits ...

let storage = Storage::new();
// It works even if omitted because it is automatically implemented as Generic
```


panarch (approved):
Nice improvement 👍 👍 
Thanks a lot!

PR #885, opened 2022-10-01
Add root workspace in members

ever0de:
ref: <https://github.com/gluesql/gluesql/pull/879/files#diff-254e37d1bdaec899062959fb14f59d4804fd9fc4886853fac7833121747f3e5a>


panarch:
ref. https://github.com/gluesql/gluesql/blob/f31a3c16f0f9421a5c71b0e3f65a8547ef605495/tests/glue.rs#L54-L72
This root test was supposed to fail but it wasn't checked by our current GitHub action.
That can be fixed by this PR.

panarch (approved):
Thanks a lot!!

PR #887, opened 2022-10-02
Update coverage.yml to ignore root examples

panarch:
`/examples` was not being tested by coverage github action but the root workspace was included to our default `cargo test` workspaces.

Update `coverage.yml` cargo test to ignore examples.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -32,7 +32,7 @@ jobs:
         env:
           RUSTFLAGS: -Cinstrument-coverage
           LLVM_PROFILE_FILE: gluesql-%p-%m.profraw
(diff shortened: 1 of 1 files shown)


.github/workflows/coverage.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         env:
           RUSTFLAGS: -Cinstrument-coverage
           LLVM_PROFILE_FILE: gluesql-%p-%m.profraw
-        run: cargo test --all-features --verbose
+        run: cargo test --all-features --examples --verbose
```
ever0de, line 35: Is it possible to exclude bench from `--all-targets`?

ever0de (approved):

PR #889, opened 2022-10-02
Update README.md; add `await` keyword.

24seconds:
## Description
I found that `await` keyword is missing in glueslq-js/README.md


panarch (approved):
Thanks a lot for the good catch!

PR #892, opened 2022-10-03
Use variable instead of closure in core/ evaluate function module

ever0de:


panarch (approved):
Looks better! Thanks a lot 👍

PR #899, opened 2022-10-03
[AST] CONCAT function concat with null return NULL

seonghun-dev:


ever0de (approved):

panarch (approved):
Looks all nice!
Good to see that `CONCAT` and `||` to have same behaviors.
Thanks a lot 👍

PR #900, opened 2022-10-03
Enable str to string clippy options

ever0de:


panarch (approved):
I am also for this decision to add `#![deny(clippy::str_to_string)]`.
This might look too strict but it makes all of us simple to write codes because now we only have a single option `to_owned` for conversion from `&str` to `String`.
Thanks a lot 👍

PR #901, opened 2022-10-03
[Test Suite] add null test in function position

zmrdltl:
from #862


panarch (approved):
Looks nice and thanks for the details!
It is certainly good to have a test case of how the function handles `NULL`.

PR #904, opened 2022-10-05
Upgrade sqlparser-rs to v0.24, migrate Interval from Literal to Expr

panarch:
Main change is that we now no longer have `AstLiteral::Interval` and that is replaced by `Expr::Interval`.

cc. @ding-young


ever0de (approved):
Looks good

PR #910, opened 2022-10-07
[AST-Builder] Impl Case

ding-young:
## Description
Implement ast-builder case expression.
There are two entry points for case expr. 
1. `ExprNode`.case()... : Case expr with operand
2. case()...  : Case expr without operand

## Detail - mini FSM
(ExprNode) -> CaseNode -> `when_then` ->  WhenThenNode 
-> 1. `(or_else, end)` ->  ExprNode 
-> 2. `when_then` -> WhenThenNode

## Reference
https://github.com/gluesql/gluesql/pull/708  @woojinnn

### DIFF
--- a/core/src/ast_builder/expr/mod.rs
+++ b/core/src/ast_builder/expr/mod.rs
@@ -1,4 +1,5 @@
 mod binary_op;
+mod case;
 mod exists;
 mod is_null;
 mod like;
@@ -12,6 +13,7 @@ pub mod extract;
 pub mod function;
 pub mod in_list;
 
+pub use case::case;
 pub use exists::{exists, not_exists};
 pub use nested::nested;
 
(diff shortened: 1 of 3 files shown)


core/src/ast_builder/expr/mod.rs
```diff
@@ -235,6 +242,31 @@ impl TryFrom<ExprNode> for Expr {
             ExprNode::Exists { subquery, negated } => Query::try_from(*subquery)
                 .map(Box::new)
                 .map(|subquery| Expr::Exists { subquery, negated }),
+            ExprNode::Case {
+                operand,
+                when_then,
+                else_result,
+            } => {
+                let operand = operand.map(|what| Expr::try_from(*what).map(Box::new).unwrap()); // (Expr::try_from(*operand).map(Box::new)?);
```
ever0de, line 250:
```suggestion
                let operand = operand
                    .map(|what| Expr::try_from(*what).map(Box::new))
                    .transpose()?;
```

panarch (approved):
It's meaningful to see that we now have a new FSM expanding ExprNode!
All looks great!
Thanks a lot 👍 👍 👍

PR #915, opened 2022-10-10
Implement macro that implements `TryBinaryOperator`,

ever0de:
1. In the binary operator of numeric , always calculate by changing `Rhs` based on the type of `Lhs`.
2. Only `try_multiply` is implemented for the `Interval` type, and the `Interval` type currently uses `std::ops::Mul` because there is no wrapping implementation for `checked_*` .
3. The `checked_*` functions only exist for integers, not floats. A float is converted to inf when it overflows. So we created a new `integer` module.


ever0de:
Since the test code can also be replaced with a macro, we will work on it.

ever0de:
And if the current dividing value of the div is 0, an Overflow error is raised.

ChobobDev:
Thank you for the fantastic work!
Now we can add more data types as we desire with ease!!!

ever0de:
rebase upstream

core/src/data/value/binary_op/integer/macros.rs
```diff
@@ -0,0 +1,569 @@
...
+                    overflow_err($variant($primitive::MAX), I128(0), Divide)
+                );
+                assert_eq!(
+                    $primitive::MAX.try_divide(&U8(0)),
+                    overflow_err($variant($primitive::MAX), U8(0), Divide)
+                );
+            }
+
+            #[test]
+            fn mod_overlfow() {
```
panarch, line 409: tiny typo correction :) 
`mod_overflow`

panarch (approved):
This is... great work that will make us keep adding numeric data types easily.
Thanks a lot for the meaningful contribution.
Merging 👍 👍 👍

PR #920, opened 2022-10-11
Add setting table_alias support to AST builder table init function

panarch:
Add `TableAliasNode` which supports setting table alias for source tables.

```rust
table("Bar").alias_as("b").select()
```
This above is identical to the below SQL query
```sql
SELECT * FROM Bar b;
```

### DIFF
--- a/core/src/ast_builder/mod.rs
+++ b/core/src/ast_builder/mod.rs
@@ -48,7 +48,7 @@ pub use {
     select_item::SelectItemNode,
     select_item_list::SelectItemList,
     show_columns::ShowColumnsNode,
-    table::TableNode,
+    table::{TableAliasNode, TableNode},
     update::UpdateNode,
 };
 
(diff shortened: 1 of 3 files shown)


devgony (approved):
LGTM!

🤔 Could we reuse alias_as() for inline_view in future?

panarch:
> LGTM!
> 
> 🤔 Could we reuse alias_as() for inline_view in future?

We can use the same name `alias_as` for both setting table_alias and `inline_view` but the structs we use may be different because the current `TableNode` is a root node for building `ast::Statement` but what we need for inline_view is to build `ast::Query`.

panarch:
Add TableAliasNode which supports setting table alias for source tables.

- AST builder example code
table("Bar").alias_as("b").select()

This above is identical to the below SQL query

- SQL
SELECT * FROM Bar b;

PR #922, opened 2022-10-11
Add ExprNode::subquery,

panarch:
Add ExprNode::subquery
Add ExprNode unit tests of syntatic sugar functions and From impls


ever0de (approved):
looks good

PR #923, opened 2022-10-11
Update AST builder between & hash_executor to accept different param …

panarch:
…types

`between`, `not_between` and `hash_executor` functions accept two `Into<ExprNode>` params.
Update these to accept different `Into<ExprNode>` type params.


ever0de (approved):
Ok, it took me a little while to figure this out.
Since `T` is determined by the caller, multiple types cannot be entered. So you have to split it into different generics.

PR #927, opened 2022-10-12
Add /pkg/rust/data/ to .gitignore

panarch:
Now the main rust package directory is `/pkg/rust/` so it would be expected for users to run `cargo run --example ..` in `/pkg/rust/` directory.
Then, temp `/pkg/rust/data` directory is automatically created but that was not in the `.gitignore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # GlueSQL project files
 /data/
+/pkg/rust/data/
 /storages/**/data/
 /storages/**/tmp/
 /reports/


.gitignore
```diff
@@ -1,5 +1,6 @@
 # GlueSQL project files
 /data/
```
ever0de, line 2: Can't you delete this?

panarch: I checked but we may need to keep both because..
```rust
cargo run --example ..
```
in the root directory still works :D

ever0de: Ah, such a case... Hmm, I'd like to recommend using it in the root.
What about `**/data` if it's hard?

panarch: :D That is also.. hard to consider, I think. because in `src/` directory `data` is commonly used as module names.

ever0de: Oh, that's right... It looks good now.

ever0de (approved):

PR #928, opened 2022-10-12
Add a new method to Row which finds value by ident,

panarch:
Rename Row::get_value to Row::get_value_by_index.
Add Row::get_value which takes columns & ident and return Option<&Value>. Apply new Row::get_value to existing codes

### DIFF
--- a/core/src/executor/context/blend_context.rs
+++ b/core/src/executor/context/blend_context.rs
@@ -33,49 +33,41 @@ impl<'a> BlendContext<'a> {
     }
 
     pub fn get_value(&'a self, target: &str) -> Option<&'a Value> {
-        let get_value = || {
-            self.columns
-                .iter()
-                .position(|column| column == target)
-                .map(|index| match &self.row {
-                    BlendContextRow::Shared(row) => row.get_value(index),
-                    BlendContextRow::Single(Some(row)) => row.get_value(index),
-                    BlendContextRow::Single(None) => Some(&Value::Null),
-                })
+        let value = match &self.row {
+            BlendContextRow::Shared(row) => row.get_value(&self.columns, target),
+            BlendContextRow::Single(Some(row)) => row.get_value(&self.columns, target),
+            BlendContextRow::Single(None) => Some(&Value::Null),
         };
 
-        match get_value() {
-            None => match &self.next {
-                None => None,
-                Some(context) => context.get_value(target),
-            },
-            Some(value) => value,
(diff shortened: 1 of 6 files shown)


core/src/executor/context/filter_context.rs
```diff
@@ -51,13 +51,15 @@ impl<'a> FilterContext<'a> {
...
+        if let Content::Some {
+            columns,
+            row: Some(row),
+            ..
+        } = &self.content
+        {
+            let value = row.get_value(columns, target);
 
-            if let Some(value) = value {
+            if value.is_some() {
```
ever0de, line 62: good

core/src/executor/context/blend_context.rs
```diff
@@ -33,49 +33,41 @@ impl<'a> BlendContext<'a> {
...
-                    BlendContextRow::Single(Some(row)) => row.get_value(index),
-                    BlendContextRow::Single(None) => Some(&Value::Null),
-                })
-        };
+            match &self.row {
+                BlendContextRow::Shared(row) => row.get_value(&self.columns, target),
+                BlendContextRow::Single(Some(row)) => row.get_value(&self.columns, target),
+                BlendContextRow::Single(None) => Some(&Value::Null),
+            }
+        })();
```
ever0de, line 62:
```rust
        let value = {
            if self.table_alias != table_alias {
                return None;
            }

            match &self.row {
                BlendContextRow::Shared(row) => row.get_value(&self.columns, target),
                BlendContextRow::Single(Some(row)) => row.get_value(&self.columns, target),
                BlendContextRow::Single(None) => Some(&Value::Null),
            }
        };
```

ever0de: There seems to be no need to use a closure (late review).

ever0de (approved):
There are only comments that are not related to this PR, so I approve them first.

core/src/executor/context/blend_context.rs
panarch: This closure is required because that causes different behavior.
closure version does not break the flow though `self.table != table_alias` happens but using block statement, it will call `return` and the rest codes will be skipped.

panarch: because of the next codes
image

ever0de: Oh sure.. That's right. Because there is a return statement in the block, it skips.
Looks like I missed it, thanks

PR #929, opened 2022-10-12
Replace plan/join.rs unit tests to use AST builder!

panarch:


devgony (approved):
It looks so neat and has way higher readability!
Thank you for all the efforts!

core/src/plan/join.rs
```diff
@@ -765,31 +596,13 @@ mod tests {
...
-                group_by: Vec::new(),
-                having: None,
-            })
-        };
-        assert_eq!(actual, expected, "complex where_clause:\n{sql}");
+        let expected = table("Player")
+            .select()
+            .left_join("PlayerItem")
+            .hash_executor("PlayerItem.user_id", "Player.id")
+            .hash_filter("PlayerItem.amount > 10 AND PlayerItem.amount * 3 <= 2")
```
ever0de, line 603: (Hope) It would be nice to have `AND` as a chaining method.

core/src/plan/join.rs
```diff
@@ -898,69 +663,20 @@ mod tests {
...
-                    left: Box::new(expr("u1.id")),
-                    op: BinaryOperator::Eq,
-                    right: Box::new(Expr::Subquery(Box::new(subquery))),
-                }),
-                group_by: Vec::new(),
-                having: None,
-            })
-        };
-        assert_eq!(actual, expected, "hash join in nested subquery:\n{sql}");
+        let expected = table("Player").alias_as("u1").select().filter(
```
zmrdltl, line 666: I am so impressed by the increase in readability through the query builder function that has such a long ast.

ever0de (approved):
I think I saw a piece of history

PR #930, opened 2022-10-12
Move Expr::Extract structure to Expr::Function::Extract

### LINKED ISSUE #912: [Translate] Refacting translate position `Cast`, `Extract`
## Description
Returns `Expr` in the form required for gluesql from `SqlExpr` provided by sqlparser.

Among them, there are functions that are translated to a lower level of `Expr` rather than `Expr::Function` e.g) cast, extract

This is because it follows the parsed result of sqlparser as it is.

It is `ambiguous` whether the parsed tokens will be classified as a `lower level of Expr` or a `lower level of Expr::function`.

In glueSQL, if an argument is received wrapped in `( )` in the sql statement, the standard is set to translate in the form of `Expr::Function::Cast` rather than `Expr::Cast`.

### lists for check
You can suggest features to move in [here](https://github.com/gluesql/gluesql/blob/372e3e70dec7df2cec72070be986560de34716c0/core/src/ast/expr.rs#L10)

```rust
pub enum Expr {
    Identifier(String),
    CompoundIdentifier{..},
    IsNull(Box<Expr>),
    IsNotNull(Box<Expr>),
    InList {..},
    InSubquery {..},
    Between {..},
    Like {..},
    ILike {..},
    BinaryOp {..},
    UnaryOp {..},
    Cast {..},
    Extract {..},
    Nested(Box<Expr>),
    Literal(AstLiteral),
    TypedString {..},
    Function(Box<Function>),
    Aggregate(Box<Aggregate>),
    Exists {..},
    Subquery(Box<Query>),
    Case {..},
}
```
## Tasks
- [x] #913
- [x] #930

zmrdltl:
resolve #912 

## Tasks
- [x] ast_builder
- [x] ast
- [x] executor
- [x] plan
- [x] translate
- [x] tests


core/src/ast/expr.rs
```diff
@@ -290,15 +283,6 @@ mod tests {
             .to_sql()
         );
 
-        assert_eq!(
```
zmrdltl, line 293: This is a unit test that will be added later for the to_sql() implementation for the enum Function.
Therefore, it will disappear temporarily.

core/src/plan/join.rs
```diff
@@ -826,6 +826,7 @@ mod tests {
         "
         );
         let actual = plan_join(&storage, &sql);
+
```
panarch, line 829: This looks not related to this PR.

core/src/ast_builder/expr/function.rs
```diff
@@ -1179,4 +1198,31 @@ mod tests {
...
+        let expected = "EXTRACT(HOUR FROM date)";
+        test_expr(actual, expected);
+
+        let actual = col("date").extract(DateTimeField::Minute);
+        let expected = "EXTRACT(MINUTE FROM date)";
+        test_expr(actual, expected);
+
+        let actual = col("date").extract(DateTimeField::Second);
+        let expected = "EXTRACT(SECOND FROM date)";
+        test_expr(actual, expected);
```
panarch, line 1226: It looks not neccessary to have `DateTimeField` enum variant cases because this unit test is for `extract` function so it is not for testing `DateTimeField`.
It would be enough to have only a single test for this.

And then, could we have a unit test which uses `pub fn extract`?
Currently, there exists `col("date"),extract(..)` but no direct `extract` fn case.

zmrdltl: Oh thanks for careful review for this i changed it

PR #933, opened 2022-10-13
Remove Option which wraps Row in FilterContext,

panarch:
There was no use case of Content::Some { row: None, .. } but Option type was unnecessarily being used. Remove Option from Content::Some { row } in FilterContext


ever0de (approved):
nice catch 👍

devgony (approved):

PR #935, opened 2022-10-13
[Dictionary] support `GLUE_INDEXES`

devgony:
## Goal
```sql
gluesql> CREATE TABLE Foo(id INT PRIMARY KEY, name TEXT);
gluesql> CREATE INDEX Foo_id_1 ON Foo (id + 1);
gluesql> CREATE INDEX Foo_name_concat ON Foo (name + '_');
gluesql> SELECT * FROM GLUE_INDEXES;
```
| TABLE_NAME | INDEX_NAME     | ORDER | EXPRESSION | UNIQUENESS |
|------------|----------------|-------|------------|------------|
| Foo        | PRIMARY        | BOTH  | id         | TRUE       |
| Foo        | Foo_id_1       | BOTH  | id + 1     | FALSE      |
| Foo        | Foo_name_concat | BOTH  | name + "_" | FALSE      |

## Todo
- [x] implement GLUE_INDEXES
- [x] PK should be included in GLUE_INDEXES as `UNIQUENESS` column

## In Next
- Unique Index
- Composite Index
- GLUE_INDEX_COLUMNS(EXPRESSIONS?)
- 🤔 Why don't ASC, DESC  work? => future works up to storage type
- adjust GLUE_INDEXES to show indexes from {TABLE_NAME}
- looks like CREATE INDEX {INDEX_NAME} ON {TABLE_NAME} (`FUNCTION(COLUMN)`) does not work

### DIFF
--- a/core/src/executor/fetch.rs
+++ b/core/src/executor/fetch.rs
@@ -164,9 +164,9 @@ pub async fn fetch_relation_rows<'a>(
             let rows = {
                 #[derive(Iterator)]
                 enum Rows<I1, I2, I3> {
-                    GlueTables(I1),
-                    GlueTabColumns(I2),
-                    GlueIndexes(I3),
+                    Tables(I1),
+                    TableColumns(I2),
+                    Indexes(I3),
                 }
                 match dict {
                     Dictionary::GlueTables => {
@@ -175,7 +175,7 @@ pub async fn fetch_relation_rows<'a>(
                             .into_iter()
                             .map(|schema| Ok(Row(vec![Value::Str(schema.table_name)])));
 
-                        Rows::GlueTables(rows)
+                        Rows::Tables(rows)
                     }
                     Dictionary::GlueTableColumns => {
                         let schemas = storage.fetch_all_schemas().await?;
@@ -192,7 +192,7 @@ pub async fn fetch_relation_rows<'a>(
                             )
                         });
 
-                        Rows::GlueTabColumns(rows)
+                        Rows::TableColumns(rows)
                     }
                     Dictionary::GlueIndexes => {
                         let schemas = storage.fetch_all_schemas().await?;
@@ -207,7 +207,7 @@ pub async fn fetch_relation_rows<'a>(
                             })
                         });
 
-                        Rows::GlueIndexes(rows)
+                        Rows::Indexes(rows)
                     }
                 }
             };
--- a/core/src/executor/fetch.rs
+++ b/core/src/executor/fetch.rs
@@ -220,7 +220,7 @@ pub async fn fetch_relation_rows<'a>(
                                 None => Vec::new(),
                             };
 
-                            let non_clustered = schema.indexes.into_iter().map(|index| {
+                            let non_clustered = schema.indexes.into_iter().map(move |index| {
                                 Ok(Row(vec![
                                     Value::Str(schema.table_name.clone()),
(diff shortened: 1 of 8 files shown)


core/src/executor/fetch.rs
```diff
@@ -227,6 +243,13 @@ pub async fn fetch_relation_columns(
                 "COLUMN_NAME".to_owned(),
                 "COLUMN_ID".to_owned(),
             ]),
+            Dictionary::GlueIndexes => Ok(vec![
+                "TABLE_NAME".to_owned(),
+                "INDEX_NAME".to_owned(),
+                "ORDER".to_owned(),
+                "EXPRESSION".to_owned(),
+                // "UNIQUENESS".to_owned(),
```
devgony, line 251: Looks like we should add index info right after creating PK

test-suite/src/dictionary_index.rs
```diff
@@ -0,0 +1,22 @@
...
+    run!("CREATE INDEX Foo_id_1 ON Foo (id + 1 DESC)");
+    run!("CREATE INDEX Foo_id_2 ON Foo (id * 2)");
+
+    let test_cases = [(
+        "SELECT * FROM GLUE_INDEXES",
+        Ok(select!(
+            TABLE_NAME       | INDEX_NAME           | ORDER             | EXPRESSION;
+            Str              | Str                  | Str               | Str;
+            "Foo".to_owned()   "Foo_id".to_owned()    "BOTH".to_owned()    "id".to_owned();
+            "Foo".to_owned()   "Foo_id_1".to_owned()  "BOTH".to_owned()   "id + 1".to_owned();
```
devgony, line 15: Looks like ASC, DESC don't work.. 🤔

ever0de: `SELECT * FROM GLUE_INDEXES ORDER BY INDEX_NAME DESC`
I see the test fails when I run it with , so it seems to work.

```shell
thread 'dictionary_index' panicked at 'assertion failed: `(left == right)`
   left: `Str("Foo_id_2")`,
  right: `Str("Foo_id")`:
[err: value] row index: 0
expected: [Str("Foo"), Str("Foo_id"), Str("BOTH"), Str("id")]
    found: [Str("Foo"), Str("Foo_id_2"), Str("BOTH"), Str("id * 2")]', test-suite/src/tester/mod.rs:81:17
```

devgony: I meant the DESC which is the value of ORDER (BOTH, ASC, DESC) not the keyword of SQL 😢
Whenever we create index, the type of order is stored in `SchemaIndex`

test-suite/src/dictionary_index.rs
```diff
@@ -0,0 +1,21 @@
+use {crate::*, gluesql_core::prelude::Value::*};
+
+test_case!(ditionary_index, async move {
+    run!("CREATE TABLE Foo (id INT PRIMARY KEY, name TEXT);");
```
panarch, line 4: There exists only a single line which is not covered.
ref.
https://coveralls.io/builds/53336746/source?filename=core%2Fsrc%2Fexecutor%2Ffetch.rs#L220

Could we fill the last one?
In here, if we add another table which does not have PRIMARY KEY before adding `Foo` and run `SELECT * FROM GLUE_INDEXES`, then we can expect the test to reach the line.

PR #937, opened 2022-10-13
Use macro to generate SQL parsers

### LINKED ISSUE #909: [Macro] Implement the `parse_*` fn generation macro
### Background 
Currently, we defined `parse_*` fn separately, which shows similar pattern; 1) tokenize 2) parse tokens with specific sql-parser fn.
By introducing new decl macro which takes `fn_name`, `parse_fn_name`, `output type` as args, `parse_*` functions can be easily generated.  
https://github.com/gluesql/gluesql/blob/10824e0c567cbd2689079ae771f59339fee0e581/core/src/parse_sql.rs#L41-L59

### TODO
1. impl parse_* macro
2. replace code blocks with macro

sa1:
Fixes #909

### DIFF
--- a/core/src/parse_sql.rs
+++ b/core/src/parse_sql.rs
@@ -18,124 +18,55 @@ pub fn parse<Sql: AsRef<str>>(sql: Sql) -> Result<Vec<SqlStatement>> {
     Parser::parse_sql(&DIALECT, sql.as_ref()).map_err(|e| Error::Parser(format!("{:#?}", e)))
 }
 
-pub fn parse_query<Sql: AsRef<str>>(sql_expr: Sql) -> Result<SqlQuery> {
-    let tokens = Tokenizer::new(&DIALECT, sql_expr.as_ref())
-        .tokenize()
-        .map_err(|e| Error::Parser(format!("{:#?}", e)))?;
+macro_rules! generate_parse_fn {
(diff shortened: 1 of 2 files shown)


core/src/parse_sql.rs
```diff
@@ -18,124 +18,55 @@ pub fn parse<Sql: AsRef<str>>(sql: Sql) -> Result<Vec<SqlStatement>> {
...
+parse_!(parse_order_by_expr, parse_order_by_expr, SqlOrderByExpr);
+parse_!(
+    parse_order_by_exprs,
+    parse_comma_separated,
+    parse_order_by_expr,
+    Vec<SqlOrderByExpr>
+);
+parse_!(parse_column_def, parse_column_def, SqlColumnDef);
+parse_!(parse_data_type, parse_data_type, SqlDataType);
+parse_!(parse_sql_assignment, parse_assignment, SqlAssignment);
```
ever0de, line 71: In this case, it seems that there is no need to enter sql in the middle, so I think you can delete it.

ever0de (commented):
Welcome

I left my little comments. This is a formatting level opinion, so please leave any additional comments.

sa1:
Thanks, I implemented those changes.

panarch (approved):
Looks quite much better! Thanks a lot 👍

PR #938, opened 2022-10-14
Fix web/ payload data_type json serialization,

panarch:
Fix payload data_type json serialization.
Fix gluesql-js web/ integration tests.

I will also add gluesql-js headless browser test to GitHub Action soon.

### DIFF
--- a/pkg/javascript/web/tests/memory_storage.rs
+++ b/pkg/javascript/web/tests/memory_storage.rs
@@ -24,5 +24,4 @@ impl Tester<MemoryStorage> for MemoryTester {
 }
 
 generate_store_tests!(wasm_bindgen_test, MemoryTester);
-generate_dictionary_tests!(wasm_bindgen_test, MemoryTester);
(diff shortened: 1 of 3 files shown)


pkg/javascript/web/tests/memory_storage.rs
```diff
@@ -24,5 +24,4 @@ impl Tester<MemoryStorage> for MemoryTester {
 }
 
 generate_store_tests!(wasm_bindgen_test, MemoryTester);
-generate_dictionary_tests!(wasm_bindgen_test, MemoryTester);
```
ever0de, line 27: Is there any reason this is missing?

panarch: We now removed `Metadata` trait so there does not exist `generate_dctionary_tests!` anymore, it is merged into `generate_store_tests!`

ever0de (approved):

PR #940, opened 2022-10-14
Feat : Apply macros in i16 data type

ChobobDev:
As we now have new casting rules and the macro,  the following PR aims to refactor the previous i16 data type under new casting rules.


core/src/data/value/binary_op/integer/i16.rs
```diff
@@ -0,0 +1,41 @@
...
+
+impl PartialEq<Value> for i16 {
+    fn eq(&self, other: &Value) -> bool {
+        match other {
+            I8(other) => (*self as i8) == *other,
+            I16(other) => self == other,
+            I32(other) => (*self as i32) == *other,
+            I64(other) => (*self as i64) == *other,
+            I128(other) => (*self as i128) == *other,
+            U8(other) => (*self as i64) == (*other as i64),
```
ever0de, line 19:
```suggestion
            U8(other) => *self == (*other as i16),
```
There doesn't seem to be any need to convert to i64. because it's a comparison

core/src/data/value/binary_op/integer/i16.rs
```diff
@@ -0,0 +1,41 @@
...
+
+impl PartialOrd<Value> for i16 {
+    fn partial_cmp(&self, other: &Value) -> Option<Ordering> {
+        match other {
+            I8(other) => (*self as i8).partial_cmp(other),
+            I16(other) => self.partial_cmp(other),
+            I32(other) => (*self as i32).partial_cmp(other),
+            I64(other) => (*self as i64).partial_cmp(other),
+            I128(other) => (*self as i128).partial_cmp(other),
+            U8(other) => (*self as i64).partial_cmp(&(*other as i64)),
```
ever0de, line 35: ditto

core/src/data/value/binary_op/integer/macros.rs
```diff
@@ -188,7 +188,7 @@ macro_rules! generate_binary_op_tests {
 
             #[test]
             fn eq() {
-                let base: $primitive = 1.into();
+                let base: $primitive = 1;
```
ever0de, line 191: good

ever0de (commented):
Looks good. I left a few small comments.
[nit-picking] Instead of adding a macro to force the "new casting rule" in the PR title, apply the macro? It seems to fit the same thing.

ever0de (approved):
LGTM!!!

panarch (approved):
Wow it is hard to believe that we can now have a new numeric type efficiently like this!
Appreciating again about the new macro @ever0de, thanks a lot @ChobobDev 
All looks nice!

PR #945, opened 2022-10-15
Support `Expr::Case` in aggregate module

ever0de:
Thanks to @24seconds


24seconds (approved):
Looks good to me with trust.

ever0de:
Added missing implementation of `Expr::Case` to `check` function.

panarch (approved):

PR #946, opened 2022-10-15
Check for `Value::Null` in `Aggregate` except `*`,

### LINKED ISSUE #944: COUNT(NULL) is 1
Hello, while I'm using the gluesql-js, I found the bug.

```
COUNT(NULL) is 1 in glueql. But it should be zero I think.
```

Could you take a look? Thank you!

ever0de:
fixed #944 

- if `Value::Null`, init count is `0`, otherwise `1`


panarch (approved):
Wow... this was a critical bug.
Thanks really a lot for the fix.

also thanks for reporting @24seconds

PR #948, opened 2022-10-15
fix: unmatced subquery should return Null (not error)

devgony:
## Goal
Currently,
```sql
gluesql> SELECT (SELECT N FROM SERIES(3) WHERE N = 4) N;
[error] nested select row not found
```
But it should be
```sql
gluesql> SELECT (SELECT N FROM SERIES(3) WHERE N = 4) N;
```
| N    |
|------|
| NULL |

It affects to scala subquery, In-List subquery.

## Todo
- [x] modify EvaluateError::NestedSelectRowNotFound to Value::Null
- [x] remove EvaluateError::NestedSelectRowNotFound

### DIFF
--- a/core/src/executor/evaluate/error.rs
+++ b/core/src/executor/evaluate/error.rs
@@ -10,9 +10,6 @@ pub enum EvaluateError {
     #[error(transparent)]
     ChronoFormat(#[from] ChronoFormatError),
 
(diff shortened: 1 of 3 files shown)


panarch (approved):
This is really a good catch. Thanks a lot! All looks nice

PR #949, opened 2022-10-15
Fix test-suite tester::test to properly print expected & found labels

panarch:
use of `found` and `expected` were wrongly used in `tester::test` test helper function.


ever0de (approved):
it was missed 
Good👍 👍 👍

PR #953, opened 2022-10-16
Rename AST builder ExprNode::not to ExprNode::negate

### LINKED ISSUE #791: Rename AST builder `ExprNode::not` to `ExprNode::negate`
Rather than `col("id").not()`, it would be better to use different name for `UnaryOperator::Not` in this case.
How about `negate`? 
e.g. `col("id").negate()`
This is only for `ExprNode::not` function.

ref. https://github.com/gluesql/gluesql/blob/06ba99eabcf38662c4fe85946b700a0f22b57b05/core/src/ast_builder/expr/unary_op.rs#L10-L13

Using `not` as a separate function looks good.
e.g. This looks ok.
```rust
not(col("id"))
// is equal to
"NOT id"
```

sa1:
Fixes #791

### DIFF
--- a/core/src/ast_builder/expr/unary_op.rs
+++ b/core/src/ast_builder/expr/unary_op.rs
@@ -8,7 +8,7 @@ impl ExprNode {
         minus(self)
     }
     #[allow(clippy::should_implement_trait)]
-    pub fn not(self) -> Self {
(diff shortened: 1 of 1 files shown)


panarch (approved):
Thanks a lot!

PR #955, opened 2022-10-17
Add AST builder integration tests of SELECT, UPDATE, INSERT and DELET…

panarch:
…E statements

Main purpose of AST builder integration tests would be providing usage guide to our users.
Add some usage examples of select, update, insert and delete queries.


ever0de (approved):
Looks good to me

PR #960, opened 2022-10-19
fix: handle Error when DROP and CREATE PRIMARY INDEX manually

devgony:
## Goal
At #935 , I missed a few edge cases like,
1. dropping index named as `PRIMARY`
```sql
DROP INDEX Bar.PRIMARY;
```
=> should get `cannot drop primary index` error

2. creating index named as `PRIMARY`
```sql
CREATE INDEX Primary ON Foo (id);
```
=> should get `cannot create index with reserved name "Primary"` error

## Todo
- [x] test: add cases for DROP and CREATE PRIMARY
- [x] feat: implement cannot drop primary
- [x] feat: implement cannot create index named as PRIMARY

## Next
- Implement Primary key with physical index
- It is expected to drop and create primary index manually in future.

### DIFF
--- a/core/src/executor/alter/error.rs
+++ b/core/src/executor/alter/error.rs
@@ -28,7 +28,4 @@ pub enum AlterError {
 
     #[error("identifier not found: {0:#?}")]
     IdentifierNotFound(Expr),
-
-    #[error("cannot create index with reserved name: {0}")]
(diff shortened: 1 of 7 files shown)


panarch (approved):
Looks all good! Thanks a lot 👍 👍

PR #962, opened 2022-10-22
Remove `macro-exports` in `data/../macros`

ever0de:
Users will never use this macro.


panarch (approved):
Awesome!
I didn't know how to make exported macro not to be visible to user-side and this looks really nice.
Thanks a lot 👍

PR #963, opened 2022-10-22
Remove unneccesary syntax in enum Evaluated

zmrdltl:
## Description
```rust
pub enum Evaluated<'a> {
    Literal(Literal<'a>),
    Value(Cow<'a, Value>),
}
```

to

```rust
pub enum Evaluated<'a> {
    Literal(Literal<'a>),
    Value(Value),
}
```
enum `Evaluated` does not need to use Cow for enum `Value`


panarch (approved):
Thanks a lot for this nice improvement :D
All looks good 👍

PR #965, opened 2022-10-22
[AST] Implement CONCAT_WS function In AST

seonghun-dev:
- Add `concat_ws` function in ast

### DIFF
--- a/core/src/executor/evaluate/function.rs
+++ b/core/src/executor/evaluate/function.rs
@@ -88,7 +88,7 @@ pub fn concat_ws<'a>(
 
     let result = exprs
         .into_iter()
-        .map(|expr| expr.try_into())
+        .map(Value::try_from)
         .filter(|value| !matches!(value, Ok(Value::Null)))
         .map(|value| Ok(String::from(value?)))
         .collect::<Result<Vec<_>>>()?
--- a/core/src/executor/evaluate/function.rs
+++ b/core/src/executor/evaluate/function.rs
@@ -79,11 +79,11 @@ pub fn concat(exprs: Vec<Evaluated<'_>>) -> Result<Evaluated> {
     }
 }
 
-pub fn concat_ws(
+pub fn concat_ws<'a>(
     name: String,
-    separator: Evaluated<'_>,
-    exprs: Vec<Evaluated<'_>>,
-) -> Result<Value> {
+    separator: Evaluated<'a>,
+    exprs: Vec<Evaluated<'a>>,
+) -> Result<Evaluated<'a>> {
     let separator = eval_to_str!(name, separator);
 
     let result = exprs
@@ -97,7 +97,7 @@ pub fn concat_ws(
         .collect::<Result<Vec<_>>>()?
         .join(&separator);
 
-    Ok(Value::Str(result))
+    Ok(Evaluated::from(Value::Str(result)))
 }
 
 pub fn lower(name: String, expr: Evaluated<'_>) -> Result<Evaluated> {
--- a/core/src/executor/evaluate/function.rs
+++ b/core/src/executor/evaluate/function.rs
@@ -90,10 +90,7 @@ pub fn concat_ws<'a>(
         .into_iter()
         .map(|expr| expr.try_into())
         .filter(|value| !matches!(value, Ok(Value::Null)))
-        .map(|value| match value {
-            Ok(value) => Ok(String::from(value)),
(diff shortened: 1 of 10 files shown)


seonghun-dev:
Same coverage Issue in https://github.com/gluesql/gluesql/pull/899

core/src/executor/evaluate/function.rs
```diff
@@ -79,6 +79,27 @@ pub fn concat(exprs: Vec<Evaluated<'_>>) -> Result<Evaluated> {
...
+) -> Result<Evaluated<'a>> {
+    let separator = eval_to_str!(name, separator);
+
+    let result = exprs
+        .into_iter()
+        .map(|expr| expr.try_into())
+        .filter(|value| !matches!(value, Ok(Value::Null)))
+        .map(|value| match value {
+            Ok(value) => Ok(String::from(value)),
+            Err(_err) => Ok(String::from(Value::Str("".to_owned()))),
```
panarch, line 95: It would be good to consider errors as terminal nodes.
Other states can be converted into error, but no case from error to non-error cases.

Though we know that it is currently not possible to write tests which reaches to this, it is a hidden context which others cannot recognize.

It is ok that we don't have a test case for this, but it can be a problem if we ignore the error because we cannot track the changes from outside.

core/src/plan/expr/function.rs
```diff
@@ -1,17 +1,18 @@
...
-        enum Exprs<I0, I1, I2, I3, I4> {
+    pub fn as_exprs(&self) -> impl Iterator<Item = &Expr> {
+        #[derive(iter_enum::Iterator)]
+        enum Exprs<I0, I1, I2, I3, I4, I5> {
             Empty(I0),
             Single(I1),
             Double(I2),
             Triple(I3),
             VariableArgs(I4),
+            TypedStringVariableArgs(I5),
```
panarch, line 15: This looks interesting that we need another name for variable args with a single expression.

nit suggestion for the name.
Though these are internal variables so it does not need to be named too carefully..
We have a variant for `ast::Expr::TypedString` and `TypedString` keyword here is not related to `Expr::TypedString` so it could be little bit confusing.
Then.. how about simply `VariableArgs2` or `VariableArgsWithSingle`...?

panarch (changes requested):

test-suite/src/function/concat_ws.rs
```diff
@@ -0,0 +1,92 @@
...
+        CREATE TABLE Concat (
+            id INTEGER,
+            rate FLOAT,
+            flag BOOLEAN,
+            text TEXT,
+            null_value TEXT NULL,
+        );
+    "
+    );
+    run!(r#"INSERT INTO Concat VALUES (1, 2.3, TRUE, "Foo", NULL);"#);
```
panarch, line 27: I missed in the previous review. Everything else looks nice.

It looks that all the current tests are only using literals.
However, you've already added values like this so could we have tests which use these `id`, `rate`, `flag`, `text` or `null_value` field?

core/src/translate/function.rs
```diff
@@ -213,6 +213,19 @@ pub fn translate_function(sql_function: &SqlFunction) -> Result<Expr> {
...
             Ok(Expr::Function(Box::new(Function::Concat(exprs))))
         }
+        "CONCAT_WS" => {
+            check_len_min(name, args.len(), 2)?;
+            let separator = translate_expr(args[0])?;
+            let exprs = args[1..]
+                .iter()
+                .copied()
+                .map(translate_expr)
+                .collect::<Result<Vec<_>>>()?;
```
panarch, line 223: Here you can use 
```rust
let exprs = args
    .into_iter()
    .skip(1)
    .map(translate_expr)
```

core/src/executor/evaluate/function.rs
```diff
@@ -79,6 +79,24 @@ pub fn concat(exprs: Vec<Evaluated<'_>>) -> Result<Evaluated> {
...
+pub fn concat_ws<'a>(
+    name: String,
+    separator: Evaluated<'a>,
+    exprs: Vec<Evaluated<'a>>,
+) -> Result<Evaluated<'a>> {
+    let separator = eval_to_str!(name, separator);
+
+    let result = exprs
+        .into_iter()
+        .map(|expr| expr.try_into())
```
panarch, line 91: nit.
```rust
    .map(Value::try_from)
```

panarch (approved):
All looks nice! Thanks a lot 👍 👍

PR #966, opened 2022-10-22
Refactor function's return type in evaluate_function()

zmrdltl:
## Description
All existing [functions](https://github.com/gluesql/gluesql/blob/main/core/src/executor/evaluate/function.rs#L60) operating in evaluate are evaluated by enum `Value` without having to have an intermediate value, 

Evaluated. However, if it is used in a future query builder, if it is a chained structure in a function such as a `substring` that should always hold an intermediate value, an overhead may occur by repeating the clone. 

In this case, it can be solved by string slicing to keep the intermediate value. Modify the return values ​​for this structure.

## Tasks
- [x] change return type `Result<Value>` to `Result<Evaluated>`

### DIFF
--- a/core/src/executor/evaluate/function.rs
+++ b/core/src/executor/evaluate/function.rs
@@ -14,7 +14,7 @@ macro_rules! eval_to_str {
         match $evaluated.try_into()? {
             Value::Str(value) => value,
             Value::Null => {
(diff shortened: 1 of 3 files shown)


panarch (approved):
Looking forward to see another big improvement coming to `Evaluated` in execution layer!

PR #967, opened 2022-10-23
Update AST builder ExprNode & ExprList to accept borrowed values

panarch:
Update `ExprNode` & `ExprList` to accept both `String` and `&str` to build `ExprNode`.
Update `ExprList` to accept both owned vector and borrowed slice.

e.g. `ExprList`
```rust
        // Vec<Into<ExprNode>>
        let actual: ExprList = vec!["id", "name"].into();
        let expected = "id, name";
        test(actual, expected);

        // Vec<ExprNode>
        let actual: ExprList = vec![col("id"), col("name")].into();
        let expected = "id, name";
        test(actual, expected);

        // &Vec<ExprNode>
        let actual = vec!["id", "name"];
        let actual: ExprList = (&actual).into();
        let expected = "id, name";
        test(actual, expected);

        // &[Into<ExprNode>]
        let actual = vec!["rate / 10", "col1"];
        let actual: ExprList = actual.as_slice().into();
        let expected = "rate / 10, col1";
        test(actual, expected);

        // &[ExprNode]
        let actual = vec![col("id"), col("name")];
        let actual: ExprList = actual.as_slice().into();
        let expected = "id, name";
        test(actual, expected);
```

### DIFF
--- a/core/src/ast_builder/expr_list.rs
+++ b/core/src/ast_builder/expr_list.rs
@@ -6,29 +6,48 @@ use {
         result::{Error, Result},
         translate::translate_expr,
     },
+    std::borrow::Cow,
 };
 
 #[derive(Clone)]
 pub enum ExprList<'a> {
-    Text(String),
-    Exprs(Vec<ExprNode<'a>>),
+    Text(Cow<'a, str>),
+    Exprs(Cow<'a, [ExprNode<'a>]>),
 }
 
-impl<'a> From<&str> for ExprList<'a> {
-    fn from(exprs: &str) -> Self {
-        ExprList::Text(exprs.to_owned())
+impl<'a> From<&'a str> for ExprList<'a> {
+    fn from(exprs: &'a str) -> Self {
+        ExprList::Text(Cow::Borrowed(exprs))
     }
 }
 
-impl<'a> From<Vec<ExprNode<'a>>> for ExprList<'a> {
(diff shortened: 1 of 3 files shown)


core/src/ast_builder/expr_list.rs
```diff
@@ -41,10 +60,74 @@ impl<'a> TryFrom<ExprList<'a>> for Vec<Expr> {
...
-                .into_iter()
-                .map(TryInto::try_into)
-                .collect::<Result<Vec<_>>>(),
+            ExprList::Exprs(exprs) => {
+                let exprs = exprs.into_owned();
+
+                exprs
+                    .into_iter()
+                    .map(TryInto::try_into)
+                    .collect::<Result<Vec<_>>>()
```
ever0de, line 69: <https://github.com/rust-lang/rust-clippy/blob/967f172e256802f9bf0ad5718d9c761da18132a9/clippy_lints/src/methods/unnecessary_to_owned.rs#L198-L239>

It gives an `unnecessary_to_owned` clippy error, and clippy tells you to fix it like this:

```
T.iter().cloned()
```

But it seems rather unnecessary referencing.
The behavior of `into_owned` is only to strip the enum when `Cow::Owned`, whereas `Iterator::cloned` calls `Clone::clone`.

So we separated it like this.

ever0de (approved):
Looks good 👍 👍 👍 👍 👍

PR #971, opened 2022-10-28
Expose wasm-pack's init function argument to users

arbfay:
Some users need to be able to use the `module_or_path` argument of the function `init` that `wasm-pack` generates. In my case, it enables loading the WASM module in an environment that is neither the browser nor node (edge). This modification can fix WASM loading issues for other users too.


panarch (approved):
Awesome, thanks that I also now got to know how wasm-pack init works.
Thanks a lot!

PR #976, opened 2022-10-29
Implement FormatType in Function::Format

### LINKED ISSUE #964: Support FormatType
## Description

Support FormatType according to lots of options(datetime, binary, hex) 

## Task

```rust
pub enum FormatType {
    Datetime(Expr),
    Binary,
    Hex,
}

impl ToSql for FormatType {
    fn to_sql(&self) -> String {
        match self {
            FormatType::Datetime(e) => e.to_sql(),
            FormatType::Binary => "BINARY".to_owned(),
            FormatType::Hex => "HEX".to_owned(),
        }
    }
}
```

CEOJINSUNG:
Resolve #964

### DIFF
--- a/core/src/translate/function.rs
+++ b/core/src/translate/function.rs
@@ -5,7 +5,7 @@ use {
         translate_data_type, translate_object_name, TranslateError,
     },
     crate::{
-        ast::{Aggregate, CountArgExpr, Expr, Function},
+        ast::{Aggregate, CountArgExpr, Expr, FormatType, Function},
         result::Result,
     },
     sqlparser::ast::{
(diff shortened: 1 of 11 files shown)


core/src/translate/function.rs
```diff
@@ -411,7 +416,11 @@ pub fn translate_function(sql_function: &SqlFunction) -> Result<Expr> {
             check_len(name, args.len(), 2)?;
 
             let expr = translate_expr(args[0])?;
-            let format = translate_expr(args[1])?;
+            let format = match args_str[1].as_str() {
```
ever0de, line 419: If you use only the first index item, it seems better to replace `args_str` with a single item.
ex)
```rust
let args = function_args_exprs
     .iter()
     .map(ToString::to_string)
     .skip(1)
     .next();
```

PR #978, opened 2022-10-29
Add enum `StrSlice` under enum `Evaluated`

zmrdltl:
## Description
after #966 
If a function that handles a string is chained and executed on the query builder, the string is copied every time by the Clone trait. This overhead is solved by always having a sliced ​​enum with a value.

## Tasks
- [x] add enum StrSlice
- [ ] modify string function structure for StrSlice
- [ ] tests


zmrdltl:
If this structure is to be used in another function, it is a structure that borrows and returns an owned string, so the implementation of `StrSlice {&str}` fails and closes.

PR #983, opened 2022-11-03
Refactor: remove `ColumnOptionDef`

### LINKED ISSUE #981: Unnecessary `name` field in `ColumnOptionDef` type
Currently when we want to add column option, we use a type called `ColumnOptionDef`.

```rust
pub struct ColumnOptionDef {
    pub name: Option<String>,
    pub option: ColumnOption,
}
```

But I see this `name` field unnecessary. I assume that field is there because of the future support for `CONSTRAINT <constrain_name> PRIMARY KEY ...`, but I think column options and constraints should be treated separately.   

When we can purge out the `name` field, we can just use `ColumnOption` directly in `ColumnDef` without intermediary `ColumnOptionDef` type.

```rust
pub struct ColumnDef {
    pub name: String,
    pub data_type: DataType,
    pub options: Vec<ColumnOption>,
}

pub enum ColumnOption {
    Null,
    NotNull,
    Default(Expr),
    Unique { is_primary: bool },
}
```

MRGRAVITY817:
## Resolves #981 

This PR removes the type `ColumnOptionDef`, replacing its role with `ColumnOption`.  (Details in #981)
Overall functionality won't change, only the translation from `sqlparser` to `gluesql` has been changed a bit.

### DIFF
--- a/core/src/translate/ddl.rs
+++ b/core/src/translate/ddl.rs
@@ -79,9 +79,8 @@ pub fn translate_column_def(sql_column_def: &SqlColumnDef) -> Result<ColumnDef> 
 fn translate_column_option_def(
     sql_column_option_def: &SqlColumnOptionDef,
 ) -> Result<Vec<ColumnOption>> {
-    let SqlColumnOptionDef { name, option } = sql_column_option_def;
+    let SqlColumnOptionDef { option, .. } = sql_column_option_def;
 
-    let name = name.as_ref().map(|name| name.value.to_owned());
     let option = match option {
         SqlColumnOption::Null => Ok(ColumnOption::Null),
         SqlColumnOption::NotNull => Ok(ColumnOption::NotNull),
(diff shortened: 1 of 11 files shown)


core/src/translate/ddl.rs
```diff
@@ -72,12 +72,15 @@ pub fn translate_column_def(sql_column_def: &SqlColumnDef) -> Result<ColumnDef>
     })
 }
 
+/// Translate `ColumnOptionDef` to `ColumnOption`.
```
ever0de, line 75: Nice comment :+1: , if you change it like this it will take the link.

```suggestion
/// Translate [`SqlColumnOptionDef`] to [`ColumnOption`].
```

image

ever0de (commented):
There are minor comments. Other changes are good.

core/src/translate/ddl.rs
MRGRAVITY817: Ah yes, good point! Thanks a lot 👍

ever0de (approved):
Looks good

panarch (approved):
Looks all nice! Thanks a lot :D

PR #986, opened 2022-11-05
Merge ColumnOption::{Null, NotNull} into a single option

### LINKED ISSUE #969: Merge `ColumnOption::{Null, NotNull}` into a single option
There exists two `Null` and `NotNull` column options in GlueSQL's AST.
However, the information we only need is that the column is whether nullable or not.
We can consider to replace `ColumnOption::Null` and `ColumnOption:NotNull` to a single option `ColumnOption::Nullable`.

https://github.com/gluesql/gluesql/blob/79e88ac1c6ff307b7e79915f0932efafbf94927b/core/src/ast/ddl.rs#L38-L48

devgony:
To resolve #969,
### Remove redundant ColumnOption::{NotNull, Null}
```diff
pub struct ColumnDef {
    pub name: String,
    pub data_type: DataType,
+   pub nullable: bool,
    pub options: Vec<ColumnOption>,
}

pub enum ColumnOption {
-   Null,
-   NotNull,
    Default(Expr),
    Unique { is_primary: bool },
}

impl ColumnDef {
-   pub fn is_nullable(&self) -> bool {
-       self.options
-           .iter()
-           .any(|option| option == &ColumnOption::Null)
-   }
..
```

## Todo
- [x] Add nullable field to ColumnDef
- [x] Remove ColumnOption::{NotNull, Null}, 
- [x] ColumnDef.is_nullable

## Next
- Modify default nullable to true
- Should handle primary_key(NotNull) and  unique only (Null) properly

### DIFF
--- a/core/src/data/row.rs
+++ b/core/src/data/row.rs
@@ -87,9 +87,8 @@ impl Row {
                     .map(|(_, value)| value);
 
                 match (value, column_def.get_default(), nullable) {
-                    (Some(&expr), _, _) | (None, Some(expr), _) => {
-                        evaluate_stateless(None, expr)?.try_into_value(data_type, nullable)
-                    }
+                    (Some(&expr), _, _) | (None, Some(expr), _) => evaluate_stateless(None, expr)?
+                        .try_into_value(data_type, nullable.to_owned()),
                     (None, None, true) => Ok(Value::Null),
                     (None, None, false) => {
                         Err(RowError::LackOfRequiredColumn(def_name.to_owned()).into())
@@ -118,7 +117,7 @@ impl Row {
             } = column_def;
 
             value.validate_type(data_type)?;
-            value.validate_null(nullable)?;
+            value.validate_null(nullable.to_owned())?;
         }
 
         Ok(())
--- a/core/src/data/row.rs
+++ b/core/src/data/row.rs
@@ -87,8 +87,9 @@ impl Row {
                     .map(|(_, value)| value);
 
                 match (value, column_def.get_default(), nullable) {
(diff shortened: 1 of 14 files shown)


core/src/data/value/mod.rs
```diff
@@ -177,7 +177,7 @@ impl Value {
         Ok(())
     }
 
-    pub fn validate_null(&self, nullable: bool) -> Result<()> {
+    pub fn validate_null(&self, nullable: &bool) -> Result<()> {
```
panarch, line 180: `bool` implements `Copy` trait so it does not need to be `&bool`, it would be simply enough to use `bool`

core/src/executor/evaluate/evaluated.rs
```diff
@@ -206,7 +206,7 @@ impl<'a> Evaluated<'a> {
         }
     }
 
-    pub fn try_into_value(self, data_type: &DataType, nullable: bool) -> Result<Value> {
+    pub fn try_into_value(self, data_type: &DataType, nullable: &bool) -> Result<Value> {
```
panarch, line 209: this, too. `bool` would be enough.

panarch (changes requested):

core/src/data/row.rs
```diff
@@ -118,7 +117,7 @@ impl Row {
             } = column_def;
 
             value.validate_type(data_type)?;
-            value.validate_null(nullable)?;
+            value.validate_null(nullable.to_owned())?;
```
panarch, line 120: `to_owned()` for `Copy` impl value might be.. too much.
you can simply use `*nullable` for these cases.

PR #988, opened 2022-11-05
Replace double quote to single quote

devgony:
### Let's use single quote only for sql literal string
```sql
SELECT 'literal';
```

## Todo
- [x] refactor: replace double quote to single quote

## Next
- fix translate to handle double qoute for Identifier


panarch (approved):
I wasn't aware of that SQL handles single quoted & double quoted differently.
Thanks a lot! Looks all nice

PR #990, opened 2022-11-06
Replace `TryFrom for AstLiteral` => `TryFrom for Expr`, add more test to dump

devgony:
## Todo
- [x] Replace TryFrom<Value> for AstLiteral => TryFrom<Value> for Expr
- [x] Add more data types to E2E test case (except F64 and Decimal due to floating point issue)
- [x] Replace double to single quote at interval

## Next
- Solve floating point issue on Bigdecimal.to_string for `to_sql`


panarch (approved):
Everything looks nice!
I wasn't also aware of using `Expr` rather than `AstLiteral`.
Now it looks that we are ready to release v0.13.1.
Thanks a lot 👍 👍 👍

PR #997, opened 2022-11-10
Modify default ColumnOption from NOT NULL to NULL

### LINKED ISSUE #974: Modify default ColumnOption from `NOT NULL` to `NULL`
## Issue
Currently, default ColumnOption is `NOT NULL` but many users would consider column as `NULL`
```sql
CREATE TABLE Foo (no INT);
```
| TABLE_NAME | COLUMN_NAME | NULLABLE |
|------------|-------------|-----------|
| Foo        | no          | NOT NULL         |
## Goal
To reduce confusion, Let's modify default ColumnOption from `NOT NULL` to `NULL`
```sql
CREATE TABLE Foo (no INT);
```
| TABLE_NAME | COLUMN_NAME | NULLABLE |
|------------|-------------|-----------|
| Foo        | no          | NULL         |

devgony:
To resolve #974 

## Goal
For better user experience, Let's modify default ColumnDef.nullable from `false` to `true`
```sql
CREATE TABLE Items (id INT);
```
### Before
| TABLE_NAME | COLUMN_NAME | NULLABLE |
|------------|-------------|-----------|
| Items        | id          | NOT NULL         |

### After
| TABLE_NAME | COLUMN_NAME | NULLABLE |
|------------|-------------|-----------|
| Items        | id          | NULL         |

## Todo
- [x] modify default ColumnDef.nullable from `false` to `true`
- [x] add Column name validation and RowErr::WrongColumnName)

### DIFF
--- a/core/src/data/row.rs
+++ b/core/src/data/row.rs
@@ -66,13 +66,13 @@ impl Row {
             return Err(RowError::TooManyValues.into());
         }
 
-        // if let Some(wrong_column_name) = columns.iter().find(|column_name| {
-        //     !column_defs
-        //         .iter()
-        //         .any(|column_def| &&column_def.name == column_name)
-        // }) {
-        //     return Err(RowError::WrongColumnName(wrong_column_name.to_owned()).into());
-        // }
+        if let Some(wrong_column_name) = columns.iter().find(|column_name| {
+            !column_defs
+                .iter()
+                .any(|column_def| &&column_def.name == column_name)
+        }) {
+            return Err(RowError::WrongColumnName(wrong_column_name.to_owned()).into());
+        }
 
         let columns = if columns.is_empty() {
             Columns::All(column_defs.iter().map(|ColumnDef { name, .. }| name))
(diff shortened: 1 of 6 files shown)


core/src/data/row.rs
```diff
@@ -63,6 +66,14 @@ impl Row {
             return Err(RowError::TooManyValues.into());
         }
 
+        if let Some(wrong_column_name) = columns.iter().find(|column_name| {
+            !column_defs
+                .iter()
+                .any(|column_def| &&column_def.name == column_name)
+        }) {
+            return Err(RowError::WrongColumnName(wrong_column_name.to_owned()).into());
```
devgony, line 74: To resovle case like
```sql
gluesql> CREATE TABLE Items (id INT NULL, name TEXT NULL);
Table created

gluesql> INSERT INTO Items (status) VALUES (1);
1 row inserted
```
It's so confused whether validate here or not

devgony: I just had missed more test case. It's reachable! even if i validate at the top!

ever0de (approved):
👍

panarch (approved):

PR #999, opened 2022-11-12
Add enum `StrSlice` under enum `Evaluated`

zmrdltl:
## Description
after #966 
If a function that handles a string is chained and executed on the query builder, the string is copied every time by the Clone trait. This overhead is solved by always having a sliced ​​enum with a value.

## Implementation plan
```rust
#[derive(Clone)]
pub enum Evaluated<'a> {
    Literal(Literal<'a>),
    StrSlice { source: String, range: Range<usize> },
    Value(Value),
}
```
If a variant called source holds the source of the string, and the string function is called continuously after that, the range is updated every time.

## Tasks
- [x] add enum StrSlice

modify string function structure for StrSlice
  - - [x] Substr
  - - [x] trim
  - - [x] ltrim
  - - [x] rtrim
 
- [x] tests


core/src/data/key.rs
```diff
@@ -105,6 +105,14 @@ impl TryFrom<&Value> for Key {
...
 }
 
+impl TryFrom<&String> for Key {
+    type Error = Error;
+
+    fn try_from(s: &String) -> Result<Self> {
+        Ok(Key::Str(s.to_owned()))
+    }
+}
+
```
ever0de, line 115: Since it is not a failing case, it seems appropriate to use `From`.

core/src/executor/evaluate/evaluated.rs
```diff
@@ -79,6 +93,9 @@ impl TryFrom<Evaluated<'_>> for HashMap<String, Value> {
             Evaluated::Value(Value::Str(v)) => HashMap::parse_json_object(v.as_str()),
             Evaluated::Value(Value::Map(v)) => Ok(v),
             Evaluated::Value(v) => Err(EvaluateError::MapOrStringValueRequired(v.into()).into()),
+            Evaluated::StrSlice { source, range } => {
```
zmrdltl, line 96: When writing StrSlice as a test, is it possible to try from that part?

core/src/executor/evaluate/evaluated.rs
```diff
@@ -210,22 +341,469 @@ impl<'a> Evaluated<'a> {
...
+            }
+        }
+    }
+
     pub fn try_into_value(self, data_type: &DataType, nullable: bool) -> Result<Value> {
         let value = match self {
-            Evaluated::Value(v) => v,
             Evaluated::Literal(v) => Value::try_from_literal(data_type, &v)?,
+            Evaluated::Value(v) => v,
+            Evaluated::StrSlice {
```
zmrdltl, line 803: I would like to know how to write a test here

zmrdltl (commented):
Failed to write part 2 test.

core/src/data/value/string.rs
```diff
@@ -0,0 +1,133 @@
...
+}
+
+impl PartialOrd<String> for Literal<'_> {
+    fn partial_cmp(&self, other: &String) -> Option<Ordering> {
+        match (self, other) {
+            (&Literal::Text(Cow::Borrowed(l)), r) => Some(l.cmp(r)),
+            _ => None,
+        }
+    }
+}
```
panarch, line 43: This would be better to be located in `data/literal.rs`

core/src/data/value/string.rs
```diff
@@ -0,0 +1,133 @@
...
+
+impl From<&String> for Key {
+    fn from(s: &String) -> Self {
+        Key::Str(s.to_owned())
+    }
+}
```
panarch, line 49: May be.. this would be too much to have `From` impl?
We can make String as Key with a single line of codes.

core/src/executor/evaluate/evaluated.rs
```diff
@@ -210,22 +313,469 @@ impl<'a> Evaluated<'a> {
...
+            source,
+            range: start..end,
+        })
+    }
+
+    pub fn trim(
+        self,
+        name: String,
+        filter_chars: Option<Evaluated<'_>>,
+        trim_where_field: &'a Option<TrimWhereField>,
```
ever0de, line 575:
```suggestion
        trim_where_field: &Option<TrimWhereField>,
```

It has nothing to do with the lifetime of `TrimWhereField`.

core/src/executor/evaluate/evaluated.rs
```diff
@@ -210,22 +313,469 @@ impl<'a> Evaluated<'a> {
...
+            Evaluated::Value(value) => match value {
+                Value::Str(string) => {
+                    let end = string.len();
+                    (string, 0..end)
+                }
+                _ => return exceptional_str_val_to_eval(name, value),
+            },
+        };
+
+        let expr_str = source.as_str();
```
ever0de, line 598: ditto above comments :)

```suggestion
```

core/src/executor/evaluate/evaluated.rs
```diff
@@ -210,22 +313,469 @@ impl<'a> Evaluated<'a> {
...
+                        .map(|(vec_idx, (_, _))| vec_idx)
+                        .next();
+
+                    let start = match pivot {
+                        Some(idx) => matched_vec[idx - 1].0 + 1,
+                        _ => matched_vec[matched_vec.len() - 1].0 + 1,
+                    };
+
+                    return Ok(Evaluated::StrSlice {
+                        source: expr_str.to_owned(),
```
ever0de, line 653:
```suggestion
                        source,
```

core/src/executor/evaluate/evaluated.rs
```diff
@@ -210,22 +313,469 @@ impl<'a> Evaluated<'a> {
...
+                        .map(|(vec_idx, (_, _))| vec_idx)
+                        .next();
+
+                    let end = match pivot {
+                        Some(idx) => matched_vec[matched_vec.len() - idx].0,
+                        _ => matched_vec[0].0,
+                    };
+
+                    return Ok(Evaluated::StrSlice {
+                        source: expr_str.to_owned(),
```
ever0de, line 678:
```suggestion
                        source,
```

core/src/executor/evaluate/evaluated.rs
```diff
@@ -210,22 +313,469 @@ impl<'a> Evaluated<'a> {
...
+                    .enumerate()
+                    .skip_while(|(vec_idx, (slice_idx, _))| vec_idx == slice_idx)
+                    .map(|(vec_idx, (_, _))| vec_idx)
+                    .next()
+                    .unwrap_or(0);
+
+                let trim_range = matched_vec[pivot - 1].0..(matched_vec[pivot].0 + range.start);
+
+                Ok(Evaluated::StrSlice {
+                    source: expr_str.to_owned(),
```
ever0de, line 695:
```suggestion
                    source,
```

core/src/executor/evaluate/evaluated.rs
```diff
@@ -210,22 +313,469 @@ impl<'a> Evaluated<'a> {
...
+                let start = match pivot {
+                    Some(idx) => match idx {
+                        0 => 0,
+                        _ => matched_vec[idx - 1].0 + 1,
+                    },
+                    _ => matched_vec[matched_vec.len() - 1].0 + 1,
+                };
+
+                Ok(Evaluated::StrSlice {
+                    source: expr_str.to_owned(),
```
ever0de, line 716:
```suggestion
                    source,
```

core/src/executor/evaluate/evaluated.rs
zmrdltl: I found the use of the function in insert.rs and filled in the test.
It was possible in the form of 
```sql
INSERT INTO a VALUES (SUBSTR(b)) 
```

core/src/executor/evaluate/evaluated.rs
zmrdltl: Added an example of using SUBSTR in the error.rs part of the schemaless test. 
Although it is for test coverage, I put a temporary test in anticipation of clear use in the future.

panarch (approved):
Awesome!! New big improvement comes into the project.
Thanks a lot 👍 👍 👍

PR #1000, opened 2022-11-12
Change chrono `from_*` methods to `from_*_opt`

ever0de:


panarch (approved):
Looks all nice!
Also looks good to specify the exact version of `chrono`.
Thanks a lot 👍

PR #1001, opened 2022-11-12
Replace `double quotes` to `identifier`

devgony:
# Goal
Currently, Both `double quotes` and `single quotes` evaluated as `literal string`.
From now, `Double quotes` will be evaluated as `identifier`.

### Before
```sql
gluesql> SELECT "name";
```
| "name" |
|--------|
| name   |

### After
```sql
gluesql> select "name";
[error] value not found: name
```
## Todo
- [x] handle double quotes as identifier at translate/expr
- [x] fix all test cases

### DIFF
--- a/core/src/data/interval/mod.rs
+++ b/core/src/data/interval/mod.rs
@@ -295,10 +295,7 @@ impl Interval {
                 format!("{:?}", to),
             )
             .into()),
-            (None, _) => {
-                println!("here");
-                Err(IntervalError::Unreachable.into())
-            }
+            (None, _) => Err(IntervalError::Unreachable.into()),
         }
     }
 }
--- a/core/src/data/interval/mod.rs
+++ b/core/src/data/interval/mod.rs
@@ -295,7 +295,10 @@ impl Interval {
                 format!("{:?}", to),
             )
             .into()),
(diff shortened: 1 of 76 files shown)


core/src/data/interval/mod.rs
```diff
@@ -287,7 +287,10 @@ impl Interval {
                 format!("{:?}", to),
             )
             .into()),
-            (None, _) => Err(IntervalError::Unreachable.into()),
+            (None, _) => {
+                println!("here");
```
ever0de, line 291:
```suggestion
```

core/src/data/interval/string.rs
```diff
@@ -15,6 +15,8 @@ impl TryFrom<&str> for Interval {
 
     fn try_from(s: &str) -> Result<Self> {
         let parsed = parse_interval(s)?;
+        println!("{parsed:?}");
+        // println!("{leading_field:?}:{last_field:?}");
```
ever0de, line 19:
```suggestion
```

ever0de (approved):
🥳 🥳 🥳

panarch (approved):
Awesome...! 👍 👍 👍 
I wasn't aware of that `double quotes` means identifier in SQL.
Looks all great, thanks a lot!

PR #1005, opened 2022-11-17
Group the import statements

### LINKED ISSUE #646: Group the import statements
We have rules for grouping import statements.
However, there are some places that are not observed from time to time.
Fix this.

## How?

There is one fine-grained rule in [rustfmt](https://github.com/rust-lang/rustfmt) that lets you fix this all at once.
<https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#imports_granularity>

Last but not least, it's still unstable (need to use nightly cargo).
Therefore, it has not been applied yet, and it would be good to use this to correct only the wrong parts.

1. `rustfmt.toml`

```toml
imports_granularity = "One"
```

2. `command`

```shell
cargo +nightly fmt
```

## Points to note

We will formally apply once this is stabilized.
Until then, you will want unstable settings to not exist.

yugeeklab:
Group statements that can group. For example, "use A::b" and
"use A::c" can group into "use A::{b, c}".

fixed #646


devgony (approved):

ever0de (commented):
```
core/src/executor/alter/mod.rs
test-suite/src/alter/mod.rs
test-suite/src/transaction/mod.rs
```

panarch (approved):
Welcome! Thanks a lot for the contribution 👍 Looks all good!

PR #1006, opened 2022-11-17
chore: Use rust-cache action to cache dependencies

jongwooo:
Signed-off-by: jongwooo <jongwooo.han@gmail.com>

## Description

Updated GitHub Action workflows to cache dependencies using [Swatinem/rust-cache](https://github.com/Swatinem/rust-cache). Caching dependencies speeds up workflow. For example, **Rustfmt** execution time was reduced from `2m 12s` to `21s`.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -17,6 +17,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
+      - uses: Swatinem/rust-cache@v2
       - run: rustup component add llvm-tools-preview
       - name: Install grcov
         uses: actions-rs/install@v0.1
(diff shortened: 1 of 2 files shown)


ever0de (commented):
Can you also add options like:

```yaml
        with:
          shared-key: cargo-test
          cache-on-failure: true
```

referece: https://github.com/dudykr/stc/blob/d8b849c672f033c1e0a04caf9737f9a6ae7c6863/.github/workflows/CI.yml#L71-L74

jongwooo (commented):
See the following [link](https://github.com/gluesql/gluesql/actions/caches) to check if dependencies are cached.

ever0de:
Could you possibly split them into different keys?

Or does this key not affect other caches?
If so, I'm sorry, can you roll back?

https://github.com/Swatinem/rust-cache
> and is stable over multiple jobs.

from the repository.

ever0de (approved):
Looks good
Our waiting time has been reduced. thanks for the first contribution

panarch (approved):
Thanks a lot for the contribution 👍 
Also thanks to the careful review comments @ever0de

PR #1007, opened 2022-11-18
Support `TableFactor::{Derived, Dictionary, Series}` in AstBuilder

devgony:
### From now, we can use `Dictionary`, `SERIES(N)` and `Derived subquery` even in AstBuilder!
```rs
let actual = glue_objects().select().into();
let expected = "SELECT * FROM GLUE_OBJECTS";
test_query(actual, expected);

let actual = glue_tables().select().into();
let expected = "SELECT * FROM GLUE_TABLES";
test_query(actual, expected);

let actual = glue_indexes().select().into();
let expected = "SELECT * FROM GLUE_INDEXES";
test_query(actual, expected);

let actual = glue_table_columns().select().into();
let expected = "SELECT * FROM GLUE_TABLE_COLUMNS";
test_query(actual, expected);

let actual = series("1 + 2").select().into();
let expected = "SELECT * FROM SERIES(1 + 2)";
test_query(actual, expected);

let actual = table("Items").select().alias_as("Sub").select().into();
let expected = "SELECT * FROM (SELECT * FROM Items) AS Sub";
test_query(actual, expected);
```

## Todo
- [x] implement Dictionary, Series with TableType
- [x] implement Derived with `Box<SelectNode>`

## Next
- impl values for ast_builder
- impl alias_as for ValuesNode


core/src/ast_builder/select/root.rs
```diff
@@ -61,29 +62,76 @@ impl SelectNode {
...
+            (TableType::Series, None) => unreachable!(),
+            (TableType::Derived { subquery, alias }, _) => match subquery.build()? {
+                Statement::Query(subquery) => TableFactor::Derived {
+                    subquery,
+                    alias: TableAlias {
+                        name: alias,
+                        columns: Vec::new(),
+                    },
+                },
+                _ => unreachable!(),
```
devgony, line 133: I got 2 unreachable cases which look like would filtered by parseError in advance.

core/src/ast_builder/table.rs
```diff
@@ -9,20 +13,32 @@ use super::AlterTableNode;
...
+#[derive(Clone)]
+pub struct TableNode<'a> {
     pub table_name: String,
+    pub table_type: TableType<'a>,
 }
 
-impl TableNode {
-    pub fn alias_as(self, table_alias: &str) -> TableAliasNode {
+impl<'a> TableNode<'a> {
+    pub fn alias_as(self, table_alias: &'a str) -> TableAliasNode {
```
ever0de, line 33:
```suggestion
    pub fn alias_as(self, table_alias: &str) -> TableAliasNode<'a> {
```

I think it is correct to declare this lifetime in TableAliasNode, not in str.

core/src/translate/query.rs
```diff
@@ -181,34 +181,27 @@ fn translate_table_factor(sql_table_factor: &SqlTableFactor) -> Result<TableFact
...
-                    name: object_name.to_owned(),
-                    columns: Vec::new(),
-                },
-            };
 
             match object_name.as_str() {
                 "SERIES" if args.is_some() => Ok(TableFactor::Series {
-                    alias: alias_or_name,
+                    alias: alias_or_name(alias, object_name),
                     size: translate_table_args(args)?,
```
ever0de, line 188: It seems that there is unreachable code inside the `translate_table_args` closure, because the value of `None` for args will never go into `translate_table_args`.

core/src/ast_builder/mod.rs
```diff
@@ -77,13 +77,51 @@ pub use expr::{
...
+        table_type: TableType::Dictionary(Dictionary::GlueIndexes),
+    }
+}
+
+pub fn glue_table_columns<'a>() -> TableNode<'a> {
+    TableNode {
+        table_name: "GLUE_TABLE_COLUMNS".to_owned(),
+        table_type: TableType::Dictionary(Dictionary::GlueTableColumns),
+    }
+}
```
panarch, line 117: 1. nit. except for `series` which takes `Into<ExprNode<'a>>`, all has static lifetimes.

e.g.
```rust
pub fn glue_table_columns() -> TableNode<'static> { .. }
```
this would be enough.

---
2. Now table sugar functions are more than a single one, then how about moving these including `series` into `table.rs` module?

core/src/ast_builder/select/root.rs
```diff
@@ -61,29 +63,63 @@ impl SelectNode {
...
+    pub fn left_join_as(self, table_name: &str, alias: &str) -> JoinNode<'a> {
         JoinNode::new(
             self,
             table_name.to_owned(),
             Some(alias.to_owned()),
             JoinOperatorType::Left,
         )
     }
+
+    pub fn alias_as(self, table_alias: &'a str) -> TableAliasNode {
```
panarch, line 79: This is too cool feature! Could we have an usage example in [test-suite/ ast_builder/select.rs](https://github.com/gluesql/gluesql/blob/main/test-suite/src/ast_builder/select.rs)?

panarch (changes requested):

panarch (approved):
Awesome 👍 👍 👍 
Now we can chain select queries as many as we want.
All looks great! Thanks a lot 👍

PR #1010, opened 2022-11-19
Change to use internal chrono errors in parsing datetime

ever0de:
We previously added an error that mimics `chrono::format::ParseErrorKind`.
You may need to react to chrono updates, but there are unreachable branches that can go unnoticed.

But here chrono doesn't provide a public constructor for `ParseError`.
https://github.com/chronotope/chrono/blob/d147e9a2152d1d961372bf5a6a261c644746efdf/src/format/mod.rs#L347-L348
Therefore, it is not compatible with the existing test method.

So we change the way we test

### DIFF
--- a/core/src/executor/evaluate/error.rs
+++ b/core/src/executor/evaluate/error.rs
@@ -1,6 +1,6 @@
 use {
     crate::ast::{Aggregate, Expr},
-    serde::Serialize,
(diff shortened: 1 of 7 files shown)


panarch (approved):
Meaningful simplification improvement! :D Looks all nice! Thanks a lot 👍

PR #1012, opened 2022-11-19
Resolve unreachable branch of `Value::position`

ever0de:
The unit test coverage is satisfied, but the error branch that did not appear in the user interface `Glue::execute` is corrected.

If it were not for this fix, the position function would probably be correct to receive a String rather than a `Value`.

### DIFF
--- a/test-suite/src/function/position.rs
+++ b/test-suite/src/function/position.rs
@@ -1,8 +1,11 @@
 use {
     crate::*,
     gluesql_core::{
-        executor::EvaluateError,
-        prelude::{Payload, Value::*},
+        data::ValueError,
+        prelude::{
+            Payload,
+            Value::{self, *},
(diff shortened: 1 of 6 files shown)


test-suite/src/function/position.rs
```diff
@@ -25,7 +28,11 @@ test_case!(position, async move {
         ),
         (
             "SELECT POSITION(1 IN 'cheese') AS test",
-            Err(EvaluateError::FunctionRequiresStringValue(String::from("POSITION")).into()),
+            Err(ValueError::UnSupportedValueByPositionFunction {
```
panarch, line 31: nit. `Unsupported` would be good.

panarch (approved):
This is a nice improvement! Thanks a lot @ever0de

PR #1013, opened 2022-11-19
Add value/ binary_op `Parital{Ord,Cmp}` impl macro

ever0de:


panarch (approved):
Beautiful! Thanks a lot 👍

PR #1014, opened 2022-11-19
Move import `ColumnOption` used only by `alter-table` feature in ast-builder

ever0de:


panarch (approved):
Oh, that's a good catch.
I'll also update rust.yml clippy actions to check these.

PR #1016, opened 2022-11-22
Add null() func which makes NULL value in AST builder

panarch:
cc. @ShaddyDC


ever0de (approved):
👍  Thanks to @ShaddyDC

devgony (approved):

PR #1017, opened 2022-11-23
Add NumericNode to handle numeric value inputs in AST builder

panarch:
Now, `num` AST builder sugar function supports multiple different types including i64, f64, and &str.

e.g.
```rust
num("1.23");
num(100_i64);
num(300_i32);
num(1.234);
```

cc. @ShaddyDC

### DIFF
--- a/core/src/ast_builder/expr/numeric.rs
+++ b/core/src/ast_builder/expr/numeric.rs
@@ -1,0 +1,127 @@
+use {
+    crate::{
+        ast::AstLiteral,
+        ast_builder::AstBuilderError,
+        result::{Error, Result},
+    },
+    bigdecimal::BigDecimal,
+    std::{borrow::Cow, str::FromStr},
+};
+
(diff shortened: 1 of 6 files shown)


core/src/ast_builder/expr/numeric.rs
```diff
@@ -0,0 +1,127 @@
...
+    bigdecimal::BigDecimal,
+    std::{borrow::Cow, str::FromStr},
+};
+
+#[derive(Clone)]
+pub enum NumericNode<'a> {
+    I8(i8),
+    I16(i16),
+    I32(i32),
+    I64(i64),
```
ever0de, line 16: Is there a reason the 128 size is missing?

panarch: For now, it is just because `bigdecimal` crate does not support from 128 size to bigdecimal

ever0de (approved):

PR #1023, opened 2022-11-26
Remove LimitOffsetNode in AST builder

### LINKED ISSUE #954: Remove `limit(n).offset(n)` support in AST builder
### To do
We support both `limit(n).offset(m)` and `offset(m).limit(n)` in AST builder.
This issue is for dropping `limit(n).offset(m)` support and make the AST builder to only have support of `offset(m).limit(n)`.

### Details
In GlueSQL, we accept both
```sql
SELECT * FROM SERIES(100) LIMIT 5 OFFSET 3;

SELECT * FROM SERIES(100) OFFSET 3 LIMIT 5;
```
and the result is same
| N |
|---|
| 4 |
| 5 |
| 6 |
| 7 |
| 8 |

This can be acceptable when we use "SQL" but in AST builder, this can be a different story.
Our AST builder provides iterator-chaining experience to users.
This means that AST builder codes should show the same order with what it is actually executed.
However, when we use iterator, limit -> offset and offset -> limit codes are obviously different.
e.g.
```rust
let a = (0..10).skip(3).take(5);
// [3, 4, 5, 6, 7]

let b = (0..10).take(5).skip(3);
// [3, 4]
```

from this example above, what we only support is `OFFSET -> LIMIT` or `skip -> take`, then it would be clear for AST builder to support only `offset(n).limit(m)`.

panarch:
resolve #954


ever0de (approved):
🥳 🥳 🥳

PR #1026, opened 2022-11-27
Update `data::Row` to contain columns

panarch:
Prior steps - Make `data::Row` struct only work in execution layer, don't expose to user via neither `Payload` or Store traits.
* [x] Change `Payload::Select` rows to use `Vec<Value>`
* [x] Replace `data::Row` to `Vec<Value>` in Store traits

Main
* [x] Put `columns` into Row struct
* [x] Remove `columns` from `FilterContext`, `BlendContext`
* [x] Remove redandunt columns field uses


core/src/data/row.rs
```diff
@@ -99,16 +107,23 @@ impl Row {
 
                 match (value, column_def.get_default(), nullable) {
                     (Some(&expr), _, _) | (None, Some(expr), _) => {
-                        evaluate_stateless(None, expr)?.try_into_value(data_type, *nullable)
+                        let value =
+                            evaluate_stateless(None, expr)?.try_into_value(data_type, *nullable)?;
+
+                        Ok(value)
```
ever0de, line 113: Hmm.. I think we can leave it as it is.

core/src/executor/fetch.rs
```diff
@@ -365,3 +419,57 @@ pub async fn fetch_join_columns<'a>(
...
+            let join_columns = Rc::clone(&join_columns);
+
+            match item {
+                SelectItem::Wildcard => {
+                    let columns = columns.iter().cloned();
+                    let join_columns = join_columns
+                        .iter()
+                        .flat_map(|(_, columns)| columns.clone())
+                        .collect::<Vec<_>>()
+                        .into_iter();
```
ever0de, line 448:
```suggestion
                    let join_columns = join_columns
                        .iter()
                        .flat_map(|(_, columns)| columns.clone());
```

core/src/executor/select/mod.rs
```diff
@@ -155,7 +78,7 @@ fn rows_with_labels(exprs_list: &[Vec<Expr>]) -> (Vec<Result<Row>>, Vec<String>)
 
 fn sort_stateless(
     rows: Vec<Result<Row>>,
-    labels: &Vec<String>,
+    labels: &[String],
```
ever0de, line 81: Good

ever0de (commented):
There are a few minor comments.
I'll see a bit more later(again today) to understand.

core/src/data/row.rs
panarch: Oh it looks that this is unnecessary change I've made while.. may be during doing some tests. original one looks enough I'll change it to the prev

core/src/executor/fetch.rs
```diff
@@ -365,3 +419,53 @@ pub async fn fetch_join_columns<'a>(
         .try_collect::<Vec<_>>()
         .await
 }
+
+pub async fn fetch_labels<'a>(
+    storage: &dyn GStore,
+    relation: &'a TableFactor,
+    joins: &'a [Join],
+    projection: &'a [SelectItem],
+) -> Result<Vec<String>> {
```
ever0de, line 428:
```suggestion
pub async fn fetch_labels (
    storage: &dyn GStore,
    relation: &TableFactor,
    joins: &[Join],
    projection: &[SelectItem],
) -> Result<Vec<String>> {
```

Lifetime sharing here doesn't seem necessary.

ever0de (approved):
We are laying a great foundation

🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀

PR #1029, opened 2022-11-29
Merge FilterContext and BlendContext into RowContext

panarch:
resolve #767 

prior issue which needs to be merged before - #1026

PR #1031, opened 2022-12-01
Merge insert logics in row.rs & execute.rs into executor/insert.rs

panarch:
…module

prior pr https://github.com/gluesql/gluesql/pull/1029

Insert execution code logics are splitted in data/row.rs and executor/execute.rs
By adding executor/insert.rs, it merges the codes into a single insert module.


ever0de (approved):
LGTM

PR #1032, opened 2022-12-01
Remove `Content::Shared` variant in executor/ `RowContext`

ever0de:


panarch (approved):
Beautiful..
I wasn't aware of setting lifetime bounds feature.

Thanks a lot!!

PR #1040, opened 2022-12-09
Remove data::RowError,

panarch:
- Move `RowError::NumberOfValuesDifferent` to `SelectError`.
- Remove `RowError::ConflictOnEmptyRow` by making evaluate module to handle this.
This `ConflictOnEmptyRow` error is now just hidden by evaluate and that might be not an ideal solution.
We need `NonEmptyVec` struct for this. (`Vec<SelectItem>` to `NonEmptyVec<SelectItem>` for projection)


ever0de (approved):
Looks good 👍

PR #1042, opened 2022-12-09
Move ast::ColumnOption::Default variant to ColumnDef

panarch:
Each `ColumnDef` can have at most one `ColumnOption::Default` but current ast structure allows `ColumnDef` to have multiple `Default`s.
Resolve this by adding `default` field to `ColumnDef` and remove `ColumnOption::Default`.

---
It can also be applied to `ColumnOption::Unique` which is the last enum variant of `ColumnOption`.


ever0de (commented):
Looks good

ever0de (approved):

PR #1046, opened 2022-12-12
Schemaless data support

panarch:
wip
Convert Row to enum with Vec and Map variants.
Make column_defs in Schema optional.
..

### DIFF
--- a/core/src/executor/validate.rs
+++ b/core/src/executor/validate.rs
@@ -17,6 +17,9 @@ pub enum ValidateError {
     #[error("conflict! storage row has no column on index {0}")]
     ConflictOnStorageColumnIndex(usize),
 
+    #[error("conflict! schemaless row found in schema based data")]
+    ConflictOnUnexpectedSchemalessRowFound,
+
     #[error("duplicate entry '{}' for unique column '{1}'", String::from(.0))]
     DuplicateEntryOnUniqueField(Value, String),
 
@@ -142,7 +145,9 @@ pub async fn validate_unique(
                 let (_, data_row) = result?;
                 let values = match data_row {
                     DataRow::Vec(values) => values,
-                    DataRow::Map(_) => todo!(),
+                    DataRow::Map(_) => {
+                        return Err(ValidateError::ConflictOnUnexpectedSchemalessRowFound.into());
+                    }
                 };
 
                 Rc::clone(&unique_constraints)
--- a/core/src/executor/validate.rs
+++ b/core/src/executor/validate.rs
@@ -3,7 +3,7 @@ use {
         ast::{ColumnDef, ColumnUniqueOption},
         data::{Key, Value},
         result::Result,
-        store::{Store, DataRow},
+        store::{DataRow, Store},
     },
     im_rc::HashSet,
     serde::Serialize,
--- a/core/src/executor/validate.rs
+++ b/core/src/executor/validate.rs
@@ -3,7 +3,7 @@ use {
         ast::{ColumnDef, ColumnUniqueOption},
         data::{Key, Value},
         result::Result,
-        store::Store,
+        store::{Store, DataRow},
     },
     im_rc::HashSet,
     serde::Serialize,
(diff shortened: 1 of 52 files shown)


core/src/executor/validate.rs
```diff
@@ -139,12 +142,19 @@ pub async fn validate_unique(
 
             let unique_constraints = Rc::new(unique_constraints);
             storage.scan_data(table_name).await?.try_for_each(|result| {
-                let (_, row) = result?;
+                let (_, data_row) = result?;
+                let values = match data_row {
```
ever0de, line 146: Looks like where you should use `into_vec`.

test-suite/src/schemaless/basic.rs
```diff
@@ -0,0 +1,115 @@
...
+            "Hello".to_owned()   3000
+        ))
+    );
+
+    // join
+    test!(
+        "SELECT
+            Player.id AS player_id,
+            Player.name AS player_name,
+            Item.obj['cost'] AS item_cost
```
devgony, line 104: Looks like we cannot unnest nested Map if identical key has both Map and other type
```sql
gluesql> CREATE TABLE Items;
Table created

gluesql> INSERT INTO Items VALUES ('{"a": {"x": 1}}');
1 row inserted

gluesql> INSERT INTO Items VALUES ('{"a": 1}');
1 row inserted

gluesql> SELECT * FROM Items;
| a     |
|-------|
| [MAP] |
| 1     |

gluesql> SELECT a['x'] FROM Items limit 1;
| a['x'] |
|--------|
| 1      |

gluesql> SELECT a["x"] FROM Items;
[error] selector requires MAP or LIST types
```
We may need some function like `get_type(columnName)`?
```sql
SELECT 
CASE 
  WHEN get_type(a) == 'MAP' THEN a['x']
  ELSE a
END
FROM Items;
```

panarch: That sounds nice to have a new function that can retrieve the type of value.
Then, how about `TYPEOF`..?
```sql
SELECT 
    CASE 
        WHEN TYPEOF(a) = 'MAP' THEN a['x']
        ELSE a
    END
FROM Items;
```

devgony: I like that!
TYPEOF sounds better 👍

test-suite/src/schemaless/basic.rs
```diff
@@ -0,0 +1,115 @@
...
+            }
+        }),
+        json!({
+            "id": 200
+        })
+    )
+    .as_str());
+
+    test!(
+        "SELECT name, dex, rare FROM Item WHERE id = 100",
```
devgony, line 35: Is there a way to filter with Map like this?
```sql
gluesql> SELECT * FROM Item WHERE obj = '{"cost": 3000}';
-- currently 0 rows
```

panarch: PartialEq between `Value::Map(_)`s is implemented so it can be supported. we can consider using `CAST` for now.
```sql
SELECT * FROM Item WHERE obj = CAST('{ "cost": 3000 }' AS MAP);
```
Not supported yet, we need to add `MAP` cast to `Value::try_cast_from_literal`.

test-suite/src/tester/macros.rs
```diff
@@ -127,3 +127,17 @@ macro_rules! concat_with_null {
...
 }
+
+#[macro_export]
+macro_rules! select_map {
+    ( $( $row: expr ),* ) => (
+        gluesql_core::executor::Payload::SelectMap(vec![
+            $(
+                match gluesql_core::data::Value::try_from($row).unwrap() {
+                    Value::Map(v) => v,
+                    _ => panic!(),
```
ever0de, line 138: When an error occurs due to a user's mistake, if panic occurs, an error is likely to occur at runtime.
How about adding a specific literal string?

ever0de (approved):
It seems good overall. It was an interesting piece of work! thank you!

Issue #1024, opened 2022-11-26
[AST] Add `Append` Function to LIST DataType


seonghun-dev:
### Description
Add `Append` Function to LIST DataType
This Append Function will append an element to the end of an array

Same Function
PostgreSQL : `ARRAY_APPEND()`

## Task
In AST module add `APPEND` Function to `LIST`  data_type

The following tests should be passed:

``` rust
test_case!(append, async move {
    run!(
        "
        CREATE TABLE Append (
            id INTEGER,
            items LIST,
            element INTEGER
        );
    "
    );
    run!(
         r#"
            INSERT INTO Append VALUES
            (1, "[1, 2, 3]", 4);
        "#
    );
    test!(
        r#"select append(items, element) as myappend from Append;"#,
        Ok(select!(
           myappend
           List;
           vec![I64(1), I64(2), I64(3), I64(4)]
        ))
    );
});

```
Reference
https://www.postgresql.org/docs/9.1/functions-array.html
https://cloud.google.com/bigquery/docs/reference/standard-sql/arrays

PR #1047, opened 2022-12-17
[AST] Add Append Function to LIST DataType

seonghun-dev:
Resolve: https://github.com/gluesql/gluesql/issues/1024

### DIFF
--- a/core/src/executor/evaluate/stateless.rs
+++ b/core/src/executor/evaluate/stateless.rs
@@ -345,5 +345,10 @@ fn evaluate_function<'a>(
 
             f::concat_ws(name, separator, exprs)
         }
+        Function::Append { expr, value } => {
+            let expr = eval(expr)?;
+            let value = eval(value)?;
+            f::append(expr, value)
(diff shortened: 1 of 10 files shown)


core/src/executor/evaluate/stateless.rs
```diff
@@ -390,5 +390,10 @@ fn evaluate_function<'a>(context: &Context<'_>, func: &'a Function) -> Result<Ev
 
             f::concat_ws(name, separator, exprs)
         }
+        Function::Append { expr, value } => {
+            let expr = eval(expr)?;
+            let value = eval(value)?;
+            f::append(expr, value)
+        }
```
panarch, line 397: to cover this, for now.. the most simple way would be this below,
```sql
CREATE TABLE Foo (
    id INTEGER,
    values LIST DEFAULT '[1,2,3]'
);
```
like this above, using `DEFAULT` can reach this lines.
Could you add this kind of test case to `test-suite/ function/append.rs`?

Issue #989, opened 2022-11-05
[AST] Implement Geometic Point Type and Geometric fucntion in AST


seonghun-dev:
# Description
We want to support the following functions that can be used to analyze geographic data, identify spatial relationships between terrain features, and construct or manipulate GEOGRAPHY.

> ### Spaital Standard Document
> spatial data types and functions define standards in the Open Geospatial Consortium
> All data types and functions are based on the this documents
> http://portal.opengeospatial.org/files/?artifact_id=25354

## Point

`POINT`
The position of a point in the coordinate space
Points are the fundamental two-dimensional building block for geometric types

### Example
DataType `POINT`

``` sql
POINT(10 20)
```

Function `ST_POINT`

Return the Point

``` sql
SELECT ST_POINT(POINT(15 20));
```

Function `ST_X`

Return the X-coordinate of point

``` sql
SELECT ST_X(POINT(15 20));
```

Function `ST_Y`

Return the Y-coordinate of point
``` sql
SELECT ST_Y(POINT(15 20));
```

# Task
- [ ] Add `POINT` DataType
- [ ] Add `ST_POINT` Fucntion
- [ ] Add `ST_X` Fucntion
- [ ] Add `ST_Y` Fucntion

The following tests should be passed:

``` rust
test_case!(point, async move {

    let test_cases = [
        ("CREATE TABLE POINT (point_field POINT)", Ok(Payload::Create)),
        (
            r#"INSERT INTO POINT VALUES(15 20);"#,
            Ok(Payload::Insert(2))
        ),
        (
             r#"SELECT point_field AS point_field FROM POINT;"#,
            Ok(select!(
                point_field
                Point;
                Point(15 20);
            ))
        ),
	(
		r#"SELECT ST_POINT(POINT(15 20));"#,
           Ok(select!(
                column1
                Point;
                Point(15 20);
		))
	),
	(
	     r#"SELECT ST_X(POINT(15 20))"#,
            Ok(select!(
                column1
                F64;
                f64::from(15);
            ))
	),
	(
	     r#"SELECT ST_Y(POINT(15 20))"#,
            Ok(select!(
                column1
                F64;
                f64::from(20);
            ))
	)
    ];

    for (sql, expected) in test_cases {
        test!(expected, sql);
    }
});
```

### Reference
http://portal.opengeospatial.org/files/?artifact_id=25354
http://postgis.net/workshops/postgis-intro/geometries.html
[...]

PR #1048, opened 2022-12-17
Implement geometric `POINT` Type and geometric functions

seonghun-dev:
Resolve: https://github.com/gluesql/gluesql/issues/989


panarch:
that failure is because of that we change the default ToSql behavior to wrap identifiers with `"`.
ref. https://github.com/gluesql/gluesql/pull/1130

image
You can change the `point` to `"point"` then the test will pass.

core/src/executor/evaluate/function.rs
```diff
@@ -596,3 +596,24 @@ pub fn cast<'a>(expr: Evaluated<'a>, data_type: &DataType) -> Result<Evaluated<'
...
 }
+
+pub fn point<'a>(x: Evaluated<'_>, y: Evaluated<'_>) -> Result<Evaluated<'a>> {
+    let x = eval_to_float!("point".to_owned(), x);
+    let y = eval_to_float!("point".to_owned(), y);
+
+    Ok(Evaluated::from(Value::Point(crate::data::Point::new(x, y))))
+}
+
+pub fn getx<'a>(name: String, expr: Evaluated<'_>) -> Result<Evaluated<'a>> {
```
panarch, line 607: it would be just good to use `get_x` because Function enum variant name is `GetX`

test-suite/src/data_type/point.rs
```diff
@@ -0,0 +1,87 @@
+use gluesql_core::translate::TranslateError;
+
+use {
```
panarch, line 3: could you merge this?

core/src/data/value/json.rs
```diff
@@ -119,6 +120,7 @@ impl TryFrom<JsonValue> for Value {
                 .map(Value::try_from)
                 .collect::<Result<Vec<Value>>>()
                 .map(Value::List),
+
```
panarch, line 123: this blank line looks unnecessary

core/src/ast/function.rs
```diff
@@ -1027,13 +1058,15 @@ mod tests {
...
         );
+
         assert_eq!(
             r#"VARIANCE("pay")"#,
             &Expr::Aggregate(Box::new(Aggregate::Variance(Expr::Identifier(
                 "pay".to_owned()
             ))))
             .to_sql()
         );
+
```
panarch, line 1069: These changes also look not related to this PR

core/src/ast/function.rs
```diff
@@ -154,6 +154,9 @@ pub enum Function {
         expr: Expr,
         value: Expr,
     },
+    GetX(Expr),
+    GetY(Expr),
+    Point(Expr, Expr),
```
panarch, line 159: It would be better to specify the fields explicitely. Then,
```rust
Point {
    x: Expr,
    y: Expr,
}
```

panarch (approved):
Looks all great!!
So good to see the new category data type 👍 👍

PR #1050, opened 2022-12-26
Add WebStorage - support localStorage & sessionStorage for web browsers

panarch:
WebStorage includes two storages
- localStorage
- sessionStorage

This PR only verifies the storage implementation.
It does not have gluesql javascript interface integration that will be done separately.


ever0de (approved):
It looks good except for improvement comments
The interface looks pretty simple 👍

PR #1053, opened 2023-01-02
Support `Jsonl` Storage

devgony:
## Goal
Support Jsonl Storage

## Todo
- [x] implement store schemaless
- [x] storeMut
- [x] implement schema by parsing `{tableName}.sql`
- [x] split test cases by namespace(directory)
- [x] exception cases (path, other filetypes..)

## Next
- CLI support
- alter-table
- file lock for single-thread mutation
- should sync manually editing `.sql` file with memory?


core/src/data/value/convert.rs
```diff
@@ -17,7 +17,7 @@ impl From<&Value> for String {
         match v {
             Value::Str(value) => value.to_owned(),
             Value::Bytea(value) => hex::encode(value),
-            Value::Bool(value) => (if *value { "TRUE" } else { "FALSE" }).to_owned(),
+            Value::Bool(value) => (if *value { "true" } else { "false" }).to_owned(),
```
panarch, line 20: It looks that you are using `Value` -> `String` for serialize data, but it would be better to use json module.
ref. https://github.com/gluesql/gluesql/blob/main/core/src/data/value/json.rs

storages/jsonl-storage/src/lib.rs
```diff
@@ -0,0 +1,396 @@
...
+                                .collect::<Result<Vec<_>>>()?
+                        }
+                        DataRow::Vec(values) => {
+                            match &schema.column_defs {
+                                Some(column_defs) => {
+                                    // todo! validate columns?
+                                    column_defs
+                                        .iter()
+                                        .map(|column_def| column_def.name.clone())
+                                        .zip(values.into_iter())
```
devgony, line 307: Is there a way to merge two different Iterators? like..
```rs
// 🦀 This code does not compile!
let json_iter: Box<dyn Iterator> = match row {
  DataRow::Map(hash_map) => hash_map.into_iter(),
  DataRow::Vec(values) => match &schema.column_defs {
    Some(column_defs) => {
        column_defs
            .iter()
            .map(|column_def| column_def.name.clone())
            .zip(values.into_iter()),
    None => todo!(),
  }
}
```
I want to merge those iterators to save below duplicated `map + collect` logic
```rs
.map(|(key, value)| {
  let value = JsonValue::try_from(value)?.to_string();

  Ok(format!("\"{key}\": {value}"))
})
.collect::<Result<Vec<_>>>()?
```

panarch: [iter_enum](https://docs.rs/iter-enum/latest/iter_enum/)

This would be useful.
You can also see some examples in core/ workspace

devgony: It works like a charm! I didn't know that was what i had done with `enum Rows` long times ago..

storages/jsonl-storage/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+        let labels = schema
+            .column_defs
+            .unwrap_or_default()
+            .into_iter()
+            .map(|column_def| column_def.name);
+
+        for row in rows {
+            let json_string = match row {
+                DataRow::Map(hash_map) => JsonIter::Map(hash_map.into_iter()),
+                DataRow::Vec(values) => JsonIter::Vec(labels.clone().zip(values.into_iter())),
```
devgony, line 308: I tried with `&label.zip()` here but looks like `iter_enum` does not work with reference

panarch: Rather than using `iter_enum`, here because we don't need to return value so how about simply using match to serialize data as jsonvalue independently?
We can deal with `DataRow::Map` and `DataRow::Vec` separately.

storages/jsonl-storage/src/lib.rs
```diff
@@ -0,0 +1,420 @@
...
+        let labels = schema.column_defs.unwrap_or_default();
+        let labels = labels
+            .iter()
+            .map(|column_def| column_def.name.as_str())
+            .collect::<Vec<_>>();
+
+        for row in rows {
+            let json_string = match row {
+                DataRow::Vec(values) => labels
+                    .iter()
```
devgony, line 302: Now it works without `clone`.
But not sure whether this `collect` and `iter` again way is better or not.

storages/jsonl-storage/src/lib.rs
```diff
@@ -296,26 +289,33 @@ impl StoreMut for JsonlStorage {
...
+                    .collect::<Result<Vec<_>>>()?
+                    .join(", "),
+                DataRow::Map(hash_map) => hash_map
+                    .into_iter()
+                    .map(|(key, value)| {
+                        let value = JsonValue::try_from(value)?.to_string();
+                        Ok(format!("\"{}\": {}", key, value))
+                    })
+                    .collect::<Result<Vec<_>>>()?
+                    .join(", "),
```
panarch, line 317: Oh, I just recognized this.
It looks that it is a kind of custom serialization from HashMap to JsonString.

```rust
let value = JsonValue::try_from(value)?.to_string();
Ok(format!("\"{}\": {}", key, value))
```
Rather than the above,
how about making JsonValue::Object and serialize as string rather than using `format` macro?

storages/jsonl-storage/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+
+                let parsed = parse(ddl)?.into_iter().next().map_storage_err(
+                    JsonlStorageError::WrongSchemaFile(
+                        schema_path
+                            .to_str()
+                            .map_storage_err(JsonlStorageError::FileNotFound.to_string())?
+                            .to_string(),
+                    )
+                    .to_string(),
+                )?;
```
devgony, line 101: https://coveralls.io/builds/56635692/source?filename=storages%2Fjsonl-storage%2Fsrc%2Flib.rs#L101
I don't get why it was not covered. Kinda bug or actually unreachable..?

storages/jsonl-storage/src/lib.rs
```diff
@@ -0,0 +1,449 @@
...
+    P: AsRef<Path>,
+{
+    let file = File::open(filename)?;
+    Ok(io::BufReader::new(file).lines())
+}
+
+type IsLeft = bool;
+struct SortMerge<
+    T: Iterator<Item = Result<(Key, DataRow)>>,
+    U: Iterator<Item = Result<(Key, DataRow)>>,
```
devgony, line 224: Is there is way to merge T and U with `single type definition` without `Box`?

storages/jsonl-storage/src/lib.rs
```diff
@@ -0,0 +1,445 @@
...
+                .partial_cmp(key_b)
+                .unwrap_or(std::cmp::Ordering::Equal)
+        });
+        let rows = rows.into_iter();
+
+        let sort_merge = SortMerge {
+            prev_rows,
+            rows,
+            current: None,
+        };
```
panarch, line 363: How about adding `SortMerge::new`?
```rust
impl<T> SortMerge<T> where T: Iterator<Item = Result<(Key, DataRow)>> {
    fn new(prev_rows: T, rows: IntoIter<(Key, DataRow)>) -> Self {
        Self {
            prev_rows,
            rows,
            current: None,
        }
    }
}
```
Beause here, we have no use case of creating `SortMerge` using `current: Some(_)` so it would be better to hide `current` instance rather than exposing it.

panarch: Then, this below would be enough.
```rust
let sort_merge = SortMerge::new(prev_rows, rows);
```

storages/jsonl-storage/src/lib.rs
```diff
@@ -0,0 +1,464 @@
...
+            (None, Some((_, row))) => {
+                self.current = None;
+
+                Some(row)
+            }
+            (Some((_, prev_row)), None) => {
+                self.current = None;
+
+                Some(prev_row)
+            }
```
panarch, line 276: Instead of these,
```rust
(left, right) => {
    self.current = None;

    left.or(right).map(|(_, row)| row)
}
```
You can put this after the `Some(_), Some(_)`.

storages/jsonl-storage/src/lib.rs
```diff
@@ -0,0 +1,464 @@
...
+    fn new(prev_rows: T, rows: IntoIter<(Key, DataRow)>) -> Self {
+        Self {
+            prev_rows,
+            rows,
+            current: None,
+        }
+    }
+}
+
+impl<T: Iterator<Item = Result<(Key, DataRow)>>> Iterator for SortMerge<T> {
```
panarch, line 246: nit.
how about this as you already used in the above?
```rust
impl<T> Iterator for SortMerge<T>
where
    T: Iterator<Item = Result<(Key, DataRow)>>,
```

storages/jsonl-storage/src/lib.rs
```diff
@@ -0,0 +1,464 @@
...
+    P: AsRef<Path>,
+{
+    let file = File::open(filename)?;
+    Ok(io::BufReader::new(file).lines())
+}
+
+type IsLeft = bool;
+struct SortMerge<T: Iterator<Item = Result<(Key, DataRow)>>> {
+    prev_rows: T,
+    rows: IntoIter<(Key, DataRow)>,
```
panarch, line 229: As we already use left | right keywords like `IsLeft`, how about having names `left_rows` & `right_rows` rather than `prev_rows` & `rows`?

storages/jsonl-storage/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+            (left, right) => {
+                self.current_row = None;
+
+                left.or(right).map(|(_, row)| row)
+            }
+        };
+
+        data_row.map(Ok)
+    }
+}
```
panarch, line 297:
```rust
struct SortMerge<T: Iterator<Item = Result<(Key, DataRow)>>> {
    left_rows: Peekable<T>,
    right_rows: Peekable<IntoIter<(Key, DataRow)>>,
}

impl<T> SortMerge<T>
where
    T: Iterator<Item = Result<(Key, DataRow)>>,
{
    fn new(left_rows: T, right_rows: IntoIter<(Key, DataRow)>) -> Self {
        let left_rows = left_rows.peekable();
        let right_rows = right_rows.peekable();

        Self {
            left_rows,
            right_rows,
        }
    }
}

impl<T> Iterator for SortMerge<T>
where
    T: Iterator<Item = Result<(Key, DataRow)>>,
{
    type Item = Result<DataRow>;

    fn next(&mut self) -> Option<Self::Item> {
        let left = self.left_rows.peek();
        let right = self.right_rows.peek();

        let (left_key, right_key) = match (left, right) {
            (Some(Ok((left_key, _))), Some((right_key, _))) => (left_key, right_key),
            (Some(_), _) => {
                return self.left_rows.next().map(|item| Ok(item?.1));
            }
            (None, Some(_)) => {
                return self.right_rows.next().map(|item| item.1).map(Ok);
            }
            (None, None) => {
                return None;
            }
        };

        match left_key.to_cmp_be_bytes().cmp(&right_key.to_cmp_be_bytes()) {
            Ordering::Less => self.left_rows.next(),
            Ordering::Greater => self.right_rows.next().map(Ok),
            Ordering::Equal => {
                self.left_rows.next();
                self.right_rows.next().map(Ok)
            }
        }
        .map(|item| Ok(item?.1))
    }
}
```

storages/jsonl-storage/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+
+fn read_lines<P>(filename: P) -> io::Result<io::Lines<io::BufReader<File>>>
+where
+    P: AsRef<Path>,
+{
+    let file = File::open(filename)?;
+    Ok(io::BufReader::new(file).lines())
+}
+
+#[test]
```
ever0de, line 131: It's a good idea to put your test code in a module separated by a cfg, so that it only builds in your test environment.

```suggestion
#[cfg(test)]
mod tests {
    #[test]
}
```

storages/jsonl-storage/src/lib.rs
```diff
@@ -0,0 +1,235 @@
...
+        .into_iter()
+        .map(ToOwned::to_owned)
+        .collect(),
+        rows: vec![vec![
+            Value::Bool(true),
+            Value::I8(93),
+            Value::I16(18550),
+            Value::I32(1726758346),
+            Value::I64(5724242792803127000),
+            Value::I128(26574198945213317000),
```
devgony, line 210: left: `Select { labels: ["boolean", "int8", "int16", "int32", "int64", "int128", "uint8", "text", "bytea", "inet", "date", "timestamp", "time", "interval", "uuid", "map", "list"], rows: [[Bool(true), I8(93), I16(18550), I32(1726758346), I64(5724242792803127000), I128(26574198945213317120), U8(159), Str("Blake Funk"), Bytea([245, 221, 66, 193, 35]), Inet(14.232.199.19), Date(2022-08-14), Timestamp(2023-02-10T00:03:21), Time(10:26:32), Interval(Microsecond(126000000000)), Uuid(112568127954684182204810458849314395617), Map({"age": I64(55), "city": Str("East Nicofurt")}), List([Str("fuchsia"), Str("olive"), Str("tan")])]] }`,
 right: `Select { labels: ["boolean", "int8", "int16", "int32", "int64", "int128", "uint8", "text", "bytea", "inet", "date", "timestamp", "time", "interval", "uuid", "map", "list"], rows: [[Bool(true), I8(93), I16(18550), I32(1726758346), I64(5724242792803127000), I128(26574198945213317000), U8(159), Str("Blake Funk"), Bytea([245, 221, 66, 193, 35]), Inet(14.232.199.19), Date(2022-08-14), Timestamp(2023-02-10T00:03:21), Time(10:26:32), Interval(Microsecond(126000000000)), Uuid(112568127954684182204810458849314395617), Map({"city": Str("East Nicofurt"), "age": I64(55)}), List([Str("fuchsia"), Str("olive"), Str("tan")])]] }`

I128(26574198945213317120) <> I128(26574198945213317000)

Looks like i128 can show different value like floating point number?

devgony: value: F64(
    2.6574198945213317e19,
)
target_column_type: Int128

devgony: Looks like.. `serde_json::from_str()` consider i128 as f64.
Since serde_json::Value::Number can handle only i64.

panarch: I think we can omit the large number case.
but in here, let's simply use small numbers.

ref. https://github.com/gluesql/gluesql/blob/3c07222834d422dddfb43cd00d5de894e9cd629f/test-suite/src/data_type/int64.rs#L120-L125

storages/jsonl-storage/src/store.rs
```diff
@@ -0,0 +1,66 @@
...
+                    .unwrap_or(false)
+            })
+            .map(|result| -> Result<_> {
+                let path = result.map_storage_err()?.path();
+                let table_name = path
+                    .file_stem()
+                    .map_storage_err(JsonlStorageError::FileNotFound.to_string())?
+                    .to_str()
+                    .map_storage_err(JsonlStorageError::FileNotFound.to_string())?
+                    .to_owned();
```
panarch, line 41: It can be simplified like this below
```rust
let table_name = path
    .file_stem()
    .and_then(OsStr::to_str)
    .map_storage_err(JsonlStorageError::FileNotFound)?
    .to_owned();
```

storages/jsonl-storage/src/error.rs
```diff
@@ -0,0 +1,42 @@
...
+        self.ok_or_else(|| error.to_string())
+            .map_err(Error::StorageMsg)
+    }
+}
+
+pub enum JsonlStorageError {
+    FileNotFound,
+    TableDoesNotExist,
+    ColumnDoesNotExist(String),
+}
```
panarch, line 28: How about using `thiserror` crate rather than iml fmt::Display manually?
ref. https://docs.rs/thiserror/latest/thiserror/

e.g.
https://github.com/gluesql/gluesql/blob/3c07222834d422dddfb43cd00d5de894e9cd629f/core/src/executor/evaluate/error.rs#L8-L18

storages/jsonl-storage/src/store_mut.rs
```diff
@@ -0,0 +1,176 @@
...
+            .filter_map(|result| {
+                result
+                    .map(|(key, data_row)| {
+                        let preservable = !keys.iter().any(|target_key| target_key == &key);
+
+                        preservable.then_some(data_row)
+                    })
+                    .unwrap_or(None)
+            })
+            .collect::<Vec<_>>();
```
panarch, line 115:
```rust
             preservable.then_some(data_row)
         })
         .transpose()
 })
```
It would be better to handle error explicitly.

storages/jsonl-storage/src/store.rs
```diff
@@ -0,0 +1,65 @@
+use {
+    crate::error::{JsonlStorageError, OptionExt, ResultExt},
+    crate::JsonlStorage,
```
panarch, line 3: nit. we can merge this in to a single `crate::{..}`.
this can be also applied to `store_mut.rs`

storages/jsonl-storage/src/store_mut.rs
```diff
@@ -0,0 +1,178 @@
...
+                    let mut json_map = Map::new();
+                    for (key, value) in hash_map {
+                        json_map.insert(key.to_string(), value.try_into()?);
+                    }
+
+                    JsonValue::Object(json_map).to_string()
+                }
+            };
+            writeln!(file, "{json_string}").map_storage_err()?;
+        }
```
panarch, line 84: nit. this would be more simple.
```rust
for row in rows {
    let json_string = match row {
        DataRow::Vec(values) => labels
            .iter()
            .zip(values.into_iter())
            .map(|(key, value)| Ok((key.to_string(), value.try_into()?)))
            .collect::<Result<Map<String, JsonValue>>>(),
        DataRow::Map(hash_map) => hash_map
            .into_iter()
            .map(|(key, value)| Ok((key.to_string(), value.try_into()?)))
            .collect(),
    }
    .map(JsonValue::Object)?
    .to_string();

    writeln!(file, "{json_string}").map_storage_err()?;
}
```

storages/jsonl-storage/tests/jsonl_stroage_sample.rs
```diff
@@ -0,0 +1,273 @@
...
+
+    let actual = glue.execute("SELECT * FROM WrongSchema");
+    let expected = Err(Error::Schema(SchemaParseError::CannotParseDDL));
+
+    assert_eq!(actual, expected);
+
+    test(
+        glue.execute("SELECT * FROM UnsupportedPrimaryKey")
+            .map(|mut payloads| payloads.remove(0)),
+        Err(Error::StorageMsg("primary key is not supported".to_owned())),
```
devgony, line 271: ## Issue
```sql
CREATE TABLE Item (id int PRIMARY KEY, name text);
INSERT INTO Item VALUES (2, 'b');
INSERT INTO Item VALUES (1, 'a');

Above query will get `duplicated key` error cuz INSERT Statement was trying to use id as primary key but internally primary key is fixed to `lineNumber`.
Line1: {id: 2, name: 'b'} already existed so that it threw `duplicated key` error.

## Todo
Throw PrimaryKeyNotSupported Error with `validata_primary_key` at `fetch_schema` and `insert_schema`.

## Another Issue
But `generate_store_tests` includes `PK tests` as well so that we might split PK test (or feature)

File: storages/jsonl-storage/src/lib.rs [hunk 21]
```diff
@@ -0,0 +1,139 @@
...
+                            JsonlStorageError::ColumnDoesNotExist(column_def.name.clone()),
+                        )?;
+
+                        if column_def
+                            .unique
+                            .map(|column_unique_option| column_unique_option.is_primary)
+                            .unwrap_or(false)
+                        {
+                            key = Some(value.clone().try_into().map_storage_err()?);
+                        }
```
panarch, line 104: nit. this might look more simple..?
```rust
if column_def.unique == Some(ColumnUniqueOption { is_primary: true }) {
    key = Some(value.clone().try_into().map_storage_err()?);
}
```

storages/jsonl-storage/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+
+            let key = match key {
+                Some(key) => key,
+                None => index.try_into().map(Key::I64).map_storage_err()?,
+            };
+
+            Ok((key, data_row))
+        });
+
+        Ok(Box::new(row_iter))
```
panarch, line 129: nit. it would be good to reduce indentation level
```rust
let get_index_key = || index.try_into().map(Key::I64).map_storage_err();
let json_row = HashMap::parse_json_object(&line.map_storage_err()?)?;

let column_defs = match &schema.column_defs {
    Some(column_defs) => column_defs,
    None => {
        let key = get_index_key()?;
        let row = DataRow::Map(json_row);

        return Ok((key, row));
    }
};

let mut key: Option<Key> = None;
let mut values = Vec::with_capacity(column_defs.len());
for column_def in column_defs {
    let value = json_row.get(&column_def.name).map_storage_err(
        JsonlStorageError::ColumnDoesNotExist(column_def.name.clone()),
    )?;

    if column_def.unique == Some(ColumnUniqueOption { is_primary: true }) {
        key = Some(value.clone().try_into().map_storage_err()?);
    }

    let value = match value.get_type() {
        Some(data_type) if data_type != column_def.data_type => {
            value.cast(&column_def.data_type)?
        }
        Some(_) | None => value.clone(),
    };

    values.push(value);
}

let key = match key {
    Some(key) => key,
    None => get_index_key()?,
};
let row = DataRow::Vec(values);

Ok((key, row))
```

storages/jsonl-storage/src/store.rs
```diff
@@ -0,0 +1,67 @@
...
+                    .as_ref()
+                    .map(|dir_entry| {
+                        dir_entry
+                            .path()
+                            .extension()
+                            .map(|os_str| os_str.to_str() == Some("jsonl"))
+                            .unwrap_or(false)
+                    })
+                    .unwrap_or(true)
+            })
```
panarch, line 36: nit.
```rust
.filter(|result| match result.as_ref() {
    Ok(dir_entry) => {
        dir_entry.path().extension().map(OsStr::to_str).flatten() == Some("jsonl")
    }
    Err(_) => true,
})
```

ever0de (approved):
LGTM

storages/jsonl-storage/src/store.rs
panarch:
```rust
.filter(|result| match result {
    Ok(entry) => entry.path().extension().and_then(OsStr::to_str) == Some("jsonl"),
    Err(_) => true,
})
```

panarch (approved):
Awesome 🥇 👍 🥇 👍 

Looks all great!!

PR #1055, opened 2023-01-07
Replace `Box::pin` to `futures_enum` in aggregate module

ever0de:


panarch (approved):
Thanks for the good catch!

PR #1061, opened 2023-01-14
Setup sql logic test

### LINKED ISSUE #1059: Setup `sqllogictest`
We need to support the logic test that we usually use to check which queries are not supported.

## 🤔 Conversation

I tried using `sqllogictest-rs`, it should be able to start the daemon. 

## Reference

- <https://github.com/jamii/hytradboi-jam-2022>
- <https://github.com/risinglightdb/sqllogictest-rs>
- <https://github.com/gregrahn/sqllogictest>

ever0de:
resolved #1059

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,7 @@
 members = [
 	"cli",
 	"core",
+	"logic-test",
 	"pkg/rust",
 	"pkg/javascript/web",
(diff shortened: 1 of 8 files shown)


ever0de:
Migrated https://github.com/gluesql/gluesql-logic-test

PR #1063, opened 2023-01-15
Implement Rand function

### LINKED ISSUE #593: support Rand() function
support the rand() function (random).  optional parameter seed.
see
[https://www.w3schools.com/sql/func_mysql_rand.asp](https://www.w3schools.com/sql/func_mysql_rand.asp)

pythonbrad:
resolve #593


core/src/executor/evaluate/stateless.rs
```diff
@@ -275,6 +275,13 @@ fn evaluate_function<'a>(context: &Context<'_>, func: &'a Function) -> Result<Ev
         Function::IfNull { expr, then } => f::ifnull(eval(expr)?, eval(then)?),
         Function::Sign(expr) => f::sign(name, eval(expr)?),
         Function::Ceil(expr) => f::ceil(name, eval(expr)?),
+        Function::Rand(expr) => {
+            let expr = match expr {
+                Some(v) => Some(eval(v)?),
+                None => None,
+            };
```
panarch, line 282: nit. there is `eval_opt` sugar closure function. so you can use
```rust
let expr = eval_opt(expr.as_ref())?;
```

core/src/ast_builder/expr/function.rs
```diff
@@ -440,6 +447,9 @@ pub fn ifnull<'a, T: Into<ExprNode<'a>>, U: Into<ExprNode<'a>>>(expr: T, then: U
 pub fn ceil<'a, T: Into<ExprNode<'a>>>(expr: T) -> ExprNode<'a> {
     ExprNode::Function(Box::new(FunctionNode::Ceil(expr.into())))
 }
+pub fn rand<'a, T: Into<ExprNode<'a>>>(expr: T) -> ExprNode<'a> {
+    ExprNode::Function(Box::new(FunctionNode::Rand(Some(expr.into()))))
+}
```
panarch, line 452: Currently, using ast builder we cannot use random function without seed.
It would be better to take `Option<ExprNode<'a>>` rather than `Into<ExprNode<'a>>`.
Like.. `rtrim` and `ltrim`.

fyi.
For now, when we use `Option<ExprNode<'a>>`, then `Into` is not supported yet.

panarch (changes requested):
I left some very nit comments.
Everything else looks nice.

Thanks a lot :)
also thanks for adding `rand` support to ast_builder.

pythonbrad:
Thanks for your review.

panarch (approved):
Looks all good. Thanks a lot 👍

Issue #864, opened 2022-09-19
Implement `INITCAP` function


kim-seo-hyun:
## Description
Add expression INITCAP.
INITCAP returns char, with the first letter of each word in uppercase, all other letters in lowercase.
Words are delimited by white space or characters that are not alphanumeric.
char can be of any of the data types CHAR, VARCHAR2, NCHAR, or NVARCHAR2. The return value is the same data type as char. 
``` sql

SELECT INITCAP('the soap') "Capitals"
  FROM DUAL; 

Capitals
---------
The Soap
```

ref : https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/INITCAP.html#GUID-9FE9E0EE-D6B6-4C2C-BDEF-4FF4E1314560

## Test to task

``` rs
"SELECT INITCAP(name) FROM Item;",
Ok(select!(
     "INITCAP(name)"
      Str              
     "Claire".to_owned()
     "Claire Kim".to_owned()
)),
```

## To do
- [ ] rename upper_lower to upper_lower_initcap 
- [ ] add test in upper_lower_initcap
- [ ] add initcap function

panarch:
resolved in #1064

PR #1064, opened 2023-01-15
add initcap function

pythonbrad:
- Linked issues
#864


core/src/executor/evaluate/function.rs
```diff
@@ -105,6 +105,22 @@ pub fn lower(name: String, expr: Evaluated<'_>) -> Result<Evaluated> {
...
+    for c in string.chars() {
+        if prev_char_alphanum {
+            char_vector.push(c.to_ascii_lowercase());
+        } else {
+            char_vector.push(c.to_ascii_uppercase());
+        }
+        prev_char_alphanum = c.is_alphanumeric();
+    }
+    let string = char_vector.iter().collect::<String>();
+    Ok(Evaluated::from(Value::Str(string)))
```
devgony, line 121: We are trying to keep convetion to avoid `mut` keyword in `core` module.
Looks like we can try to refactor like below. 
(though `scan` use internal state as mut ref)
```suggestion
let string = string
    .chars()
    .scan(false, |prev_char_alphanum, current| {
        let char = match prev_char_alphanum {
            true => current.to_ascii_lowercase(),
            false => current.to_ascii_uppercase(),
        };
        *prev_char_alphanum = current.is_alphanumeric();

        Some(char)
    })
    .collect::<String>();

Ok(Evaluated::from(Value::Str(string)))
```

pythonbrad: Thanks

devgony (approved):
Look good! Thank you for contribution!

panarch:
Sorry for the late review on this. I'll review and leave comments at most in a few days.

panarch (changes requested):
I have an one nit request. Could you make a `initcap` test in test-suite rather than merging `initcap` with `upper_lower` because `initcap` has some different behaviors so it would be better to be separated.
Everything else looks great. Thanks a lot.

pythonbrad:
done ✅

panarch (approved):
Looks all great 👍 👍 
Thanks a lot!

PR #1066, opened 2023-01-16
add custom function for custom storage developer [DRAFT]

### LINKED ISSUE #660: Implement `Function::Custom` for custom storage developer
We found a need to easily create custom functions in the references below.
- <https://github.com/KOBA789/gh-sql/issues/12>

Function parsing in `sqlparser-rs` crate is already supported now.
https://github.com/sqlparser-rs/sqlparser-rs/blob/f74753436a57c961a17863168b84056847f88edf/src/ast/mod.rs#L1039-L1044

### Example

If it is a function name that is not supported by `translate`, an error is returned. There is a way to pass it to Custom and use it if the name is in the store trait for the custom function implemented by the user.

```rust
/// declare `function` enum variant
Function::Custom {
    name: String,
    args: Vec<Expr>
}

// The name has not been decided.
trait CustomFunction {
    fn custom_function(self, name: String, args: Vec<Expr>) -> Result<Evaluated>
}
```

- Why the result value is `Result<Evaluated>`

https://github.com/gluesql/gluesql/blob/3551c6f6e9d395aa3fd884e9230582c430f0f5d7/core/src/executor/evaluate/mod.rs#L203-L208

pythonbrad:
resolve #660


pythonbrad:
@ever0de, I don't know if i understand well the issue but me i see something like that
```rust
use gluesql::{
    prelude::{Glue, Payload, Value},
    sled_storage::SledStorage,
    ...
};

impl CustomFunction for Function {
    fn call(&self, name: String, args: Vec<Value>) -> Result<Value, TranslateError> {
        match name.as_str() {
            "HELLO" => {
                if args.is_empty() {
                    Ok(Value::Str(String::from("Hello!")))
                } else {
                    Err(TranslateError::FunctionArgsLengthNotMatching {
                        name,
                        expected: 0,
                        found: args.len(),
                    })
                }
            }
            _ => Err(TranslateError::UnsupportedFunction(name)),
        }
    }
}

fn main() {
    let storage = SledStorage::new("/tmp/gluesql/hello_world").expect("Something went wrong!");
    let queries = "
      CREATE TABLE greet (name TEXT);
      INSERT INTO greet VALUES (hello());
    ";

    glue.execute(queries).expect("Execution failed");
}
```

pythonbrad:
I don't think that it's possible to implement it in using trait.
[How do I implement a trait I don't own for a type I don't own?](https://stackoverflow.com/questions/25413201/how-do-i-implement-a-trait-i-dont-own-for-a-type-i-dont-own)

ever0de:
> I don't think that it's possible to implement it in using trait. [How do I implement a trait I don't own for a type I don't own?](https://stackoverflow.com/questions/25413201/how-do-i-implement-a-trait-i-dont-own-for-a-type-i-dont-own)

One of the traits or implementings can be internal.
Because now both are foreign.

However, from the standpoint of the library, the implementing medium does not have to be the library.
This is because we can either create a dyn object for the trait or receive it as a generic.

I think that something that implements CustomFunction should go at or below the Glue level.
And I think the trait for CustomFunction can be implemented by users who create custom storage.

pythonbrad:
@ever0de, What do you think about this implementation ?

### PROPOSAL

- User: *main.rs*
```rs
use gluesql::{memory_storage::MemoryStorage, prelude::{Glue, Value, FunctionProxy}};

fn translate(_args: Vec<Value>) -> Result<Value> {
    if let Value::Str(v) = &_args[0] {
        Ok(Value::Str(v.replace("Python", "Rust")))
    } else {
        Ok(Value::Null)
    }
}

fn main() {
    let storage = MemoryStorage::default();

    let proxy = FunctionProxy { func: translate };

    glue.storage.register_function("translate", FunctionProxy { proxy, [], Value::Str });

    glue.execute("SELECT translate('I like Python!')").unwrap();
}
```

- Core: *executor/evaluate/function.rs*
```rs
pub struct FunctionProxy {
    func: Box<fn(Vec<Value>) -> Value>,
    args: Vec<Value>,
    return: Value
}

impl FunctionProxy {
    fn _call(&self, args: Vec<Evaluated<'_>>) -> Result<Evaluated> {
        let args = args
            .into_iter()
            .map(|expr| Value::try_from(expr).unwrap())
            .collect::<Vec<_>>();

        let value = self.call(args)?;

        Ok(Evaluated::from(value))
    }
    fn call(&self, args: &Vec<Value>) -> Result<Value> {
        // TODO
        // Verify len or len range
        // Verify args types
        self.func(args)
    }
}
```

- Storage: *core/store/function.rs*
```rs
#[async_trait(?Send)]
pub trait Function {
    fn register_function(&mut self, _name: &str, _proxy: FunctionProxy) -> Result<()> {
        let msg = "[Storage] Function::register_function is not supported".to_owned();

        Err(Error::StorageMsg(msg))
    }
    async fn get_function(&self, _name: &str) -> Result<&FunctionProxy> {
        let msg = "[Storage] Function::get_function is not supported".to_owned();

        Err(Error::StorageMsg(msg))
    }
    fn unregister_function(&mut self, _name: &str) -> Result<()> {
        let msg = "[Storage] Function::unregister_function is not supported".to_owned();

        Err(Error::StorageMsg(msg))
    }
}
```

- Core: *executor/evalute/mod.rs*
```rs
Function::Custom { name, exprs } => {
    let custom_func = storage
        .get_function(name)
        .await
        .map_err(|_| TranslateError::UnsupportedFunction(name.to_string()))?;
    custom_func._call(exprs)
}
```

ever0de:
Interesting. It looks good.

pythonbrad:
> Interesting. It looks good.

I implemented it for MemoryStorage. The tests work. 🆗 
But I have a problem about how to store correctly functions.

https://github.com/gluesql/gluesql/pull/1066/files#diff-df3394cd73d05b609521ab3bc36e8e344362aff556006f125880d6868eb60ba0R29-R43

panarch:
btw. I saw you are planning to implement this Function trait for all storages.
but it would be enough with supporting this for `MemoryStorage` first because this is optional trait so you don't need to do all at once.

panarch:
fyi. https://github.com/gluesql/gluesql/pull/1091
It would be good to proceed after merging the main branch.
Now you don't get bothered by handling cfg feature combinations :D

pythonbrad:
> fyi. #1091 It would be good to proceed after merging the main branch. Now you don't get bothered by handling cfg feature combinations :D

I am lost, how to proceed now?
```rs
cfg_if! {
    if #[cfg(feature = "function")] {
        pub trait GStore: Store + Index + Function {}
        impl<S: Store + Function> GStore for S {}
    } else {
        pub trait GStore: Store + Index {}
        impl<S: Store> GStore for S {}
    }
}

cfg_if! {
    if #[cfg(feature = "function")] {
        pub trait GStoreMut: StoreMut + IndexMut + AlterTable + Transaction + Function {}
        impl<S: StoreMut + IndexMut + AlterTable + Transaction + Function> GStoreMut for S {}
    } else {
        pub trait GStoreMut: StoreMut + IndexMut + AlterTable + Transaction {}
        impl<S: StoreMut + IndexMut + AlterTable + Transaction> GStoreMut for S {}
    }
}
```

pythonbrad:
I try to correct

ever0de:
Is the implementation complete?
I ask because there is a comment left

pythonbrad:
I can say that it's complete.
I just implement for memory storage.

pythonbrad:
The custom function can take any argument types and return any type supported by GlueSql.
I would added checking of arguments length, types and return type, but since this api will be used by a developer, i think it will be a limitation. I prefer leave the control to the developer.
```rs
```

pythonbrad:
Please @panarch, I have been working on a pull request and I want to check with you whether I should continue working on it or close it.

panarch:
@pythonbrad Ok, I think that we'd better to focus on https://github.com/gluesql/gluesql/pull/1095 because that is almost close to the complete.
How about to close this for now and consider it again later?

marshauf:
I am interested in custom functions written in Rust like implemented in this PR.
Would it be possible to revive this PR? In addition add the possibility to use async custom functions?

ever0de:
> I am interested in custom functions written in Rust like implemented in this PR. Would it be possible to revive this PR? In addition add the possibility to use async custom functions?

The implementation is complete with another PR.
I am attaching the link below
https://github.com/gluesql/gluesql/pull/1095

panarch:
@marshauf for the custom functions written in Rust, it would be better to do approach based on extension of `CustomFunction` trait.
We can already have extra custom functions written in Rust by implementing custom `CustomFunction` trait for storages but that is possible only in compile time.
So to make database users to extend custom functions to existing storages, what we need to provide is a kind of function injection feature.
I'll also think about the design, and if you have an idea, please make the issue and we can discuss based on that.

PR #1067, opened 2023-01-18
Add IndexedDB storage support

panarch:
Initial IndexedDB storage support for GlueSQL.
It uses https://github.com/devashishdxt/idb
Thanks @devashishdxt

### DIFF
--- a/storages/idb-storage/src/lib.rs
+++ b/storages/idb-storage/src/lib.rs
@@ -11,11 +11,12 @@ use {
     gloo_utils::format::JsValueSerdeExt,
     gluesql_core::{
         data::{Key, Schema, Value},
-        result::{MutResult, Result, TrySelf},
+        result::{Error, MutResult, Result, TrySelf},
         store::{DataRow, RowIter, Store, StoreMut},
     },
     idb::{CursorDirection, Database, Factory, ObjectStoreParams, Query, TransactionMode},
     serde_json::Value as JsonValue,
+    std::sync::{Arc, Mutex},
     wasm_bindgen::JsValue,
 };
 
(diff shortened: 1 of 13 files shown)


storages/idb-storage/src/lib.rs
```diff
@@ -0,0 +1,454 @@
...
+        let store = transaction.object_store(table_name).err_into()?;
+        let cursor = store
+            .open_cursor(None, Some(CursorDirection::Next))
+            .await
+            .err_into()?;
+
+        let mut cursor = match cursor {
+            Some(cursor) => cursor,
+            None => {
+                return Ok(Box::new(Vec::new().into_iter()));
```
ever0de, line 169: https://doc.rust-lang.org/stable/std/iter/fn.empty.html use this
```suggestion
                return Ok(Box::new(std::iter::empty()));
```

storages/idb-storage/src/lib.rs
```diff
@@ -0,0 +1,454 @@
...
+impl Store for IdbStorage {
+    async fn fetch_all_schemas(&self) -> Result<Vec<Schema>> {
+        let transaction = self
+            .database
+            .transaction(&[SCHEMA_STORE], TransactionMode::ReadOnly)
+            .err_into()?;
+
+        let store = transaction.object_store(SCHEMA_STORE).err_into()?;
+        let schemas = store.get_all(None, None).await.err_into()?;
+
```
ever0de, line 103: Missing `commit`

storages/idb-storage/src/lib.rs
```diff
@@ -0,0 +1,454 @@
...
+            let key: Key = Value::try_from(key)?.try_into()?;
+
+            let row = convert(current_row, column_defs.as_deref())?;
+
+            rows.push((key, row));
+
+            let result = cursor.advance(1).await;
+            if result.is_err() {
+                break;
+            }
```
ever0de, line 188:
```suggestion
            if let Err(err) = result {
                match err {
                    idb::Error::CursorFinished => break,
                    _ => // temp: return Err(err.into()),
                }
            }
```

When a variant other than `CursorFinished` comes out, I think we need to propagate an error.

storages/idb-storage/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+    }
+
+    pub async fn append_data(&mut self, table_name: &str, new_rows: Vec<DataRow>) -> Result<()> {
+        let transaction = self
+            .database
+            .transaction(&[table_name], TransactionMode::ReadWrite)
+            .err_into()?;
+        let store = transaction.object_store(table_name).err_into()?;
+
+        for data_row in new_rows.into_iter() {
```
ever0de, line 345:
```suggestion
        for data_row in new_rows {
```

storages/idb-storage/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+        table_name: &str,
+        new_rows: Vec<(Key, DataRow)>,
+    ) -> Result<()> {
+        let transaction = self
+            .database
+            .transaction(&[table_name], TransactionMode::ReadWrite)
+            .err_into()?;
+        let store = transaction.object_store(table_name).err_into()?;
+
+        for (key, data_row) in new_rows.into_iter() {
```
ever0de, line 372: ditto

storages/idb-storage/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+    }
+
+    pub async fn delete_data(&mut self, table_name: &str, keys: Vec<Key>) -> Result<()> {
+        let transaction = self
+            .database
+            .transaction(&[table_name], TransactionMode::ReadWrite)
+            .err_into()?;
+        let store = transaction.object_store(table_name).err_into()?;
+
+        for key in keys.into_iter() {
```
ever0de, line 398: ditto

ever0de (approved):
Looks good 👍 
Took a look at IndexedDB's interesting interface. It's interesting.

storages/idb-storage/src/lib.rs
ever0de: Alas, I missed it in the double check, but I need to do error propagation here.

PR #1071, opened 2023-01-26
Change StoreMut interface to use &mut self, not to take ownership

panarch:
…self

New `StoreMut` interface which does not take ownership of self.
Rename previous `StoreMut` trait to `IStoreMut` which works for core internal codes.
Custom storage makers only needs to impl this `StoreMut` below.

```rust
#[async_trait(?Send)]
pub trait StoreMut {
    async fn insert_schema(&mut self, schema: &Schema) -> Result<()>;

    async fn delete_schema(&mut self, table_name: &str) -> Result<()>;

    async fn append_data(&mut self, table_name: &str, rows: Vec<DataRow>) -> Result<()>;

    async fn insert_data(&mut self, table_name: &str, rows: Vec<(Key, DataRow)>) -> Result<()>;

    async fn delete_data(&mut self, table_name: &str, keys: Vec<Key>) -> Result<()>;
}
```

### DIFF
--- a/core/src/result.rs
+++ b/core/src/result.rs
@@ -9,7 +9,7 @@ use {
             SelectError, SortError, UpdateError, ValidateError,
         },
         plan::PlanError,
-        store::{GStore, GStoreMut},
+        store::GStoreMut,
         translate::TranslateError,
     },
     serde::Serialize,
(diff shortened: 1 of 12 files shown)


core/src/store/mod.rs
```diff
@@ -87,17 +87,71 @@ pub trait Store {
...
+        table_name: &str,
+        rows: Vec<(Key, DataRow)>,
+    ) -> MutResult<Self, ()>;
 
-    async fn delete_data(self, table_name: &str, keys: Vec<Key>) -> MutResult<Self, ()>;
+    async fn delete_data(mut self, table_name: &str, keys: Vec<Key>) -> MutResult<Self, ()>;
+}
+
+#[async_trait(?Send)]
+impl<T: GStoreMut> IStoreMut for T {
```
ever0de, line 123:
```suggestion
impl<T: StoreMut> IStoreMut for T {
```

core/src/result.rs
```diff
@@ -130,11 +130,11 @@ pub trait TrySelf<V>
 where
     Self: Sized,
 {
-    fn try_self<T: GStore + GStoreMut>(self, storage: T) -> MutResult<T, V>;
+    fn try_self<T: GStoreMut>(self, storage: T) -> MutResult<T, V>;
```
ever0de, line 133: You can remove constraints on `TrySelf`

ever0de (approved):
Looks good :+1:

PR #1072, opened 2023-01-28
Update all store mut trait methods to take &mut self

panarch:
Remove IStoreMut, core only uses StoreMut with &mut self.
Update AlterTable, IndexMut and Transaction trait methods to take &mut self.

### DIFF
--- a/core/src/executor/alter/table.rs
+++ b/core/src/executor/alter/table.rs
@@ -23,7 +23,7 @@ pub async fn create_table<T: GStore + GStoreMut>(
     if_not_exists: bool,
     source: &Option<Box<Query>>,
 ) -> Result<()> {
-    let target_columns_defs = match source.as_ref().map(AsRef::as_ref) {
+    let target_columns_defs = match source.as_deref() {
         Some(Query { body, .. }) => match body {
             SetExpr::Select(select_query) => match &select_query.from.relation {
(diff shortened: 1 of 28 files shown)


core/src/executor/alter/table.rs
```diff
@@ -5,184 +5,158 @@ use {
...
-                    .map_ok(Into::into)
-                    .try_collect()
-                    .await
-            }
-            .await
-            .try_self(storage)?;
-
-            storage.append_data(target_table_name, rows).await
+        Some(query) => {
+            let rows = select(storage, query, None)
```
ever0de, line 131: The `select` function is getting the dyn reference from GStore. I need to fix it here too.

core/src/executor/insert.rs
```diff
@@ -46,47 +46,44 @@ enum RowsData {
...
                 .insert_data(table_name, rows)
                 .await
-                .map(|(storage, _)| (storage, num_rows))
+                .map(|_| num_rows)
         }
     }
 }
 
-async fn fetch_vec_rows<T: GStore + GStoreMut>(
+async fn fetch_vec_rows<T: GStore>(
```
ever0de, line 86: nice catch !

ever0de (approved):
[nit-picking]
Other than minor comments, it looks good

PR #1073, opened 2023-01-28
Remove `MutResult`

panarch:
Change core/ execute functions not to take storage ownership and make it return Result, not MutResult

ref. #1072


ever0de (approved):
LGTM

PR #1074, opened 2023-01-29
Fix `plan::validate` to handle `CTAS` and `ITAS` adding unit test

devgony:
## Goal
Include source queries of `CTAS` and `ITAS` at `plane::validate`
```sql
CREATE TABLE Ids AS SELECT id FROM Users A JOIN Users B on A.id = B.id;
INSERT INTO Users SELECT id FROM Users A JOIN Users B on A.id = B.id;
```
## Todo
- [x] add unit test cases
- [x] add E2E test cases
- [x] fix plan::validate to handle `CTAS` and `ITAS`
- [x] fix fetch_schema_map to handle `CTAS`

## Next
- Right after implementing implicit join or update join, they should be validated as well

### DIFF
--- a/core/src/plan/validate.rs
+++ b/core/src/plan/validate.rs
@@ -202,8 +202,16 @@ fn validate_test() {
                 amount INTEGER
             );
         ");
-    let sql = "INSERT INTO Player VALUES(1, 'a');";
-    let (schema_map, statement) = plan(&storage, sql);
 
+    let sql = "INSERT INTO Player VALUES(1, 'a')";
+    let (schema_map, statement) = plan(&storage, sql);
+    assert!(contextualize_stmt(&schema_map, &statement).is_some());
+
+    let sql = "DROP TABLE Player";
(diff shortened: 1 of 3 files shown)


core/src/plan/validate.rs
```diff
@@ -5,25 +5,34 @@ use {
...
-                expr: Expr::Identifier(ident),
-                ..
-            } = select_item
-            {
-                if let Some(context) = contextualize_stmt(schema_map, statement) {
-                    context.validate_duplicated(ident)?;
+    let query = match statement {
+        Statement::Query(query) => Some(query),
+        Statement::Insert { source, .. } => Some(source),
+        Statement::CreateTable { source, .. } => source.as_ref().map(Deref::deref),
```
panarch, line 17: nit. without using `std::ops::Deref` you can simply use `source.as_deref()` here

panarch (approved):
image

Looks all great! Thanks a lot 👍 👍

PR #1075, opened 2023-01-29
Replace `dyn object` to generic

ever0de:


panarch (approved):
wow, nice improvement! Thanks a lot! All looks nice

PR #1076, opened 2023-01-29
Remove `?Send` option of `async-trait`

ever0de:
Not every `async_trait` we implement requires `?Send`.


ever0de:
<https://docs.rs/async-trait/latest/async_trait/#dyn-traits>
I understood with that document.
It looks like we can remove this constraint by replacing the default implementation with declare_macro.

ever0de:
depends on #1078

ever0de:
Close, because it's deprecated code

PR #1082, opened 2023-02-03
Update `StringExt` implementation to use `str`

ever0de:
context: https://github.com/gluesql/gluesql/pull/999/files#diff-f4f27532f21aa04e9550528e5d3922ece5b9ac8d3056096eacfd11c7da2a7ed0R344-R345

### DIFF
--- a/core/src/data/literal.rs
+++ b/core/src/data/literal.rs
@@ -62,8 +62,8 @@ impl<'a> TryFrom<&'a AstLiteral> for Literal<'a> {
     }
 }
 
-impl PartialEq<String> for Literal<'_> {
-    fn eq(&self, other: &String) -> bool {
+impl PartialEq<str> for Literal<'_> {
+    fn eq(&self, other: &str) -> bool {
         match (self, other) {
             (Literal::Text(l), r) => l.as_ref() == r,
             _ => false,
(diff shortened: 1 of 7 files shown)


core/src/data/literal.rs
```diff
@@ -62,24 +62,48 @@ impl<'a> TryFrom<&'a AstLiteral> for Literal<'a> {
...
+    fn eq(&self, other: &&str) -> bool {
+        PartialEq::<str>::eq(self, *other)
+    }
+}
+
+impl PartialOrd<&str> for Literal<'_> {
+    fn partial_cmp(&self, other: &&str) -> Option<Ordering> {
+        PartialOrd::<str>::partial_cmp(self, *other)
+    }
+}
```
panarch, line 105: Can we use `AsRef<str>` here?
```rust
impl<T: AsRef<str>> PartialEq<T> for Literal<'_> {
     fn eq(&self, other: &T) -> bool {             
         PartialEq::<str>::eq(self, other.as_ref())
     }                                             
 }                                                 
```

ever0de: Oh.. `&&str` doesn't seem to be included here.

ever0de: No, I was impatient. (just before check success)

panarch (approved):
nice improvement! Thanks a lot 👍

PR #1083, opened 2023-02-04
Replace `async-recursion` crate to `LocalBoxFuture`

ever0de:
In the process of attaching `Send` and `Sync`, it is difficult to solve if there are lifetime and `async-recursion` macros together.

Replace with `LocalBoxFuture`.


ever0de:
Close, because it's deprecated code

PR #1086, opened 2023-02-09
Refactor CreateTable.columns from `Vec<ColumnDef>` to `Option<Vec<ColumnDef>>`

devgony:
## Goal
```diff
CreateTable {                                                                                     
    if_not_exists: bool,                                                                          
    name: String,                                                                                 
    Optional schema                                                                           
-  columns: Vec<ColumnDef>,
+  columns: Option<Vec<ColumnDef>>,                                                              
    source: Option<Box<Query>>,                                                                   
    engine: Option<String>,                                                                       
},                          
```

## Todo
Refactor CreateTable.columns from Vec<ColumnDef> to Option<Vec<ColumnDef>>

## Next
- implement `from_ddl()`

### DIFF
--- a/core/src/executor/alter/table.rs
+++ b/core/src/executor/alter/table.rs
@@ -102,7 +102,7 @@ pub async fn create_table<T: GStore + GStoreMut>(
                 Some(column_defs)
             }
         },
-        None if column_defs.is_some() => column_defs.map(|s| s.to_vec()),
+        None if column_defs.is_some() => column_defs.map(<[ColumnDef]>::to_vec),
         None => None,
     };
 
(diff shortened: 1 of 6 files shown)


core/src/executor/alter/table.rs
```diff
@@ -102,7 +102,7 @@ pub async fn create_table<T: GStore + GStoreMut>(
                 Some(column_defs)
             }
         },
-        None if !column_defs.is_empty() => Some(column_defs.to_vec()),
+        None if column_defs.is_some() => column_defs.map(|s| s.to_vec()),
```
devgony, line 105: is there any way like `.map(slice::to_vec)`?

panarch: that's interesting.
I just got to know that.. this works!
```rust
 column_defs.map(<[_]>::to_vec)
```

panarch (approved):
Thanks for the good catch! All looks good :D

PR #1089, opened 2023-02-09
Support `Schema::from_ddl(ddl: &str) -> String`

devgony:
## Goal
Currently we can convert `Schema` to  `DDL String` by `Schema::to_ddl(self) -> String`
In this PR, reverse conversion  `DDL String`(CREATE TABLE & CREATE INDEX) to `Schema` will be supported by
```rs
Schema::from_ddl(ddl: &str) -> String
```

## Todo
- [x] replace Schema::to_ddl(`self`) to Schema::to_ddl(`&self`)
- [x] Add SchemaParseError::CannotParseDDL 
- [x] implement `Schema::from_ddl(ddl: &str) -> String`

### DIFF
--- a/core/src/data/schema.rs
+++ b/core/src/data/schema.rs
@@ -271,6 +271,14 @@ mod tests {
     }
 
     #[test]
+    fn invalid_ddl() {
+        let invalid_ddl = "DROP TABLE Users";
+        let actual = Schema::from_ddl(invalid_ddl);
+        assert_eq!(actual, Err(SchemaParseError::CannotParseDDL.into()));
+    }
+
+    #[test]
+    #[cfg(feature = "index")]
     fn table_with_index() {
         let schema = Schema {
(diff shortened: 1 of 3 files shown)


core/src/data/schema.rs
```diff
@@ -62,17 +70,124 @@ impl Schema {
...
+                    }
+                    _ => Err(SchemaParseError::CannotParseDDL.into()),
+                }
+            })
+            .collect::<Result<Vec<_>>>()?;
+
+        let create_table = statements
+            .get(0)
+            .map(Ok::<_, Error>)
+            .unwrap_or_else(|| Err(SchemaParseError::CannotParseDDL.into()))?;
```
ever0de, line 110:
```suggestion
        let create_table = statements.get(0).ok_or(SchemaParseError::CannotParseDDL)?;
```

How about this?

panarch (approved):

PR #1091, opened 2023-02-10
Remove Store trait related cfg features,

panarch:
Enable all by default.

Compile time feature control causes too much management cost.
We already have test cases in `test-suite` splitted based on store trait features.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -11,7 +11,7 @@ env:
 
 jobs:
   clippy_workspace:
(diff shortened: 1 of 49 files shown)


ever0de (approved):
This code looks good. Would it be possible to update the README as there are some outdated sections present

panarch:
> This code looks good. Would it be possible to update the README as there are some outdated sections present

v0.14 release will be .. next month so it would be too early to update README at this moment.
We also have plan to write the user guide document so I'll update README later with the doc.

PR #1094, opened 2023-02-13
Add "cli" unittest

pythonbrad:


pythonbrad:
```
[Select { labels: ["name"], rows: [] }]
thread 'main' panicked at 'index out of bounds: the len is 0 but the index is 0', pkg/rust/examples/sled_multi_threaded.rs:59:26
```
This error can be manage if we wait the thread to terminate before get the row.

panarch (approved):
Thanks a lot 👍 👍

PR #1095, opened 2023-02-17
Support user level sql function

pythonbrad:
### Linked to 
#660
### TODO
- [x] Show function statement
- [x] Drop function statement
- [x] Create function statement
### Example
```sql
CREATE FUNCTION nothing();
CREATE FUNCTION hello() RETURN 'hello!';
CREATE FUNCTION add(x int) RETURN x+1;
CREATE FUNCTION div(a int, b int) RETURN a/b;
CREATE FUNCTION incr(a int, b int default 1) RETURN (SELECT bar from foo);
SELECT nothing();
SELECT incr(1);
SHOW FUNCTIONS;
```


core/src/executor/evaluate/mod.rs
```diff
@@ -311,10 +311,12 @@ async fn evaluate_function<'a, 'b: 'a, 'c: 'a, T: GStore>(
             if let Some(custom_func) = custom_func {
                 let args: Vec<Evaluated<'_>> = stream::iter(exprs).then(eval).try_collect().await?;
 
-                let args = args
-                    .into_iter()
-                    .map(|expr| Value::try_from(expr).unwrap())
-                    .collect::<Vec<_>>();
+                let fargs: Vec<Evaluated<'_>> = if let Some(fargs) = &custom_func.args {
+                    let fargs: Vec<Expr> = fargs.iter().filter_map(|y| y.default.clone()).collect();
+                    stream::iter(&fargs).then(eval).try_collect().await?
```
pythonbrad, line 316: @panarch , please can ou help me to debug this part ?
i got this error `returns a value referencing data owned by the current function`

core/src/executor/evaluate/mod.rs
```diff
@@ -323,25 +325,28 @@ async fn evaluate_function<'a, 'b: 'a, 'c: 'a, T: GStore>(
...
+                            } else {
+                                &arg
+                            };
+                            hm.insert(farg.name.to_owned(), value.to_owned());
+                            Ok(())
+                        })?;
 
                     let row = Row::Map(hm);
                     let rowcontext = RowContext::new(name, Cow::Borrowed(&row), None);
                     let context = Rc::new(rowcontext);
```
pythonbrad, line 346: @panarch, i try to use context set variables in the body by their values passed in argument
i try to implement the context, i got this error
`returning this value requires that "'3" must outlive "'2"`

core/src/executor/evaluate/mod.rs
```diff
@@ -19,11 +19,13 @@ use {
         stream::{self, StreamExt, TryStreamExt},
     },
     im_rc::HashMap,
-    std::{borrow::Cow, rc::Rc},
+    std::{borrow::Cow, collections::HashMap as StdHashMap, rc::Rc},
 };
 
 pub use {error::EvaluateError, evaluated::Evaluated, stateless::evaluate_stateless};
 
+use crate::translate::TranslateError;
```
panarch, line 27: Here we are in `evaluate` module, so rather than using `TranslateError` we can simply use `EvaluateError`.
To decide what type of `Error` to use, we can simply choose based on location and name of the module.

core/src/store/function.rs
```diff
@@ -0,0 +1,32 @@
...
+
+#[async_trait(?Send)]
+pub trait FunctionMut {
+    async fn create_function(&mut self, _func: CustomFunction) -> Result<()> {
+        Err(Error::StorageMsg(
+            "[Storage] Function is not supported".to_owned(),
+        ))
+    }
+
+    async fn drop_function(&mut self, _func_name: &str) -> Result<()> {
```
panarch, line 27: As we use `insert` and `delete` prefixes in store traits, (e.g. StoreMut), how about `insert_function` and `delete_function`?

core/src/store/function.rs
```diff
@@ -0,0 +1,32 @@
...
+use async_trait::async_trait;
+
+#[async_trait(?Send)]
+pub trait Function {
+    async fn fetch_function(&self, _func_name: &str) -> Result<Option<&CustomFunction>> {
+        Err(Error::StorageMsg(
+            "[Storage] Function is not supported".to_owned(),
+        ))
+    }
+    async fn show_functions(&self) -> Result<Vec<String>> {
```
panarch, line 12: this also implies fetching data from storage, as we use `fetch_schema` and `fetch_all_schemas`, it looks that we can also follow the patterns like.. `fetch_function` and `fetch_all_functions`?

core/src/store/function.rs
```diff
@@ -0,0 +1,32 @@
+use crate::data::CustomFunction;
+use crate::result::{Error, Result};
+use async_trait::async_trait;
+
+#[async_trait(?Send)]
+pub trait Function {
```
panarch, line 6: It would be good if we can use name as short as possible but `Function` looks that it has too broad meaning.
In SQL, we are already familiar with `CustomFunction` keyword so how about simply to use `pub trait CustomFunction` and `pub trait CustomFunctionMut` for store traits?

core/src/store/function.rs
```diff
@@ -0,0 +1,32 @@
+use crate::data::CustomFunction;
+use crate::result::{Error, Result};
+use async_trait::async_trait;
```
panarch, line 3: nit. this is the project specific code convention.
we prefer bundling `use` statement into a single one. it is for letting `cargo fmt` to format all by itself.

then here
```rust
use {
    crate::{
        data::CustomFunction,
        result::{Error, Result},
    },
    async_trait::async_trait,
};
```

storages/memory-storage/tests/memory_storage.rs
```diff
@@ -94,3 +100,101 @@ fn memory_storage_transaction() {
     test!(glue "COMMIT", Ok(vec![Payload::Commit]));
     test!(glue "ROLLBACK", Ok(vec![Payload::Rollback]));
 }
+
+#[test]
+fn memory_storage_function() {
```
panarch, line 105: This test doesn't need to be only for `memory_storage`.
Could we move this to `test-suite`? Then all the storage makers can use this test.

panarch (changes requested):

core/src/executor/evaluate/mod.rs
```diff
@@ -294,6 +301,90 @@ async fn evaluate_function<'a, 'b: 'a, 'c: 'a, T: GStore>(
             let exprs = stream::iter(exprs).then(eval).try_collect().await?;
             f::concat(exprs)
         }
+        Function::Custom { name, exprs } => {
+            let custom_func = storage
+                .fetch_function(name)
+                .await
+                .map_err(|_| EvaluateError::UnsupportedFunction(name.to_string()))?;
+            if let Some(custom_func) = custom_func {
```
panarch, line 309:
```rust
let custom_func = storage
    .fetch_function(name)
    .await?
    .ok_or_else(|| EvaluateError::UnsupportedFunction(name.to_string()))?;
```
it would be enough to write like this. then we don't need to use `if let Some(custon_func) = custom_func` to wrap remaining codes.

core/src/executor/evaluate/mod.rs
```diff
@@ -294,6 +301,90 @@ async fn evaluate_function<'a, 'b: 'a, 'c: 'a, T: GStore>(
...
+                } else {
+                    Err(EvaluateError::FunctionArgsLengthNotWithinRange {
+                        name: custom_func.func_name.to_owned(),
+                        expected_minimum: min,
+                        expected_maximum: max,
+                        found: args.len(),
+                    })
+                };
+
+                Ok(value?)
```
panarch, line 383:
```rust
    if let Some(v) = &custom_func.return_ {
        eval_with_context(v, context).await
    } else {
        Ok(Evaluated::from(Value::Null))
    }
} else {
    Err(EvaluateError::FunctionArgsLengthNotWithinRange {
        name: custom_func.func_name.to_owned(),
        expected_minimum: min,
        expected_maximum: max,
        found: args.len(),
    }.into())
}
```

nit. you can omit `Ok(value)` assignment.

core/src/executor/evaluate/mod.rs
```diff
@@ -294,6 +301,90 @@ async fn evaluate_function<'a, 'b: 'a, 'c: 'a, T: GStore>(
...
+                    .map(|v| Value::try_from(v).unwrap())
+                    .collect();
+
+                let empty = vec![];
+
+                let fargs = if let Some(fargs) = &custom_func.args {
+                    fargs
+                } else {
+                    &empty
+                };
```
panarch, line 322: nit. this would be enough.
```rust
let fargs = custom_func.args.as_deref().unwrap_or(&[]);
```

core/src/executor/evaluate/mod.rs
```diff
@@ -294,6 +301,90 @@ async fn evaluate_function<'a, 'b: 'a, 'c: 'a, T: GStore>(
...
+            let custom_func = storage
+                .fetch_function(name)
+                .await
+                .map_err(|_| EvaluateError::UnsupportedFunction(name.to_string()))?;
+            if let Some(custom_func) = custom_func {
+                let args: Vec<Evaluated<'_>> = stream::iter(exprs).then(eval).try_collect().await?;
+                let args: Vec<Value> = args
+                    .into_iter()
+                    .map(|v| Value::try_from(v).unwrap())
+                    .collect();
```
panarch, line 314: You can remove `unwrap` by using `collect::<Result<Vec<_>>>()`

core/src/translate/ddl.rs
```diff
@@ -88,3 +89,22 @@ pub fn translate_column_def(sql_column_def: &SqlColumnDef) -> Result<ColumnDef>
...
+        v.value.to_owned()
+    } else {
+        return Err(TranslateError::UnNamedFunctionArgNotSupported.into());
+    };
+    let data_type = translate_data_type(&arg.data_type)?;
+    let default = if let Some(default) = &arg.default_expr {
+        Some(translate_expr(default)?)
+    } else {
+        None
+    };
```
panarch, line 104: nit.
```rust
let name = arg
    .name
    .as_ref()
    .map(|v| v.value.to_owned())
    .ok_or(TranslateError::UnNamedFunctionArgNotSupported)?;
let data_type = translate_data_type(&arg.data_type)?;
let default = arg.default_expr.as_ref().map(translate_expr).transpose()?;
```

core/src/translate/mod.rs
```diff
@@ -185,6 +199,29 @@ pub fn translate(sql_statement: &SqlStatement) -> Result<Statement> {
...
+        } => {
+            let args = if let Some(args) = args {
+                Some(
+                    args.iter()
+                        .map(translate_operate_function_arg)
+                        .collect::<Result<Vec<_>>>()?,
+                )
+            } else {
+                None
+            };
```
panarch, line 217: nit. `as_ref()` would be useful.
```rust
let args = args
    .as_ref()
    .map(|args| {
        args.iter()
            .map(translate_operate_function_arg)
            .collect::<Result<Vec<_>>>()
    })
    .transpose()?;
```

panarch (changes requested):
Wow, now it's really close to the completion!

With some comments, it would be nice that we can add a few more tests to reach uncovered lines.
ref. you can ignore `await` uncovered lines which we cannot make it work.
https://coveralls.io/builds/57904316
image

core/src/executor/evaluate/mod.rs
```diff
@@ -294,6 +301,81 @@ async fn evaluate_function<'a, 'b: 'a, 'c: 'a, T: GStore>(
...
+
+            let dargs: Vec<Value> = if let Some(fargs) = &custom_func.args {
+                let dargs: Vec<&Expr> =
+                    fargs.iter().filter_map(|y| y.default.as_ref()).collect();
+                let dargs: Vec<Evaluated<'_>> =
+                    stream::iter(dargs).then(eval).try_collect().await?;
+                dargs
+                    .into_iter()
+                    .map(|expr| Value::try_from(expr).unwrap())
+                    .collect()
```
ever0de, line 327: [nit-picking]

```suggestion
            let dargs = if let Some(fargs) = &custom_func.args {
                let dargs =
                    fargs.iter().filter_map(|y| y.default.as_ref()).collect::<Vec<_>>();
                let dargs =
                    stream::iter(dargs).then(eval).try_collect::<Vec<_>>().await?;
                dargs
                    .into_iter()
                    .map(|expr| Value::try_from(expr).unwrap())
                    .collect::<Vec<_>>()
```

When using collect, it is more convenient for future code modifications to pass Vec<_> as an argument instead of explicitly declaring the type of the variable being bound with let.

pythonbrad:
@ever0de , finally i done a rebase. I apply reviews.

ever0de:
Could you please confirm and apply CI?

pythonbrad:
@panarch, i fixed it

test-suite/src/function/custom.rs
```diff
@@ -0,0 +1,103 @@
...
+        (
+            "CREATE FUNCTION add_none ()",
+            Err(AlterError::FunctionAlreadyExists("add_none".to_owned()).into()),
+        ),
+        (
+            "CREATE FUNCTION add_one (n INT, x INT DEFAULT 1) RETURN n + x",
+            Ok(Payload::Create),
+        ),
+        (
+            "CREATE FUNCTION add_two (n INT, x INT DEFAULT 1, y INT) RETURN n + x + y",
```
panarch, line 23: Here, if the arg value is `NULL` then it looks that the default expr is used.
But when we use `value.validate_type`, `NULL` is always valid because all the params in SQL are nullable.

How about allowing `DEFAULT` for the last arguments.

e.g.
```sql
CREATE FUNCTION add_one (n INT, x INT DEFAULT 1) RETURN n + x
```
Then, we can call `add_one` like..
```sql
add_one(1);
add_one(1, 2);
```

and in the case of two defaults,
```sql
CREATE FUNCTION add_two (n INT, x INT DEFAULT 1, y INT DEFAULT 23) RETURN n + x + y
```
Then,
```sql
add_two(1);
add_two(1, 2);
add_two(1, 2, 3);
```
Users can omit the last arguments if they want.

---
and 
```sql
CREATE FUNCTION add_two (n INT, x INT DEFAULT 1, y INT) RETURN n + x + y
```
the above would be invalid because there's no `DEFAULT` expr for the last param `y INT`.

core/src/data/function.rs
```diff
@@ -0,0 +1,28 @@
+use {
+    crate::ast::{Expr, OperateFunctionArg},
+    serde::{Deserialize, Serialize},
+};
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct CustomFunction {
+    pub func_name: String,
+    pub args: Option<Vec<OperateFunctionArg>>,
+    pub return_: Option<Expr>,
```
panarch, line 10: 1. how about `body` rather than `return_`?
2. is that any use case of using `None` for `return_`?
I'm not sure about supporting custom function without function body expr.
if it doesn't, then we can filter out `None` case and remove `Option`.

core/src/data/function.rs
```diff
@@ -0,0 +1,28 @@
+use {
+    crate::ast::{Expr, OperateFunctionArg},
+    serde::{Deserialize, Serialize},
+};
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct CustomFunction {
+    pub func_name: String,
+    pub args: Option<Vec<OperateFunctionArg>>,
```
panarch, line 9: Would it be enough with `pub args: Vec<OperateFunctionArg>`?
like.. `ADD_ZERO()`, `ADD_ONE(3), `ADD_TWO(1, 2)`
it does not look that the custom function has explicit `None` state.

core/src/ast/ddl.rs
```diff
@@ -102,9 +110,26 @@ impl ToSql for ColumnUniqueOption {
...
+impl ToSql for OperateFunctionArg {
+    fn to_sql(&self) -> String {
+        let OperateFunctionArg {
+            name,
+            data_type,
+            default,
+        } = self;
+        let default = default
+            .as_ref()
+            .map(|expr| format!(" DEFAULT {}", expr.to_sql_unquoted()))
```
panarch, line 122: Here, we need to use `expr.to_sql` rather than `to_sql_unquoted`
ref. https://github.com/gluesql/gluesql/pull/1130

core/src/ast/function.rs
```diff
@@ -194,6 +198,14 @@ impl ToSql for Function {
                     .join(", ");
                 format!("CONCAT({items})")
             }
+            Function::Custom { name, exprs } => {
+                let exprs = exprs
+                    .iter()
+                    .map(ToSqlUnquoted::to_sql_unquoted)
```
panarch, line 204: We also need to use `ToSql::to_sql` in here.

In fact, we need to add `ToSqlUnquoted` impl for `ast::Function` but I think that we can do that impl in another PR, not in here.

panarch (changes requested):

core/src/executor/evaluate/error.rs
```diff
@@ -100,6 +100,20 @@ pub enum EvaluateError {
...
     UnsupportedUnaryFactorial(String),
+
+    #[error("unsupported custom function in subqueries")]
+    UnsupportedCustomFunction,
+
+    #[error("function args.length not matching: {name}, expected: {expected_minimum} ~ {expected_maximum}, found: {found}")]
+    FunctionArgsLengthNotWithinRange {
+        name: String,
+        expected_minimum: usize,
+        expected_maximum: usize,
```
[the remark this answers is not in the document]
panarch: @ever0de current custom function does not support open range so it would be better to use `expected_minimum` & `expected_maximum`.
`Range` can describe `(3..)` or `(..=4)` like open ranges that are not supported by our custom function args

test-suite/src/function/custom.rs
```diff
@@ -0,0 +1,127 @@
...
+            }
+            .into()),
+        ),
+        (
+            "SELECT add_two(1, null, 2) as r",
+            Ok(select!(
+                r
+                I64;
+                4
+            )),
```
panarch, line 83: default is only applied when the argument is not specified.
e.g.
```bash
gluesql> create table foo (id integer, c integer null default 1);
gluesql> insert into foo (id) values (100);
1 row inserted

gluesql> select * from foo;
| id  | c |
|-----|---|
| 100 | 1 |

gluesql> insert into foo (id, c) values (100, NULL);
1 row inserted

gluesql> select * from foo;
| id  | c    |
|-----|------|
| 100 | 1    |
| 100 | NULL |
```

So, the expected value would be `Value::Null`

core/src/executor/evaluate/mod.rs
```diff
@@ -294,6 +301,75 @@ async fn evaluate_function<'a, 'b: 'a, 'c: 'a, T: GStore>(
...
+                Err((EvaluateError::FunctionArgsLengthNotWithinRange {
+                    name: custom_func.func_name.to_owned(),
+                    expected_minimum: min,
+                    expected_maximum: max,
+                    found: args.len(),
+                })
+                .into())
+            };
+
+            Ok(value?)
```
panarch, line 371: We can simplify the codes.
I added some comments for your information. when you use this, you can remove the comments.
Behaviors are mostly same but there's slight difference in handling default expr.

```rust
let CustomFunction {
    func_name,
    args,
    body,
} = storage
    .fetch_function(name)
    .await?
    .ok_or_else(|| EvaluateError::UnsupportedFunction(name.to_string()))?;

// it would make easy to read code by splitting
// validation and evaluation.
let min = args.iter().filter(|arg| arg.default.is_none()).count();
let max = args.len();
if !(min..=max).contains(&exprs.len()) {
    return Err((EvaluateError::FunctionArgsLengthNotWithinRange {
        name: func_name.to_owned(),
        expected_minimum: min,
        expected_maximum: max,
        found: exprs.len(),
    })
    .into());
}

let exprs = exprs.iter().chain(
    args.iter()
        .skip(max - exprs.len())
        .filter_map(|arg| arg.default.as_ref()),
);

let context = stream::iter(args.iter().zip(exprs))
    .then(|(arg, expr)| async move {
        // if user provides NULL explicitly,
        // then NULL should be used, not replaced by default expr.
        eval(expr)
            .await?
            .try_into_value(&arg.data_type, true) // Evaluated::try_into_value
            .map(|value| (arg.name.to_owned(), value))
    })
    .try_collect()
    .await
    .map(|row| {
        let row = Cow::Owned(Row::Map(row));
        let context = RowContext::new(name, row, None);

        Some(Rc::new(context))
    })?;

// function body should not capture outer context so
// aggregated also needs to be `None`
evaluate(storage, context, None, body).await
```

panarch (changes requested):

test-suite/src/function/custom.rs
```diff
@@ -0,0 +1,127 @@
+use {
```
panarch, line 1: Could we move this `function/custom.rs` to the root with a name of `custom_function.rs`?
Because unlike other normal functions, this is special case even with option store trait, so it would be good to emphasis more.

PR #1096, opened 2023-02-18
Support Metadata trait

devgony:
## Goal
- Support Metadata trait
- Metadata uses Schemaless so that storage developer can put anything in it but basically it contains `CREATED` field of Table and Index

## Todo
- [x] Implement Metadata trait
- [x] remove `CREATED` from GLUE_OBJECTS of core
- [x] split test case of `CREATED` to generate_metadata_tests


storages/memory-storage/src/lib.rs
```diff
@@ -68,21 +72,52 @@ impl Store for MemoryStorage {
     }
 }
 
+#[async_trait(?Send)]
+impl Metadata for MemoryStorage {
+    async fn scan_meta(&self) -> HashMap<String, Value> {
```
panarch, line 77:
```rust
async function scan_metadata(&self) -> Result<Box<dyn Iterator<Item = Result<(String, Value)>>>>
```

panarch: it looks that we can simply use `HashMap<String, Value>` rather than `Value`.

storages/memory-storage/src/lib.rs
```diff
@@ -24,10 +26,11 @@ pub struct Item {
     pub rows: BTreeMap<Key, DataRow>,
 }
 
-#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
 pub struct MemoryStorage {
     pub id_counter: i64,
     pub items: HashMap<String, Item>,
+    pub metadata: HashMap<String, Value>,
```
panarch, line 33:
```rust
pub metadata: HashMap<String, HashMap<String, Value>>,
```
this would be more precise..?

test-suite/src/metadata/mod.rs
```diff
@@ -0,0 +1,2 @@
+pub mod table;
+pub use table::table;
```
panarch, line 2: These can provide two different ways to use module.

Like other modules we are using,
ref. https://github.com/gluesql/gluesql/blob/main/test-suite/src/data_type/mod.rs

How about only using `pub mod table;`?

core/src/store/mod.rs
```diff
@@ -1,15 +1,17 @@
...
-pub trait GStore: Store + Index {}
-impl<S: Store + Index> GStore for S {}
+pub trait GStore: Store + Index + Metadata {}
+impl<S: Store + Index + Metadata> GStore for S {}
 
 pub trait GStoreMut: StoreMut + IndexMut + AlterTable + Transaction {}
 impl<S: StoreMut + IndexMut + AlterTable + Transaction> GStoreMut for S {}
 
 pub use {
+    self::metadata::{MetaIter, Metadata},
```
panarch, line 14: Like other modules, `alter_table`, `data_row` .. 
we can omit `self::`

core/src/executor/fetch.rs
```diff
@@ -206,31 +206,40 @@ pub async fn fetch_relation_rows<'a, T: GStore>(
...
+                                ("OBJECT_NAME".to_owned(), Value::Str(schema.table_name)),
+                                ("OBJECT_TYPE".to_owned(), Value::Str("TABLE".to_owned())),
+                            ]);
+
+                            let table_rows = match meta {
+                                Some(Value::Map(meta)) => {
+                                    table_row.into_iter().chain(meta).collect()
+                                }
+                                _ => table_row,
+                            };
```
panarch, line 231: It looks that we can reduce use of `collect`.

```rust
let meta = table_metas
    .iter()
    .find_map(|(table_name, hash_map)| {
        (table_name == &schema.table_name).then(|| hash_map.clone())
    })
    .unwrap_or_default();

let table_rows = [
    ("OBJECT_NAME".to_owned(), Value::Str(schema.table_name)),
    ("OBJECT_TYPE".to_owned(), Value::Str("TABLE".to_owned())),
]
.into_iter()
.chain(meta);
```

storages/memory-storage/src/lib.rs
```diff
@@ -71,18 +75,30 @@ impl Store for MemoryStorage {
 #[async_trait(?Send)]
 impl StoreMut for MemoryStorage {
     async fn insert_schema(&mut self, schema: &Schema) -> Result<()> {
+        let created = HashMap::from([
+            ("OBJECT_TYPE".to_owned(), Value::Str("TABLE".to_owned())),
```
panarch, line 79: This would not be necessary because which the same `OBJECT_TYPE` property is being used in execution layer.

devgony: I removed OBJECT_TYPE in this PR but it may be needed if we implement metadata of index later since only OBJECT_NAME is used as a key here.
Just in case TABLE_NAME and INDEX_NAME are identical (but object_types are different), both OBJECT_NAME and OBJECT_TYPE should be a key

panarch (approved):
This is great!
All looks nice 👍 Thanks a lot 👍 👍

PR #1102, opened 2023-02-18
Replace `TryStream` to `Stream`

ever0de:
While `Stream<Item = Result<T, E>>` implements `TryStream<Ok = T, Error = E>`,
https://docs.rs/futures/latest/futures/prelude/trait.TryStream.html#impl-TryStream-for-S

`TryStream<Ok = T, Error = E>` does not implement `Stream<Item = Result<T, E>>`.
That's why we were sometimes writing `TryStream<Ok=T, Error = E, Item = Result<T, E>>`.
Replace this with.

### DIFF
--- a/core/src/executor/aggregate/mod.rs
+++ b/core/src/executor/aggregate/mod.rs
@@ -11,11 +11,11 @@ use {
     crate::{
         ast::{Expr, SelectItem},
         data::Key,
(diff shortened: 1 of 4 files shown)


panarch (approved):
This is also nice improvement! Thanks a lot 👍

PR #1103, opened 2023-02-18
Replace `Box::pin` to `futures_enum::Stream`

ever0de:
depend on: #1102


panarch (approved):
Also the nice one! Thanks a lot 👍 👍

PR #1105, opened 2023-02-18
Remove `Error::Storage` variant

ever0de:


panarch (approved):
I'm also for this change!
It wasn't worth enough to manage `Error::Storage` which does not impl PartialEq.

Thanks a lot 🥇

PR #1107, opened 2023-02-18
Fix error handling for `ilike` function on `Literal` being treated as…

ever0de:
… `like`


ever0de:
Is this close to a bug?

panarch:
> Is this close to a bug?

very..... small bug may be
I think both ok. it includes both improvement (merge like error) and bug fix (literal ilike handling).

panarch (approved):
Nice catch! Thanks a lot 👍

PR #1117, opened 2023-03-11
Pass data even when `deleted_by` is not present

ever0de:
### Before

```sql
gluesql> create table foo(id int primary key);
Table created

gluesql> insert into foo values(1),(2);
2 rows inserted

gluesql> delete from foo where id =1 ;
1 row deleted

gluesql> insert into foo values(1);
data: SnapshotItem { data: Vec([I64(1)]), created_by: 3, deleted_by: Some(4) }
1 row inserted

gluesql> select * from foo;
| id |
|----|
| 2  |

gluesql> insert into foo values(1);
data: SnapshotItem { data: Vec([I64(1)]), created_by: 3, deleted_by: Some(4) }
1 row inserted

gluesql> select * from foo;
| id |
|----|
| 2  |
```

### DIFF
--- a/storages/sled-storage/src/snapshot.rs
+++ b/storages/sled-storage/src/snapshot.rs
@@ -23,13 +23,13 @@ impl<T: Clone> Snapshot<T> {
     }
 
     pub fn update(mut self, txid: u64, data: T) -> (Self, Option<T>) {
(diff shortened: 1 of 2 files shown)


panarch (approved):
Good fix! Thanks a lot 👍 👍 👍

PR #1118, opened 2023-03-11
Replace closure to variable in `data/interval` module

ever0de:


panarch (approved):
Looks nice to see the good catch :D Thanks a lot!

PR #1119, opened 2023-03-11
fix column alias with identifer

### LINKED ISSUE #1060: Column alias should be supported in identifier
## Description 
`SELECT * FROM (VALUES (1, 'a'), (2, 'b')) AS Derived(id, name)` 
`SELECT id, name FROM (VALUES (1, 'a'), (2, 'b')) AS Derived(id, name)` 
It seems that above queries would produce same result, but below one does not work. 

## To reproduce 
Add below test case to `values.rs`  
```rust 
        (
            "SELECT id, name FROM (VALUES (1, 'a'), (2, 'b')) AS Derived(id, name)",
            Ok(select!(
                id      | name;
                I64     | Str;
                1         "a".to_owned();
                2         "b".to_owned()
            )),
        ),

```

test result
```
thread 'values' panicked at 'assertion failed: `(left == right)`
  left: `Err(Evaluate(ValueNotFound("id")))`,
 right: `Ok(Select { labels: ["id", "name"], rows: [[I64(1), Str("a")], [I64(2), Str("b")]] })`'
```

ding-young:
## Issue 
fix #1060

### DIFF
--- a/core/src/executor/fetch.rs
+++ b/core/src/executor/fetch.rs
@@ -519,7 +519,7 @@ pub async fn fetch_labels<T: GStore>(
 
     let columns = columns.unwrap_or_default();
     let join_columns = join_columns.unwrap_or_default();
-    
+
     projection
         .iter()
         .flat_map(|item| match item {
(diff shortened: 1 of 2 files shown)


panarch (approved):
This looks the best one we can choose for now. looks all good! Thanks a lot 👍

PR #1121, opened 2023-03-11
Keep `Cargo.lock`

### LINKED ISSUE #1120: Keep `Cargo.lock` file
https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html
https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

We currently have a binary project called `gluesql-cli`.
Other libraries say that `Cargo.lock` is ignored by default when adding dependencies.

> For libraries the situation is somewhat different. A library is not only used by the library developers, but also any downstream consumers of the library. Users dependent on the library will not inspect the library’s Cargo.lock (even if it exists). This is precisely because a library should not be deterministically recompiled for all users of the library.

ever0de:
resolved #1120


panarch (approved):
We maintain `cli` binary so this looks reasonable! Thanks a lot 👍

PR #1126, opened 2023-03-25
Simplify JsonlStorage codes

panarch:
1. Simplify SortMerge::next codes
2. Simplify Store::fetch_data impl by replacing from find_map to for loop


ever0de (approved):

devgony (approved):
👍 Incredibly simplified!

PR #1130, opened 2023-03-27
Wrap identifiers with double qoute (`"`) at `to_sql`

devgony:
## Goal
```sql
gluesql> CREATE TABLE "1" (";" INT);
Table created
gluesql> SHOW TABLES;
```
| tables |
|--------|
| 1      |
```sql
gluesql> SHOW COLUMNS FROM "1";
```
| Field | Type |
|-------|------|
| ;     | INT  |

- Currently Table with non-word identifier (like above `Table 1`) can be created but will get `parse error`  at `dump_database` and `from_ddl`(especially at `json_storage`)
- To fix it, this PR will wrap all identifiers(TABLE_NAME, COLUMN_NAME, ALIAS..) with double qoute (`"`) at `to_sql`
## Todo
- [x] wrap all kind of identifiers with `"` at `to_sql`
- [x] correct test cases (QuotedStrings were confused with Identifiers)
- [x] implement  `impl From<Expr> for String` to solve below cases
  - [x] correct SelectItemNode and GLUE_INDEXES to use `expr.into()`
## Next
- fix CTAS to infer types from not only VALUES but also SELECT clause

### DIFF
--- a/core/src/ast/expr.rs
+++ b/core/src/ast/expr.rs
@@ -262,10 +262,11 @@ impl Expr {
 
 #[cfg(test)]
 mod tests {
+
     use {
         crate::ast::{
             AstLiteral, BinaryOperator, DataType, DateTimeField, Expr, Query, Select, SelectItem,
-            SetExpr, TableFactor, TableWithJoins, ToSql, UnaryOperator,
+            SetExpr, TableFactor, TableWithJoins, ToSql, ToSqlUnquoted, UnaryOperator,
         },
         bigdecimal::BigDecimal,
         regex::Regex,
@@ -304,6 +305,15 @@ mod tests {
                 ident: "column".into()
             }
             .to_sql()
+        );
+
+        assert_eq!(
+            "alias.column",
+            Expr::CompoundIdentifier {
+                alias: "alias".into(),
+                ident: "column".into()
+            }
+            .to_sql_unquoted()
         );
 
         let id_expr: Box<Expr> = Box::new(Expr::Identifier("id".to_owned()));
--- a/core/src/ast/expr.rs
+++ b/core/src/ast/expr.rs
@@ -1,7 +1,7 @@
 use {
     super::{
         Aggregate, AstLiteral, BinaryOperator, DataType, DateTimeField, Function, Query, ToSql,
-        UnaryOperator,
+        ToSqlUnquoted, UnaryOperator,
     },
     serde::{Deserialize, Serialize},
 };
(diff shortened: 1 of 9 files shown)


core/src/ast/expr.rs
```diff
@@ -221,6 +221,159 @@ impl ToSql for Expr {
     }
 }
 
+impl From<Expr> for String {
```
devgony, line 224: From now, to_sql will have two usages:
1. To dump DDL with `to_ddl`
2. To use as label (`SelectItemNode`) or to just show expr at `GLUE_INDEXES`

I'm not sure implementing `From<Expr> for String` again is best or not.
So many lines are duplicated and this way still have potential problem cuz it uses `to_sql` as well at some part (like BinOperator or Query)

panarch: This looks interesting.
How about adding an extra function that can be shared for both?
We can also consider adding `to_ddl` function.

```rust
impl Expr {
    fn to_ddl(&self) -> String {
        ..
    }
}

// and shared to_..sql(&self, ident_type?) function which is used by both `ToSql::to_sql` and `Expr::to_ddl`
```

devgony: That sounds good idea! and we don't even need `to_ddl` if shared `to_string_with(&self, qouted: bool)` takes `qoute` to match each cases.
But it is hard to decide at which level to distinguish each cases.
Once method is decided between qouted or not, it should be kept to use equal method due to recursion.
Looks like eventually it will withdraw similar duplication
What about adding `qoute: bool`  to `ToSql::to_sql()`?
```rs
impl Expr {
▏   fn to_string_with(&self, qouted: bool) -> String { 
▏   ▏   match self {
▏   ▏   ▏   Expr::Identifier(s) => match qouted {
▏   ▏   ▏   ▏   true => format! {r#""{s}""#},
▏   ▏   ▏   ▏   false => self.to_sql(),
▏   ▏   ▏   },
▏   ▏   ▏   Expr::BinaryOp { left, op, right } => format!(
▏   ▏   ▏   ▏   "{} {} {}",
▏   ▏   ▏   ▏   left.to_string_with(qouted),
▏   ▏   ▏   ▏   op.to_sql(),
▏   ▏   ▏   ▏   right.to_string_with(qouted)
▏   ▏   ▏   ),
▏   ▏   ▏   Expr::CompoundIdentifier { alias, ident } => match qouted {
▏   ▏   ▏   ▏   true => format!(r#""{alias}"."{ident}""#),
▏   ▏   ▏   ▏   false => self.to_sql(),
▏   ▏   ▏   },
▏   ▏   ▏   Expr::IsNull(s) => format!("{} IS NULL", s.to_string_with(qouted)),
            // .. so many duplications
▏   ▏   }
▏   }
```

core/src/ast/query.rs
```diff
@@ -1,5 +1,5 @@
 use {
-    super::{Expr, IndexOperator},
+    super::{Expr, IndexOperator, ToSqlUnquoted},
```
panarch, line 2: In `ast/query.rs`, `ToSqlUnquoted` impl for `IndexItem` is missing.
It looks that only except `IndexItem`, all looks well implemented.

Could you also add some unit tests for new added lines?

panarch (approved):
Pleased to see this meaningful improvement 👍 
All looks nice! Thanks a lot 👍 👍

PR #1131, opened 2023-03-29
Remove memory-storage dep from JsonStorage/ Cargo.toml

panarch:


devgony (approved):
Sorry for my mistake 😭
It was added when we tried to implement Store with MemoryStorage

panarch:
> Sorry for my mistake 😭 It was added when we tried to implement Store with MemoryStorage

No problem at all :D I just found so did quick fix

PR #1133, opened 2023-03-30
fix: infer types from select clause in `CTAS`

devgony:
## Goal
```sql
CREATE TABLE Target AS SELECT Source.*, 'literal', .. FROM Source
```
Currently, This SQL copies only Source table's schema to Target table, not including 'literal' columns 
Therefore we need to iterate over result of SELECT query and infer the Target schema column types

### DIFF
--- a/core/src/executor/alter/table.rs
+++ b/core/src/executor/alter/table.rs
@@ -1,3 +1,6 @@
+use crate::{ast::SelectItem, executor::select::select_with_labels};
+use futures::StreamExt;
+
 use {
     super::{validate, validate_column_names, AlterError},
     crate::{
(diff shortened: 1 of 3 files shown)


core/src/executor/insert.rs
```diff
@@ -40,29 +40,29 @@ pub enum InsertError {
...
     Append(Vec<DataRow>),
     Insert(Vec<(Key, DataRow)>),
 }
 
+pub async fn fetch_insert_rows<T: GStore + GStoreMut>(
+    storage: &mut T,
+    table_name: Option<&str>,
+    columns: &[String],
+    query: &Query,
+    column_defs: Option<&Vec<ColumnDef>>,
```
panarch, line 53: We can use `column_defs: Option<&[ColumnDef]>` in here.
Then, we can also replace `as_ref` to `as_deref` for the codes which use this.

core/src/executor/insert.rs
```diff
@@ -85,8 +85,8 @@ pub async fn insert<T: GStore + GStoreMut>(
 
 async fn fetch_vec_rows<T: GStore>(
     storage: &T,
-    table_name: &str,
-    column_defs: Vec<ColumnDef>,
+    table_name: Option<&str>,
+    column_defs: &Vec<ColumnDef>,
```
panarch, line 89: Here, `column_defs: &[ColumnDef]` would be also enough.

PR #1134, opened 2023-04-02
Fix/remove unnecessary UUID test case

### LINKED ISSUE #1129: Fix misleadingmethod for type casting of UUID when its string
During my work on implementing the unsigned integer, I came across a conversion function that converts a string data type to a UUID data type. However, as we are now focusing on providing support for the u128 data type, I believe that converting string to u128 type with using parse_uuid should be deprecated.

AS-IS
```rust
impl TryFrom<&Value> for u128 {
    type Error = Error;

    fn try_from(v: &Value) -> Result<u128> {
        match v {
            Value::Uuid(value) => Ok(*value),
            Value::Str(value) => parse_uuid(value),
            Value::Inet(IpAddr::V6(v)) => Ok(u128::from(*v)),
            _ => Err(ValueError::ImpossibleCast.into()),
        }
    }
}
```

TO-BE
```rust
impl TryFrom<&Value> for u128 {
    type Error = Error;

    fn try_from(v: &Value) -> Result<u128> {
        match v {
            Value::Uuid(value) => Ok(*value),
            Value::Inet(IpAddr::V6(v)) => Ok(u128::from(*v)),
            Value::Str(value) => u128::from_str_radix(value, 10).map_err(|_| ValueError::FailedToParseNumber.into()),
            _ => Err(ValueError::ImpossibleCast.into()),
        }
    }
}
```

ChobobDev:
resolves #1129

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -907,7 +907,6 @@ dependencies = [
  "gluesql-core",
  "gluesql-test-suite",
  "gluesql-utils",
- "gluesql_memory_storage",
  "hex",
  "iter-enum",
  "serde",
--- a/core/src/data/value/convert.rs
+++ b/core/src/data/value/convert.rs
@@ -542,10 +542,9 @@ impl TryFrom<&Value> for u128 {
     fn try_from(v: &Value) -> Result<u128> {
         match v {
             Value::Uuid(value) => Ok(*value),
-             Value::Str(value) => value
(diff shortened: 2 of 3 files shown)


ChobobDev:
I think I am ready for the review :)

panarch (approved):
Looks all nice!!
Thanks a lot :D

PR #1135, opened 2023-04-02
Add JsonStorage support to CLI

devgony:
## Goal
Support JsonStorage on CLI

## Todo
- [x] JsonStorage on CLI
- [x] enhance clap to use ValueEnum 

## Next
- enhance `dump_database` to support all storages
  - create `trait Dump`?

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -19,8 +19,7 @@ use {
     gluesql_memory_storage::MemoryStorage,
     gluesql_sled_storage::SledStorage,
     itertools::Itertools,
-    std::{fmt::Debug, fs::File, io::Write, path::PathBuf, str::FromStr},
-    strum_macros::EnumString,
+    std::{fmt::Debug, fs::File, io::Write, path::PathBuf},
 };
 
 #[derive(Parser, Debug)]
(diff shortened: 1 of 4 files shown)


cli/src/lib.rs
```diff
@@ -24,40 +25,72 @@ use {
...
+#[derive(clap::ValueEnum, Debug, Clone)]
+enum Storage {
+    Memory,
+    Sled,
+    Json,
 }
 
 pub fn run() -> Result<()> {
     let args = Args::parse();
+    let (path, storage) = match (args.path, args.storage) {
```
panarch, line 54: It looks that two separate `match` statements can be merged into a single one.

```rust
let path = args
    .path
    .as_ref()
    .map(PathBuf::as_path)
    .and_then(Path::to_str);

match (path, args.storage) {
    (None, None) | (None, Some(Storage::Memory)) => {
        println!("[memory-storage] initialized");

        run(MemoryStorage::default(), args.execute);
    }
    (Some(_), Some(Storage::Memory)) => {
        panic!("failed to load memory-storage: it should be without path");
    }
    (Some(path), Some(Storage::Sled)) => {
        println!("[sled-storage] connected to {}", path);

        run(
            SledStorage::new(path).expect("failed to load sled-storage"),
            args.execute,
        );

        if let Some(dump_path) = args.dump {
            let mut storage = SledStorage::new(path).expect("failed to load sled-storage");
            dump_database(&mut storage, dump_path)?;
        }
    }
    (Some(path), Some(Storage::Json)) => {
        println!("[json-storage] connected to {}", path);

        run(
            JsonStorage::new(path).expect("failed to load json-storage"),
            args.execute,
        );
    }
    (None, Some(_)) | (Some(_), None) => {
        panic!("both path and storage should be specified");
    }
}
```

devgony: Thanks!

I realized current logic does not make dump_database works so that i used `match with 3 elements` including `args.dump`

panarch (approved):
Nice improvement!
Looks all good :D Thanks a lot 👍

Issue #439, opened 2021-12-15
Let's move the const in the file into the struct according to the usage


ever0de:
## Before

https://github.com/gluesql/gluesql/blob/3776b1b59fa9922a458bd34847277efd8c965a1e/src/storages/sled_storage/metadata.rs#L14

## After

```rust
impl SledStorage {
    const SCHEMA_PREFIX: &'static str = "schema/";

// use case
SledStorage::SCHEMA_PREFIX
```

## Effect

Reuse is a bit easier, and the meaning becomes clearer.

PR #1151, opened 2023-04-22
Move the SCHEMA_PREFIX const into an impl in SledStorage

garypen:
fixes: #439


panarch (approved):
Looks good! Thanks a lot 👍

Issue #1152, opened 2023-04-23
Implement CALC_DISTANCE in AST For Geometic Type


seonghun-dev:
# Description
In two Point Calculate two point distance.

## CALC_DISTANCE

`CALC_DISTANCE`

calc_distance is a function used in spatial analysis to calculate the distance between two geometries. It is commonly used in geographic information systems and spatial databases to determine the proximity of objects and perform various spatial analyses. The function takes two geometries as inputs and returns the shortest distance between them. It can operate on several geometry types, including points, lines, polygons, and multi-geometries. The result of calc_distance is expressed in the same unit of measurement as the input geometries, making it easy to compare distances between different objects.

### Example
Function `CALC_DISTANCE`

```sql
SELECT CALC_DISTANCE(point1, point2) As dis FROM Foo;
```

# Task
- [ ] Add `CALC_DISTANCE` Function

The following tests should be passed:

``` rust
test_case!(calc_distance, async move {
    run!(
        "
        CREATE TABLE Foo (
            id INTEGER,
            geo1 POINT,
            geo2 POINT
        );
    "
    );
    run!(
         r#"
            INSERT INTO Prepend VALUES
            (1, POINT(0.3134, 3.156), POINT(1.415, 3.231));
        "#
    );
    test!(
        r#"select calc_distance(geo1, geo2) as georesult from Foo"#,
        Ok(select!(
           georesult
           F64;
           1.10415_f64
        ))
    );
});
```

### Reference
https://postgis.net/docs/ST_Distance.html

PR #1153, opened 2023-04-23
Feat : add calc_distance function

seonghun-dev:
resolve :  https://github.com/gluesql/gluesql/issues/1152


test-suite/src/function/geometry/calc_distance.rs
```diff
@@ -0,0 +1,49 @@
+use gluesql_core::translate::TranslateError;
+
+use {
```
panarch, line 3: could you merge this?

panarch (approved):

PR #1162, opened 2023-04-30
Implement optional storage for the execution layer

### LINKED ISSUE #1160: Make all storage param as optional in execution layer
In the context of the SQL Database project, we have recently merged `evaluate` and `evaluate_stateless` as per [#1132](https://github.com/gluesql/gluesql/pull/1132). The distinction between `evaluate` and `evaluate_stateless` in the execution layer now depends on whether the storage parameter is present or not.

However, aside from evaluate, all other code requires the storage parameter mandatorily. This design is currently limiting because it prevents stateless evaluation for queries that don't need storage. For instance, when creating a subquery using values such as `VALUES (1), (2)`, the query doesn't utilize storage, and thus, should support stateless evaluation. With the current structure, it's impossible as the function always demands storage.

To address this issue, I propose we change all functions in the execution layer that require storage as a parameter. We should make the storage parameter optional (`Option`). This change will increase flexibility and accommodate for stateless evaluation where necessary.

pythonbrad:
resolve #1160

### DIFF
--- a/core/src/executor/alter/table.rs
+++ b/core/src/executor/alter/table.rs
@@ -12,13 +12,15 @@ use {
 };
 
 pub async fn create_table<T: GStore + GStoreMut>(
-    storage: &mut T,
+    storage: Option<&mut T>,
     target_table_name: &str,
     column_defs: Option<&[ColumnDef]>,
     if_not_exists: bool,
     source: &Option<Box<Query>>,
     engine: &Option<String>,
 ) -> Result<()> {
+    let storage = storage.ok_or(AlterError::StatelessOperation)?;
+
     let target_columns_defs = match source.as_deref() {
         Some(Query { body, .. }) => match body {
             SetExpr::Select(select_query) => match &select_query.from.relation {
@@ -115,7 +117,7 @@ pub async fn create_table<T: GStore + GStoreMut>(
 
     match source {
         Some(query) => {
-            let rows = select(storage, query, None)
+            let rows = select(Some(storage), query, None)
                 .await?
                 .map_ok(Into::into)
                 .try_collect()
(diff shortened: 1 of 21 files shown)


pythonbrad:
@panarch , i tried to do step by step, then each commit can be compile successfully

I would like to explain an error that occurred in this pull request. While attempting to compile this last commit, I encountered the following error: "expected enum `std::option::Option<&T>` found enum `std::option::Option<&mut T>`."

It appears that I passed an `Option<mut &T>` from one function's signature to another function that only accepted `Option<&T>`. However, in the previous commit, changing the storage type from `&mut T` to `Option<_>`, before give to another function with signature `Option<&T>` was Ok. Maybe the `mut` cannot be removed indirectly 🤔 .

I think that i can't bypass it, because of the Store mutability (StoreMut).

core/src/executor/alter/table.rs
```diff
@@ -23,7 +23,7 @@ pub async fn create_table<T: GStore + GStoreMut>(
         Some(Query { body, .. }) => match body {
             SetExpr::Select(select_query) => match &select_query.from.relation {
                 TableFactor::Table { name, .. } => {
-                    let schema = storage.fetch_schema(name).await?;
+                    let schema = storage.ok_or(AlterError::AlterUnsupported)?.fetch_schema(name).await?;
```
panarch, line 26: `storage.ok_or(AlterError::AlterUnsupported)?` codes takes ownership of storage. Then it means that we cannot pass that to `select` which requires `Option<&T>`.

We can unwrap Option in the beginning.
```rust
let storage = storage.ok_or(AlterError::AlterUnsupported)?;
```

Then for `select`
```rust
let rows = select(Some(storage), query, None)
 ....
```

will work.

core/src/executor/execute.rs
```diff
@@ -89,9 +93,11 @@ pub async fn execute<T: GStore + GStoreMut>(
 }
 
 async fn execute_inner<T: GStore + GStoreMut>(
-    storage: &mut T,
+    storage: Option<&mut T>,
     statement: &Statement,
 ) -> Result<Payload> {
+    let storage = storage.ok_or(ExecuteError::StatelessOperation)?;
```
panarch, line 99: The purpose of making the execution layer functions to take `storage` param as optional is for handling `None` case.
Then, we can execute `SELECT 1;` like queries which does not require storage by not using storage.
However, in this case, because
```rust
let storage = storage.ok_or(ExecuteError::StatelessOperation)?;
```
the above line requires `storage` not to be optional. we can't still handle `SELECT 1` without storage.

if the entire function content requires `storage` then we don't need to make the function `storage` param optional.
once we make the storage optional, then it means that we also have some cases which `None` works for that.

PR #1163, opened 2023-05-01
Add docusaurus deployment github action setup

panarch:
this action deploys `gluesql.org/docs/dev` when main is updated.


devgony (approved):
👍🏼

PR #1166, opened 2023-05-02
Add SQL function list with categories to docs/

panarch:
Add the list of SQL functions with categories to the docs/ directory. The SQL functions are currently listed without details, and the categories are Text, Math, Date & Time, List & Map, Geometry, and Others. This commit lays the groundwork for documenting each SQL function in detail.

image


ever0de (changes requested):
PI, POINT functions are missing.

ever0de (approved):

PR #1169, opened 2023-05-02
Add DESCRIBE statement

### LINKED ISSUE #1113: Support DESCRIBE statement
https://jakewheat.github.io/sql-overview/sql-2016-foundation-grammar.html#_20_10_describe_statement

The describe statement is particularly useful for finding out information about a table without digging in deep.

pythonbrad:
resolve #1113


pkg/javascript/web/src/payload.rs
```diff
@@ -80,6 +80,26 @@ fn convert_payload(payload: Payload) -> Json {
                 "columns": Json::Array(columns),
             })
         }
+        Payload::ExplainTable(columns) => {
+            let columns = columns
+                .into_iter()
+                .map(|(field, data_type, nullable, key, default, extra)| {
```
devgony, line 86: Currently we don't have any fields to show on `extra`.
Looks like it would be better to remove `extra`.

core/src/executor/execute.rs
```diff
@@ -256,6 +257,29 @@ async fn execute_inner<T: GStore + GStoreMut>(
 
             Ok(Payload::ShowColumns(output))
         }
+        Statement::ExplainTable { table_name } => {
+            let Schema { column_defs, .. } = storage
+                .fetch_schema(table_name)
+                .await?
+                .ok_or_else(|| ExecuteError::TableNotFound(table_name.to_owned()))?;
+
+            let output: Vec<(String, DataType, bool, String, String, String)> = column_defs
```
devgony, line 266: It would be better to wrap verbose tuple with struct like..
```rs
struct ExplainTableRow {
   name: String,
   data_type: DataType,
   nullable: bool,
   key: String,
   default: String,
}
```

pythonbrad:
@devgony , I applied your reviews

devgony:
Thank you for quick commit.
Can you do the `cargo fmt` and push again?

pythonbrad:
> Thank you for quick commit.
> 
> Can you do the `cargo fmt` and push again?

Please, can you check the `web/src/payload.rs` ? The json result

devgony:
> > Thank you for quick commit.
> > Can you do the `cargo fmt` and push again?
> 
> Please, can you check the `pkg/javascript/web/src/payload.rs` ? The json result

It seems like working well on web as well 👍

image

devgony (approved):

pythonbrad:
hi @panarch 
is the pr ok ?

PR #1174, opened 2023-05-05
Refactor SQL function tests in test-suite,

panarch:
This commit includes the following refactoring changes to the SQL function tests in the GlueSQL test-suite:

Removed unnecessary CREATE TABLE statements
Fixed tabular alignment issues in expected values
Simplified numerical calculations in math functions


ever0de (approved):
There are some SQL statements in which semicolons are missing at the end.
(If this doesn't matter, please ignore the comment)

PR #1177, opened 2023-05-06
Add nullable, key, default to `GLUE_TABLE_COLUMNS`

devgony:
## Goal
```sql
gluesql> CREATE TABLE T1 (id INT PRIMARY KEY, name TEXT NULL);
Table created

gluesql> CREATE TABLE T2 (id INT UNIQUE DEFAULT 0 NULL);
Table created

gluesql> SELECT * FROM GLUE_TABLE_COLUMNS;
```
gluesql> SELECT * FROM GLUE_TABLE_COLUMNS;
| TABLE_NAME | COLUMN_NAME | COLUMN_ID | NULLABLE | KEY         | DEFAULT |
|------------|-------------|-----------|----------|-------------|---------|
| T1         | id          | 1         | FALSE    | PRIMARY KEY |         |
| T1         | name        | 2         | TRUE     |             |         |
| T2         | id          | 1         | TRUE     | UNIQUE      | 0       |

## Todo
- [x] add nullable, key, default to `GLUE_TABLE_COLUMNS`

### DIFF
--- a/core/src/executor/fetch.rs
+++ b/core/src/executor/fetch.rs
@@ -1,11 +1,13 @@
-use crate::ast::ToSql;
-
 use {
     super::{context::RowContext, evaluate::evaluate_stateless, filter::check_expr},
     crate::{
         ast::{
-            ColumnDef, ColumnUniqueOption, Dictionary, Expr, IndexItem, Join, Query, Select,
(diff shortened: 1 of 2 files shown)


panarch (approved):
Looks all nice!
Thanks a lot 👍

PR #1185, opened 2023-05-13
Fix InvalidJsonString error message replacing payload to fileName

devgony:
## Todo
- [x] Fix InvalidJsonString error message replacing payload to fileName
- [x] rename InvalidJsonString to InvalidJsonContent


panarch (approved):
Looks all nice! Thanks for the quick fix 👍

PR #1189, opened 2023-05-13
Add math conversion test case for ast_builder

changi1122:
# Goal

- Add test cases about "DEGREES", "RADIANS" for AST Builder

# Todo

- [x]  Add test cases about "DEGREES", "RADIANS" for AST Builder


panarch (approved):
Looks all nice!! Thanks a lot 👍 👍

PR #1193, opened 2023-05-13
Add Testcase/case conversion

sooyeonyim-t:
- some test code of case conversion(upper, lower, initCapped)


panarch (approved):

PR #1198, opened 2023-05-14
Add testcase/formatting for ast_builder

sooyeonyim-t:
- some test code of datetime formatting(date, time, timestamp) for ast_builder


test-suite/src/ast_builder/function/datetime/formatting.rs
```diff
@@ -0,0 +1,96 @@
+use chrono::{NaiveDate, NaiveTime};
```
devgony, line 1: It would be better to merge imports

test-suite/src/ast_builder/function/datetime/mod.rs
```diff
@@ -1,3 +1,5 @@
-mod conversion;
-
-pub use conversion::conversion;
+mod conversion;
+mod formatting;
+
+pub use conversion::conversion;
+pub use formatting::formatting;
```
devgony, line 5: you can merge `pub use` as well

panarch (approved):
Looks all nice! Thanks a lot 👍 🔢

PR #1201, opened 2023-05-14
Add math basic arithmetic test case for ast_builder

changi1122:
## Goal

Add math basic arithmetic test case about "ABS", "DIV", "MOD", "GCD", "LCM" for ast_builder


panarch (approved):

PR #1202, opened 2023-05-14
Update math basic arithmetic docs for ast_builder

changi1122:
## Goal

Update math basic arithmetic docs for ast_builder


panarch (approved):
Thanks a lot!!

PR #1210, opened 2023-05-17
Remove pkg/js rollup custom build,

panarch:
Simplify pkg/javascript build settings - remove rollup custom build.
Update corresponding docs - getting-started/javascript-web.md
Update pkg/javascript/ rollup examples.
Update javascript.yml GitHub Action


.github/workflows/javascript.yml
```diff
@@ -28,9 +28,10 @@ jobs:
         uses: jetli/wasm-pack-action@v0.3.0
       - run: |
           cd pkg/javascript/web
-          yarn install
-          yarn build
-          yarn test:firefox
+          npm install
```
ever0de, line 31: Is there a reason the package manager changed?

panarch: our current directory shape (`/pkg/javascript` includes `examples` and `package.json` in `examples/` uses `/pkg/javascript`) causes something.. like infinite loop loading when I use `yarn` but `npm` does not make any that kind of noise.
That was the main reason.
and another small one is `npm` is default tool but `yarn` is a tool that requires a single more step to install it. so if there's no big different between them, then it would be good enough to use default installed `npm`.

panarch: Below is revised by GPT4 :D

The primary reason for the change from yarn to npm is that our current directory structure (/pkg/javascript, which includes examples, and package.json in examples/ uses /pkg/javascript) leads to a kind of infinite loop loading when yarn is used. This issue was not encountered with npm.

Additionally, npm is the default package manager and doesn't require an extra installation step like yarn does. So unless there's a significant advantage to using yarn, sticking with the default npm is a more straightforward choice.

ever0de (approved):

PR #1211, opened 2023-05-19
Add a case for insert with source

devgony:
## Goal 
Enhance test cases of `insert`

## Todo
- [x] Add a case for insert with source


panarch (approved):
Thanks a lot 👍

PR #1212, opened 2023-05-19
Setup blog based on docusaurus,

panarch:
Update docusaurus.config.js to handle both docs/ and blog/ by using GLUESQL_DOC_TYPE env variable.
Add test blog article.

blog/
image

docs/ - blog link on the top-right side
image

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -71,7 +71,18 @@ const config = {
     navbar: {
       title: 'GlueSQL',
       items: [
-        ...(!isBlog ? [
+        ...(isBlog ? [
+          {
+            to: '/',
+            label: 'Blog',
+            position: 'left',
+          },
+          {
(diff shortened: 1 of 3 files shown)


docs/docusaurus.config.js
```diff
@@ -57,30 +71,48 @@ const config = {
...
-          position: 'left',
-          label: 'AST Builder',
-        },
-        {
-          type: 'doc',
-          docId: 'storages/intro',
-          position: 'left',
-          label: 'Storages',
-        },
+        ...(!isBlog ? [
```
ever0de, line 74: Unless there's a good reason not to, why not use a conditional like the one on L42?

panarch: though there's no special reason for that but L42 is for making js object and here is for array so I just intended to make it to look more simple.

ever0de (approved):

Issue #1220, opened 2023-05-26
Support unnesting List and Array in CLI


devgony:
## Goal
Currently in CLI, MAP or List types show like this.
```
gluesql> CREATE TABLE T (map MAP, list LIST);
Table created
gluesql> INSERT INTO T VALUES('{"a": 1, "b": 2}', '[1, 2]');
1 row inserted

gluesql> SELECT * FROM T;
```
| map   | list   |
|-------|--------|
| [MAP] | [LIST] |

Let's unnest result like this.
| map   | list   |
|-------|--------|
| {"a": 1, "b": 2} | [1, 2] |

PR #1226, opened 2023-05-27
feat: fmt list and map

Mehrbod2002:
fixes: #1220 
fix fmt on list and map Value and tests added

### DIFF
--- a/core/src/data/value/convert.rs
+++ b/core/src/data/value/convert.rs
@@ -871,7 +871,7 @@ mod tests {
         let mut map = HashMap::new();
         map.insert("a".to_owned(), Value::I32(1));
         map.insert("b".to_owned(), Value::I32(2));
-        test!(Value::Map(map), "{\"a\":1,\"b\":2}");
+        test!(Value::Map(map), "{\"b\":2,\"a\":1}");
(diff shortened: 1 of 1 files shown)


Mehrbod2002:
@ever0de Look out for the tests's result .

Mehrbod2002:
@ever0de Sometime HashMap switch the entries

Mehrbod2002:
@ever0de Please re-run the workflows

panarch (changes requested):
Oh sorry for the late review.

the current code will generate
```rust
let actual = Value::List(vec![Value::I32(1), Value::I32(2), Value::Str("hello".to_owned())]);
let expected = "[1,2,hello]";
```
that the string "hello" is not wrapped by `"`.

to fix this, we actually don't need to implement stringify by yourself.
we can use it from serde.
https://github.com/gluesql/gluesql/blob/0e13939662f6acac7091e96943ed93fed3f05e5d/core/src/data/value/json.rs#L53-L58

and that `serde_json::Value` implements `Display` so we can use it.

Mehrbod2002:
@panarch you can review it .

panarch (approved):
Looks all nice 👍 
Thanks a lot for the contribution!

PR #1233, opened 2023-06-02
Split custom Partial{Eq|Ord} impl of Value & Literal into evaluate_{eq|cmp}

### LINKED ISSUE #1157: Value PartialEq lacks support for Null.
When comparing Same Payload, there will be inequality when there is Value::Null  in the row.
```rust
                       let c_payload = Payload::Select {
                            labels: a_labels.clone(),
                            rows: vec![Row(a_values.clone())],
                        };
                        let c1_payload = Payload::Select {
                            labels: a_labels.clone(),
                            rows: vec![Row(a_values.clone())],
                        };
                        println!("c_payload:{:?}",c_payload);
                        let bool = c1_payload == c_payload;
                        println!("bool:{:?}",bool);
```
```shell
c_payload:Select { labels: ["role0id", "role0role_name", "role0remark", "role0create_time", "role0modify_time", "role0create_user_id", "role0modify_user_id", "role0status"], rows: [Row([I64(1), Str("test1"), Null, Null, Null, Null, Null, Null])] }
c1_payload:Select { labels: ["role0id", "role0role_name", "role0remark", "role0create_time", "role0modify_time", "role0create_user_id", "role0modify_user_id", "role0status"], rows: [Row([I64(1), Str("test1"), Null, Null, Null, Null, Null, Null])] }
bool:false
```
Could it be my operational problem?

### LINKED ISSUE #1158: Separating PartialEq Implementation for Value::Null in Payload Comparison for SQL Evaluation
When comparing two Payloads with the same values, the comparison returns false if either of the rows has a Value::Null. This is because SQL treats NULL differently from other values. While this behavior is intentional and reflects SQL's evaluation rules, it can cause inconvenience in practice.

Currently, we are using a custom implementation of PartialEq for comparing instances of Value enum. However, we plan to separate this implementation into a separate method that is specific to SQL evaluation. This will allow us to implement PartialEq using the default implementation, without any customizations.

Once we separate NULL handling in evaluation, the following assertion will work as expected:
```rust
assert_eq!(Value::Null, Value::Null);
```

panarch:
…ate_eq,

Update Value & Literal to use default derived PartialEq impl. Replace previous custom PartialEq impl to evaluate_eq. For evaluation (in Evaluated, ..), rather than using PartialEq, update all to use evaluate_eq method.

resolve #1157 
resolve #1158 

```rust
assert_eq!(Value::Null, Value::Null);
assert_eq!(Value::Null.evaluate_eq(&Value::Null), false);

assert_eq!(Literal::Null, Literal::Null);
assert_eq!(Literal::Null.evaluate_eq(&Literal::Null), false);
```


core/src/data/literal.rs
```diff
@@ -96,31 +96,27 @@ impl<T: AsRef<str>> PartialOrd<T> for Literal<'_> {
...
+            (Null, Null) => Some(Ordering::Equal),
             _ => None,
         }
     }
 }
 
 impl<'a> Literal<'a> {
+    pub fn evaluate_eq(&self, other: &Literal<'_>) -> bool {
+        match (self, other) {
+            (Null, Null) => true,
```
ever0de, line 115: The `evaluate_eq` should have the same implementation as the previous `PartialEq`.

```suggestion
            (Null, Null) => false,
```

test-suite/src/ast_builder/function/text/position_and_indexing.rs
```diff
@@ -18,7 +18,7 @@ test_case!(position_and_indexing, async move {
     let actual = table("Item")
         .create_table()
         .add_column("id INTEGER PRIMARY KEY")
-        .add_column("index TEXT")
+        .add_column("index INTEGER")
```
ever0de, line 21: Since this is a test module for text-related functions, it would make sense to either make it text, or change the module location.

ever0de (approved):

Issue #1011, opened 2022-11-19
Support `ESCAPE` keyword in WHERE LIKE clause


devgony:
## Goal
```sql
SELECT * FROM Items
WHERE name LIKE 'GLUE_SQL'
```
| id | name     |
|----|----------|
| 1  | GLUE+SQL |
| 2  | GLUE_SQL |
```sql
SELECT * FROM Items
WHERE name LIKE 'GLUE#_SQL' ESCAPE '#'
```
| id | name     |
|----|----------|
| 2  | GLUE_SQL |

=> `#` can escape reserved character `_`

## Ref
https://dev.mysql.com/doc/refman/8.0/en/string-comparison-functions.html#operator_like

Sung-Heon:
May I try this?

panarch:
> May I try this?

Sure you can, thanks :)

PR #1237, opened 2023-06-05
feat: add escape-char to like()

Mehrbod2002:
Fixes : #1011 
add ESCAPE option to `like` command .

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -4,11 +4,11 @@ on:
   push:
     branches: [main, release-*]
     paths-ignore:
-      - 'docs/**'
+      - "docs/**"
   pull_request:
     branches: [main, release-*]
     paths-ignore:
-      - 'docs/**'
+      - "docs/**"
 
 env:
   CARGO_TERM_COLOR: always
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -4,11 +4,11 @@ on:
   push:
     branches: [main, release-*]
     paths-ignore:
-      - 'docs/**'
+      - "docs/**"
   pull_request:
     branches: [main, release-*]
     paths-ignore:
-      - 'docs/**'
+      - "docs/**"
 
 env:
   CARGO_TERM_COLOR: always
@@ -61,6 +61,8 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: Swatinem/rust-cache@v2
+      - name: MongoDB in GitHub Actions
+        uses: supercharge/mongodb-github-action@v1.10.0
       - name: Redis in GitHub Actions
         uses: supercharge/redis-github-action@1.7.0
         with:
@@ -82,4 +84,4 @@ jobs:
           cargo run --package gluesql --example memory_storage_usage
           cargo run --package gluesql --example sled_multi_threaded
           cargo run --package gluesql --example using_config
-          cargo run --package gluesql --example hello_ast_builder 
+          cargo run --package gluesql --example hello_ast_builder
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -35,6 +35,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "91429305e9f0a25f6205c5b8e0d2db09e0708a7a6df0f42212bb56c32c8ac97a"
 dependencies = [
  "cfg-if",
+ "getrandom",
  "once_cell",
  "version_check",
  "zerocopy",
@@ -92,7 +93,7 @@ dependencies = [
  "polling",
  "rustix 0.37.27",
  "slab",
- "socket2",
+ "socket2 0.4.10",
  "waker-fn",
 ]
 
@@ -158,6 +159,18 @@ dependencies = [
  "object",
  "rustc-demangle",
 ]
+
+[[package]]
+name = "base64"
+version = "0.13.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9e1b586273c5702936fe7b7d6896644d8be71e6314cfe09d3167c95f712589e8"
+
+[[package]]
+name = "base64"
+version = "0.21.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "35636a1494ede3b646cc98f74f8e62c773a38a659ebc777a2cf26b9b74171df9"
 
 [[package]]
 name = "bigdecimal"
@@ -267,6 +280,28 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 1.0.109",
+]
+
+[[package]]
+name = "bson"
+version = "2.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "58da0ae1e701ea752cc46c1bb9f39d5ecefc7395c3ecd526261a566d4f16e0c2"
+dependencies = [
+ "ahash 0.8.6",
+ "base64 0.13.1",
+ "bitvec",
+ "chrono",
+ "hex",
+ "indexmap 1.9.3",
+ "js-sys",
+ "once_cell",
+ "rand",
+ "serde",
+ "serde_bytes",
+ "serde_json",
+ "time",
+ "uuid",
 ]
 
 [[package]]
@@ -442,10 +477,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "convert_case"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6245d59a3e82a7fc217c5828a6692dbc6dfb63a0c8c90495621f7b9d79704a0e"
+
+[[package]]
 name = "core-foundation-sys"
 version = "0.8.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e496a50fda8aacccc86d7529e2c1e0892dbd0f898a6b5645b5561b89c3210efa"
+
+[[package]]
+name = "cpufeatures"
+version = "0.2.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ce420fe07aecd3e67c5f910618fe65e94158f6dcc0adf44e00d69ce2bdfe0fd0"
+dependencies = [
+ "libc",
+]
 
 [[package]]
 name = "crc32fast"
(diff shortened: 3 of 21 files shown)


core/src/ast/expr.rs
```diff
@@ -35,11 +35,13 @@ pub enum Expr {
         expr: Box<Expr>,
         negated: bool,
         pattern: Box<Expr>,
+        escape_char: Option<char>,
```
devgony, line 38: I thought it would be better to follow mysql's way which sets default value of escape_char as `\` so that it does not need to be `Option`

Cuz `\` is the most commonly used.

> To test for literal instances of a wildcard character, precede it by the escape character. If you do not specify the ESCAPE character, \ is assumed

https://dev.mysql.com/doc/refman/8.0/en/string-comparison-functions.html#operator_like

Mehrbod2002: @devgony 
The changes pushed.

panarch:
> @panarch what do you think?

It took me some more time to do look around, I'll review and leave comments till tomorrow.

core/src/translate/expr.rs
```diff
@@ -73,21 +73,31 @@ pub fn translate_expr(sql_expr: &SqlExpr) -> Result<Expr> {
...
+            escape_char,
         } => Ok(Expr::Like {
             expr: translate_expr(expr).map(Box::new)?,
             negated: *negated,
             pattern: translate_expr(pattern).map(Box::new)?,
+            escape_char: if let Some(escape) = escape_char {
+                *escape
+            } else {
+                '\\'
+            },
```
panarch, line 85:
```rust
escape_char: escape_char.unwrap_or('\\'),
```
It would be good to use `unwrap_or`.
We can apply this to `SqlExpr::ILike`, too.

core/src/data/value/mod.rs
```diff
@@ -635,11 +635,17 @@ impl Value {
...
         use Value::*;
 
         match (self, other) {
-            (Str(a), Str(b)) => a.like(b, case_sensitive).map(Bool),
+            (Str(a), Str(b)) => {
+                if escape_char == '%' || escape_char == '#' {
+                    a.like_with_escape(b, case_sensitive, escape_char).map(Bool)
+                } else {
+                    a.like(b, case_sensitive).map(Bool)
+                }
```
panarch, line 647: It would be better to handle escape_char in only one place. (now it becomes two: `value/mod.rs` and `data/string_ext.rs`.
currently that is handled in `data/string_ext.rs` so we'd better to make `StringExt::like` to always take `escape_char`. rather than adding a new function `StringExpr::like_with_escape`.

panarch (changes requested):

core/src/executor/evaluate/evaluated.rs
```diff
@@ -290,27 +290,39 @@ impl<'a> Evaluated<'a> {
...
-    pub fn like(&self, other: Evaluated<'a>, case_sensitive: bool) -> Result<Evaluated<'a>> {
+    pub fn like(
+        &self,
+        other: Evaluated<'a>,
+        case_sensitive: bool,
+        escape_char: char,
+    ) -> Result<Evaluated<'a>> {
         let evaluated = match (self, other) {
             (Evaluated::Literal(l), Evaluated::Literal(r)) => {
                 Evaluated::Literal(l.like(&r, case_sensitive)?)
```
devgony, line 301: 1. Escape should be implemented at Literal::like as well.
2. Then, you can write test case on `SELECT clause` with simple comparison like below.
3. We created new `test!` macro which includes `name` field to help understanding. you can just copy this test cases to `like_ilike.rs`.
```rs
    test! {
        name: "`_` should match any single character",
        sql: r#"SELECT 'GLUE+SQL' LIKE 'GLUE_SQL'"#,
        expected: Ok(select!(column1 Bool; true))
    };

    test! {
        name: r#"`\` should escape `_` to normal character - false"#,
        sql: r#"SELECT 'GLUE+SQL' LIKE 'GLUE\_SQL'"#,
        expected: Ok(select!(column1 Bool; false))
    };

    test! {
        name: r#"`\` should escape `_` to normal character - true"#,
        sql: r#"SELECT 'GLUE_SQL' LIKE 'GLUE\_SQL'"#,
        expected: Ok(select!(column1 Bool; true))
    };

    test! {
        name: "`%` should match any sequence of zero or more characters",
        sql: r#"SELECT 'GLUE+-SQL' LIKE 'GLUE%SQL'"#,
        expected: Ok(select!(column1 Bool; true))
    };

    test! {
        name: r#"`\` should escape `%` to normal character - false"#,
        sql: r#"SELECT 'GLUE+-SQL' LIKE 'GLUE\%SQL'"#,
        expected: Ok(select!(column1 Bool; false))
    };

    test! {
        name: r#"`\` should escape `%` to normal character - true"#,
        sql: r#"SELECT 'GLUE%SQL' LIKE 'GLUE\%SQL'"#,
        expected: Ok(select!(column1 Bool; true))
    };

    test! {
        name: r#"`\` should escape `\` to normal character"#,
        sql: r#"SELECT 'GLUE\SQL' LIKE 'GLUE\\SQL'"#,
        expected: Ok(select!(column1 Bool; true))
    };

    test! {
        name: r#"`\` should escape `'` to normal character"#,
        sql: r#"SELECT 'GLUE''SQL' LIKE 'GLUE\'SQL'"#,
        expected: Ok(select!(column1 Bool; true))
    };

    test! {
        name: r#"ESCAPE clause should set escape character (default: `\\`)"#,
        sql: r#"SELECT 'GLUE_SQL' LIKE 'GLUE\_SQL' ESCAPE '\\'"#,
        expected: Ok(select!(column1 Bool; true))
    };

    test! {
        name: r#"ESCAPE clause should set escape character to `#`"#,
        sql: r#"SELECT 'GLUE_SQL' LIKE 'GLUE#_SQL' ESCAPE '#'"#,
        expected: Ok(select!(column1 Bool; true))
    };

    test! {
        name: r#"`#` should escape `#` to normal character"#,
        sql: r#"SELECT 'GLUE#SQL' LIKE 'GLUE##SQL' ESCAPE '#'"#,
        expected: Ok(select!(column1 Bool; true))
    };
```

devgony: In `Like clause of SQL`, `_` is special character which means `match any single character`.
But in second test (`SELECT 'GLUE+SQL' LIKE 'GLUE\_SQL'`), `_` was escaped by `\` so that it became just normal character.
Then, the test result should be `false` as `+` is different with normal character `_`

Sixth test (`SELECT 'GLUE%SQL' LIKE 'GLUE\%SQL'`) has similar goal.
`%` is special character which means `match any sequence of zero or more characters`
But `%` was escaped by `\` so that it became just normal character.
Then, the test result should be `true` as `%` is equal to normal character `%`

That was why i wanted to test `escaping` all the special characters(`_`, `%`, `\`, `'`) in `Like clause of SQL`

devgony: The reason why CI test was failed is that the test cases i gave you have wrong column name.
It looks like each sql should use explicit alias

```diff
test! {
        name: "`_` should match any single character",
-        sql: r#"SELECT 'GLUE+SQL' LIKE 'GLUE_SQL'"#,
+        sql: r#"SELECT 'GLUE+SQL' LIKE 'GLUE_SQL' AS column1"#,
        expected: Ok(select!(column1 Bool; true))
    };
```

Mehrbod2002: @devgony the changes pushed but in this case : `SELECT 'GLUE\SQL' LIKE 'GLUE\\SQL'` . the regex should be `^GLUE\SQL$` and match string is `"GLUE\\SQL"` and not match . I am new regex :) hope you can solve it and commit the learn about it . thanks .

test-suite/src/like_ilike.rs
```diff
@@ -124,4 +131,70 @@ test_case!(like_ilike, async move {
...
+        name: r#"`\` should escape `\` to normal character"#,
+        sql: r#"SELECT 'GLUE\SQL' LIKE 'GLUE\\SQL'"#, ////
+        expected: Ok(select!(column1 Bool; true))
+    };
+
+    test! {
+        name: r#"`\` should escape `'` to normal character"#,
+        sql: r#"SELECT 'GLUE''SQL' LIKE 'GLUE\'SQL'"#,
+        expected: Ok(select!(column1 Bool; true))
+    };
```
devgony, line 181: Currently, I realized that this SQL is not supported by sqlparser so that let's remove it for now.

core/src/data/string_ext.rs
```diff
@@ -7,28 +7,68 @@ pub enum StringExtError {
...
+
+        match regex {
+            Ok(regex) => {
+                let escaped_match_string = regex::escape(match_string.as_str());
+                Ok(regex.is_match(escaped_match_string.as_str()))
+            }
+            Err(_) => Err(StringExtError::UnreachablePatternParsing.into()),
+        }
     }
 }
```
devgony, line 74: We are trying to avoid using `mut` keyword in GlueSQL engine codes.
This is my implementation.
```suggestion
    fn like(&self, pattern: &str, case_sensitive: bool, escape_char: char) -> Result<bool> {
        let pattern = pattern
            .chars()
            .scan(false, |escaped, char| {
                if !*escaped && char == escape_char {
                    *escaped = true;

                    return Some("".to_owned());
                }

                let regex = match (&escaped, &char) {
                    (false, '_') => ".".to_owned(),
                    (false, '%') => ".*".to_owned(),
                    (true, '\\') => "\\\\".to_owned(),
                    _ => char.to_string(),
                };

                *escaped = false;

                Some(regex)
            })
            .collect::<String>();

        let matched = RegexBuilder::new(&format!("^{pattern}$"))
            .case_insensitive(!case_sensitive)
            .build()
            .map_err(|_| StringExtError::UnreachablePatternParsing)?
            .is_match(self);

        Ok(matched)
    }
```

devgony (changes requested):

test-suite/src/like_ilike.rs
```diff
@@ -124,4 +131,70 @@ test_case!(like_ilike, async move {
...
+        name: r#"ESCAPE clause should set escape character to `#`"#,
+        sql: r#"SELECT 'GLUE_SQL' LIKE 'GLUE#_SQL' ESCAPE '#'"#,
+        expected: Ok(select!(column1 Bool; true))
+    };
+
+    test! {
+        name: r#"`#` should escape `#` to normal character"#,
+        sql: r#"SELECT 'GLUE#SQL' LIKE 'GLUE##SQL' ESCAPE '#'"#,
+        expected: Ok(select!(column1 Bool; true))
+    };
```
devgony, line 199: To pass test, you should do `AS column1` in each cases
```rs
    test! {
        name: "`_` should match any single character",
        sql: r#"SELECT 'GLUE+SQL' LIKE 'GLUE_SQL' AS column1"#,
        expected: Ok(select!(column1 Bool; true))
    };

    test! {
        name: r#"`\` should escape `_` to normal character - false"#,
        sql: r#"SELECT 'GLUE+SQL' LIKE 'GLUE\_SQL' AS column1"#,
        expected: Ok(select!(column1 Bool; false))
    };

    test! {
        name: r#"`\` should escape `_` to normal character - true"#,
        sql: r#"SELECT 'GLUE_SQL' LIKE 'GLUE\_SQL' AS column1"#,
        expected: Ok(select!(column1 Bool; true))
    };

    test! {
        name: "`%` should match any sequence of zero or more characters",
        sql: r#"SELECT 'GLUE+-SQL' LIKE 'GLUE%SQL' AS column1"#,
        expected: Ok(select!(column1 Bool; true))
    };

    test! {
        name: r#"`\` should escape `%` to normal character - false"#,
        sql: r#"SELECT 'GLUE+-SQL' LIKE 'GLUE\%SQL' AS column1"#,
        expected: Ok(select!(column1 Bool; false))
    };

    test! {
        name: r#"`\` should escape `%` to normal character - true"#,
        sql: r#"SELECT 'GLUE%SQL' LIKE 'GLUE\%SQL' AS column1"#,
        expected: Ok(select!(column1 Bool; true))
    };

    test! {
        name: r#"`\` should escape `\` to normal character"#,
        sql: r#"SELECT 'GLUE\SQL' LIKE 'GLUE\\SQL' column1"#, ////
        expected: Ok(select!(column1 Bool; true))
    };

    test! {
        name: r#"ESCAPE clause should set escape character (default: `\`)"#,
        sql: r#"SELECT 'GLUE_SQL' LIKE 'GLUE\_SQL' ESCAPE '\' AS column1"#,
        expected: Ok(select!(column1 Bool; true))
    };

    test! {
        name: r#"ESCAPE clause should set escape character to `#`"#,
        sql: r#"SELECT 'GLUE_SQL' LIKE 'GLUE#_SQL' ESCAPE '#' AS column1"#,
        expected: Ok(select!(column1 Bool; true))
    };

    test! {
        name: r#"`#` should escape `#` to normal character"#,
        sql: r#"SELECT 'GLUE#SQL' LIKE 'GLUE##SQL' ESCAPE '#' AS column1"#,
        expected: Ok(select!(column1 Bool; true))
    };
```

devgony (changes requested):

Mehrbod2002:
@devgony Thanks a lot for your very helpful guides :)

devgony:
I thought it was ready to merge.
Could you resolve conflict one more?

Mehrbod2002:
@devgony you can merge it now

Issue #1241, opened 2023-06-09
Implement MD5 Function in AST


seonghun-dev:
# Description
Computes the MD5 Hash with the result written in hexadecimal.

### Example
Function `MD5`

```sql
SELECT MD5('Bar')  FROM Foo;
```

# Task
- [x] Add `MD5` Function

The following tests should be passed:

``` rust
test_case!(md5_hash, async move {
    run!(
        "
        CREATE TABLE Foo (
            id INTEGER,
            data Text
        );
        "
    );
    run!(
        r#"
        INSERT INTO Foo VALUES
        (1, 'Hello, world!'),
        (2, 'OpenAI is awesome!'),
        (3, 'Rust programming is fun!');
        "#
    );
    test!(
        "SELECT MD5(data) AS md5hash FROM Foo",
        Ok(select!(
            md5hash
            String;
            "6cd3556deb0da54bca060b4c39479839".to_owned(),
            "5704cd1303d04a1390c270b37d195008".to_owned(),
            "f1aeb0b4346a0e45a54ca0bf9c0e00b7".to_owned()
        ))
    );
});

```

### Reference
https://www.postgresql.org/docs/15/functions-string.html#FUNCTIONS-STRING-OTHER

PR #1242, opened 2023-06-09
Implement MD5 Function

seonghun-dev:
Resolve: https://github.com/gluesql/gluesql/issues/1241


panarch (approved):
Looks all nice! Good to have a new function 👍 Thanks a lot!

PR #1247, opened 2023-06-10
Remove sync methods in core/ Glue struct,

panarch:
Update Glue struct to provide async methods only.
Apply the changes to existing codes (mainly test codes)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1062,6 +1062,7 @@ dependencies = [
  "async-trait",
  "bincode",
  "criterion",
+ "futures",
  "gluesql-core",
  "gluesql-test-suite",
  "gluesql-utils",
--- a/cli/src/cli.rs
+++ b/cli/src/cli.rs
@@ -5,6 +5,7 @@ use {
         print::Print,
     },
     edit::{edit_file, edit_with_builder, Builder},
+    futures::executor::block_on,
     gluesql_core::{
         prelude::Glue,
         store::{GStore, GStoreMut},
@@ -161,7 +162,7 @@ where
     }
 
     fn execute(&mut self, sql: impl AsRef<str>) -> Result<()> {
-        match self.glue.execute(sql) {
+        match block_on(self.glue.execute(sql)) {
             Ok(payloads) => self.print.payloads(&payloads)?,
             Err(e) => {
                 println!("[error] {}\n", e);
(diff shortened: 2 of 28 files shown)


ever0de (approved):

serprex:
@panarch I created a PR to update pgwire examples to use latest gluesql, but I found that calling `.await` on result of execute continued to fail even after moving to storing glue in `tokio::sync::Mutex`. I even tried getting around having a mutex at all using channels: https://github.com/serprex/pgwire/tree/fix1 to no avail

Are you aware of gluesql having compatibility issues with tokio?

ever0de:
> @panarch I created a PR to update pgwire examples to use latest gluesql, but I found that calling `.await` on result of execute continued to fail even after moving to storing glue in `tokio::sync::Mutex`. I even tried getting around having a mutex at all using channels: [serprex/pgwire@`fix1`](https://github.com/serprex/pgwire/tree/fix1?rgh-link-date=2023-12-24T17%3A21%3A44Z) to no avail
> 
> Are you aware of gluesql having compatibility issues with tokio?

hello :)
I've checked the pgwire code, and it appears that the [`SimpleQueryHandler`](https://github.com/sunng87/pgwire/blob/9e5caafa6c0167df55f840f1c4137a5245dcdd71/src/api/query.rs#L30-L95) requires the `Send` + `Sync` supertraits.

It doesn't really seem to be a compatibility issue with tokio, 
> I think what you mean by "compatibility with `tokio`" is nested calls to `futures::block_on`.

I think it's due to the failure to implement `Send` and `Sync` in `gluesql::{GStore, GStoreMut}`.
- Related issue: https://github.com/gluesql/gluesql/issues/1265
- <https://docs.rs/async-trait/latest/async_trait/#non-threadsafe-futures>

You can check by simply implementing the following `async_trait`.
```rust
pub struct GluesqlProcessor {
    db: SharedMemoryStorage,
}

#[async_trait(?Send)]
trait SendTest {
    async fn query(&self, query: &str);
}

#[async_trait(?Send)] // success, but `#[async_trait]` fails.
impl SendTest for GluesqlProcessor {
    async fn query(&self, query: &str) {
        let db = self.db.clone();
        let mut glue = Glue::new(db.clone());

        let result = glue.execute(query).await;
        println!("{:?}", result);
    }
}
```

We are working on some ideas to address this issue, but have not yet completed a functional implementation of `gluesql` that supports the Optional Send combination of `async_trait`. 

Would appreciate your understanding on this point 😢

PR #1253, opened 2023-06-21
ci: Automatically assign a PR to its author

### LINKED ISSUE #667: Add Assignees automatically when someone creates PR

rapsealk:
This PR resolves #667, using a well-known GitHub Actions workflow [toshimaru/auto-author-assign](https://github.com/toshimaru/auto-author-assign).

Please refer [this](https://github.com/rapsealk/github-actions-swing/pull/1) to see how it will work.

### DIFF
--- a/.github/auto-author-assign-config.yml
+++ b/.github/auto-author-assign-config.yml
@@ -1,0 +1,10 @@
+addAssignees: author
+
+addReviewers: true
+reviewers:
+  - panarch
+  - ever0de
+  - devgony
+numberOfReviewers: 0
+
+runOnDraft: true
--- a/.github/workflows/auto-author-assign.yml
+++ b/.github/workflows/auto-author-assign.yml
@@ -1,8 +1,8 @@
 name: auto-author-assign
 
 on:
-  pull_request_target:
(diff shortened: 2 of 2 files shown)


ever0de:
Maybe you could consider [kentaro-m/auto-assign-action](https://github.com/kentaro-m/auto-assign-action)?
It's not a huge difference, but it has additional features like assigning reviewers, etc. and it seems like a reasonable choice for maintenance to use the workflow that has more users.
What do you think?

rapsealk:
@ever0de Update completed. Please refer [this](https://github.com/rapsealk/github-actions-swing/pull/2) to see how it works.

ever0de (approved):
That's good 👍, Now we don't have to specify it manually, which is nice 😃

panarch (approved):
That looks really good! Thanks a lot for the contribution 👍 👍

devgony (approved):

PR #1263, opened 2023-07-08
Bump rust version to 1.70

zmrdltl:
## Reference
new rust stable version 1.70.0 (1 June, 2023) released. https://releases.rs/docs/1.70.0/


ever0de:
1.70 The PR you're getting is great, but the title you've written about, separate from the current issue, is something to think about.
reference: https://github.com/starkat99/half-rs/issues/97#issue-1779545608
Because of this, we may need to set the MSRV to 1.70. 
But I don't know if this is the right direction

zmrdltl:
> 1.70 The PR you're getting is great, but the title you've written about, separate from the current issue, is something to think about. reference: [starkat99/half-rs#97 (comment)](https://github.com/starkat99/half-rs/issues/97#issue-1779545608) Because of this, we may need to set the MSRV to 1.70. But I don't know if this is the right direction

In another issue I'm working on, there is a case of using a crate that uses a half crate. Fortunately, since it is a storage implementation issue that requires separating the workspace, msrv will be specified only in that case.

ever0de (approved):
### Organize what was said verbally in discord.

I think we are using [edition 2021 (first version 1.56.0)](https://doc.rust-lang.org/edition-guide/rust-2021/index.html)
> It actually asks for a higher version because of the `workspace.package` feature.
It may be correct to set MSRV to a higher version, but the current issue is not about core, it's about the bottleneck in implementing reference storage.
Therefore, I believe we should keep the MSRV for edition 2021.

There are two ways to do this.
1. Separate the `storages/*` workspace from the current project and change it to its own workspace. 2.
2. implement a stroage implementation as a sub-repository.
    - ex: https://github.com/ever0de/gluesql-discord

ever0de:
Since rust-toolchain is a value that overrides rustup version, it does not depend on rustc alone.
So I changed the name from `rustc` to `rust`.

> rustup automatically determines which toolchain to use when one of the installed commands **like rustc**(cargo[clippy, check, fmt, ..], rustc, ..) is executed.

- https://rust-lang.github.io/rustup/overrides.html

ever0de:
closed #1388

PR #1264, opened 2023-07-09
Simplity JsonStorage Store::fetch_all_schemas codes,

panarch:
Merge filter -> map to a single map and add filter_map with Result::transpose


devgony (approved):
Looks more concise 👍

PR #1280, opened 2023-07-13
Implement CSV Storage,

panarch:
Currently it only fails on schemaless tests

### DIFF
--- a/storages/csv-storage/src/lib.rs
+++ b/storages/csv-storage/src/lib.rs
@@ -1,0 +1,178 @@
+pub mod error;
+mod store;
+mod store_mut;
+
+use {
+    error::{CsvStorageError, ResultExt},
+    gluesql_core::{
+        ast::{ColumnUniqueOption, DataType},
+        data::{Key, Schema, Value},
+        error::{Error, Result},
+        store::{
+            AlterTable, CustomFunction, CustomFunctionMut, DataRow, Index, IndexMut, Metadata,
+            RowIter, Transaction,
+        },
+    },
+    std::{
+        collections::HashMap,
+        fs::{self, File},
+        io::Read,
+        path::PathBuf,
+    },
+};
+
+pub struct CsvStorage {
+    pub path: PathBuf,
+}
+
+impl CsvStorage {
+    pub fn new(path: &str) -> Result<Self> {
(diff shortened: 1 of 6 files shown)


storages/csv-storage/src/lib.rs
```diff
@@ -0,0 +1,246 @@
...
+
+    fn scan_data(&self, table_name: &str) -> Result<(Option<Vec<String>>, RowIter)> {
+        let data_path = self.data_path(table_name);
+        let (schema, generated) = match (self.fetch_schema(table_name)?, data_path.exists()) {
+            (None, _) | (_, false) => return Ok((None, Box::new(std::iter::empty()))),
+            (Some(v), true) => v,
+        };
+
+        let mut data_rdr = csv::Reader::from_path(data_path).map_storage_err()?;
+        let mut fetch_data_header_columns = || -> Result<Vec<String>> {
```
devgony, line 114: Looks like it is okay to use just `immutable` and  `variable`
```rs
let fetch_data_header_columns = data_rdr
    .headers()
    .map_storage_err()?
    .into_iter()
    .map(|header| header.to_string())
    .collect::<Vec<_>>();
```

devgony: Oh... If it was intended to run only in 2 of 3 conditions, mutable closure would be right.

storages/csv-storage/src/store_mut.rs
```diff
@@ -0,0 +1,247 @@
...
+            Ok(())
+        } else {
+            let rows = prev_rows
+                .collect::<Result<Vec<_>>>()?
+                .into_iter()
+                .map(|(_, row)| row)
+                .chain(rows)
+                .collect();
+
+            self.write(table_name, columns, rows)
```
devgony, line 94: How about using mut and extend?
```rs
let mut appended_rows = prev_rows
    .map(|result| Ok(result?.1))

appended_rows.extend(rows);

self.write(table_name, columns, appended_rows)
```

panarch: how about
```rust
let rows = prev_rows
    .map(|item| item.map(|(_, row)| row))
    .chain(rows.into_iter().map(Ok))
```
this? then we can merge two rows with a single collect.

devgony (approved):

PR #1283, opened 2023-07-13
[Function] Implement TAKE function 

### LINKED ISSUE #1273: Implement TAKE function
# Description
Implementation `TAKE()`
Returns the LIST taken only the first n from the LIST

## Example
Function `TAKE()`
```sql
SELECT TAKE(items, size) from Foo;
```

ding-co:
resolves #1273

### DIFF
--- a/core/src/executor/evaluate/function.rs
+++ b/core/src/executor/evaluate/function.rs
@@ -511,9 +511,9 @@ pub fn take<'a>(expr: Evaluated<'_>, size: Evaluated<'_>) -> Result<Evaluated<'a
         }
     };
 
-    match (expr, size) {
-        (Value::List(mut l), _) => {
-            l = l.into_iter().take(size).collect();
+    match expr {
+        Value::List(l) => {
+            let l = l.into_iter().take(size).collect();
             Ok(Evaluated::Value(Value::List(l)))
         }
         _ => Err(EvaluateError::ListTypeRequired.into()),
--- a/core/src/executor/evaluate/function.rs
+++ b/core/src/executor/evaluate/function.rs
@@ -501,13 +501,14 @@ pub fn prepend<'a>(expr: Evaluated<'_>, value: Evaluated<'_>) -> Result<Evaluate
     }
 }
 
-pub fn take<'a>(expr: Evaluated<'_>, size: Evaluated<'_>) -> Result<Evaluated<'a>> {
+pub fn take<'a>(name: String, expr: Evaluated<'_>, size: Evaluated<'_>) -> Result<Evaluated<'a>> {
     let expr: Value = expr.try_into()?;
     let size = match size.try_into()? {
-        Value::I64(number) => usize::try_from(number)
-            .map_err(|_| EvaluateError::FunctionRequiresUSizeValue("take".to_owned()))?,
+        Value::I64(number) => {
+            usize::try_from(number).map_err(|_| EvaluateError::FunctionRequiresUSizeValue(name))?
+        }
         _ => {
-            return Err(EvaluateError::FunctionRequiresIntegerValue("take".to_owned()).into());
+            return Err(EvaluateError::FunctionRequiresIntegerValue(name).into());
         }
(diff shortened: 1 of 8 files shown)


test-suite/src/function/take.rs
```diff
@@ -0,0 +1,63 @@
+use gluesql_core::executor::EvaluateError;
+
+use {crate::*, gluesql_core::prelude::Value::*};
```
panarch, line 3: In our repo, we recommend to wrap imports by using single `use`, then the cargo fmt can automatically do formatting.

core/src/executor/evaluate/function.rs
```diff
@@ -501,6 +501,25 @@ pub fn prepend<'a>(expr: Evaluated<'_>, value: Evaluated<'_>) -> Result<Evaluate
...
+        }
+    };
+
+    match (expr, size) {
+        (Value::List(mut l), _) => {
+            l = l.into_iter().take(size).collect();
+            Ok(Evaluated::Value(Value::List(l)))
+        }
+        _ => Err(EvaluateError::ListTypeRequired.into()),
+    }
```
panarch, line 520: 1. `size` does not need to be used in `match` statement.
2. rather than using `mut l`, it would be better to use variable shadowing
then,
```rust
match expr {
    Value::List(l) => {
        let l = l.into_iter().take(size).collect();

        Ok(Evaluated::Value(Value::List(l)))
    }
    _ => Err(EvaluateError::ListTypeRequired.into()),
}
````

File: core/src/plan/expr/function.rs [hunk 3]
```diff
@@ -140,6 +140,7 @@ impl Function {
             }
             | Self::Append { expr, value: expr2 }
             | Self::Prepend { expr, value: expr2 }
+            | Self::Take { expr, size: expr2 }
```
panarch, line 143: ref. https://github.com/gluesql/gluesql/pull/1282/commits/6b4aa8a7cd2d6196110ab0317610263ef66955db

could you add an unit test for this `Self::Take`?

core/src/executor/evaluate/function.rs
```diff
@@ -501,6 +501,25 @@ pub fn prepend<'a>(expr: Evaluated<'_>, value: Evaluated<'_>) -> Result<Evaluate
     }
 }
 
+pub fn take<'a>(expr: Evaluated<'_>, size: Evaluated<'_>) -> Result<Evaluated<'a>> {
```
panarch, line 504: you can consider to get `name` param like other functions.
Then the `name` param can be used in `EvaluateError::FunctionRequiresUSizeValue(name)` like errors.

test-suite/src/function/take.rs
```diff
@@ -0,0 +1,61 @@
+use {crate::*, gluesql_core::executor::EvaluateError, gluesql_core::prelude::Value::*};
+
```
panarch, line 2: You can consider a few more tests which handles `NULL`
either `expr` or `size` field is null, then the result is null in sql evaluation.
e.g.
```sql
TAKE(NULL, 0)
-- NULL

TAKE(items, NULL)
-- NULL
```

panarch (changes requested):

test-suite/src/function/take.rs
ding-co: Null has more high priority than size or list validation?

e.g.
```rust
TAKE(NULL, -5)
-- NULL? or FunctionRequiresUSizeValue?
```

panarch: that's really a good point.
for now, we don't have strict policy about the priority on what to handle earlier, yet.
it may depend on the arrange of implementation codes so it's all up to you.

ding-co: I think about that size or list validation has more priority,
I have to choose one more option to give higher priority validation between them.

So I adapt the way for just returning NULL when NULL value exists in query.

panarch (approved):

Issue #1062, opened 2023-01-14
Scala Subquery should contain only 1 column


devgony:
Scala Subquery should contain only 1 column or throw error.
![image](https://user-images.githubusercontent.com/51254761/212463210-326115c3-9930-43e8-b552-14f84f76e41d.png)

Currently it returns the first column rather than throws error
```sql
gluesql> select (select 1, 2);
```
| (SELECT 1, 2) |
|---------------|
| 1             |

ChobobDev:
I will work on this issue ,

Thanks in advance :)

PR #1284, opened 2023-07-13
Fix Scala Subquery should contain only 1 column

ChobobDev:
resolves : #1062


ChobobDev:
I think I am ready for the review !

core/src/executor/evaluate/error.rs
```diff
@@ -74,6 +74,9 @@ pub enum EvaluateError {
     #[error("subquery returns more than one row")]
     MoreThanOneRowReturned,
 
+    #[error("subquery recieved more than one column")]
+    MoreThanOneColumnReceived,
```
devgony: Looks like MoreThanOneColumnReturned sounds more easy to understand.

devgony (commented):
Looks like MoreThanOneColumnReturned sounds more easy to understand.

core/src/executor/evaluate/error.rs
ChobobDev: yeah, i totally agree with you,
I have made changes as you guided. Thank you

ever0de (approved):

panarch (approved):
Looks nice! Thanks for the contribution 👍

Issue #1271, opened 2023-07-11
Implement BIT_SHIFT function


codernineteen:
# Description
<br>
Support bit shift operation for specific usages

# Example
When a user stores ip address as 32bit integer type, It will be helpful if there is a way to convert it into network address again by using bit shift operation. Shift operations enable  the user to find the information of each octets.
```Bash
mysql> SELECT * FROM ipdata;
+------------+
| ipaddress  |
+------------+
| 2130706433 |
| 2921711528 |
| 3056763172 |
+------------+
3 rows in set (0.00 sec)

mysql> SELECT ipaddress, (ipaddress>>24) as FirstOctet, (ipaddress>>16<<16)-(ipaddress>>24<<24)>>16 as SecondOctet, (ipaddress>>8<<8)-(ipaddress>>16<<16)>>8 as ThirdOctet, ipaddress-(ipaddress>>8<<8) as FourthOctet FROM ipdata;
+------------+------------+-------------+------------+-------------+
| ipaddress  | FirstOctet | SecondOctet | ThirdOctet | FourthOctet |
+------------+------------+-------------+------------+-------------+
| 2130706433 |        127 |           0 |          0 |           1 |
| 2921711528 |        174 |          37 |        199 |         168 |
| 3056763172 |        182 |          50 |        129 |          36 |
+------------+------------+-------------+------------+-------------+
3 rows in set (0.02 sec)
```

# Reference
<br>
https://www.w3resource.com/mysql/bit-functions/left-right-shift.php

PR #1281, opened 2023-07-13
Feat: impl bitwise-and operation

jinlee0:
resolve #1277

PR #1286, opened 2023-07-14
Implement BIT_SHIFT_LEFT operation

codernineteen:
From the issue #1271,
Added below key features to several modules in core directory.
1. a way to translate the result of sqlparser
2. a execute path for shift_left operation
3. a definition of the operator '<<' as 'ShiftLeft'
4. new variant in LiteralError
5. test cases for above implementations


test-suite/src/lib.rs
```diff
@@ -34,15 +35,14 @@ pub mod schemaless;
 pub mod series;
 pub mod show_columns;
 pub mod synthesize;
+pub mod tester;
```
panarch, line 38: This change does not look related to the issue that implements bit_shift_left.

test-suite/src/lib.rs
```diff
@@ -5,6 +5,7 @@ pub mod alter;
 pub mod arithmetic;
 pub mod ast_builder;
 pub mod basic;
+pub mod bit_shift;
```
panarch, line 8: Not only adding a bit_shift module, we also need to use it.
you can see the code below, there's `glue!` macro and you can consider to use that in `generate_store_tests`

panarch (changes requested):

core/src/ast/operator.rs
```diff
@@ -59,6 +60,7 @@ impl ToSql for BinaryOperator {
             BinaryOperator::And => "AND".to_owned(),
             BinaryOperator::Or => "OR".to_owned(),
             BinaryOperator::Xor => "XOR".to_owned(),
+            BinaryOperator::ShiftLeft => "<<".to_owned(),
```
panarch, line 63: though it may look too long name, but it would be important to mention that the operation is "bitwise".
so it would be better to use explicit name: `BitwiseShiftLeft`

ref. https://github.com/gluesql/gluesql/pull/1281

core/src/data/value/mod.rs
```diff
@@ -555,6 +555,218 @@ impl Value {
...
+            Err(_) => Err(ValueError::ImpossibleCast.into()),
+        }
+    }
+
+    pub fn shift_left(&self, rhs: &Value) -> Result<Value> {
+        use Value::*;
+
+        match (self, rhs) {
+            (I8(a), I64(b)) => {
+                let conversion_result = Self::validate_shift_rhs(*b);
```
panarch, line 570: most of primitive integer types already have `checked_shl` so you don't need to convert the value into `u32`.
you can simply run `checked_shl` for `i8`, `i16`, ..

core/src/executor/evaluate/evaluated.rs
```diff
@@ -220,6 +220,10 @@ impl<'a> Evaluated<'a> {
         binary_op(self, other, |l, r| l.modulo(r), |l, r| l.modulo(r))
     }
 
+    pub fn shift_left<'b>(&'a self, other: &Evaluated<'b>) -> Result<Evaluated<'b>> {
```
panarch, line 223: as we'd better to use more explicit name for binary operator `BitwiseShiftLeft`, this method name would be better to use `bitwise_shift_left`.

panarch (changes requested):

codernineteen (commented):
https://docs.rs/num-traits/latest/num_traits/ops/checked/trait.CheckedShl.html

According to above docs, it is true that 'checked_shl' was implemented for most of primitive integer types. However it seems that 'rhs' parameter for checked_shl is 'u32' type.
When i ran the test cases i wrote in suite, i always got the I64 Value type as a result. That's why i added this conversion funtionality into this module.
If there is anything i misunderstood , pls feedback )

panarch:
> https://docs.rs/num-traits/latest/num_traits/ops/checked/trait.CheckedShl.html

We don't need to use `num-traits` because the primitive types already support `checked_shl` by default.
ref. https://doc.rust-lang.org/std/primitive.i32.html#method.checked_shl

it would be better to see the doc of rust primitive integer types.
each of them returns the same type of itself.

core/src/data/value/mod.rs
```diff
@@ -555,6 +555,157 @@ impl Value {
...
+            },
+            _ => Err(ValueError::ImpossibleCast.into()),
+        }
+    }
+
+    pub fn bitwise_shift_left(&self, rhs: &Value) -> Result<Value> {
+        use Value::*;
+
+        let try_rhs_u32 = Self::validate_bitwise_shift_rhs(rhs);
+        match try_rhs_u32 {
```
panarch, line 575: we can simplify the codes like these below,
`validate_bitwise_shift_rhs` is now only being used in a single place so no more as a separate function is required.

```rust
let rhs = u32::try_from(rhs)?;
match self {
    Value::I8(lhs) => lhs.checked_shl(rhs).map(Value::I8),
    Value::I16(lhs) => lhs.checked_shl(rhs).map(Value::I16),
    Value::I32(lhs) => lhs.checked_shl(rhs).map(Value::I32),
    Value::I64(lhs) => lhs.checked_shl(rhs).map(Value::I64),
    Value::I128(lhs) => lhs.checked_shl(rhs).map(Value::I128),
    Value::U8(lhs) => lhs.checked_shl(rhs).map(Value::U8),
    Value::U16(lhs) => lhs.checked_shl(rhs).map(Value::U16),
    Value::U32(lhs) => lhs.checked_shl(rhs).map(Value::U32),
    Value::U64(lhs) => lhs.checked_shl(rhs).map(Value::U64),
    Value::U128(lhs) => lhs.checked_shl(rhs).map(Value::U128),
    Value::Null => Some(Value::Null),
    _ => {
        return Err(ValueError::NonNumericMathOperation {
            lhs: self.clone(),
            rhs: Value::U32(rhs),
            operator: NumericBinaryOperator::BitwiseShiftLeft,
        }
        .into());
    }
}
.ok_or_else(|| {
    ValueError::BinaryOperationOverflow {
        lhs: self.clone(),
        rhs: Value::U32(rhs),
        operator: NumericBinaryOperator::BitwiseShiftLeft,
    }
    .into()
})
```

core/src/data/value/mod.rs
```diff
@@ -1591,6 +1628,142 @@ mod tests {
         test!(modulo I128(6),   F64(2.0) => F64(0.0));
         test!(modulo I128(6),   F32(2.0_f32) => F32(0.0_f32));
 
+        test!(bitwise_shift_left I8(1),     I64(2) => I8(4));
```
panarch, line 1631: at the time that we combine arithmetic binary operator unit tests into a single unit tests, we didn't know that the size of tests will grow like this big.

now rather than making this unit tests bigger, it would be better to make the unit test only for `bitwise_shift_left`.
Whether to use re-use this macro or not is up to you.

core/src/data/value/mod.rs
```diff
@@ -555,6 +555,41 @@ impl Value {
         }
     }
 
+    pub fn bitwise_shift_left(&self, rhs: &Value) -> Result<Value> {
+        use Value::*;
+
+        let rhs = u32::try_from(rhs)?;
```
panarch, line 561: Oh, I missed one.
in the literal
the result of `SELECT 10 << NULL` is `NULL`.
but for the value, it will return error.
for the consistency, it would be better to add logic to return `NULL` when `rhs` is null

panarch (changes requested):

core/src/data/literal.rs
```diff
@@ -190,6 +197,28 @@ impl<'a> Literal<'a> {
...
+                _ => Err(LiteralError::UnreachableBinaryArithmetic.into()),
+            },
+            (Null, Number(_)) | (Number(_), Null) | (Null, Null) => Ok(Literal::Null),
+            _ => Err(LiteralError::IncompatibleBitOperation(
+                format!("{:?}", self),
+                format!("{:?}", other),
+            )
+            .into()),
+        }
+    }
```
panarch, line 220: here we can remove the use of unreachable error.

still we need to think about the proper error names, but briefly we can re-arrange codes like this below.

```rust
match (self, other) {
    (Number(l), Number(r)) => {
        let l = l.to_i64().unwrap(); // non-integer -> error
        let r = r.to_u32().unwrap();
        let res = l.checked_shl(r).unwrap(); // overflow -> error
        Ok(Number(Cow::Owned(BigDecimal::from(res)))
    }
    (Null, Number(_)) | (Number(_), Null) | (Null, Null) => Ok(Literal::Null),
    // it would be better to have more explicit error name because it is caused by non-integer literals
    _ => Err(LiteralError::IncompatibleBitOperation(
        format!("{:?}", self),
        format!("{:?}", other),
    )
    .into()),
}
```

panarch: image

panarch (approved):
Looks all great! Thanks a lot for the contribution 👍 👍

PR #1288, opened 2023-07-14
Implement VALUES function for Map type

### LINKED ISSUE #1276: Implement `VALUES` Function for `Map` type.
# Description
GlueSql provides Map Type as a data type and we can use this as follows :
Gluesql_Map_Type

I want to implement `VALUES` Function that is currently not supported.
does_not_support_values

### VALUES Function
  - syntax : `VALUES(map)`
  - argument : Map type
  - return : List of all values from map

# Example
### Clickhouse
- Query
  ```
  SELECT mapValues(mapdata)
  ```
- Result
  ```
  [1,'taegi']
  ```

# Task
- write test-suite codes for `VALUES` function.
- add `VALUES` to enum `Function`.
- implement `VALUES` function.

# Reference
- clickhouse
  https://clickhouse.com/docs/en/sql-reference/functions/tuple-map-functions#mapvalues

tgsong827:
# Description

resolve #1276

implement `VALUES` to get list of all values of Map.

# Example

```
CREATE TABLE USER (id INTEGER, data MAP);

INSERT INTO USER VALUES 
  (1, '{"id": 1, "name": "alice"}'),
  (2, '{"name": "bob"}'),
  (3, '{}');

SELECT VALUES(data) as result FROM USER WHERE id=1;

| result      |
|-------------|
| [1,"alice"] |
```

### DIFF
--- a/core/src/executor/evaluate/function.rs
+++ b/core/src/executor/evaluate/function.rs
@@ -1,3 +1,5 @@
+use itertools::Itertools;
+
 use {
     super::{EvaluateError, Evaluated},
     crate::{
@@ -569,6 +571,21 @@ pub fn is_empty<'a>(expr: Evaluated<'_>) -> Result<Evaluated<'a>> {
     };
 
     Ok(Evaluated::from(Value::Bool(length == 0)))
+}
+
+pub fn values<'a>(expr: Evaluated<'_>) -> Result<Evaluated<'a>> {
+    let expr: Value = expr.try_into()?;
+    match expr {
+        Value::Map(m) => {
+            let mut values = Vec::new();
+            // list will be sorted by key value that is String type.
+            for key in m.keys().sorted() {
+                values.push(m[key].clone());
+            }
+            Ok(Evaluated::from(Value::List(values)))
+        }
+        _ => Err(EvaluateError::MapTypeRequired.into()),
+    }
 }
 
 // --- etc ---
--- a/core/src/executor/evaluate/function.rs
+++ b/core/src/executor/evaluate/function.rs
@@ -5,7 +5,6 @@ use {
         data::{Key, Point, Value, ValueError},
         result::Result,
     },
-    itertools::Itertools,
     md5::{Digest, Md5},
     rand::{rngs::StdRng, Rng, SeedableRng},
     std::ops::ControlFlow,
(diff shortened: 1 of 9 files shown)


core/src/executor/evaluate/function.rs
```diff
@@ -1,3 +1,5 @@
+use itertools::Itertools;
```
ever0de, line 1: Combine with the use syntax on line 3

core/src/executor/evaluate/function.rs
```diff
@@ -538,6 +540,21 @@ pub fn is_empty<'a>(expr: Evaluated<'_>) -> Result<Evaluated<'a>> {
...
+    let expr: Value = expr.try_into()?;
+    match expr {
+        Value::Map(m) => {
+            let mut values = Vec::new();
+            // list will be sorted by key value that is String type.
+            for key in m.keys().sorted() {
+                values.push(m[key].clone());
+            }
+            Ok(Evaluated::from(Value::List(values)))
+        }
```
ever0de, line 553:
```suggestion
        Value::Map(m) => Ok(Evaluated::from(Value::List(
            m.into_iter()
                .sorted_by(|(k1, _), (k2, _)| k1.cmp(k2))
                .map(|(_, v)| v)
                .collect(),
        ))),
```

You can clear the `clone` logic by using ownership of the hashmap, sorting using `sorted_by`, and doing `collect` with the results.

ever0de (changes requested):

core/src/executor/evaluate/function.rs
```diff
@@ -538,6 +539,19 @@ pub fn is_empty<'a>(expr: Evaluated<'_>) -> Result<Evaluated<'a>> {
     Ok(Evaluated::from(Value::Bool(length == 0)))
 }
 
+pub fn values<'a>(expr: Evaluated<'_>) -> Result<Evaluated<'a>> {
+    let expr: Value = expr.try_into()?;
+    match expr {
+        Value::Map(m) => Ok(Evaluated::from(Value::List(
+            m.into_iter()
+                .sorted_by(|(k1, _), (k2, _)| k1.cmp(k2))
```
devgony, line 547: Finally, SORT function (#1300)  was merged so that we can remove internal sort here.
Then, let's wrap test cases with `SORT()`

```rs
sql: r#"SELECT SORT(VALUES(data)) as result FROM USER WHERE id=1"#,
```

tgsong827: Thanks for reviews.
I updated values function to remove sorting logic and added `SORT()` function into test codes. But if the list consists of different types, it can not be ordered correctly every time.
So I want to handle this in another issue #1324.
Could you please review this PR again?

ever0de (approved):

panarch (approved):

PR #1289, opened 2023-07-15
[Draft]replace HashMap to IndexMap

devgony:
# 🚧 Draft yet

## Goal
- Currently, our Value::Map takes HashMap which does not guarantee insertion order.
- To guarantee insertion order, replace HashMap to IndexMap

## Todo
- [x] replace HashMap to IndexMap
- [ ] impl enum GlueMap to include both of HashMap to IndexMap
- [ ] recover some part of IndexMaps to HashMap


core/src/executor/evaluate/evaluated.rs
```diff
@@ -89,7 +93,7 @@ impl TryFrom<Evaluated<'_>> for HashMap<String, Value> {
                 Err(EvaluateError::TextLiteralRequired(format!("{v:?}")).into())
             }
             Evaluated::Value(Value::Str(v)) => HashMap::parse_json_object(v.as_str()),
-            Evaluated::Value(Value::Map(v)) => Ok(v),
+            Evaluated::Value(Value::Map(v)) => Ok(HashMap::from_iter(v.into_iter())),
```
devgony, line 96: To convert from IndexMap to HashMap, currently it collects again.
I'm considering which is better between this way and using enum like this.
```rs
// core/src/store/data_row.rs
enum GluMap {
  IndexMap(IndexMap<String, Value>
  HashMap(HashMap<String,Value>
}

pub enum DataRow {
  Vec(Vec<Value>),
  Map(GlueMap)
}
```

cli/src/lib.rs
```diff
@@ -132,7 +134,9 @@ pub fn dump_database(storage: &mut SledStorage, dump_path: PathBuf) -> Result<()
                         result.map(|data_row| {
                             let values = match data_row {
                                 DataRow::Vec(values) => values,
-                                DataRow::Map(values) => vec![Value::Map(values)],
+                                DataRow::Map(values) => {
+                                    vec![Value::Map(IndexMap::from_iter(values))]
```
devgony, line 138: ditto

test-suite/src/tester/macros.rs
```diff
@@ -134,7 +134,7 @@ macro_rules! select_map {
         gluesql_core::executor::Payload::SelectMap(vec![
             $(
                 match gluesql_core::data::Value::try_from($row).unwrap() {
-                    Value::Map(v) => v,
+                    Value::Map(v) => std::collections::HashMap::from_iter(v.into_iter()),
```
devgony, line 137: ditto

devgony:
There is no way to save order infomation of IndexMap in Storage so that close this PR for now.

PR #1299, opened 2023-07-15
[RFC] Feature/func bit not

gurugio:
Just a draft PR to check the test coverage.

I don't write full description for each patch because I don't understand well 
what ast, ast_builder, evaluator, executor, plan and translate do and how they co-operate.
I will add description later.

I will add more test cases to increase the test coverage.
- Creating a table and run bit_not function and operator with the values in the table

### DIFF
--- a/core/src/ast/function.rs
+++ b/core/src/ast/function.rs
@@ -9,6 +9,7 @@ use {
 #[strum(serialize_all = "SCREAMING_SNAKE_CASE")]
 pub enum Function {
     Abs(Expr),
+    BitNot(Expr),
     Lower(Expr),
     Initcap(Expr),
     Upper(Expr),
@@ -186,6 +187,7 @@ impl ToSql for Function {
     fn to_sql(&self) -> String {
         match self {
             Function::Abs(e) => format!("ABS({})", e.to_sql()),
+            Function::BitNot(e) => format!("BIT_NOT({})", e.to_sql()),
             Function::Initcap(e) => format!("INITCAP({})", e.to_sql()),
             Function::Lower(e) => format!("LOWER({})", e.to_sql()),
             Function::Upper(e) => format!("UPPER({})", e.to_sql()),
@@ -457,6 +459,14 @@ mod tests {
         assert_eq!(
             r#"ABS("num")"#,
             &Expr::Function(Box::new(Function::Abs(Expr::Identifier("num".to_owned())))).to_sql()
+        );
+
+        assert_eq!(
+            r#"BIT_NOT("num")"#,
+            &Expr::Function(Box::new(Function::BitNot(Expr::Identifier(
+                "num".to_owned()
+            ))))
+            .to_sql()
         );
 
         assert_eq!(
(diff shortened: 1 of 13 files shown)


panarch:
though it does not need to be well written, it would be better to use english for all others to understand the pr content.
could you write the pr description in english?

devgony (commented):
Could you please run `cargo fmt`?
Or you could install `rust-analyzer` and use `format on save`

panarch:
```rust
pub enum UnaryOperator {
    Plus,
    Minus,
    Not,
    /// Bitwise Not, e.g. `~9` (PostgreSQL-specific)
    PGBitwiseNot,
    /// Square root, e.g. `|/9` (PostgreSQL-specific)
    PGSquareRoot,
    ..
}
```
It looks that sqlparser-rs supports bitwise not operator using `~` symbol.
could we consider to use this rather than adding as a function?

ref.
https://github.com/sqlparser-rs/sqlparser-rs/blob/a50671d95d34688529fbc3d4452bec164ff21497/src/ast/operator.rs#L30C17-L46

gurugio:
> though it does not need to be well written, it would be better to use english for all others to understand the pr content. could you write the pr description in english?

Sure.

gurugio:
> ```rust
> pub enum UnaryOperator {
>     Plus,
>     Minus,
>     Not,
>     /// Bitwise Not, e.g. `~9` (PostgreSQL-specific)
>     PGBitwiseNot,
>     /// Square root, e.g. `|/9` (PostgreSQL-specific)
>     PGSquareRoot,
>     ..
> }
> ```
> 
> It looks that sqlparser-rs supports bitwise not operator using `~` symbol. could we consider to use this rather than adding as a function?
> 
> ref. https://github.com/sqlparser-rs/sqlparser-rs/blob/a50671d95d34688529fbc3d4452bec164ff21497/src/ast/operator.rs#L30C17-L46

Yes, I will add operator.
I thought the sqlparser-rs does not support that operator.

One question.
Do you think it would be better to add only operator?
Or both of operator and function?

panarch:
> One question.
Do you think it would be better to add only operator?
Or both of operator and function?

Let's make it simple for this PR :)
It would be good enough for only having an operator.

devgony:
> > Could you please run `cargo fmt`? Or you could install `rust-analyzer` and use `format on save`
> 
> Sure, thank you for your review. Actually I already have been using the rust-analyzer in the vscode. I don't understand why there is mis-formatted code but will double-check for next PR.

You could try in VSCode: Settings (ctrl + ,) > search "format on save" > enable

gurugio:
> > One question.
> > Do you think it would be better to add only operator?
> > Or both of operator and function?
> 
> Let's make it simple for this PR :) It would be good enough for only having an operator.

Got it.

gurugio:
> > > Could you please run `cargo fmt`? Or you could install `rust-analyzer` and use `format on save`
> > 
> > 
> > Sure, thank you for your review. Actually I already have been using the rust-analyzer in the vscode. I don't understand why there is mis-formatted code but will double-check for next PR.
> 
> You could try in VSCode: Settings (ctrl + ,) > search "format on save" > enable

Done. Thank you

Issue #1272, opened 2023-07-12
Implement SORT function


Jaehui-Lee:
# Description
Implementation `SORT()`

## Example
Function `SORT()`
```sql
SELECT SORT(Bar) FROM Foo;
```

# Task

# Reference

PR #1300, opened 2023-07-16
implement `SORT` function

Jaehui-Lee:
resolves https://github.com/gluesql/gluesql/issues/1272

I am worried about checking whether all elements in the list are comparable. Is there any better way?

### DIFF
--- a/core/src/executor/evaluate/function.rs
+++ b/core/src/executor/evaluate/function.rs
@@ -9,6 +9,7 @@ use {
     rand::{rngs::StdRng, Rng, SeedableRng},
     std::ops::ControlFlow,
     uuid::Uuid,
+    std::cmp::Ordering,
 };
 
 macro_rules! eval_to_str {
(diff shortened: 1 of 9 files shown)


core/src/executor/evaluate/function.rs
```diff
@@ -7,6 +7,7 @@ use {
     },
     md5::{Digest, Md5},
     rand::{rngs::StdRng, Rng, SeedableRng},
+    std::cmp::Ordering,
     std::ops::ControlFlow,
```
panarch, line 11: nit. it would be good to merge this
```rust
std::{cmp::Orderinb, ops::ControlFlow},
```

core/src/executor/evaluate/function.rs
```diff
@@ -501,6 +502,45 @@ pub fn prepend<'a>(expr: Evaluated<'_>, value: Evaluated<'_>) -> Result<Evaluate
     }
 }
 
+pub fn sort<'a>(expr: Evaluated<'_>, order: Evaluated<'_>) -> Result<Evaluated<'a>> {
+    let expr: Value = expr.try_into()?;
+    let order: Value = order.try_into()?;
+
+    match expr {
+        Value::List(mut l) => match order {
```
panarch, line 510: how about adapting the similar strategy which we do in `sort` module?
in the `sort.rs` execution module, (https://github.com/gluesql/gluesql/blob/main/core/src/executor/sort.rs)

`Value` does not impl `Ord` but `Key` does, so for `Value::List(Vec<Value>)`,
1. convert the list value to contain `Key` - `Vec<(Key, Value)>`
2. then sort it.
3. remove `Key`

fyi. the ideal way to resolve this would be.. to use `Value::evaluate_cmp` but currently the `evaluate_cmp` impl in `Value` returns `Option<Ordering>` so it is not proper to be used for sorting.
for now, the realistic option would be to use `Key`. but I will consider to update `Value::evaluate_cmp` to return `Ordering`

panarch (changes requested):

core/src/executor/evaluate/function.rs
Jaehui-Lee: https://github.com/gluesql/gluesql/pull/1300/commits/0f860a04c4e149ec3ec7429d0e92026380d93907

I understood that adding each `Value` of the list as a `Key`, and sorting using `Key::cmp`
Is it right what you intended?

Jaehui-Lee: Is there anything else I need to fix?

panarch (approved):
Looks all great! well organized test-suite cases also look nice.

PR #1303, opened 2023-07-18
Add Expr::Array support for LIST data type

### LINKED ISSUE #1294: Add `Expr::Array` support for `LIST` data type
e.g.
```bash
gluesql> select ARRAY[1,2,3];
[error] translate: unsupported expr: ARRAY[1, 2, 3]

gluesql> select [1,2,3];
[error] translate: unsupported expr: [1, 2, 3]
```

`ARRAY[1,2,3]` or `[1,2,3]` like inputs are supported by sqlparser-rs but we don't.
we can provide this functionality by supporting `Expr::Array(Array)` in sqlparser-rs

ref.
https://github.com/sqlparser-rs/sqlparser-rs/blob/a50671d95d34688529fbc3d4452bec164ff21497/src/ast/mod.rs#L558

ChobobDev:
resolves #1294


ChobobDev:
I am deeply sorry for the delay.
I think I am now ready for the review !

core/src/plan/expr/mod.rs
```diff
@@ -69,6 +69,11 @@ impl<'a> From<&'a Expr> for PlanExpr<'a> {
                 let exprs = indexes.iter().chain(once(obj.as_ref())).collect();
                 PlanExpr::MultiExprs(exprs)
             }
+            Expr::Array { elem } => {
+                let exprs = elem.iter().collect();
+                PlanExpr::MultiExprs(exprs)
+                //fix
```
panarch, line 75: we can remove this unused comment

core/src/executor/evaluate/mod.rs
```diff
@@ -292,6 +292,10 @@ async fn evaluate_inner<'a, 'b: 'a, 'c: 'a, T: GStore>(
             let indexes = try_join_all(indexes.iter().map(eval)).await?;
             expr::array_index(obj, indexes)
         }
+        Expr::Array { elem } => {
+            let elem = try_join_all(elem.iter().map(eval)).await?;
+            expr::array(elem)
+        }
```
panarch, line 298: now it doesn't need to split the evaluation code in `evaluate/expr.rs`.
then it would be simply enough with
```rust
Expr::Array { elem } => {
            try_join_all(elem.iter().map(eval))
                .await?
                .into_iter()
                .map(Value::try_from)
                .collect::<Result<Vec<_>>>()
                .map(Value::List)
                .map(Evaluated::Value)
        }
```

core/src/translate/expr.rs
```diff
@@ -1,3 +1,5 @@
+use sqlparser::ast::Array;
+
 use {
```
panarch, line 3: could you merge this to use single `use`?

core/src/executor/evaluate/mod.rs
ChobobDev: Thx, i will apply the following

core/src/plan/expr/mod.rs
ChobobDev: Oh, i forgot to remove following, thx for pointing this out

test-suite/src/array.rs
```diff
@@ -0,0 +1,43 @@
+use {
+    crate::*,
+    gluesql_core::prelude::{Payload, Value::*},
+};
+
+test_case!(array, {
+    let g = get_tester!();
+    g.run("CREATE TABLE Test (id INTEGER DEFAULT 1,surname TEXT NULL, name LIST NOT NULL);")
```
panarch, line 8: for `array` test case, 'surname' field use looks unnecessary which is not related to `LIST` data type.
and..
could we have also simple `SELECT` query based test cases?
e.g.
```sql
SELECT ['hello', 1, 2, True];

SELECT ['a', 'b', 'c'][1];
```
thanks to array expression, now we can provide the above cases support.

panarch (changes requested):

core/src/translate/expr.rs
ChobobDev: oh, i will apply this ASAP

test-suite/src/array.rs
ChobobDev: I have applied this changes

panarch (approved):
It's nice to have `SELECT [1, 2];` support!
Thanks a lot for the contribution :D

Issue #1311, opened 2023-07-21
Implement GREATEST function


TheMan1697:
> # Description
> The GREATEST function in SQL is used to select the largest (maximum) value from a list of expressions. It compares each value and returns the highest value. The function can accept numbers, strings, dates.
> # Syntax
> ```
> GREATEST (expr1, expr2, expr3, ..., expr_n)
> ```
> Where expr1, expr2, expr3, ..., expr_n are the expressions to be compared. This could be column values or literal values. These expressions must all be of the same type, or types that SQL can implicitly convert.
> 
> # Example
> ```
> SELECT GREATEST (5, 10, 15);
> ```

PR #1312, opened 2023-07-21
Implement GREATEST function

TheMan1697:
#1311


core/src/executor/evaluate/function.rs
```diff
@@ -605,6 +605,37 @@ pub fn generate_uuid<'a>() -> Evaluated<'a> {
...
+    if !exprs.iter().all(|value| value.get_type() == *first_type) {
+        return Err(EvaluateError::CannotCompareDifferentTypes.into());
+    }
+
+    let greatest_value = exprs.into_iter().max_by(|a, b| match (a, b) {
+        (Value::I64(a), Value::I64(b)) => a.cmp(b),
+        (Value::F64(a), Value::F64(b)) => a.partial_cmp(b).unwrap_or(std::cmp::Ordering::Equal),
+        (Value::Str(a), Value::Str(b)) => a.cmp(b),
+        (Value::Date(a), Value::Date(b)) => a.cmp(b),
+        _ => panic!("Unsupported type in comparison"),
```
zmrdltl, line 630: We do not recommend using panic. Therefore, it would be better to implement a separate error object for unsupported types.

core/src/executor/evaluate/function.rs
```diff
@@ -605,6 +605,37 @@ pub fn generate_uuid<'a>() -> Evaluated<'a> {
...
+    });
+
+    match greatest_value {
+        Some(value) => Ok(Evaluated::from(value)),
+        None => Ok(Evaluated::from(Value::Null)),
```
zmrdltl, line 635: If there is a test for if the result is `None` it would be really nice

core/src/ast/function.rs
```diff
@@ -335,6 +336,19 @@ impl ToSql for Function {
...
             Function::GenerateUuid() => "GENERATE_UUID()".to_owned(),
+            Function::Greatest(items) => {
+                let items = items
+                    .iter()
+                    .filter(|&item| !matches!(item, Expr::IsNull(_)))
+                    .map(ToSql::to_sql)
+                    .collect::<Vec<_>>()
+                    .join(", ");
+                if items.is_empty() {
+                    "NULL".to_owned()
```
zmrdltl, line 347: It would be nice to have a test that returns NULL for empty items.

zmrdltl (changes requested):

core/src/executor/evaluate/function.rs
```diff
@@ -645,6 +645,41 @@ pub fn generate_uuid<'a>() -> Evaluated<'a> {
     Evaluated::from(Value::Uuid(Uuid::new_v4().as_u128()))
 }
 
+pub fn greatest(_name: String, exprs: Vec<Evaluated<'_>>) -> Result<Evaluated<'_>> {
```
panarch, line 648: to handle this function, it looks that we can do a bit different approach than other functions.

`Evaluated` provides `evaluate_cmp` method which can compare with other `Evaluated` value.
Then, rather than converting the evaluated into `Value`, we can simply use `Evaluated::evaluate_cmp` to compare each other.
`Evaluated::evaluate_cmp` returns `None` on non-comparable pair so we can still use `CannotCompareDifferentTypes` error on `None` return value.

core/src/ast/function.rs
panarch: may be this causes some confusion..?

though `is_empty` it does not mean that the result of `to_sql` is `NULL`.
we can simply use `GREATEST()` for that.

panarch (changes requested):

core/src/executor/evaluate/function.rs
```diff
@@ -645,6 +645,22 @@ pub fn generate_uuid<'a>() -> Evaluated<'a> {
...
+
+    for expr in iter {
+        match greatest.evaluate_cmp(&expr) {
+            Some(std::cmp::Ordering::Less) => greatest = expr,
+            Some(_) => (),
+            None => return Err(EvaluateError::ComparisonOperationError.into()),
+        }
+    }
+
+    Ok(greatest)
```
panarch, line 661: then, with new error names and using `try_fold`, this below may work.

```rust
exprs
    .into_iter()
    .try_fold(None, |greatest, expr| -> Result<_> {
        let greatest = match greatest {
            Some(greatest) => greatest,
            None => return Ok(Some(expr)),
        };

        match greatest.evaluate_cmp(&expr) {
            Some(Ordering::Less) => Ok(Some(expr)),
            Some(_) => Ok(Some(greatest)),
            None => Err(EvaluateError::NonComparableArgumentError(name.to_owned()).into()),
        }
    })?
    .ok_or(EvaluateError::FunctionRequiresAtLeastOneArgument(name.to_owned()).into())
```

panarch (changes requested):

core/src/executor/evaluate/function.rs
```diff
@@ -645,6 +645,22 @@ pub fn generate_uuid<'a>() -> Evaluated<'a> {
     Evaluated::from(Value::Uuid(Uuid::new_v4().as_u128()))
 }
 
+pub fn greatest(_name: String, mut exprs: Vec<Evaluated<'_>>) -> Result<Evaluated<'_>> {
+    exprs.retain(|expr| !expr.is_null() && !expr.is_bool());
```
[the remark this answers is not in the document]
TheMan1697: I intended to ignore NULL values when they come in and proceed with the comparison. Also, I thought it was strange that GREATEST can compare Boolean values, so I intended to exclude it.

core/src/ast/function.rs
```diff
@@ -351,6 +352,15 @@ impl ToSql for Function {
                 format!("UNWRAP({}, {})", expr.to_sql(), selector.to_sql())
             }
             Function::GenerateUuid() => "GENERATE_UUID()".to_owned(),
+            Function::Greatest(items) => {
+                let items = items
+                    .iter()
+                    .filter(|&item| !matches!(item, Expr::IsNull(_)))
```
panarch, line 358: Oh, I missed the this one.
we don't need to filter out the `NULL` arguments. `NULL` also needs to be showed.
```
GREATEST(NULL, 9, 7)
GREATEST(1, 2, 3)
GREATEST(NULL, NULL, 2)
```
all above expressions are valid.

panarch (approved):
Looks all great! Thanks a lot for the contribution 👍 👍

PR #1330, opened 2023-07-23
Update core/src/executor/evaluate/evaluated.rs

### LINKED ISSUE #1328: Remove `From<Value> for Evaluated<'a>`
https://github.com/gluesql/gluesql/blob/5d96a9aca9fdbd4bd457a7831db708fe29402c0c/core/src/executor/evaluate/evaluated.rs#L21-L25

This `From` impl only shorten 
```rust
Evaluated::Value(value)
```
to
```rust
Evaluate::from(value)
```

Now there's no reason to keep using this code. we can simply use `Evaluated::Value` instead.
Let's remove this.
Once we remove the declaration codes and run `cargo check`, then `cargo` will lead us what to do.

viktoriussuwandi:
fix #1328

Hi [panarch](https://github.com/panarch),
just trying to jump in this issue, please hcek it out


panarch:
Rather than changing the codes, you can simply removing the entire block.
image
Then if you run `cargo check`, the compiler will kindly introduce what you need to do.

ever0de:
closed #1399

PR #1331, opened 2023-07-24
Refactor GCD and LCM functions

### LINKED ISSUE #1310: Function GCD, LCM need to handle more edge cases
note that the general definition of GCD, LCM is **positive** integer.

> the greatest common divisor (GCD) of two or more [ntegers, which are not all zero, is the largest **positive** integer that divides each of the integers

> In [arithmetic](https://en.wikipedia.org/wiki/Arithmetic) and [number theory](https://en.wikipedia.org/wiki/Number_theory), the least common multiple, lowest common multiple, or smallest common multiple of two [integers](https://en.wikipedia.org/wiki/Integer) a and b, usually denoted by lcm(a, b), is the smallest **positive** integer that is [divisible](https://en.wikipedia.org/wiki/Divisible) by both a and b.

References: (https://en.wikipedia.org/wiki/Greatest_common_divisor), (https://en.wikipedia.org/wiki/Least_common_multiple)
(I apologize for using Wikipedia references, but I'm sure that the same definitions can be found in any number theory book)

Therefore, the output of the following queries is incorrect:

```rs
gluesql> SELECT GCD(-2, 0);
| GCD(-2, 0) |
|------------|
| -2         |

gluesql> SELECT GCD(-14, -7);
| GCD(-14, -7) |
|--------------|
| -7           |

gluesql> SELECT LCM(-12, 2);
| LCM(-12, 2) |
|-------------|
| -12         |
```
it should be
GCD(-2, 0) = 2
GCD(-14, -7) = 7
LCM(-12, 2) = 12

Furthermore, the LCM function should handle the case of LCM(0, 0)

```rs
gluesql> SELECT LCM(0, 0);
thread 'main' panicked at 'attempt to divide by zero', core/src/executor/evaluate/function.rs:450:9
```

cake-monotone:
Resolves #1310

Hi! I have refactored the GCD and LCM functions for improved safety and error handling.

With these changes:
### GCD
- The GCD function is now implemented using iteration not recursive function.
- An overflow error is prevented when one of the arguments is `i64::MIN`

### LCM
- An overflow error is prevented when one of the arguments is `i64::MIN`.
- An overflow error is prevented when dealing with large prime numbers, i.e) `LCM(large prime, large prime)`.
- The case where both arguments are zero, ex) `LCM(0, 0)` is now properly handled.

### DIFF
--- a/core/src/executor/evaluate/function.rs
+++ b/core/src/executor/evaluate/function.rs
@@ -3,7 +3,7 @@ use {
     crate::{
         ast::{DataType, DateTimeField},
         data::{Key, Point, Value, ValueError},
-        result::Result,
+        result::{Error, Result},
     },
     md5::{Digest, Md5},
     rand::{rngs::StdRng, Rng, SeedableRng},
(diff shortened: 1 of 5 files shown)


core/src/executor/evaluate/function.rs
```diff
@@ -452,26 +452,42 @@ pub fn gcd<'a>(name: String, left: Evaluated<'_>, right: Evaluated<'_>) -> Resul
...
+        let gcd_val: i128 = gcd_i64(a, b)?.into();
+
+        let a: i128 = a.into();
+        let b: i128 = b.into();
+
+        // lcm(a, b) = abs(a * b) / gcd(a, b)   if gcd(a, b) != 0
+        // lcm(a, b) = 0                        if gcd(a, b) == 0
+        let result = (a * b).abs().checked_div(gcd_val).unwrap_or(0);
+
+        TryInto::<i64>::try_into(result).map_err(|_| Error::Value(ValueError::LcmResultOutOfRange))
```
panarch, line 472: Looks all nice!
this is just a very nit comment.

```rust
i64::try_from(result).map_err(|_| Error::Value(ValueError::LcmResultOutOfRange))
```

panarch (approved):
Looks nice! Thanks a lot 👍

PR #1333, opened 2023-07-26
Implement COALESCE function

### LINKED ISSUE #1332: Implement COALESCE function
Implementation for COALESCE function

## Example

```sql
SELECT COALESCE(NULL, 3, "test");
```

## Implemtation Spec

- At least one argument is required (this specification is same with PostgreSQL, MySQL, while SQLite requires a minimum of two arguments).
- Returns `NULL` if all arguments in the COALESCE function are `NULL`

cake-monotone:
Implementation for COALESCE function

Resolves #1332

### DIFF
--- a/core/src/ast/function.rs
+++ b/core/src/ast/function.rs
@@ -43,6 +43,7 @@ pub enum Function {
         data_type: DataType,
     },
     Ceil(Expr),
+    Coalesce(Vec<Expr>),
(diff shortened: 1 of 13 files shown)


cake-monotone:
✨  resolved conflicts ✨

panarch (approved):

PR #1340, opened 2023-07-28
Implement `SLICE` function

### LINKED ISSUE #1307: Implement `SLICE` function
# Description
Implement `SLICE` function that return subsets of list.

If start value is positive, start from array index (start at `0`).
If start value is negative, start from the end (last index is `-1`).

# Example
```
SLICE(<list>, <start>, <length>)
```

Kwontaehwon:
resolved #1307 
# Description

Implement `SLICE` function that return subsets of list.
**Support both positive and negative `start` value**

+ If start value over the list length, it returns empty list : `[]`
```
INSERT INTO Test VALUES ('[1,2,3,4]');
SELECT SLICE(list, 100, 5) AS value FROM Test;
```

+ If `start` value is negative and its absolute value is over the list length, it convert to index 0.

# Example

`SLICE(<list>, <start>, <length>)`


core/src/executor/evaluate/function.rs
```diff
@@ -574,6 +574,40 @@ pub fn sort<'a>(expr: Evaluated<'_>, order: Evaluated<'_>) -> Result<Evaluated<'
...
+        }
+    };
+    match expr {
+        Value::List(l) => {
+            if start < 0 {
+                start += l.len() as i64;
+            }
+            if start < 0 {
+                start = 0;
+            }
```
zmrdltl, line 600: really good! The slice function will soon become a new feature. It would be nice if only very small changes were possible. Couldn't these parts be put together?

Kwontaehwon: I intended that when a negative index is given, if index remain negative even after adding the size of vector to convert it to a positive index, it falls outside the range of vector. So I change index to 0.

Issue #1351, opened 2023-07-29
[AST Builder] Implement Index by Expression in Select


seonghun-dev:
# Description
Implements ast builders for `index_by` Expression

They take a similar form to index hints in traditional SQL, 
but while in SQL these index hints can be ignored at the option of the optimizer, 
in GlueSQL Query Builder they are generated rather than ignored 
because they are part of the actual query generation.

#### MySQL
- https://dev.mysql.com/doc/refman/8.0/en/index-hints.html

#### Oracle
- Write a hint comment on the first line of the SELECT clause to give it the proper index
```sql
SELECT /*+ INDEX(a EMP_IDX02) */ a.empno , a.ename , a.hiredate FROM emp a
```

#### Current FSM Chart (Exclude Detail)
![image](https://github.com/gluesql/gluesql/assets/80201773/8db109fe-bca2-4a7a-981c-f6f385c7501a)

#### New FSM Chart (Exclude Detail)
![image](https://github.com/gluesql/gluesql/assets/80201773/7d9a9c67-c2db-4c45-bbfa-ba0cbe88fda2)

#### In index Node FSM
The interior of the node in the index consists of another FSM.
Prevent chaining from becoming too long.

![image](https://github.com/gluesql/gluesql/assets/80201773/7533eb54-0a13-4b97-b24e-4748b2f6b1d3)

## Task
_Index Hint Works With Update Expression, But in this Issue will be limit to Select Expression_
- [x] Add Index Node to Select FSM
- [x] Add Use Index Node
- [x] Add Primary Key Node
- [x] Add Non clustered Node

The following tests should be passed:
### Index By
``` rust
    #[test]
    fn test_index() {
        let sql = "
        SELECT * FROM Player
        WHERE name IN (
            SELECT * FROM Player WHERE id = 1
        )";

        let index_node = use_index().primary_key("id");
        let query_node = table("Player")
            .select()
            .index_by(index_node)
            .filter("id".eq(1));
        let query_builder = table("Player")
            .select()
            .filter("name".in_list(subquery))
            .build();

        let actual = plan(&storage, query_builder);
        let expected = {
            let subquery = Query {
                body: SetExpr::Select(Box::new(Select {
                    projection: vec![SelectItem::Wildcard],
                    from: TableWithJoins {
                        relation: TableFactor::Table {
                            name: "Player".to_owned(),
                            alias: None,
                            index: Some(IndexItem::PrimaryKey(expr("1"))),
                        },
[...]

PR #1355, opened 2023-07-31
feat: add index_by node

seonghun-dev:
- Resolve : https://github.com/gluesql/gluesql/issues/1351


core/src/ast_builder/table_factor.rs
```diff
@@ -19,6 +19,7 @@ pub struct TableFactorNode<'a> {
     pub table_name: String,
     pub table_type: TableType<'a>,
     pub table_alias: Option<String>,
+    pub index: Option<IndexItem>,
```
panarch, line 22: When we use ast builder, all the ast building process can be made without failure which means that all a kind of `build` can be done in lazy.
let's consider to have `IndexItemNode`.

core/src/ast_builder/table_name.rs
```diff
@@ -75,6 +79,16 @@ impl<'a> TableNameNode {
     pub fn alter_table(self) -> AlterTableNode {
         AlterTableNode::new(self.table_name)
     }
+
+    pub fn index_by(self, index: IndexItem) -> IndexNode<'a> {
```
panarch, line 83: With `IndexItemNode`,
```rust
pub fn index_by<T: Into<IndexItemNode>>(self, index_item: T) -> TableFactorNode {
..
```

this kind of approach may work.

core/src/ast_builder/index_by.rs
```diff
@@ -0,0 +1,121 @@
+use crate::{
+    ast::IndexItem,
+    ast_builder::{SelectNode, TableFactorNode},
+};
+
+#[derive(Clone, Debug)]
+pub struct IndexNode<'a> {
```
panarch, line 7: it would be enough with `pub enum IndexItem` which contains `ast::IndexItem` and some nodes in `index_item/**` because `TableFactorNode` can have `IndexItemNode`.

panarch (changes requested):

core/src/ast_builder/index_item/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+mod non_clustered;
+mod order;
+mod primary_key;
+mod primary_key_expr;
+
+pub use {
+    crate::ast::IndexItem, crate::result::Result, cmp_expr::CmpExprNode,
+    non_clustered::non_clustered, non_clustered::NonClusteredNode, order::OrderNode,
+    primary_key::primary_key, primary_key::PrimaryKeyNode, primary_key_expr::PrimaryKeyCmpExprNode,
+};
```
panarch, line 11: nit. 
```rust
pub use {
    crate::{ast::IndexItem, result::Result},
    cmp_expr::CmpExprNode,
    non_clustered::{non_clustered, NonClusteredNode},
    order::OrderNode,
    primary_key::{primary_key, PrimaryKeyNode},
    primary_key_expr::PrimaryKeyCmpExprNode,
};
```

core/src/ast_builder/index_item/primary_key.rs
```diff
@@ -0,0 +1,15 @@
+use {super::PrimaryKeyCmpExprNode, crate::ast_builder::ExprNode};
+
+#[derive(Clone, Debug)]
+pub struct PrimaryKeyNode;
+
+impl<'a> PrimaryKeyNode {
+    pub fn eq<T: Into<ExprNode<'a>>>(self, expr: T) -> PrimaryKeyCmpExprNode<'a> {
+        PrimaryKeyCmpExprNode::new(expr.into())
```
panarch, line 8: It currently goes to `PrimaryKeyCmpExpNode` but could we consider to use `IndexItemNode`?
may be we don't need `PrimaryKeyCmpExprNode` which can be merged into `IndexItemNode`

core/src/ast_builder/index_item/non_clustered.rs
```diff
@@ -0,0 +1,101 @@
...
+    };
+
+    #[test]
+    fn test() {
+        let index_node: IndexItemNode = non_clustered("idx".to_owned()).gt("1").into();
+        let actual = index_node.prebuild().unwrap();
+        let expected = IndexItem::NonClustered {
+            name: "idx".to_owned(),
+            asc: None,
+            cmp_expr: Some((IndexOperator::Gt, to_expr("1"))),
```
panarch, line 52: we convert `actual` into `AST` and compare with `expected`.
Then, rather than using `to_expr` which parses and translates, how about simply using AST Expr?
this below would be enough
```rust
Expr::Literal(AstLiteral::Number(1.into()))
```

core/src/ast_builder/index_item/non_clustered.rs
```diff
@@ -0,0 +1,101 @@
...
+pub fn non_clustered(index_name: String) -> NonClusteredNode {
+    NonClusteredNode { index_name }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::{
+        ast::IndexOperator,
+        ast_builder::{index_item::non_clustered::non_clustered, select::Prebuild, to_expr},
+        ast_builder::{index_item::IndexItem, IndexItemNode},
```
panarch, line 42: nit.
```rust
ast_builder::{
            index_item::{non_clustered, IndexItem},
            IndexItemNode,
            select::Prebuild,
}
```

core/src/ast_builder/index_item/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+        .map(|(index_operator, expr)| (index_operator, expr));
+
+    Ok(IndexItem::NonClustered {
+        name,
+        asc,
+        cmp_expr: cmp_expr_result,
+    })
+}
+
+fn build_primary_key(expr: ExprNode) -> Result<IndexItem> {
```
panarch, line 78: `build_non_clustered` and `build_primary_key` functions are only used in `prebuild` and the amount of codes are small. then it would be enough to put the code into prebuild

panarch (approved):
Now all looks great!! Thanks a lot for the contribution 👍 👍

Issue #480, opened 2022-02-09
Adding (Optional) Python Bindings To GlueSQL


the-wondersmith:
Hello again! I've been plugging away at a project that uses GlueSQL. It's a [PEP 249](https://www.python.org/dev/peps/pep-0249/)-compliant DB API v2 module for reading and writing a semi-obscure embedded, fixed-width / flat-file based "database" that's been EOL'd for... probably _decades_ at this point. 

That's beside the point though - in the process of working on this project I've realized how _incredibly_ useful / helpful / convenient it would be if GlueSQL included (optional) bindings for [pyo3](https://pyo3.rs/). It would make it almost _trivial_ to use custom or obscure storage types in Python without being forced to sacrifice being able to use SQL to query the stored records (as opposed to having to damn near reinvent the wheel by _effectively_ implementing a query engine in Python.) Even better, using GlueSQL in Rust means that you could interact with that same storage backend with _anything_ else that Rust supports as a compilation target or has FFI bindings to 🤩. (Case in point, see [gluesql-js](https://github.com/gluesql/gluesql-js)).

After a bit more thought though, I came to the conclusion that if GlueSQL _did have_ `pyo3` bindings, it's not like the Python-side DB API stuff would really ever need to change from project to project. Once it's written and tested, it's pretty much drop in whatever your storage struct is and you're good to go. Long story short, I [took a crack at it](https://github.com/the-wondersmith/gluesql/tree/python-bindings) just to see what it'd look like. I've got the new `pyo3`-oriented code behind a cargo config flag (`py-bindings`), and still need to write the tests for the new additions, _BUT_ as it turns out it works a treat so far 😅.

**SO**, having already done a decent chunk of the legwork I have a proposal and a question - I'd like to write an official "extension" (or whatever the proper term is) that adds drop-in support for using GlueSQL with Python, including a `PEP 249`-compliant interface module _and_ accompanying [SQLAlchemy](https://www.sqlalchemy.org/) dialect. I will sign on to write and maintain all of the code that lives at the intersection of GlueSQL and `pyo3`, assuming I can get and necessary guidance and feedback to ensure that everything keeps to GlueSQL's style and quality standards. The question would be, are y'all down with that?

Regardless of whatever the answer turns out to be, I _seriously_ can't thank y'all enough for your work on this project. It's _fantastic_, and I'd (obviously) love to help out however I can.

the-wondersmith:
[...]

PR #1357, opened 2023-07-31
feat: Implement python binding

jopemachine:
Related issue: https://github.com/gluesql/gluesql/issues/480.

This PR implements Python bindings for the following storages:

- [x] MemoryStorage
- [x] JsonStorage
- [x] SharedMemoryStorage
- [x] SledStorage

I think GlueSQL in Python will be a powerful tool.

Future improvements:

- [ ] Convert the query function to an asynchronous function.
- [ ] Optimize the convert_payload function for better performance.
- [ ] Add bindings for the Composite Storage.


pkg/python/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "gluesql-py"
+authors = ["Taehoon Moon <taehoon.moon@outlook.com>"]
```
panarch, line 3: you can become the first author here :)
ref.
https://github.com/gluesql/gluesql/blob/bac635e489dcf617a4f47a752ab94f6470731f71/storages/json-storage/Cargo.toml#L3-L6

jopemachine: Thanks!

panarch (commented):
now it looks all great!
just before merging it, could you add a github action for python build?
we need a way to verify that other future changes not to break the python build.

panarch (approved):
It's so pleasure to see the new meaningful functionality.
Everything looks great!! Let's ship this to the main 👍 👍 👍

PR #1360, opened 2023-08-03
Implement TRANSACTION for memory storage

codernineteen:
I wrote detailed explanation in commit.
Because there should be more efficient algorithm to implement transaction functionality, This will be just draft pull request.


storages/memory-storage/src/transaction.rs
```diff
@@ -10,20 +10,41 @@ use {
...
+            Some(snapshot) => {
+                let data = snapshot.to_owned().data;
+                self.functions = data.functions;
+                self.id_counter = data.id_counter;
+                self.items = data.items;
+                self.metadata = data.metadata;
+                self.state = StorageState::Idle;
+                self.snapshot = None;
+            }
+            None => println!("TODO: non snapshot error handling"),
```
devgony, line 40: Are you on way to implement this part?
And you can remove other `TODO` comments if was done.

codernineteen: Apologize for incomplete pull request. 
I'm working on optimizing transaction functionality now and will update this pull request based on your review soon!

PR #1361, opened 2023-08-04
Remove eval_to_{int|float..} macro uses in core/../evaluate/function.rs,

panarch:
Update all function return types from Result to ControlFlow.
Replace eval_to_{..} macros to normal functions.
Except for NULL break handling, this does not affect existing business logics.


core/src/executor/evaluate/function.rs
```diff
@@ -1,128 +1,193 @@
...
+pub fn concat(exprs: Vec<Evaluated<'_>>) -> ControlFlow<Evaluated> {
+    let value = exprs
+        .into_iter()
+        .try_fold(None, |left: Option<Evaluated>, right| {
+            let left = match left {
+                None => return Continue(Some(right)),
+                Some(left) => left,
+            };
+
+            left.concat(right).break_if_null().map(Some)
```
ever0de, line 157:
```suggestion
        .try_fold(None, |left: Option<Evaluated>, right| match left {
            None => Continue(Some(right)),
            Some(left) => left.concat(right).break_if_null().map(Some),
        })?;
```

core/src/executor/evaluate/function.rs
```diff
@@ -470,218 +530,210 @@ pub fn lcm<'a>(name: String, left: Evaluated<'_>, right: Evaluated<'_>) -> Resul
...
+    Continue(a)
 }
 
 // --- list ---
-pub fn append<'a>(expr: Evaluated<'_>, value: Evaluated<'_>) -> Result<Evaluated<'a>> {
-    let expr: Value = expr.try_into()?;
-    let value: Value = value.try_into()?;
+pub fn append<'a>(expr: Evaluated<'_>, value: Evaluated<'_>) -> ControlFlow<Evaluated<'a>> {
+    let expr: Value = expr.try_into().break_if_null()?;
+    let value: Value = value.try_into().break_if_null()?;
```
ever0de, line 559:
```suggestion
    let expr = expr.try_into().break_if_null()?;
    let value = value.try_into().break_if_null()?;
```

ever0de: Please also remove the similar `type annotation` below

core/src/executor/evaluate/mod.rs
```diff
@@ -484,13 +487,14 @@ async fn evaluate_function<'a, 'b: 'a, 'c: 'a, T: GStore>(
...
             };
+
             f::rand(name, expr)
         }
         Function::Round(expr) => f::round(name, eval(expr).await?),
         Function::Floor(expr) => f::floor(name, eval(expr).await?),
         Function::Radians(expr) => f::radians(name, eval(expr).await?),
         Function::Degrees(expr) => f::degrees(name, eval(expr).await?),
-        Function::Pi() => Ok(Evaluated::from(Value::F64(std::f64::consts::PI))),
+        Function::Pi() => return Ok(Evaluated::from(Value::F64(std::f64::consts::PI))),
```
ever0de, line 497:
```suggestion
        Function::Pi() => return Ok(Evaluated::Value(Value::F64(std::f64::consts::PI))),
```

ever0de (changes requested):

devgony (approved):

PR #1364, opened 2023-08-05
Implement entries in ast builder

### LINKED ISSUE #1362: Implement `entries` in ast builder
proposal for implementing `entries` in ast builder for Map type in gluesql

2-NOW:
resolve #1362


ever0de (approved):

core/src/ast_builder/expr/function.rs
```diff
@@ -1695,4 +1704,11 @@ mod tests {
         let expected = "LAST_DAY(TIMESTAMP '2023-07-29 11:00:00')";
         test_expr(actual, expected);
     }
+
+    #[test]
+    fn function_entries() {
+        let actual = entries(col("map"));
+        let expected = "ENTRIES(map)";
+        test_expr(actual, expected);
```
panarch, line 1712: cause we have two entry points for `entries`
1. `entries(col("test_col"))`
2. `col("test_col").entries()`

it would be nice to cover all two cases.
everything else looks all good.

2-NOW: Thank you. I've applied the feedback you mentioned :)

panarch (approved):

PR #1365, opened 2023-08-05
feat: Implement select without table function in ast_builder

### LINKED ISSUE #1339: Implement `SELECT without Table` in AST Builder
## Goal
- Currently in SQL, we support `SELECT without Table` like this
```sql
SELECT 1, 'A';
```
- But in AST builder, it is not supported yet.
- We may implement like this
```rs
select() // should be able to use `select` method without `table()`
.project(num(1))
.project("A")
```
- The way to implement can be changed if you have better idea

## Good to know
- `SELECT without Table` internally uses `SERIES(1)`
- https://github.com/gluesql/gluesql/pull/733
- you may start to implement from `core/src/ast_builder/table_factor.rs`

ding-co:
resolves #1339


panarch (approved):
Looks nice! Thanks for the contribution 👍

PR #1371, opened 2023-08-12
Implement `SPLICE` function

### LINKED ISSUE #1348: Implement `SPLICE` function
# Description
SPLICE(list_data, begin_index, end_index, values...)
Remove data in `list_data` from `begin_index`(inclusive) to `end_index`(exclusive).
If `values` exists, insert the values at that position.
# Examples
```sql
SELECT SPLICE('[0, 1, 2, 3, 4, 5]', 1, 3)
> [0, 3, 4, 5]
```
```sql
SELECT SPLICE('[0, 1, 2, 3, 4, 5]', 1, 3, 100)
> [0, 100, 3, 4, 5]
```
```sql
SELECT SPLICE('[0, 1, 2, 3, 4, 5]', 1, 3, 100, 99, 98)
> [0, 100, 99, 98 3, 4, 5]
```

jinlee0:
resolves #1348


jinlee0:
@panarch 
I think I'm ready for a review!

test-suite/src/function/splice.rs
```diff
@@ -0,0 +1,60 @@
+use {
+    crate::*,
+    gluesql_core::{executor::EvaluateError, prelude::Value::*},
+};
+
+test_case!(splice, async move {
```
panarch, line 6: now all the test-suite refactoring works which I was planning to do at this time are finished and merged into main branch.
there's some changes on writing tests.
1. `async move {` -> `{`
2. `run!`, `test!` macros -> `g.run(..`, `g.test(..` normal functions

could you merge & apply these changes?

core/src/plan/expr/function.rs
```diff
@@ -6,13 +6,14 @@ use {
...
         #[derive(iter_enum::Iterator)]
-        enum Exprs<I0, I1, I2, I3, I4, I5> {
+        enum Exprs<I0, I1, I2, I3, I4, I5, I6> {
             Empty(I0),
             Single(I1),
             Double(I2),
             Triple(I3),
             VariableArgs(I4),
             VariableArgsWithSingle(I5),
+            Quadruple(I6),
```
ever0de, line 16: Is it correct to create a `Quadruple` instead of a `VariableArgs`? @panarch

ever0de (approved):
Looks good to me

core/src/plan/expr/function.rs
panarch: for now, both can be accepted so this decision which adds Quadruple looks also good.

nit. it would be better to add Quadruple in between Triple and VariableArgs rather than adding it to the last.

panarch (approved):

PR #1375, opened 2023-08-14
feat: implement ast_builder for values function

tgsong827:
It's about function `values` related with #1288

### DIFF
--- a/core/src/ast_builder/mod.rs
+++ b/core/src/ast_builder/mod.rs
@@ -78,7 +78,7 @@ pub use expr::{
         get_x, get_y, greatest, ifnull, initcap, is_empty, last_day, lcm, left, length, ln, log,
         log10, log2, lower, lpad, ltrim, md5, modulo, now, pi, point, position, power, radians,
         rand, repeat, replace, reverse, right, round, rpad, rtrim, sign, sin, skip, sqrt, substr,
-        take, tan, to_date, to_time, to_timestamp, upper, FunctionNode,
+        take, tan, to_date, to_time, to_timestamp, upper, values as fn_values, FunctionNode,
     },
 };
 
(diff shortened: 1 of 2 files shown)


ever0de (approved):

core/src/ast_builder/mod.rs
```diff
@@ -78,7 +78,7 @@ pub use expr::{
         get_x, get_y, greatest, ifnull, initcap, is_empty, last_day, lcm, left, length, ln, log,
         log10, log2, lower, lpad, ltrim, md5, modulo, now, pi, point, position, power, radians,
         rand, repeat, replace, reverse, right, round, rpad, rtrim, sign, sin, skip, sqrt, substr,
-        take, tan, to_date, to_time, to_timestamp, upper, FunctionNode,
+        take, tan, to_date, to_time, to_timestamp, upper, values as fn_values, FunctionNode,
```
ever0de, line 81: One thing to think about, @panarch 
1. `fn_values`
2. `function::values` (should we modularize all functions?)

Do you think there are options other than `fn_values`?

tgsong827: I don't know well, is it possible to move the modules related to functions in `mod.rs(ast_builder)` to another file(ex. `mod_functions.rs`). 
If possible, we can use the values function with its original name and `ast_bulder::expr::functions::tests` can use this like `use ast_builder::mod_functions::values`.

ever0de: That's not possible if the names overlap where you do the `use`.
Number 2 above suggests exporting to `module::fn`, do you have an answer? @panarch

panarch: > Number 2 above suggests exporting to module::fn

this looks reasonable.

e.g.
```rust
use gluesql::ast_builder::fn::values;
```

panarch:
```rust
use gluesql::ast_builder::function::values;

use gluesql::ast_builder::aggregate;

use gluesql::ast_builder::function as f;
```

panarch (approved):
Thanks a lot!

PR #1379, opened 2023-08-20
chore: add example of convert from payload to custom struct

ever0de:


panarch (approved):
Looks really nice! 
Thanks a lot 👍 👍 👍

PR #1388, opened 2023-08-28
chore: bump rust versino to 1.72

ever0de:
If the current rust version is below 1.70, the coverage ci will fail (see clap-derive MSRV).

resolver ref: <https://doc.rust-lang.org/cargo/reference/resolver.html#feature-resolver-version-2>


panarch (approved):
👍 👍

PR #1396, opened 2023-09-02
Feature: redis storage

gurugio:
I'm working for Redis storage.
There are many patches at the moment. I will rebase them into one patch later.
Please check the latest code.


storages/redis-storage/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+pub struct Item {
+    pub schema: Schema,
+}
+
+pub struct RedisStorage {
+    pub namespace: String,
+    pub items: HashMap<String, Item>,
+    pub conn: RefCell<Connection>,
+    pub functions: HashMap<String, StructCustomFunction>,
+}
```
panarch, line 31: here `items` or `functions` fields might be considered as a cache but because it is redis storage. it would be too much for the start.
let's make the storage implementation simple. it would be enough to only have `namespace` and `conn` fields.
in the case of `items` we can always store & fetch from redis rather than cloning it in here.

and.. `functions`, you may not be able to implement custom function trait for now. (it is because of the limitation of the current interface)

gurugio: Got it. I will remove items and functions fields.

storages/redis-storage/src/lib.rs
```diff
@@ -0,0 +1,459 @@
...
+        store::{CustomFunction, CustomFunctionMut, DataRow, RowIter, Store, StoreMut},
+    },
+    redis::{Commands, Connection},
+    std::cell::RefCell,
+    std::collections::BTreeMap,
+};
+
+pub struct Item {
+    pub schema: Schema,
+}
```
panarch, line 24: This code block looks currently not being used. it would be good to remove this

storages/redis-storage/src/lib.rs
gurugio: Done

storages/redis-storage/src/lib.rs
gurugio: Done

storages/redis-storage/src/lib.rs
```diff
@@ -0,0 +1,494 @@
...
+                            ))
+                        })?;
+                } else {
+                    return Err(Error::StorageMsg(format!(
+                        "[RedisStorage] Wrong key format: key={}",
+                        redis_key
+                    )));
+                }
+            }
+        }
```
panarch, line 381: nit.
```rust
let rows = redis_keys
    .into_iter()
    .filter_map(|redis_key| {
        redis::cmd("GET")
            .arg(&redis_key)
            .query::<String>(&mut self.conn.borrow_mut())
            .ok()
            .map(|value| (redis_key, value))
    })
    .map(|(redis_key, value)| {
        let key = Self::redis_parse_key(&redis_key).map_err(|_| {
            Error::StorageMsg(format!(
                "[RedisStorage] failed to deserialize key: key={}",
                redis_key,
            ))
        })?;

        let value = serde_json::from_str::<DataRow>(&value).map_err(|e| {
            Error::StorageMsg(format!(
                "[RedisStorage] failed to deserialize value={} error={:?}",
                value, e
            ))
        })?;

        Ok((key, value))
    })
    .collect::<Result<BTreeMap<Key, DataRow>>>()?;
```

gurugio: done

panarch (changes requested):

storages/redis-storage/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+redis = "0.23.3"
+serde_json = "1.0.105"
+chrono = { version = "=0.4.31", features = ["serde", "wasmbind"] }
+futures = "0.3"
+toml = "0.8.6"
+
+[dev-dependencies]
+test-suite.workspace = true
+tokio = { version = "1", features = ["rt", "macros"] }
+futures = "0.3"
```
panarch, line 24: `futures` is already declared in `[dependencies]` so we can remove it here.

storages/redis-storage/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+
+[dependencies]
+gluesql-core.workspace = true
+async-trait = "0.1"
+serde = { version = "1", features = ["derive"] }
+redis = "0.23.3"
+serde_json = "1.0.105"
+chrono = { version = "=0.4.31", features = ["serde", "wasmbind"] }
+futures = "0.3"
+toml = "0.8.6"
```
panarch, line 19: `toml` is currently only used in `tests/` so we can declare it in `dev-dependencies` rather than `dependencies`.

storages/redis-storage/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+repository.workspace = true
+documentation.workspace = true
+
+[dependencies]
+gluesql-core.workspace = true
+async-trait = "0.1"
+serde = { version = "1", features = ["derive"] }
+redis = "0.23.3"
+serde_json = "1.0.105"
+chrono = { version = "=0.4.31", features = ["serde", "wasmbind"] }
```
panarch, line 17: is there any reason to specify `=0.4.31`? it would be enough to use `0.4.31`.

panarch (changes requested):

panarch (approved):
Great work 👍 Looks all nice. Thanks a lot for the contribution 👍 👍 Let's ship this to the main!

PR #1397, opened 2023-09-02
fix: Literal comparison with BinaryOperator

### LINKED ISSUE #1306: Fix Literal comparison with BinaryOperator
## Issue
- Literal comparison returns wrong result
```sql
gluesql> select 1 <= 'a';
| 1 <= 'a' |
|----------|
| TRUE     |

gluesql> select 1 >= 'a';
| 1 >= 'a' |
|----------|
| TRUE     |

gluesql> select 1 = 'a';
| 1 = 'a' |
|---------|
| FALSE   |
```

## Solve
- BinaryOperator should handle `None` case
```rs
// core/src/executor/evaluate/expr.rs
    match op {
        BinaryOperator::Plus => l.add(&r),
        BinaryOperator::Minus => l.subtract(&r),
        BinaryOperator::Multiply => l.multiply(&r),
        BinaryOperator::Divide => l.divide(&r),
        BinaryOperator::Modulo => l.modulo(&r),
        BinaryOperator::StringConcat => l.concat(r),
        BinaryOperator::Eq => cmp!(l.evaluate_eq(&r)),
        BinaryOperator::NotEq => cmp!(!l.evaluate_eq(&r)),
        BinaryOperator::Lt => cmp!(l.evaluate_cmp(&r) == Some(Ordering::Less)),
        BinaryOperator::LtEq => cmp!(l.evaluate_cmp(&r) != Some(Ordering::Greater)),
        BinaryOperator::Gt => cmp!(l.evaluate_cmp(&r) == Some(Ordering::Greater)),
        BinaryOperator::GtEq => cmp!(l.evaluate_cmp(&r) != Some(Ordering::Less)),
        BinaryOperator::And => cond!(l && r),
        BinaryOperator::Or => cond!(l || r),
        BinaryOperator::Xor => cond!(l ^ r),
    }
```
- It is recommanded to write test cases at the end of `test-suite/src/ordering.rs`
- And write `test!` macro with named parameters
```rs
// `test!` macro with named parameters
    test! (
        name: "select all from table",
        sql : "SELECT * FROM TestB",
        expected : Ok(select!(id I64; 1; 1; 3; 4))
    );

### PR DESCRIPTION
**Pull request by ding-co**
resolves #1306

### CODE REVIEW

**Comment by ever0de** [comment]
It seems better to create a branch from the latest `upstream/main` than to work from `origin/main`.
If you do this, the commit included in the PR will not have the old commit in it, which is good for review.

**Comment by ding-co** [comment]
> It seems better to create a branch from the latest `upstream/main` than to work from `origin/main`.
> 
> If you do this, the commit included in the PR will not have the old commit in it, which is good for review.

Thank you!

File: core/src/executor/evaluate/expr.rs [hunk 1]
```diff
@@ -49,9 +49,19 @@ pub fn binary_op<'a>(
         BinaryOperator::Eq => cmp!(l.evaluate_eq(&r)),
         BinaryOperator::NotEq => cmp!(!l.evaluate_eq(&r)),
         BinaryOperator::Lt => cmp!(l.evaluate_cmp(&r) == Some(Ordering::Less)),
-        BinaryOperator::LtEq => cmp!(l.evaluate_cmp(&r) != Some(Ordering::Greater)),
+        BinaryOperator::LtEq => {
+            if l.evaluate_cmp(&r).is_none() {
+                return Ok(Evaluated::from(Value::Bool(false)));
+            }
+            cmp!(l.evaluate_cmp(&r) != Some(Ordering::Greater))
+        }
```
panarch, line 57: `matches!` would be useful. same approach can be applied for `BinaryOperator::GtEq`
```rust
BinaryOperator::LtEq => cmp!(matches!(
    l.evaluate_cmp(&r),
    Some(Ordering::Less) | Some(Ordering::Equal)
)),
```

panarch (changes requested):

core/src/executor/evaluate/expr.rs
ding-co: Wow! matches! macro thank you~

panarch (approved):
Looks nice! Thanks a lot 👍

PR #1399, opened 2023-09-02
refactor: remove implementation of from<Value> trait for Evaluated

### LINKED ISSUE #1328: Remove `From<Value> for Evaluated<'a>`

tgsong827:
resolves #1328
then #1330 can be closed

### DIFF
--- a/core/src/executor/evaluate/evaluated.rs
+++ b/core/src/executor/evaluate/evaluated.rs
@@ -18,12 +18,6 @@ pub enum Evaluated<'a> {
     Value(Value),
 }
 
-impl<'a> From<Value> for Evaluated<'a> {
-    fn from(value: Value) -> Self {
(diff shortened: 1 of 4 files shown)


ever0de (approved):

panarch (approved):
That's also the nice refactoring! Thanks a lot 👍

PR #1400, opened 2023-09-02
Support Foreign key constraint

devgony:
## Disclaimer
Some Databases like mysql allow to create FK on non-unique index.
But Here I implemented to allow only unique index including PK

## Todo
- [x] add constraints to CreateTable, Schema
- [x] If there is no parent, insert should fail
- [x] check if parent has PK or unique
- [x] convert enum TableConstraint to ForeignKey. (There whould be separated field for each constraints)
- [x] merge NO ACTION and RESTRICT to NO ACTION
  - currenlty NO ACTION is default and final action on Update or Delete
- [x] on update, on delete => impl NO ACTION only in this PR
- [x] can't drop table if child exists
- [x] drop table {} cascade should drop table and dependent contraint
- [x] should fk name be required? - required
- [x] arrage errors on proper module: {delete, insert}
- [x] add test cases for self reference
- [x] handle Drop and rename column
## In Next
- optimize `fetch` to use `fetch_relation_rows`
  - implement get_table_factor
- implement Cascade actions on  UPDATE and DELETE
- ast builder for drop table if exists cascades
- with ALTER TABLE trait 
  - alter table drop column gets error
  - alter table alter column should validate type


devgony:
Thanks to @panarch,
1. convert `enum TableConstraint` to each filds of schema like fk_constraint, pk_constraint, ...
2. merge `NO ACTION` and `RESTRICT` to `NO ACTION`

core/src/ast_builder/table_name.rs
```diff
@@ -54,11 +54,19 @@ impl<'a> TableNameNode {
...
-        DropTableNode::new(self.table_name, true)
+        DropTableNode::new(self.table_name, true, false)
+    }
+
+    pub fn drop_table_cascade(self) -> DropTableNode {
+        DropTableNode::new(self.table_name, false, true)
+    }
+
+    pub fn drop_table_if_exists_cascade(self) -> DropTableNode {
+        DropTableNode::new(self.table_name, true, true)
```
devgony, line 69: should it be implemented by builder pattern?

panarch: that sounds nice.
something like these below...?

```rust
drop_table("Foo");
drop_table("Foo").if_exists();
drop_table("Foo").cascade();
drop_table("Foo").if_exists().cascade();

```

core/src/executor/alter/error.rs
```diff
@@ -47,4 +48,30 @@ pub enum AlterError {
...
     #[error("non-default argument should not follow the default argument")]
     NonDefaultArgumentFollowsDefaultArgument,
+
+    #[error("foreign table not found: {0}")]
+    ForeignTableNotFound(String),
+
+    #[error("foreign key column not found: {0}")]
+    ForeignKeyColumnNotFound(String),
+
+    #[error("foreign key column '{column}' of data type '{column_type:?}' does not match foreign column '{foreign_column}' of data type '{foreign_column_type:?}'")]
```
panarch, line 58: '{column_type:?}' and '{foreign_column_type:?}' here, `ast::DataType` has Display trait impl so we don't need to use ':?'

core/src/executor/alter/table.rs
```diff
@@ -146,8 +219,84 @@ pub async fn drop_table<T: GStore + GStoreMut>(
...
+                                    });
+                                }
+
+                                None
+                            },
+                        )
+                        .collect::<Vec<_>>()
+                },
+            )
+            .collect();
```
panarch, line 254:
```rust
let referring_children: Vec<ReferringChild> = schemas
    .into_iter()
    .flat_map(move |schema| {
        let Schema {
            table_name: referring_table_name,
            foreign_keys,
            ..
        } = schema;

        foreign_keys.into_iter().filter_map(move |foreign_key| {
            let ForeignKey {
                name: constraint_name,
                referred_table,
                ..
            } = foreign_key;

            (&referred_table == table_name && &referring_table_name != table_name)
                .then_some(ReferringChild {
                    table_name: referring_table_name.clone(),
                    constraint_name,
                })
        })
    })
    .collect();
```

I think that we can simply the codes more on this, I will try. it looks possible to simplify the below codes, too.

core/src/executor/alter/table.rs
```diff
@@ -146,8 +219,84 @@ pub async fn drop_table<T: GStore + GStoreMut>(
...
+                        .await?
+                        .ok_or_else(|| AlterError::TableNotFound(table_name.to_owned()))?;
+                    schema
+                        .foreign_keys
+                        .retain(|foreign_key| foreign_key.name != constraint_name);
+                    storage.insert_schema(&schema).await?;
+                }
+            }
+            (false, _) => {}
+        }
```
panarch, line 280: cascade check is only required at the first, so.. simple if statement would be enough.
```rust
 if !referring_children.is_empty() && !cascade {
    return Err(AlterError::CannotDropTableParentOnReferringChildren {
        parent: table_name.into(),
        referring_children,
    }
    .into());
}

for ReferringChild {
    constraint_name, ..
} in referring_children
{
    let mut schema = storage
        .fetch_schema(table_name)
        .await?
        .ok_or_else(|| AlterError::TableNotFound(table_name.to_owned()))?;
    schema
        .foreign_keys
        .retain(|foreign_key| foreign_key.name != constraint_name);
    storage.insert_schema(&schema).await?;
}
```

core/src/executor/alter/table.rs
```diff
@@ -103,12 +118,69 @@ pub async fn create_table<T: GStore + GStoreMut>(
...
+            referred_column,
+            ..
+        } = foreign_key;
+        let foreign_schema =
+            storage
+                .fetch_schema(referred_table)
+                .await?
+                .ok_or_else(|| -> Error {
+                    AlterError::ForeignTableNotFound(referred_table.to_owned()).into()
+                })?;
```
panarch, line 134:
```rust
.ok_or_else(|| AlterError::ForeignTableNotFound(referred_table.to_owned()))?
```
we can simplify the codes like this.

core/src/executor/update.rs
```diff
@@ -113,6 +120,38 @@ impl<'a, T: GStore> Update<'a, T> {
             .try_collect::<Vec<(&str, Value)>>()
             .await?;
 
+        for (id, value) in assignments.iter() {
```
panarch, line 123: these lines can be attached to the above assignments stream.
```rust
        .and_then(|(id, value)| async move {
                for foreign_key in foreign_keys {
                    let ForeignKey {
                        name,
                        column,
                        referred_table,
                        referred_column,
                        ..
                    } = foreign_key;

                    if column != id || value == Value::Null {
                        return Ok((id, value));
                    }

                    let no_parent = self
                        .storage
                        .fetch_data(referred_table, &Key::try_from(&value)?)
                        .await?
                        .is_none();

                    if no_parent {
                        return Err(ValidateError::ForeignKeyViolation {
                            name: name.to_owned(),
                            table: table_name.to_owned(),
                            column: column.to_owned(),
                            referred_table: referred_table.to_owned(),
                            referred_column: referred_column.to_owned(),
                        }
                        .into());
                    }
                }

                Ok((id, value))
            })
```

core/src/executor/insert.rs
```diff
@@ -38,6 +38,13 @@ pub enum InsertError {
 
     #[error("map type required: {0}")]
     MapTypeValueRequired(String),
+
+    #[error("cannot find referenced value on {table_name}.{column_name} with value {referenced_value:?}")]
+    CannotFindReferencedValue {
```
devgony, line 43: Is it better to separate Error even though it's identical with UpdateError::CannotFindReferencedValue ?

panarch: though the name is same, if the location is different, it needs to be separated. this module based error location makes us simple to add new errors.
and they are also different because the one is from INSERT and the other is from UPDATE.

core/src/executor/execute.rs
```diff
@@ -245,9 +269,50 @@ async fn execute_inner<T: GStore + GStoreMut>(
             selection,
         } => {
             let columns = fetch_columns(storage, table_name).await?.map(Rc::from);
+            let referencings = storage.fetch_referencings(table_name).await?;
```
panarch, line 272: now `delete` also look big enough to be splitted into a separate module as we do for `insert` and `update`. how about adding `delete` module and move these codes to that?

core/src/ast/mod.rs
```diff
@@ -7,13 +7,15 @@ mod operator;
 mod query;
 
 pub use {
+    crate::error::Result,
```
panarch, line 10: this and `Parser` are unnecessary.

core/src/executor/mod.rs
```diff
@@ -16,6 +16,7 @@ mod validate;
 pub use {
     aggregate::AggregateError,
     alter::AlterError,
+    alter::Referencing,
```
panarch, line 19: nit. we can merge these

core/src/executor/update.rs
```diff
@@ -4,8 +4,9 @@ use {
         evaluate::{evaluate, Evaluated},
     },
     crate::{
-        ast::{Assignment, ColumnDef, ColumnUniqueOption},
+        ast::{Assignment, ColumnDef, ColumnUniqueOption, ForeignKey},
         data::{Row, Value},
+        prelude::Key,
```
panarch, line 9: we may not need to use `prelude` in internal codes. we can use `data::Key` instead

storages/mongo-storage/src/utils.rs
```diff
@@ -1,8 +1,10 @@
 use {
-    crate::ResultExt,
+    crate::{description::TableDescription, error::ResultExt},
     bson::{doc, Document},
+    gluesql_core::ast::ForeignKey,
     gluesql_core::{ast::ColumnDef, error::Result},
```
panarch, line 5: nit. these can be merged

core/src/executor/alter/table.rs
```diff
@@ -103,12 +118,79 @@ pub async fn create_table<T: GStore + GStoreMut>(
...
+
+        let referenced_column_def = column_defs
+            .and_then(|column_defs| {
+                column_defs
+                    .into_iter()
+                    .find(|column_def| column_def.name == *referenced_column_name)
+            })
+            .ok_or_else(|| -> Error {
+                AlterError::ForeignKeyColumnNotFound(referenced_column_name.to_owned()).into()
+            })?
```
panarch, line 150: ```.ok_or_else(|| AlterError::ForeignKeyColumnNotFound(referenced_column_name.to_owned()))?```

core/src/executor/alter/table.rs
```diff
@@ -103,12 +118,79 @@ pub async fn create_table<T: GStore + GStoreMut>(
...
+        let referencing_column_def = target_columns_defs
+            .as_deref()
+            .and_then(|column_defs| {
+                column_defs
+                    .iter()
+                    .find(|column_def| column_def.name == *referencing_column_name)
+            })
+            .ok_or_else(|| -> Error {
+                AlterError::ForeignKeyColumnNotFound(referencing_column_name.to_owned()).into()
+            })?;
```
panarch, line 162: this, too.
```rust
.ok_or_else(|| AlterError::ForeignKeyColumnNotFound(referenced_column_name.to_owned()))?
```

core/src/executor/validate.rs
```diff
@@ -26,6 +26,15 @@ pub enum ValidateError {
 
     #[error("duplicate entry '{0:?}' for primary_key field")]
     DuplicateEntryOnPrimaryKeyField(Key),
+
+    #[error("foreign key `{name}` violated - referencing: {table}.{column}, referenced: {referenced_table}.{referenced_column}")]
+    ForeignKeyViolation {
```
panarch, line 31: it looks that this error is not used.

core/src/executor/execute.rs
```diff
@@ -28,6 +30,12 @@ use {
 pub enum ExecuteError {
     #[error("table not found: {0}")]
     TableNotFound(String),
+
+    #[error("referencing column exists: {0}")]
+    ReferencingColumnExists(String),
+
+    #[error("Value not found on column: {0}")]
+    ValueNotFound(String),
```
panarch, line 38: now we can add `DeleteError` and move these errors into that.

core/src/executor/insert.rs
```diff
@@ -178,6 +200,53 @@ async fn fetch_vec_rows<T: GStore>(
     }
 }
 
+async fn validate_foreign_key<T: GStore>(
+    foreign_keys: Vec<ForeignKey>,
+    rows: &[Vec<Value>],
+    storage: &T,
```
panarch, line 206: it would be more consistent to set storage as the first parameter.

panarch: then, from environmental params to target params
storage, column_defs, foreign_keys and rows

core/src/executor/insert.rs
```diff
@@ -178,6 +200,53 @@ async fn fetch_vec_rows<T: GStore>(
...
+        let ForeignKey {
+            referencing_column_name,
+            referenced_table_name,
+            referenced_column_name,
+            ..
+        } = &foreign_key;
+        if let Some(target_index) = column_defs
+            .iter()
+            .enumerate()
+            .find(|(_, c)| &c.name == referencing_column_name)
```
panarch, line 219: this needs to be error than filtering out.

core/src/executor/delete.rs
```diff
@@ -0,0 +1,73 @@
...
+                        op: BinaryOperator::Eq,
+                        right: Box::new(value.to_owned().try_into()?),
+                    };
+
+                    let columns = Some(vec![referencing_column_name.to_owned()]).map(Rc::from);
+                    let referencing_rows =
+                        fetch(storage, referencing_table_name, columns, Some(expr)).await?;
+
+                    let len = referencing_rows.count().await;
+                    if len > 0 && on_delete == &ReferentialAction::NoAction {
```
panarch, line 55: what we need is only the existence of referencing_row. then we dont need to count. rather than count, it would be enough to use `next` instead.

core/src/executor/delete.rs
```diff
@@ -0,0 +1,73 @@
...
+                        ExecuteError::ValueNotFound(referenced_column_name.to_owned()),
+                    )?;
+
+                    let expr = &Expr::BinaryOp {
+                        left: Box::new(Expr::Identifier(referencing_column_name.clone())),
+                        op: BinaryOperator::Eq,
+                        right: Box::new(value.to_owned().try_into()?),
+                    };
+
+                    let columns = Some(vec![referencing_column_name.to_owned()]).map(Rc::from);
```
panarch, line 50: we don't need fetching actual data here so
```rust
let columns = Some(Rc::from(Vec::new()));
```
would be enough.

core/src/executor/delete.rs
```diff
@@ -0,0 +1,73 @@
...
+                {
+                    let value = row.get_value(referenced_column_name).ok_or(
+                        ExecuteError::ValueNotFound(referenced_column_name.to_owned()),
+                    )?;
+
+                    let expr = &Expr::BinaryOp {
+                        left: Box::new(Expr::Identifier(referencing_column_name.clone())),
+                        op: BinaryOperator::Eq,
+                        right: Box::new(value.to_owned().try_into()?),
+                    };
```
panarch, line 48:
```rust
let value = row
    .get_value(referenced_column_name)
    .ok_or(ExecuteError::ValueNotFound(
        referenced_column_name.to_owned(),
    ))?
    .clone();

let expr = &Expr::BinaryOp {
    left: Box::new(Expr::Identifier(referencing_column_name.clone())),
    op: BinaryOperator::Eq,
    right: Box::new(Expr::try_from(value)?),
};
```

for value, use of `clone` would be better than `to_owned` because in this case, what we do is converting `&Value` to `Value` which is `&T` to `T`.
we can use to_owned for `&T` to `U` case.

core/src/executor/update.rs
```diff
@@ -110,6 +117,36 @@ impl<'a, T: GStore> Update<'a, T> {
...
+                for foreign_key in foreign_keys {
+                    let ForeignKey {
+                        referencing_column_name,
+                        referenced_table_name,
+                        referenced_column_name,
+                        ..
+                    } = foreign_key;
+                    if referencing_column_name != id || value == Value::Null {
+                        return Ok((id, value));
+                    }
```
panarch, line 130: 1. `value == Value::Null` check can be done before the foreign key for loop?
2. we may need.. `continue` instead of `return`?

core/src/executor/alter/table.rs
```diff
@@ -103,12 +118,79 @@ pub async fn create_table<T: GStore + GStoreMut>(
...
+                referenced_column_type: referenced_column_def.data_type.to_owned(),
+            }
+            .into());
+        }
+
+        if referenced_column_def
+            .unique
+            .map(|x| !x.is_primary)
+            .unwrap_or(true)
+        {
```
panarch, line 178: would it be simple with..
```rust
        if referenced_column_def.unique != Some(ColumnUniqueOption { is_primary: true }) {
```
like this above?

core/src/executor/insert.rs
```diff
@@ -178,6 +203,60 @@ async fn fetch_vec_rows<T: GStore>(
...
+    rows: &[Vec<Value>],
+) -> Result<()> {
+    for foreign_key in foreign_keys {
+        let ForeignKey {
+            referencing_column_name,
+            referenced_table_name,
+            referenced_column_name,
+            ..
+        } = &foreign_key;
+        if let Some(target_index) = column_defs
```
panarch, line 219:
```rust
let target_index = column_defs
            .iter()
            .enumerate()
            .find(|(_, c)| &c.name == referencing_column_name)
            .ok_or_else(|| {
                InsertError::UnreachableReferencingColumnName(referencing_column_name.to_owned())
            })?;
```

nit. we may use less indent here.

panarch (commented):
wow.. I only left nit comment about renaming error name prefixes.
Everything else looks great!
now it really comes to the merge soon. awesome

panarch (approved):
It is such a pleasure to see this! Amazing work you've done. Now we support FOREIGN KEY!
Thanks really a lot for the contribution. Looks all nice! Let's ship this to the main 🥇 👍 👍

PR #1401, opened 2023-09-02
chore: implement `ConvertError`,

### LINKED ISSUE #1390: Make `impossible cast` error more meaningful
During an AST constructed select I am getting `value: impossible cast` without any further hints which makes it hard to debug.

```rust
let payload = table(TABLE_NAME)
    .select()
    .filter(format!("year = {year}"))
    .execute(glue)
    .await;
```

ever0de:
resolved #1390 

### TODO 

Create a parent error that also tracks which column, ast, ... it occurs in.


panarch (approved):
That's brilliant to split errors by adding a new local one.
ConvertError looks great! Thanks a lot 👍 👍

PR #1404, opened 2023-09-05
chore: remove `Result` from `ast_builder::transaction` return type

ever0de:


panarch (approved):
This looks nice 👍 👍 Thanks a lot for the good catch!

PR #1415, opened 2023-09-28
Reflect Deprecation of `from_utc` in Crate `chrono`

zmrdltl:
Resolved Clippy warning
Ref. https://docs.rs/chrono/latest/src/chrono/datetime/mod.rs.html#130


panarch (approved):
Thanks a lot 👍 👍

PR #1416, opened 2023-09-28
Fix parsing of BigDecimal literals with zero fraction part as floats, not ints

### LINKED ISSUE #1309: Float literals without fractional part (ex. 42.0, 0.00) are intepreted as integers
Currently, all float literals without fractional part are interpreted as integer, not floats.

More specifically, the following tests have passed:
```rs
    test!("SELECT 0.0", Ok(select_with_null!("0.0"; I64(0))));
    test!("SELECT 1.0", Ok(select_with_null!("1.0"; I64(1))));
    test!("SELECT 42.0", Ok(select_with_null!("42.0"; I64(42))));
    test!("SELECT 1.1", Ok(select_with_null!("1.1"; F64(1.1))));
```

The issue becomes more apprent when using the GCD(i64, i64) function, which further confirms that float literals are being treated as integers:

```sql
gluesql> SELECT gcd(1, 1);
| gcd(1, 1) |
|-----------|
| 1         |

gluesql> SELECT gcd(1.0, 1);
| gcd(1.0, 1) |
|-------------|
| 1           |

gluesql> SELECT gcd(1.0, 1.1);
[error] evaluate: function requires integer value: GCD
```

zmrdltl:
## Background
Resolves #1309 

## Problem
Direct use of `to_i64` on `bigdecimal::BigDecimal` literals resulted in floating-point numbers like `2.0` being mistakenly interpreted as integers.

## Solution
We improved the mechanism to determine if a given `BigDecimal` is an integer or a floating-point number:

### Initial Approach:
Initially, we determined the type of number by comparing its digit count with the length of its string representation.

```rust
fn is_integer_representation(bigdecimal_val: &BigDecimal) -> bool {
    let bigdecimal_val_str = bigdecimal_val.abs().to_string();
    bigdecimal_val.abs().digits() == bigdecimal_val_str.len() as u64
}
```

### Enhanced Solution (Already Applied):
With the introduction of the [get_scale](https://github.com/akubera/bigdecimal-rs/pull/116) method, we optimized the process by directly accessing the scale of a `BigDecimal`. This method efficiently discerns between integers and numbers with fractional components.

```rust
fn is_integer_representation(bigdecimal_val: &BigDecimal) -> bool {
    bigdecimal_val.get_scale() == 0
}
```

## Code Update
The parsing approach for `Literal::Number` was adjusted based on the enhanced solution:

```rust
impl TryFrom<&Literal<'_>> for Value {
    // ... rest of the implementation ...
    fn try_from(literal: &Literal<'_>) -> Result<Self> {
        match literal {
            Literal::Number(bigdecimal_val) => {
                if is_integer_representation(bigdecimal_val) {
                    convert_to_i64(bigdecimal_val)
                } else {
                    convert_to_f64(bigdecimal_val)
                }
            },
            // ... other match arms ...
        }
    }
}
```

## Tests
Tests have been added, as referenced in the original issue, to validate the applied solution.

## Further Enhancement

In our initial refactoring efforts, we had made use of an `is_integer()` method to determine if a `BigDecimal` was an integer or not. Post our code enhancements, we have now replaced the usage of the `is_integer()` function with the more efficient `is_integer_representation()` method. 

The new method's implementation is as follows:

```rust
fn is_integer_representation(&self) -> bool {
    self.fractional_digit_count() == 0
}
```

This refined approach ensures a more direct and accurate check for integer representation, optimizing both readability and performance. We have updated all relevant parts of our codebase to reflect this change, ensuring a consistent and efficient mechanism to handle `BigDecimal` numbers.

## Conclusion

By leveraging the capabilities of the `BigDecimal` crate and our improved detection mechanism, we have enhanced the accuracy and efficiency of our number type determination process. The updates and tests ensure that our implementation correctly interprets floating-point numbers and integers, addressing the issues identified in #1309.


panarch (approved):
That's nice improvement! Thanks a lot 👍 👍

PR #1419, opened 2023-10-02
Change store RowIter type from Iterator to Stream

### LINKED ISSUE #1410: Change store `RowIter` type from `Iterator` to `Stream`
#### Background:
Currently, the return type of the data retrieval in the storage layer of the gluesql project is an `Iterator`. This design has limitations when integrating with asynchronous storage systems like Redis, where using a `Stream` would be more appropriate.

#### Problem:
The current `Iterator` return type doesn't allow us to pass data as a stream directly from storage layers that support streaming, such as Redis. As a result, the performance and efficiency of the system can be compromised.

#### Proposed Solution:
It would be beneficial to update the `RowIter` type from `Iterator` to `Stream` to enable better integration with storage systems that support asynchronous data retrieval and streaming.

#### Benefits:
- Improved performance by leveraging async capabilities of storage systems.
- More flexibility in storage implementations, allowing for more efficient data retrieval methods.
- Easier to integrate with modern async libraries and frameworks.

This change would make the project more flexible and better optimized for various storage backends.

panarch:
resolve #1410

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -956,7 +956,6 @@ dependencies = [
  "gluesql-json-storage",
  "gluesql_memory_storage",
  "gluesql_sled_storage",
- "itertools",
  "rustyline",
  "rustyline-derive",
  "tabled",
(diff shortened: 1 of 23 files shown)


devgony (approved):
Thank you ! It could improve the performance of mongodb storage a lot !

PR #1423, opened 2023-10-15
docs:Add doc about SPLICE function

jinlee0:
Add doc about SPLICE function

### DIFF
--- a/docs/docs/sql-syntax/functions/list-map/splice.md
+++ b/docs/docs/sql-syntax/functions/list-map/splice.md
@@ -3,44 +3,30 @@
 The "SPLICE" function in GlueSQL is used to modify elements in a list. It allows you to remove or replace elements in a list. The syntax for the "SPLICE" function is as follows:
 
 ```sql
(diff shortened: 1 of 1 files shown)


docs/docs/sql-syntax/functions/list-map/splice.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+Now, we can use the "SPLICE" function to modify the list in various ways:
+
+1. Remove elements from a list:
+
+```sql
+-- Remove elements at positions 1 to 3 from the list '[1, 2, 3]'
+SELECT SPLICE(CAST('[1, 2, 3, 4, 5]' AS List), 1, 4) AS result;
+-- Output: '[1, 4, 5]'
```
panarch, line 37: it looks.. it should be `1, 3`, not `1, 4`

jinlee0: Thank you!

panarch (approved):
Looks all nice. Thanks a lot 👍

PR #1427, opened 2023-10-15
Upgrade to chrono v0.4.31 and adjust millisecond value in Timestamp Creation

zmrdltl:
## Cause
After upgrading to version 0.4.31, it was observed that the `and_hms_milli_opt()` function considers values of 1000 milliseconds (equivalent to 1 second) as potentially invalid unless representing a leap second with `sec == 59`. For the date `1989-03-20`, 1000 milliseconds was not a valid input.

## Resolution
Fixed the millisecond value in the `NaiveDate::from_ymd_opt()` function by changing it from `1000` to `999`, adjusting to the maximum valid millisecond value of `999`.


panarch (approved):
👍 👍  Thanks a lot!

PR #1433, opened 2023-10-29
test: add test cases for astb-padding

devgony:
## Goal
- LPAD: Left-pads a string with a specified string up to a certain length.
- RPAD: Right-pads a string with a specified string up to a certain length.

## Todo
- [x] test: add test cases for astb-padding


panarch (approved):
Looks nice!! Thanks a lot 👍

PR #1434, opened 2023-10-29
docs: write docmentation for padding

devgony:
## Todo
- [x] write docmentation for ast-builder/functions/text/padding


panarch (approved):
This looks also nice :) Thanks a lot!

Issue #1402, opened 2023-09-04
Support StoreMut trait to Optional


seonghun-dev:
## Goal
Support StoreMut trait to Optional
If sql is implemented read-only, there is no need to implement storemut, so we will change it to allow you to selectively implement StoreMut trait

PR #1435, opened 2023-10-29
Support StoreMut trait to Optional 

seonghun-dev:
- Resolve : https://github.com/gluesql/gluesql/issues/1402


panarch (approved):
Looks nice :)
By adding default impl codes, we can provide clear guide that `StoreMut` is also optional.
Thanks a lot!

PR #1440, opened 2023-11-18
chore: bump `[sqlparser, itertools, stum_macros, ordered-float]`

ever0de:
https://github.com/ever0de/gluesql/blob/68795078d85068f2b63d8e652e4c3f8f9940caf8/core/src/translate/expr.rs#L116-L119
> https://github.com/sqlparser-rs/sqlparser-rs/blob/ff8312bfd8983abac26df1eb60b4c77133b1901a/src/parser/mod.rs#L1436-L1446

This can only be reached by Snowflake, BigQuery, and Generic. Currently we are using [PostgreSqlDialect](https://github.com/ever0de/gluesql/blob/68795078d85068f2b63d8e652e4c3f8f9940caf8/core/src/parse_sql.rs#L15-L16), so it is not reachable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2363,9 +2363,9 @@ dependencies = [
 
 [[package]]
 name = "sqlparser"
-version = "0.36.1"
+version = "0.39.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


panarch (approved):
👍 👍 Looks all nice! Thanks a lot!

Issue #1444, opened 2023-12-16
Implement Elixir binding for GlueSQL storages


MRGRAVITY817:
## Background

Elixir, a functional programming language which runs above BEAM(Erlang virtual machine), has earned its reputation for being scalable, fault-tolerant and reliable in distributed systems. The language itself has ranked 2nd most loved programming language, and web framework Phoenix is ranked as the most loved framework in 2023 StackOverflow Developer Survey.   

Discord, which is one of the pioneers who adopted Elixir from day one, has posted a blog article about how they had leveraged the performance of their Elixir-based infrastructure by adopting Rust library into it. For further details, check this [link](https://discord.com/blog/using-rust-to-scale-elixir-for-11-million-concurrent-users).    

As more Rust libraries being adopted in Elixir projects like this case, I strongly believe introducing GlueSQL storages in Elixir ecosystem will show another creative way to manage data in web/and various other projects.  

## Suggestion

It would be nice to start with memory storage. Elixir/Erlang already has it's own default memory storage called [ETS](https://www.erlang.org/doc/man/ets.html), which can be used with the query builder [ETSO](https://hexdocs.pm/etso/readme.html). It's a very handy stuff, but not that performant. Making comparison between ETS and memory storage in BEAM will be very interesting.

PR #1445, opened 2023-12-16
feat: Implement Elixir Binding

MRGRAVITY817:
This PR resolves issue #1444 

## The goal of this PR
- Implement Elixir binding for GlueSQL memory storage
- Build a good foundation to add more storages in near future


pkg/elixir/test/memory_storage_test.exs
```diff
@@ -0,0 +1,167 @@
+defmodule GlueSQLTest do
+  use ExUnit.Case
+  doctest GlueSQL
+
+  alias GlueSQL.Storages.MemoryStorage
+  alias GlueSQL
```
dkuku, line 6: This is not needed in elixir.

dkuku:
I love this idea. Have you maybe checked serde_rustler for sending the data back to the beam?
That's the most updated fork I found
https://github.com/benhaney/serde_rustler/
I did try it in my first rust code and it looks promising - I can even receive elixir structs from rustler:
https://github.com/dkuku/fitparser_ex/

MRGRAVITY817:
> I love this idea. Have you maybe checked serde_rustler for sending the data back to the beam instead of serializing it to json? That's the most updated fork I found https://github.com/benhaney/serde_rustler/ I did try it in my first rust code and it looks promising - I can even receive elixir structs from rustler: https://github.com/dkuku/fitparser_ex/

@dkuku Oh, actually the main reason that I didn't used `serde_rustler` was because original repo wasn't being maintained for a while. Thanks, I'll try update my code with this.

lu-zero:
Why this cannot live in a separate repo?

Also, unreproducible binaries on the repo would get the crate flagged.

MRGRAVITY817:
Hi @lu-zero !

> Why this cannot live in a separate repo?

Actually, it can live in a separate repo. 

I first thought it was a good idea to include this in this repo (since javascript/python version is also here as well), but that is not mandatory.

> Also, unreproducible binaries on the repo would get the crate flagged.

Yeah, this also makes me think if it should be created in separate repo. Thanks for the advice 👍

MRGRAVITY817:
@panarch What are your thoughts about creating a separate repo for this rather than putting inside `pkg` folder?  

@lu-zero has pointed out some valid insight, and I also think that if we keep adding language bindings it would be too much of a burden for maintainer & reviewers because they also have to learn a new language.

panarch:
@lu-zero raised a valid concern about increasing complexity in the gluesql repository with additional language bindings. Given the resource constraints, continuously managing new bindings like Elixir internally seems challenging.

I propose creating a separate repository for Elixir support to maintain a clean and focused main repository. Documentation can be linked from gluesql.org/docs for consistency.

@MRGRAVITY817, could you share your thoughts on this approach?

MRGRAVITY817:
@panarch Yup, I'll just create a new repo and add a link in README once it's done. Thanks!

PR #1463, opened 2024-03-02
Exclude tests {mongo, redis} with features

devgony:
## Todo
- [x] exclude test cases of redis-storage with `test-redis` feature
- [x] fix partial excluding test cases of mongo-storage to exclude all cases


panarch (approved):
Looks nice! thanks a lot 👍

PR #1464, opened 2024-03-02
Fix insert_schema to overwrite if exsits on {MongoStorage, IdbStorage}

devgony:
## Todo
- [x] Fix insert_schema to overwrite if exsits on MongoStorage
- [x] refactor: impl Validator 
- [x] add test-suite/store case to test insert_schema consecutively

### DIFF
--- a/storages/idb-storage/src/lib.rs
+++ b/storages/idb-storage/src/lib.rs
@@ -287,18 +287,18 @@ impl StoreMut for IdbStorage {
             return Err(e);
         }
 
-        let transaction = self
-            .database
-            .transaction(&[SCHEMA_STORE], TransactionMode::ReadWrite)
-            .err_into()?;
-        let store = transaction.object_store(SCHEMA_STORE).err_into()?;
-
         let schema_exists = self
             .fetch_schema(&schema.table_name)
             .await
             .map_err(|e| e.to_string())
             .map_err(Error::StorageMsg)?
             .is_some();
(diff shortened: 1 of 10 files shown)


storages/mongo-storage/tests/mongo_insert_schema.rs
```diff
@@ -0,0 +1,57 @@
...
+            data_type: DataType::Text,
+            nullable: false,
+            default: None,
+            unique: None,
+        });
+
+        column_defs
+    });
+
+    storage.insert_schema(&schema).await.unwrap();
```
devgony, line 48: where should i put this `consecutive insert_schema test` to run on all storages? 🤔

storages/idb-storage/src/lib.rs
```diff
@@ -476,6 +376,100 @@ impl StoreMut for IdbStorage {
...
 
+enum RequestType {
+    InsertSchema,
+    DeleteSchema,
+}
+impl IdbStorage {
+    async fn upgrade_version(
+        &mut self,
+        table_name: String,
+        request_type: RequestType,
```
devgony, line 387: Is there other way to recognize the RequestType without using enum?

panarch (approved):
Thanks a lot for the contribution!
it is also meaningful that we now have test cases for verifying store trait impl directly without using sql.
looks all great, thanks!

PR #1467, opened 2024-03-30
Update ast_builder/insert.rs to export only node structs

panarch:


devgony (approved):
It might be possible to add client-side clippy check to CI?

PR #1469, opened 2024-04-09
Add column comment support on CREATE TABLE,

panarch:
Add comment to ast::ColumnDef.
Update translate_column_def to handle comment ColumnOption.
Update MongoStorage validator description field to support both DEFAULT and COMMENT.
Update ParquetStorage to support column comment.
Modify test-suite store/insert_schema test case to use comment field.

e.g.
```sql
CREATE TABLE Foo (
  id TEXT NULL COMMENT 'comment something'
);
```

### DIFF
--- a/storages/mongo-storage/src/store.rs
+++ b/storages/mongo-storage/src/store.rs
@@ -21,6 +21,7 @@ use {
         options::{FindOptions, ListIndexesOptions},
         IndexModel,
     },
+    regex::Regex,
     std::{collections::HashMap, future},
 };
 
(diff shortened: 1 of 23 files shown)


storages/mongo-storage/src/store.rs
```diff
@@ -231,9 +232,25 @@ impl MongoStorage {
...
+
+                                (default, comment)
+                            })
+                        })
+                        .unwrap_or((None, None));
+
+                    let default = default
                         .map(parse_expr)
                         .map(|expr| expr.and_then(|expr| translate_expr(&expr)))
                         .transpose()?;
```
devgony, line 256: How about using serde with new `ColumnDescription` struct?
In #1400 , I was using serde for `TableDescription` (There are two kinda description fields; table and column).
It would be also helpful to extend more features of column someday.

```rust
      let column_description = doc.get_str("description");
      let ColumnDescription { default, comment } = match column_description {
          Ok(desc) => {
              serde_json::from_str::<ColumnDescription>(desc).map_storage_err()?
          }
          Err(ValueAccessError::NotPresent) => ColumnDescription {
              default: None,
              comment: None,
          },
          Err(_) => {
              return Err(Error::StorageMsg(
                  MongoStorageError::InvalidGlueType.to_string(),
              ))
          }
      };
```
```rust
// storages/mongo-storage/src/description.rs
use gluesql_core::ast::Expr;
use serde::{Deserialize, Serialize};

#[derive(Serialize, Deserialize)]
pub struct ColumnDescription {
    pub default: Option<Expr>,
    pub comment: Option<String>,
}
```
and then,

(⚠️ 🦀 `fold` need to return result.. )
```rust
// storages/mongo-storage/src/store_mut.rs
    let column_description = description::ColumnDescription {
        default: column_def.default.clone(),
        comment: column_def.comment.clone(),
    };

    let description = to_string(&column_description).map_storage_err()?;

    property.extend(doc! {
        "description": description,
    });
```

devgony (approved):
👍👍👍

PR #1470, opened 2024-04-13
Replace rust-toolchain to use rust-toolchain.toml, TOML based

panarch:


ever0de:
https://github.com/gluesql/gluesql/blob/bc5b7cbcbd00f0cc3e1cd2a3111c4c0932c5de55/.github/workflows/rust.yml#L22-L29

You have an action step that reads the `rust-toolchain` file to set the Rust version. You need to make modifications here.

ever0de (approved):

Issue #1474, opened 2024-04-17
Payload does not implement Clone


LucaCappelletti94:
Hi - at this time [`Payload`](https://github.com/gluesql/gluesql/blob/fccf1333f4da00dc25baeaa944635063cd585e29/core/src/executor/execute.rs#L34) does not implement the trait [`Clone`](https://doc.rust-lang.org/core/clone/index.html).

I do not see any reason why this should be the case, therefore I will be adding this in a pull request shortly.

PR #1475, opened 2024-04-17
Added Clone trait to Payload

LucaCappelletti94:
Added the trait [`Clone`](https://doc.rust-lang.org/core/clone/index.html) to [`Payload`](https://github.com/gluesql/gluesql/blob/fccf1333f4da00dc25baeaa944635063cd585e29/core/src/executor/execute.rs#L34).

This pull request resolves issue #1474


panarch (approved):
There is no reason not to have Clone for Payload.
Thanks a lot!

PR #1476, opened 2024-04-19
Add table comment support on CREATE TABLE

panarch:


devgony (commented):
It would be better if it supports `GLUE_TABLES.COMMENT` as well 👍🏼

core/src/executor/fetch.rs
```diff
@@ -260,7 +260,10 @@ pub async fn fetch_relation_rows<'a, T: GStore>(
                         let rows = schemas.into_iter().map(move |schema| {
                             Ok(Row::Vec {
                                 columns: Rc::clone(&columns),
-                                values: vec![Value::Str(schema.table_name)],
+                                values: vec![
+                                    Value::Str(schema.table_name),
+                                    Value::Str(schema.comment.unwrap_or_default()),
+                                ],
```
devgony, line 266: currently user may be confused with NULL and `""` .

gluesql> create table x (id int) comment = 'a';
Table created

gluesql> create table y (id int);
Table created

gluesql> select * from glue_tables;
| TABLE_NAME | COMMENT |
|------------|---------|
| x          | a       |
| y          |         |

gluesql> select * from glue_tables where COMMENT is null;
| TABLE_NAME | COMMENT |
|------------|---------|

If we direclty fetch NULL, it would be more clear

```rust
let rows = schemas.into_iter().map(move |schema| {
    let comment = match schema.comment {
      Some(comment) => Value::Str(comment),
      None => Value::Null,
    };
    
    Ok(Row::Vec {
      columns: Rc::clone(&columns),
      values: vec![
          Value::Str(schema.table_name),
          comment
      ],
    })
```

devgony: We missed this point in #1469 as well.. 😢

panarch: oh thanks! for now, I updated GLUE_TABLES and I will also work on GLUE_TABLE_COLUMNS in another PR. it would be also good to apply same stuff to DEFAULT.

devgony (approved):

PR #1479, opened 2024-04-28
Make some columns in GLUE_TABLE_COLUMNS nullable,

panarch:
apply to KEY, DEFAULT and COMMENT columns.


devgony (approved):
Thank you for fixing my mistake on `default` as well 👍🏼

Issue #1481, opened 2024-05-09
Adding bytea method for the AST builder


LucaCappelletti94:
Hi - I believe it is necessary to add an additional AST builder method analogous to the `uuid` one reported below for BYTEA data types.

I will be adding it in a pull request shortly.

https://github.com/gluesql/gluesql/blob/4917931e643a9a1b3d91dc413a2af2ba71906ea5/core/src/ast_builder/expr/mod.rs#L365-L371

PR #1482, opened 2024-05-09
Added bytea to AST builder

LucaCappelletti94:
This pull request resolves issue #1481.

Looking how bytea is encoded/decoded internally, I hope I have used the correct method but I was not able to figure out where I should put a test to actually evaluate whether this method inserts the expected values.


LucaCappelletti94:
@panarch all passed, added the generic as requested

panarch (approved):

PR #1498, opened 2024-05-25
refactor: simplify `Value::validate_type()`

MRGRAVITY817:
Most of the match case in function `Value::validate_type()` in `core/src/data/value/mod.rs` file can be simplified.


panarch (approved):
Nice refactoring work! Thanks a lot 👍 👍

PR #1504, opened 2024-06-14
Fix panic in SledStorage tx_timeout calculation by re-arranging terms

### LINKED ISSUE #1499: Failed to deal with non-monotonic SystemTime. attempt to subtract with overflow
thread panicked at 'attempt to subtract with overflow':gluesql_sled_storage-0.15.0/src/lock.rs:137
```rust
 let now = SystemTime::now()
        .duration_since(UNIX_EPOCH)
        .map_err(err_into)
        .map_err(ConflictableTransactionError::Abort)?
        .as_millis();
    if tx_timeout.map(|tx_timeout| now - created_at >= tx_timeout) == Some(true) {

```

here `now` is an unsized integer, and it may overflow and cause panic or false-positive timeout

[std](https://doc.rust-lang.org/std/index.html)::[time](https://doc.rust-lang.org/std/time/index.html)::[SystemTime](https://doc.rust-lang.org/std/time/struct.SystemTime.html#)

Distinct from the [Instant](https://doc.rust-lang.org/std/time/struct.Instant.html) type, this time measurement is not monotonic. This means that you can save a file to the file system, then save another file to the file system, and the second file has a SystemTime measurement earlier than the first. In other words, an operation that happens after another operation in real time may have an earlier SystemTime!

JakkuSakura:
fixes #1499

### DIFF
--- a/storages/sled-storage/src/index.rs
+++ b/storages/sled-storage/src/index.rs
@@ -93,7 +93,7 @@ impl Index for SledStorage {
                 ));
             }
         };
-        let lock_txid = lock::fetch(&self.tree, txid, created_at as i64, self.tx_timeout)?;
+        let lock_txid = lock::fetch(&self.tree, txid, created_at, self.tx_timeout)?;
 
         let prefix_len = build_index_key_prefix(table_name, index_name).len();
         let tree = self.tree.clone();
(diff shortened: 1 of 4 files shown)


panarch (approved):
Looks all nice! 

Issue #1514, opened 2024-07-10
Multi-key primary keys are not supported


LucaCappelletti94:
Hi all,

I just received the following error when trying to create a table that has a composite primary key.

Am I correct in assuming that multiple primary keys are not supported at this time?

What would be necessary to add support for this feature? I am willing to try and add it.

```
Translate(UnsupportedConstraint("PRIMARY KEY (table_id, user_id)"))
```

panarch:
Yes, you are correct that composite primary keys are not supported at this time. Implementing this feature requires significant changes in the core areas, including the AST, planner, and execution layer. This can be a challenging issue, especially for someone just starting out. However, feel free to give it a try, and if you have any questions, leave a comment, and we will support you.

There are multiple storage options, but I recommend focusing on the memory storage first and ignoring the other storage types for now. If you can successfully implement the feature for one storage, it will not be too difficult to extend it to the others.

LucaCappelletti94:
Ok! Which files should I check out first?

LucaCappelletti94:
Fortunately, in the SQL parser, I see that the multi-column primary keys are already supported. Hopefully, no edits will be needed there.

LucaCappelletti94:
In this enum, shouldn't there be more options? 

https://github.com/gluesql/gluesql/blob/25309dd15a50ee061e714576079b2cd8c1fcca9c/core/src/ast/mod.rs#L43-L46

LucaCappelletti94:
Where in the code is the unique constraint enforced currently?

panarch:
> In this enum, shouldn't there be more options?
> 
> https://github.com/gluesql/gluesql/blob/25309dd15a50ee061e714576079b2cd8c1fcca9c/core/src/ast/mod.rs#L43-L46

it can have more, but not yet supported.

panarch:
> Where in the code is the unique constraint enforced currently?

here. it includes both clustered (primary key) and non-clustered index validation codes. it would be enough to focus on primary key related part.
https://github.com/gluesql/gluesql/blob/main/core/src/executor/validate.rs

leejw51:
0.15.0 is fine for `Multi-key primary key`, is it ok to use multikey primary key on 0.15.0?

devgony:
> 0.15.0 is fine for `Multi-key primary key`, is it ok to use multikey primary key on 0.15.0?

We don't support the feature yet, can you share your schema? (or CREATE TABLE SQL)

PR #1515, opened 2024-07-11
Support for multiple primary keys

LucaCappelletti94:
Pull request to address the issue #1514

### DIFF
--- a/core/src/ast/ddl.rs
+++ b/core/src/ast/ddl.rs
@@ -84,11 +84,7 @@ impl ToSql for ColumnDef {
             let default = default
                 .as_ref()
                 .map(|expr| format!("DEFAULT {}", expr.to_sql()));
-            let unique = if *unique {
-                Some("UNIQUE".to_owned())
-            } else {
-                None
-            };
+            let unique = unique.then_some("UNIQUE".to_owned());
             let comment = comment
                 .as_ref()
                 .map(|comment| format!("COMMENT '{}'", comment));
--- a/core/src/ast/ddl.rs
+++ b/core/src/ast/ddl.rs
@@ -33,13 +33,6 @@ pub struct ColumnDef {
     pub comment: Option<String>,
 }
 
-impl ColumnDef {
-    /// Returns whether the column must be unique.
-    pub fn is_unique(&self) -> bool {
-        self.unique
-    }
-}
-
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
 pub struct OperateFunctionArg {
     pub name: String,
--- a/core/src/ast/ddl.rs
+++ b/core/src/ast/ddl.rs
@@ -91,7 +91,11 @@ impl ToSql for ColumnDef {
             let default = default
                 .as_ref()
                 .map(|expr| format!("DEFAULT {}", expr.to_sql()));
-            let unique = if *unique { Some("UNIQUE".to_owned()) } else { None };
+            let unique = if *unique {
+                Some("UNIQUE".to_owned())
+            } else {
+                None
+            };
             let comment = comment
                 .as_ref()
                 .map(|comment| format!("COMMENT '{}'", comment));
@@ -122,9 +126,7 @@ impl ToSql for OperateFunctionArg {
 
 #[cfg(test)]
 mod tests {
-    use crate::ast::{
-        AstLiteral, ColumnDef, DataType, Expr, OperateFunctionArg, ToSql,
-    };
+    use crate::ast::{AstLiteral, ColumnDef, DataType, Expr, OperateFunctionArg, ToSql};
 
     #[test]
     fn to_sql_column_def() {
--- a/core/src/ast/ddl.rs
+++ b/core/src/ast/ddl.rs
@@ -34,9 +34,41 @@ pub struct ColumnDef {
     pub comment: Option<String>,
 }
 
+impl ColumnDef {
+    /// Returns whether the column is primary key.
+    pub fn is_primary(&self) -> bool {
+        self.unique
+            .as_ref()
+            .map(ColumnUniqueOption::is_primary)
+            .unwrap_or(false)
+    }
+
+    /// Returns whether the column must be unique.
+    pub fn is_unique(&self) -> bool {
+        self.unique.is_some()
+    }
+
+    /// Returns whether the column is unique but not primary key.
+    pub fn is_unique_not_primary(&self) -> bool {
+        self.is_unique() && !self.is_primary()
+    }
+
+    /// Sets the column as primary key.
+    pub fn set_primary(&mut self) {
+        self.unique = Some(ColumnUniqueOption { is_primary: true });
+    }
+}
+
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
 pub struct ColumnUniqueOption {
     pub is_primary: bool,
+}
+
+impl ColumnUniqueOption {
+    /// Returns true if the unique option is primary key.
+    pub fn is_primary(&self) -> bool {
+        self.is_primary
+    }
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
--- a/core/src/ast/ddl.rs
+++ b/core/src/ast/ddl.rs
@@ -29,45 +29,14 @@ pub struct ColumnDef {
     pub nullable: bool,
     /// `DEFAULT <restricted-expr>`
     pub default: Option<Expr>,
-    /// `{ PRIMARY KEY | UNIQUE }`
-    pub unique: Option<ColumnUniqueOption>,
+    pub unique: bool,
     pub comment: Option<String>,
 }
 
 impl ColumnDef {
-    /// Returns whether the column is primary key.
-    pub fn is_primary(&self) -> bool {
-        self.unique
-            .as_ref()
-            .map(ColumnUniqueOption::is_primary)
-            .unwrap_or(false)
-    }
-
     /// Returns whether the column must be unique.
     pub fn is_unique(&self) -> bool {
-        self.unique.is_some()
-    }
-
-    /// Returns whether the column is unique but not primary key.
-    pub fn is_unique_not_primary(&self) -> bool {
-        self.is_unique() && !self.is_primary()
-    }
-
-    /// Sets the column as primary key.
-    pub fn set_primary(&mut self) {
-        self.unique = Some(ColumnUniqueOption { is_primary: true });
-    }
-}
-
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
-pub struct ColumnUniqueOption {
-    pub is_primary: bool,
-}
-
-impl ColumnUniqueOption {
-    /// Returns true if the unique option is primary key.
-    pub fn is_primary(&self) -> bool {
-        self.is_primary
+        self.unique
     }
 }
 
@@ -122,13 +91,7 @@ impl ToSql for ColumnDef {
             let default = default
                 .as_ref()
                 .map(|expr| format!("DEFAULT {}", expr.to_sql()));
-            let unique = unique.as_ref().and_then(|unique| {
-                if unique.is_primary() {
-                    None
-                } else {
-                    Some(unique.to_sql())
-                }
-            });
+            let unique = if *unique { Some("UNIQUE".to_string()) } else { None };
             let comment = comment
                 .as_ref()
                 .map(|comment| format!("COMMENT '{}'", comment));
@@ -139,17 +102,6 @@ impl ToSql for ColumnDef {
                 .collect::<Vec<_>>()
                 .join(" ")
         }
-    }
-}
-
-impl ToSql for ColumnUniqueOption {
-    fn to_sql(&self) -> String {
-        if self.is_primary {
-            "PRIMARY KEY"
-        } else {
-            "UNIQUE"
-        }
-        .to_owned()
     }
 }
 
@@ -171,7 +123,7 @@ impl ToSql for OperateFunctionArg {
 #[cfg(test)]
 mod tests {
     use crate::ast::{
-        AstLiteral, ColumnDef, ColumnUniqueOption, DataType, Expr, OperateFunctionArg, ToSql,
+        AstLiteral, ColumnDef, DataType, Expr, OperateFunctionArg, ToSql,
     };
 
     #[test]
@@ -183,7 +135,7 @@ mod tests {
                 data_type: DataType::Text,
                 nullable: false,
                 default: None,
-                unique: Some(ColumnUniqueOption { is_primary: false }),
+                unique: true,
                 comment: None,
             }
             .to_sql()
@@ -196,7 +148,7 @@ mod tests {
                 data_type: DataType::Boolean,
                 nullable: true,
                 default: None,
-                unique: None,
+                unique: false,
                 comment: None,
             }
             .to_sql()
@@ -209,7 +161,7 @@ mod tests {
                 data_type: DataType::Int,
                 nullable: false,
                 default: None,
-                unique: Some(ColumnUniqueOption { is_primary: true }),
+                unique: false,
                 comment: None,
             }
             .to_sql()
@@ -222,7 +174,7 @@ mod tests {
                 data_type: DataType::Boolean,
                 nullable: false,
                 default: Some(Expr::Literal(AstLiteral::Boolean(false))),
-                unique: None,
+                unique: false,
                 comment: None,
             }
             .to_sql()
@@ -235,7 +187,7 @@ mod tests {
                 data_type: DataType::Boolean,
                 nullable: false,
                 default: Some(Expr::Literal(AstLiteral::Boolean(false))),
-                unique: Some(ColumnUniqueOption { is_primary: false }),
+                unique: true,
                 comment: None,
             }
             .to_sql()
@@ -248,7 +200,7 @@ mod tests {
                 data_type: DataType::Boolean,
                 nullable: false,
                 default: None,
-                unique: None,
+                unique: false,
                 comment: Some("this is comment".to_owned()),
             }
             .to_sql()
--- a/core/src/ast/ddl.rs
+++ b/core/src/ast/ddl.rs
@@ -91,7 +91,7 @@ impl ToSql for ColumnDef {
             let default = default
                 .as_ref()
                 .map(|expr| format!("DEFAULT {}", expr.to_sql()));
-            let unique = if *unique { Some("UNIQUE".to_string()) } else { None };
+            let unique = if *unique { Some("UNIQUE".to_owned()) } else { None };
             let comment = comment
                 .as_ref()
                 .map(|comment| format!("COMMENT '{}'", comment));
--- a/core/src/ast/ddl.rs
+++ b/core/src/ast/ddl.rs
@@ -122,7 +122,13 @@ impl ToSql for ColumnDef {
             let default = default
                 .as_ref()
                 .map(|expr| format!("DEFAULT {}", expr.to_sql()));
-            let unique = unique.as_ref().map(ToSql::to_sql);
+            let unique = unique.as_ref().and_then(|unique| {
+                if unique.is_primary() {
+                    None
+                } else {
+                    Some(unique.to_sql())
+                }
+            });
             let comment = comment
                 .as_ref()
                 .map(|comment| format!("COMMENT '{}'", comment));
@@ -197,7 +203,7 @@ mod tests {
         );
 
         assert_eq!(
-            r#""id" INT NOT NULL PRIMARY KEY"#,
+            r#""id" INT NOT NULL"#,
             ColumnDef {
                 name: "id".to_owned(),
                 data_type: DataType::Int,
--- a/core/src/ast/expr.rs
+++ b/core/src/ast/expr.rs
@@ -94,18 +94,6 @@ impl ToSqlUnquoted for Expr {
     }
 }
 
-impl<'a> TryInto<&'a str> for &'a Expr {
-    type Error = ();
-
-    fn try_into(self) -> Result<&'a str, Self::Error> {
-        match self {
-            Expr::Identifier(s) => Ok(s),
-            Expr::CompoundIdentifier { ident, .. } => Ok(ident),
-            _ => Err(()),
-        }
-    }
-}
-
 impl Expr {
     fn to_sql_with(&self, quoted: bool) -> String {
         match self {
--- a/core/src/ast/expr.rs
+++ b/core/src/ast/expr.rs
@@ -94,6 +94,25 @@ impl ToSqlUnquoted for Expr {
     }
 }
 
+impl<'a> TryInto<Vec<&'a str>> for &'a Expr {
+    type Error = ();
+
+    fn try_into(self) -> Result<Vec<&'a str>, Self::Error> {
+        match self {
+            Expr::Identifier(s) => Ok(vec![&s]),
+            Expr::CompoundIdentifier { ident, .. } => Ok(vec![&ident]),
+            Expr::Array { elem } => {
+                let mut columns = Vec::new();
+                for e in elem {
+                    columns.extend(<&Expr as TryInto<Vec<&str>>>::try_into(e)?);
+                }
+                Ok(columns)
+            }
+            _ => Err(()),
+        }
+    }
+}
+
 impl Expr {
     fn to_sql_with(&self, quoted: bool) -> String {
         match self {
--- a/core/src/ast/expr.rs
+++ b/core/src/ast/expr.rs
@@ -95,6 +95,40 @@ impl ToSqlUnquoted for Expr {
 }
 
 impl Expr {
+    /// Create a new AND binary expression.
+    pub fn and(self, other: Expr) -> Expr {
+        Expr::BinaryOp {
+            left: Box::new(self),
+            op: BinaryOperator::And,
+            right: Box::new(other),
+        }
+    }
+
+    /// Create a new equals binary expression.
+    pub fn equal(self, other: Expr) -> Expr {
+        Expr::BinaryOp {
+            left: Box::new(self),
+            op: BinaryOperator::Eq,
+            right: Box::new(other),
+        }
+    }
+
+    /// Commute the AND binary expression.
+    pub fn commute_and(self) -> Expr {
+        match self {
+            Expr::BinaryOp {
+                left,
+                op: BinaryOperator::And,
+                right,
+            } => Expr::BinaryOp {
+                left: right,
+                op: BinaryOperator::And,
+                right: left,
+            },
+            _ => self,
+        }
+    }
+
     fn to_sql_with(&self, quoted: bool) -> String {
         match self {
             Expr::Identifier(s) => match quoted {
--- a/core/src/ast/expr.rs
+++ b/core/src/ast/expr.rs
@@ -94,20 +94,13 @@ impl ToSqlUnquoted for Expr {
     }
 }
 
-impl<'a> TryInto<Vec<&'a str>> for &'a Expr {
+impl<'a> TryInto<&'a str> for &'a Expr {
     type Error = ();
 
-    fn try_into(self) -> Result<Vec<&'a str>, Self::Error> {
+    fn try_into(self) -> Result<&'a str, Self::Error> {
         match self {
-            Expr::Identifier(s) => Ok(vec![&s]),
-            Expr::CompoundIdentifier { ident, .. } => Ok(vec![&ident]),
-            Expr::Array { elem } => {
-                let mut columns = Vec::new();
-                for e in elem {
-                    columns.extend(<&Expr as TryInto<Vec<&str>>>::try_into(e)?);
-                }
-                Ok(columns)
-            }
+            Expr::Identifier(s) => Ok(s),
+            Expr::CompoundIdentifier { ident, .. } => Ok(ident),
             _ => Err(()),
         }
     }
--- a/core/src/ast/mod.rs
+++ b/core/src/ast/mod.rs
@@ -209,8 +209,12 @@ impl ToSql for Statement {
                 comment,
             } => {
                 let if_not_exists = if_not_exists.then_some("IF NOT EXISTS");
+
                 let primary_key = columns.as_ref().and_then(|cols| {
+                    // If the columns are defined, we check whether there exists a primary key.
                     if cols.iter().any(|col| col.is_primary()) {
+                        // If there is, we format the primary key constraint as a string, which
+                        // will be placed at the end of the CREATE TABLE statement.
                         Some(format!(
                             "PRIMARY KEY ({})",
                             cols.iter()
@@ -234,6 +238,8 @@ impl ToSql for Statement {
                             .chain(foreign_keys)
                             .collect::<Vec<_>>();
 
+                        // If we identified earlier that there is a primary key, we add it
+                        // at the end of the CREATE TABLE statement.
                         if let Some(primary_key) = primary_key {
                             body.push(primary_key);
                         }
--- a/core/src/ast/mod.rs
+++ b/core/src/ast/mod.rs
@@ -218,7 +218,7 @@ impl ToSql for Statement {
                         indices
                             .iter()
                             .copied()
-                            .map(|i| columns[i].name.clone())
+                            .map(|i| columns[i].name.as_ref())
                             .collect::<Vec<_>>()
                             .join(", ")
                     )),
--- a/core/src/ast/mod.rs
+++ b/core/src/ast/mod.rs
@@ -221,6 +221,7 @@ impl ToSql for Statement {
                             .filter_map(|(i, column)| {
                                 indices.contains(&i).then_some(column.name.as_str())
                             })
+                            .map(|s| format!(r#""{}""#, s))
                             .collect::<Vec<_>>()
                             .join(", ")
                     )),
--- a/core/src/ast/mod.rs
+++ b/core/src/ast/mod.rs
@@ -212,8 +212,9 @@ impl ToSql for Statement {
             } => {
                 let if_not_exists = if_not_exists.then_some("IF NOT EXISTS");
 
-                let primary_key =
-                    primary_key.as_ref().map(|cols| format!("PRIMARY KEY ({})", cols.join(", ")));
+                let primary_key = primary_key
+                    .as_ref()
+                    .map(|cols| format!("PRIMARY KEY ({})", cols.join(", ")));
                 let body = match (source, columns) {
                     (Some(query), _) => Some(format!("AS {}", query.to_sql())),
                     (None, None) => None,
--- a/core/src/ast/mod.rs
+++ b/core/src/ast/mod.rs
@@ -215,10 +215,12 @@ impl ToSql for Statement {
                 let primary_key = match (primary_key.as_ref(), columns.as_ref()) {
                     (Some(indices), Some(columns)) => Some(format!(
                         "PRIMARY KEY ({})",
-                        indices
+                        columns
                             .iter()
-                            .copied()
-                            .map(|i| columns[i].name.as_ref())
+                            .enumerate()
+                            .filter_map(|(i, column)| {
+                                indices.contains(&i).then_some(column.name.as_str())
+                            })
                             .collect::<Vec<_>>()
                             .join(", ")
                     )),
--- a/core/src/ast/mod.rs
+++ b/core/src/ast/mod.rs
@@ -220,19 +220,13 @@ impl ToSql for Statement {
                     (None, None) => None,
                     (None, Some(columns)) => {
                         let foreign_keys = foreign_keys.iter().map(ToSql::to_sql);
-                        let mut body = columns
+                        let body = columns
                             .iter()
                             .map(ToSql::to_sql)
                             .chain(foreign_keys)
-                            .collect::<Vec<_>>();
-
-                        // If we identified earlier that there is a primary key, we add it
-                        // at the end of the CREATE TABLE statement.
-                        if let Some(primary_key) = primary_key {
-                            body.push(primary_key);
-                        }
-
-                        let body = body.join(", ");
+                            .chain(primary_key)
+                            .collect::<Vec<_>>()
+                            .join(", ");
 
                         Some(format!("({body})"))
                     }
--- a/core/src/ast/mod.rs
+++ b/core/src/ast/mod.rs
@@ -87,6 +87,7 @@ pub enum Statement {
         source: Option<Box<Query>>,
         engine: Option<String>,
         foreign_keys: Vec<ForeignKey>,
+        primary_key: Option<Vec<String>>,
         comment: Option<String>,
     },
     /// CREATE FUNCTION
@@ -206,27 +207,13 @@ impl ToSql for Statement {
                 source,
                 engine,
                 foreign_keys,
+                primary_key,
                 comment,
             } => {
                 let if_not_exists = if_not_exists.then_some("IF NOT EXISTS");
 
-                let primary_key = columns.as_ref().and_then(|cols| {
-                    // If the columns are defined, we check whether there exists a primary key.
-                    if cols.iter().any(|col| col.is_primary()) {
-                        // If there is, we format the primary key constraint as a string, which
-                        // will be placed at the end of the CREATE TABLE statement.
-                        Some(format!(
-                            "PRIMARY KEY ({})",
-                            cols.iter()
-                                .filter(|col| col.is_primary())
-                                .map(|col| col.name.clone())
-                                .collect::<Vec<_>>()
-                                .join(", ")
-                        ))
-                    } else {
-                        None
-                    }
-                });
+                let primary_key =
+                    primary_key.as_ref().map(|cols| format!("PRIMARY KEY ({})", cols.join(", ")));
                 let body = match (source, columns) {
                     (Some(query), _) => Some(format!("AS {}", query.to_sql())),
                     (None, None) => None,
@@ -500,6 +487,7 @@ mod tests {
                 source: None,
                 engine: None,
                 foreign_keys: Vec::new(),
+                primary_key: None,
                 comment: None,
             }
             .to_sql()
@@ -514,6 +502,7 @@ mod tests {
                 source: None,
                 engine: None,
                 foreign_keys: Vec::new(),
+                primary_key: None,
                 comment: None,
             }
             .to_sql()
@@ -529,12 +518,13 @@ mod tests {
                     data_type: DataType::Boolean,
                     nullable: false,
                     default: None,
-                    unique: None,
+                    unique: false,
                     comment: None,
                 },]),
                 source: None,
                 engine: None,
                 foreign_keys: Vec::new(),
+                primary_key: None,
                 comment: Some("this is comment".to_owned()),
             }
             .to_sql()
@@ -551,7 +541,7 @@ mod tests {
                         data_type: DataType::Int,
                         nullable: false,
                         default: None,
-                        unique: None,
+                        unique: false,
                         comment: None,
                     },
                     ColumnDef {
@@ -559,7 +549,7 @@ mod tests {
                         data_type: DataType::Int,
                         nullable: true,
                         default: None,
-                        unique: None,
+                        unique: false,
                         comment: None,
                     },
                     ColumnDef {
@@ -567,13 +557,14 @@ mod tests {
                         data_type: DataType::Text,
                         nullable: false,
                         default: None,
-                        unique: None,
+                        unique: false,
                         comment: None,
                     }
                 ]),
                 source: None,
                 engine: None,
                 foreign_keys: Vec::new(),
+                primary_key: None,
                 comment: None,
             }
             .to_sql()
@@ -618,6 +609,7 @@ mod tests {
                 })),
                 engine: None,
                 foreign_keys: Vec::new(),
+                primary_key: None,
                 comment: None,
             }
             .to_sql()
@@ -639,6 +631,7 @@ mod tests {
                 })),
                 engine: None,
                 foreign_keys: Vec::new(),
+                primary_key: None,
                 comment: None,
             }
             .to_sql()
@@ -656,6 +649,7 @@ mod tests {
                 source: None,
                 engine: Some("MEMORY".to_owned()),
                 foreign_keys: Vec::new(),
+                primary_key: None,
                 comment: None,
             }
             .to_sql()
@@ -671,12 +665,13 @@ mod tests {
                     data_type: DataType::Boolean,
                     nullable: false,
                     default: None,
-                    unique: None,
+                    unique: false,
                     comment: None,
                 },]),
                 source: None,
                 engine: Some("SLED".to_owned()),
                 foreign_keys: Vec::new(),
+                primary_key: None,
                 comment: None,
             }
             .to_sql()
@@ -727,7 +722,7 @@ mod tests {
                         default: Some(Expr::Literal(AstLiteral::Number(
                             BigDecimal::from_str("10").unwrap()
                         ))),
-                        unique: None,
+                        unique: false,
                         comment: None,
                     }
                 }
--- a/core/src/ast/mod.rs
+++ b/core/src/ast/mod.rs
@@ -209,17 +209,36 @@ impl ToSql for Statement {
                 comment,
             } => {
                 let if_not_exists = if_not_exists.then_some("IF NOT EXISTS");
+                let primary_key = columns.as_ref().and_then(|cols| {
+                    if cols.iter().any(|col| col.is_primary()) {
+                        Some(format!(
+                            "PRIMARY KEY ({})",
+                            cols.iter()
+                                .filter(|col| col.is_primary())
+                                .map(|col| col.name.clone())
+                                .collect::<Vec<_>>()
+                                .join(", ")
+                        ))
+                    } else {
+                        None
+                    }
+                });
                 let body = match (source, columns) {
                     (Some(query), _) => Some(format!("AS {}", query.to_sql())),
                     (None, None) => None,
                     (None, Some(columns)) => {
                         let foreign_keys = foreign_keys.iter().map(ToSql::to_sql);
-                        let body = columns
+                        let mut body = columns
                             .iter()
                             .map(ToSql::to_sql)
                             .chain(foreign_keys)
-                            .collect::<Vec<_>>()
-                            .join(", ");
+                            .collect::<Vec<_>>();
+
+                        if let Some(primary_key) = primary_key {
+                            body.push(primary_key);
+                        }
+
+                        let body = body.join(", ");
 
                         Some(format!("({body})"))
                     }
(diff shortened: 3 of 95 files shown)


LucaCappelletti94:
Now several storages pass the test suite. How would I go about and test the remaining storage, like redis and IDB storage?

panarch:
> @panarch now several storages pass the test suite. How would I go about and test the remaining storage, like redis and IDB storage?

idb storage - https://github.com/gluesql/gluesql/blob/main/storages/idb-storage/README.md
you need to use wasm-pack for running test for idb storage
```bash
WASM_BINDGEN_TEST_TIMEOUT=60 wasm-pack test --headless --firefox --chrome
```

redis and mongodb storage,
```bash
cargo test --features test-redis
cargo test --features test-mongo
```
redis and mongodb test are not triggered by default because they require some pre installed dependencies (redis & mongodb) which cannot be managed by cargo.
you need to use "test-redis" feature for testing redis, and "test-mongo" feature for testing mongodb.

fyi. https://github.com/gluesql/gluesql/blob/main/.github/workflows/rust.yml

LucaCappelletti94:
Thanks @panarch ! Now I got to the point of all storages working except for Mongo, which I am very unfamiliar with. I am struggling to understand where it is encoding / decoding the primary key.

panarch:
Great start! The initial tests look good.

Could we also include unit tests in the primary_key planner found [here](https://github.com/gluesql/gluesql/blob/main/core/src/plan/primary_key.rs)?
This will help us precisely verify the functionality of composite primary keys.

LucaCappelletti94:
Ok, so I got up until this point where the message I receive from MongoDB is not easy to convert to the error produced by the other endpoints in the same test. How do you suggest I proceed?

```
Diff < left / right > :
 Err(
<    StorageMsg(
<        "Kind: An error occurred when trying to execute a write operation: BulkWriteFailure { write_errors: Some([BulkWriteError { index: 0, code: 11000, code_name: None, message: \"E11000 duplicate key error collection: multiple_primary_keys.Allegro index: Allegro_PRIMARY_KEY_MONGO_GLUESQL dup key: { table_id: 1, user_id: 1 }\", details: None }]), write_concern_error: None, inserted_ids: {} }, labels: {}",
>    Validate(
>        DuplicateEntryOnPrimaryKeyField(
>            Composite(
>                [
>                    I64(
>                        1,
>                    ),
>                    I64(
>                        1,
>                    ),
>                ],
>            ),
>        ),
     ),
 )
```

LucaCappelletti94:
I have added some tests with selects involving where statements and they are currently failing. Will have to investigate what may be the cause.

LucaCappelletti94:
> Great start! The initial tests look good.
> 
> Could we also include unit tests in the primary_key planner found [here](https://github.com/gluesql/gluesql/blob/main/core/src/plan/primary_key.rs)? This will help us precisely verify the functionality of composite primary keys.

I have added also these tests.

core/src/plan/primary_key.rs
```diff
@@ -389,6 +389,142 @@ mod tests {
...
+
+        let sql = "SELECT * FROM Player WHERE id = 1;";
+        let actual = plan(&storage, sql);
+        let expected = select(Select {
+            projection: vec![SelectItem::Wildcard],
+            from: TableWithJoins {
+                relation: TableFactor::Table {
+                    name: "Player".to_owned(),
+                    alias: None,
+                    index: None,
```
panarch, line 410: if the primary key planner works, then `index` field should exist.
based on that `index` field, execution layer handles whether to call `Store::fetch_data` or `Store::scan_data`.

panarch:
Looks good.

It seems you're familiar with the concepts of storage and the test suite.

Now, it would be more efficient to focus solely on the core workspace.

Once we complete the core tasks, we can move on to storage and the test suite.

The only way to accurately verify whether the primary key index is correctly identified by the planner is through unit tests in the primary key planner.

Since GlueSQL currently supports only a rule-based logical planner, it's easier to write tests because the outcomes are predictable. You can start by modifying the test content in the same manner you started.

If the planner successfully identifies the composite primary key, the index field in the TableFactor of the planner unit test will contain an entry.

Let's focus on that.

The planner uses a MockStorage, which is a test-specific storage that provides schema information from memory, so running tests solely within the core workspace will suffice.

core/src/plan/primary_key.rs
LucaCappelletti94: Ok, I will add this to the documentation of the planner.

LucaCappelletti94: Should the primary key planner work for select statements that are not operating on a `TableFactor::Table`? Without knowing which table is involved, we cannot know which expression entries are the primary keys as we cannot extract their schema.

panarch: > Should the primary key planner work for select statements that are not operating on a TableFactor::Table

we only need to handle `TableFactor::Table`. we can ignore other TableFactor variants.

LucaCappelletti94: Can you point me to where the index is used afterwards? I am still trying to understand how to represent the multi-primary key variant and seeing how it is used may certainly help.

LucaCappelletti94:
Ah I suppose because it can be symbolic in some cases, okay.

panarch:
ref. https://github.com/gluesql/gluesql/blob/25309dd15a50ee061e714576079b2cd8c1fcca9c/core/src/ast/ddl.rs#L32-L40

ColumnDef currently holds information for a single column. Therefore, if we enable composite primary keys, the primary key field should not remain within ColumnDef. It seems necessary to modify the AST to move the primary key field outside of ColumnDef. What are your thoughts on this approach?

LucaCappelletti94:
> ref.
> 
> https://github.com/gluesql/gluesql/blob/25309dd15a50ee061e714576079b2cd8c1fcca9c/core/src/ast/ddl.rs#L32-L40
> 
> ColumnDef currently holds information for a single column. Therefore, if we enable composite primary keys, the primary key field should not remain within ColumnDef. It seems necessary to modify the AST to move the primary key field outside of ColumnDef. What are your thoughts on this approach?

I believe my current solution of allowing multiple ColumnDef to have primary key set to True to be a viable one, but we will see if it turns out to be shortsighted. At this time, I am still finishing the refactoring to add the composite primary key to the planner.

panarch (commented):
I believe that with a slight modification to the current structure, particularly the `Schema` and `ColumnDef` structure, we can simplify the code significantly. This could eliminate the need for many of the helper functions we recently added, such as `is_primary`, `is_unique`, etc.

Given that the primary key now supports composite keys, it might be more efficient to store the primary key information as a simple boolean value within `ColumnDef`, rather than embedding primary key details there. This way, we can use the `unique` field in `ColumnDef` as a simple boolean, like this:

```rust
pub struct ColumnDef {
    pub name: String,
    pub data_type: DataType,
    pub nullable: bool,
    /// `DEFAULT <restricted-expr>`
    pub default: Option<Expr>,
    pub unique: bool,
    pub comment: Option<String>,
}
```

By simplifying the `unique` field to a boolean, the code becomes more straightforward and easier to manage. Next, we need to store the primary key information elsewhere, and the logical place for this would be in the `Schema` structure. Here is how we could do that:

```rust
#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
pub struct Schema {
    pub table_name: String,
    pub column_defs: Option<Vec<ColumnDef>>,
    pub primary_key: Vec<String>, // primary key column names
    pub indexes: Vec<SchemaIndex>,
    pub engine: Option<String>,
    pub foreign_keys: Vec<ForeignKey>,
    pub comment: Option<String>,
}
```

In this schema, the `primary_key` is a list of column names. This approach supports schemaless configurations because, in schemaless scenarios, there won't be `ColumnDef` objects, but we can still identify primary key columns by their names. If we were only considering schemas with defined structures, we could simply use the indices of `column_defs` for the primary key. However, to accommodate both structured and schemaless setups, using column names is more flexible and future-proof. Although the current PR doesn't address schemaless support, this design provides a clear path for future enhancements, ensuring we can eventually support primary keys in a schemaless context.

Here is the rationale behind these changes:

1. **Simplification of `ColumnDef`**: By storing the `unique` attribute as a simple boolean, we reduce the complexity within each column definition. This avoids the need to check for multiple attributes and simplifies logic that interacts with column definitions.

2. **Separation of Concerns**: Moving primary key information to the `Schema` level separates column-specific details from table-wide constraints. This enhances clarity and maintainability, as primary key constraints inherently apply to the table rather than individual columns.

In conclusion, these proposed changes aim to streamline the codebase, reduce unnecessary complexity, and lay a foundation for future enhancements. What are your thoughts on this approach?

core/src/data/key.rs
```diff
@@ -50,6 +50,7 @@ pub enum Key {
     Interval(Interval),
     Uuid(u128),
     Inet(IpAddr),
+    Composite(Vec<Key>),
```
panarch, line 53: The idea of adding a `Composite` variant to the `Key` enum is fantastic. I was thinking of a more complex solution, but this approach resolves the problem in a much simpler and more straightforward way. 

The reason `Key` didn't previously support `Value::List` wasn't due to any specific reason; it just wasn't implemented. In this case, how about naming the variant `Key::List` instead of `Key::Composite` to match the `Value` naming? This way, we explicitly support `List` and also handle composite primary keys.

LucaCappelletti94: Ok!

LucaCappelletti94:
> I believe that with a slight modification to the current structure, particularly the `Schema` and `ColumnDef` structure, we can simplify the code significantly. This could eliminate the need for many of the helper functions we recently added, such as `is_primary`, `is_unique`, etc.
> 
> Given that the primary key now supports composite keys, it might be more efficient to store the primary key information as a simple boolean value within `ColumnDef`, rather than embedding primary key details there. This way, we can use the `unique` field in `ColumnDef` as a simple boolean, like this:
> 
> ```rust
> #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
> pub struct ColumnDef {
>     pub name: String,
>     pub data_type: DataType,
>     pub nullable: bool,
>     /// `DEFAULT <restricted-expr>`
>     pub default: Option<Expr>,
>     pub unique: bool,
>     pub comment: Option<String>,
> }
> ```
> 
> By simplifying the `unique` field to a boolean, the code becomes more straightforward and easier to manage. Next, we need to store the primary key information elsewhere, and the logical place for this would be in the `Schema` structure. Here is how we could do that:
> 
> ```rust
> #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
> pub struct Schema {
>     pub table_name: String,
>     pub column_defs: Option<Vec<ColumnDef>>,
>     pub primary_key: Vec<String>, // primary key column names
>     pub indexes: Vec<SchemaIndex>,
>     pub engine: Option<String>,
>     pub foreign_keys: Vec<ForeignKey>,
>     pub comment: Option<String>,
> }
> ```
> 
> In this schema, the `primary_key` is a list of column names. This approach supports schemaless configurations because, in schemaless scenarios, there won't be `ColumnDef` objects, but we can still identify primary key columns by their names. If we were only considering schemas with defined structures, we could simply use the indices of `column_defs` for the primary key. However, to accommodate both structured and schemaless setups, using column names is more flexible and future-proof. Although the current PR doesn't address schemaless support, this design provides a clear path for future enhancements, ensuring we can eventually support primary keys in a schemaless context.
> 
> Here is the rationale behind these changes:
> 
> 1. **Simplification of `ColumnDef`**: By storing the `unique` attribute as a simple boolean, we reduce the complexity within each column definition. This avoids the need to check for multiple attributes and simplifies logic that interacts with column definitions.
> 2. **Separation of Concerns**: Moving primary key information to the `Schema` level separates column-specific details from table-wide constraints. This enhances clarity and maintainability, as primary key constraints inherently apply to the table rather than individual columns.
> 
> In conclusion, these proposed changes aim to streamline the codebase, reduce unnecessary complexity, and lay a foundation for future enhancements. What are your thoughts on this approach?

Take a look at my approach for the PR #1519 - maybe we can fully remove the unique keyword and handle the unique only as constraints there?

test-suite/src/foreign_key.rs
```diff
@@ -148,7 +148,7 @@ test_case!(foreign_key, {
     .await;
 
     g.named_test(
-        "Creating table with foreign key should be succeeded if referenced table has primary key. NO ACTION(=RESTRICT) is default",
+        "Creating table with foreign key should succeeded if referenced table has primary key. NO ACTION(=RESTRICT) is default",
```
panarch, line 151: could you make a separate PR for this? then I can merge right after.
It looks not related to this PR so..

and
"should be succeeded" -> "should succeeded" -> "should succeed" ?

core/src/ast/ddl.rs
```diff
@@ -29,45 +29,14 @@ pub struct ColumnDef {
...
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
-pub struct ColumnUniqueOption {
-    pub is_primary: bool,
-}
-
-impl ColumnUniqueOption {
-    /// Returns true if the unique option is primary key.
-    pub fn is_primary(&self) -> bool {
-        self.is_primary
+        self.unique
```
panarch, line 39: nit. it is just a single accessor method, we may not need this..? we can simply use `column_def.unique` rather than `column_def.is_unique`

LucaCappelletti94: Which method are you referring to? The `is_primary` method no longer exists.

LucaCappelletti94: Ah the `is_unique` - removed.

core/src/data/key.rs
LucaCappelletti94: Done

storages/parquet-storage/src/lib.rs
```diff
@@ -134,25 +146,25 @@ impl ParquetStorage {
         let mut key_counter: u64 = 0;
 
         if let Some(column_defs) = &fetched_schema.column_defs {
+            let primary_key_indices = fetched_schema.primary_key.as_ref().map(|primary_key| {
+                gluesql_core::executor::get_primary_key_column_indices(column_defs, primary_key)
+            });
```
panarch, line 151: could we have this method in `data::Schema`? then we can use via `fetched_schema.get_primary_key_column_indices()`

LucaCappelletti94: Reasonable, on it.

panarch (commented):
changes in `storages/memory_storage/src/{alter_table}|{lib}.rs` does not look related to this PR. could we have this in a different pr?

LucaCappelletti94:
> changes in `storages/memory_storage/src/{alter_table}|{lib}.rs` does not look related to this PR. could we have this in a different pr?

Could you kindly reference the changes?

panarch:
> > changes in `storages/memory_storage/src/{alter_table}|{lib}.rs` does not look related to this PR. could we have this in a different pr?
> 
> Could you kindly reference the changes?

sure,
these two files - 
image

https://github.com/gluesql/gluesql/pull/1515/files#diff-598c47895adf9154366b60755ec0650b6045af4f6695679fc4d734878bfe654a
https://github.com/gluesql/gluesql/pull/1515/files#diff-df3394cd73d05b609521ab3bc36e8e344362aff556006f125880d6868eb60ba0

LucaCappelletti94:
> > > changes in `storages/memory_storage/src/{alter_table}|{lib}.rs` does not look related to this PR. could we have this in a different pr?
> > 
> > 
> > Could you kindly reference the changes?
> 
> sure, these two files - image
> 
> https://github.com/gluesql/gluesql/pull/1515/files#diff-598c47895adf9154366b60755ec0650b6045af4f6695679fc4d734878bfe654a https://github.com/gluesql/gluesql/pull/1515/files#diff-df3394cd73d05b609521ab3bc36e8e344362aff556006f125880d6868eb60ba0

Without those changes, the clippy test will rant and cause this PR to fail.

panarch:
> > > > changes in `storages/memory_storage/src/{alter_table}|{lib}.rs` does not look related to this PR. could we have this in a different pr?
> > > 
> > > 
> > > Could you kindly reference the changes?
> > 
> > 
> > sure, these two files - image
> > https://github.com/gluesql/gluesql/pull/1515/files#diff-598c47895adf9154366b60755ec0650b6045af4f6695679fc4d734878bfe654a https://github.com/gluesql/gluesql/pull/1515/files#diff-df3394cd73d05b609521ab3bc36e8e344362aff556006f125880d6868eb60ba0
> 
> Without those changes, the clippy test will rant and cause this PR to fail.

oh, could I know the version you are using?
ok, if you don't mind then, could we have these two file changes in the separate pr?
I will merge it right after.
clippy does not complain in 1.76, it looks that the clippy becomes smarter in the recent version :)

core/src/data/key.rs
```diff
@@ -130,12 +132,31 @@ impl TryFrom<Value> for Key {
...
+impl TryFrom<Vec<Value>> for Key {
+    type Error = Error;
+
+    fn try_from(values: Vec<Value>) -> Result<Self> {
+        Ok(Key::List(
+            values
+                .into_iter()
+                .map(Key::try_from)
+                .collect::<Result<Vec<_>>>()?,
+        ))
```
panarch, line 150: nit.
```rust
values
    .into_iter()
    .map(Key::try_from)
    .collect::<Result<Vec<_>>>()
    .map(Key::List)
```

core/src/data/key.rs
```diff
@@ -130,12 +132,31 @@ impl TryFrom<Value> for Key {
...
+    }
+}
+
+impl From<Vec<Key>> for Key {
+    fn from(keys: Vec<Key>) -> Self {
+        Key::List(keys)
+    }
+}
+
```
panarch, line 159: nit. we might not need this.
only existing use case is in test-suite primary_key case but we can simply wrap `Key::List` rather than using from impl.

LucaCappelletti94:
> > > > > changes in `storages/memory_storage/src/{alter_table}|{lib}.rs` does not look related to this PR. could we have this in a different pr?
> > > > 
> > > > 
> > > > Could you kindly reference the changes?
> > > 
> > > 
> > > sure, these two files - image
> > > https://github.com/gluesql/gluesql/pull/1515/files#diff-598c47895adf9154366b60755ec0650b6045af4f6695679fc4d734878bfe654a https://github.com/gluesql/gluesql/pull/1515/files#diff-df3394cd73d05b609521ab3bc36e8e344362aff556006f125880d6868eb60ba0
> > 
> > 
> > Without those changes, the clippy test will rant and cause this PR to fail.
> 
> oh, could I know the version you are using? ok, if you don't mind then, could we have these two file changes in the separate pr? I will merge it right after. clippy does not complain in 1.76, it looks that the clippy becomes smarter in the recent version :)

(base) [lucac@pippo gluesql]$ cargo clippy --version
clippy 0.1.79 (129f3b9 2024-06-10)

storages/parquet-storage/src/lib.rs
LucaCappelletti94: Done

core/src/data/key.rs
LucaCappelletti94: Done

core/src/data/key.rs
LucaCappelletti94: Updated

panarch:
> > > > > > changes in `storages/memory_storage/src/{alter_table}|{lib}.rs` does not look related to this PR. could we have this in a different pr?
> > > > > 
> > > > > 
> > > > > Could you kindly reference the changes?
> > > > 
> > > > 
> > > > sure, these two files - image
> > > > https://github.com/gluesql/gluesql/pull/1515/files#diff-598c47895adf9154366b60755ec0650b6045af4f6695679fc4d734878bfe654a https://github.com/gluesql/gluesql/pull/1515/files#diff-df3394cd73d05b609521ab3bc36e8e344362aff556006f125880d6868eb60ba0
> > > 
> > > 
> > > Without those changes, the clippy test will rant and cause this PR to fail.
> > 
> > 
> > oh, could I know the version you are using? ok, if you don't mind then, could we have these two file changes in the separate pr? I will merge it right after. clippy does not complain in 1.76, it looks that the clippy becomes smarter in the recent version :)
> 
> (base) [lucac@pippo gluesql]$ cargo clippy --version clippy 0.1.79 (129f3b9 2024-06-10)

We currently use 1.76
https://github.com/gluesql/gluesql/blob/main/rust-toolchain.toml

that is weird. rustup should automatically recognize the rust-toolchain.toml file and use matched version but it seems that it didnt.
some of unrelated changes look caused by 1.79 clippy. could you use 1.76?
we need to handle it separately.

LucaCappelletti94:
> > > > > > > changes in `storages/memory_storage/src/{alter_table}|{lib}.rs` does not look related to this PR. could we have this in a different pr?
> > > > > > 
> > > > > > 
> > > > > > Could you kindly reference the changes?
> > > > > 
> > > > > 
> > > > > sure, these two files - image
> > > > > https://github.com/gluesql/gluesql/pull/1515/files#diff-598c47895adf9154366b60755ec0650b6045af4f6695679fc4d734878bfe654a https://github.com/gluesql/gluesql/pull/1515/files#diff-df3394cd73d05b609521ab3bc36e8e344362aff556006f125880d6868eb60ba0
> > > > 
> > > > 
> > > > Without those changes, the clippy test will rant and cause this PR to fail.
> > > 
> > > 
> > > oh, could I know the version you are using? ok, if you don't mind then, could we have these two file changes in the separate pr? I will merge it right after. clippy does not complain in 1.76, it looks that the clippy becomes smarter in the recent version :)
> > 
> > 
> > (base) [lucac@pippo gluesql]$ cargo clippy --version clippy 0.1.79 (129f3b9 2024-06-10)
> 
> We currently use 1.76 https://github.com/gluesql/gluesql/blob/main/rust-toolchain.toml
> 
> that is weird. rustup should automatically recognize the rust-toolchain.toml file and use matched version but it seems that it didnt. some of unrelated changes look caused by 1.79 clippy. could you use 1.76? we need to handle it separately.

I will revert it shortly so to handle it afterwards in a separate PR. Is there anything else you would like me to change?

panarch:
big ones all look great so I think from now, like this, I may suggest some small cleanups or refactoring stuffs. I will leave comments.

LucaCappelletti94:
> > > > > > > > changes in `storages/memory_storage/src/{alter_table}|{lib}.rs` does not look related to this PR. could we have this in a different pr?
> > > > > > > 
> > > > > > > 
> > > > > > > Could you kindly reference the changes?
> > > > > > 
> > > > > > 
> > > > > > sure, these two files - image
> > > > > > https://github.com/gluesql/gluesql/pull/1515/files#diff-598c47895adf9154366b60755ec0650b6045af4f6695679fc4d734878bfe654a https://github.com/gluesql/gluesql/pull/1515/files#diff-df3394cd73d05b609521ab3bc36e8e344362aff556006f125880d6868eb60ba0
> > > > > 
> > > > > 
> > > > > Without those changes, the clippy test will rant and cause this PR to fail.
> > > > 
> > > > 
> > > > oh, could I know the version you are using? ok, if you don't mind then, could we have these two file changes in the separate pr? I will merge it right after. clippy does not complain in 1.76, it looks that the clippy becomes smarter in the recent version :)
> > > 
> > > 
> > > (base) [lucac@pippo gluesql]$ cargo clippy --version clippy 0.1.79 (129f3b9 2024-06-10)
> > 
> > 
> > We currently use 1.76 https://github.com/gluesql/gluesql/blob/main/rust-toolchain.toml
> > that is weird. rustup should automatically recognize the rust-toolchain.toml file and use matched version but it seems that it didnt. some of unrelated changes look caused by 1.79 clippy. could you use 1.76? we need to handle it separately.
> 
> I will revert it shortly so to handle it afterwards in a separate PR. Is there anything else you would like me to change?

Ok reverted these changes so we may handle them in a subsequent PR.

core/src/data/value/selector.rs
```diff
@@ -53,7 +53,7 @@ impl Value {
             .try_fold(self, |selectable, key| {
                 selectable.get_value_from_compound_type(&key)
             })
-            .map(Clone::clone)
+            .cloned()
```
panarch, line 56: nit. this also looks that is caused by clippy 1.79.

core/src/ast/ddl.rs
```diff
@@ -90,7 +84,11 @@ impl ToSql for ColumnDef {
             let default = default
                 .as_ref()
                 .map(|expr| format!("DEFAULT {}", expr.to_sql()));
-            let unique = unique.as_ref().map(ToSql::to_sql);
+            let unique = if *unique {
+                Some("UNIQUE".to_owned())
+            } else {
+                None
+            };
```
panarch, line 91: nit. `bool::then_some` would be useful.
```rust
let unique = unique.then_some("UNIQUE".to_owned());
```

core/src/ast/expr.rs
```diff
@@ -94,6 +94,18 @@ impl ToSqlUnquoted for Expr {
...
+
+    fn try_into(self) -> Result<&'a str, Self::Error> {
+        match self {
+            Expr::Identifier(s) => Ok(s),
+            Expr::CompoundIdentifier { ident, .. } => Ok(ident),
+            _ => Err(()),
+        }
+    }
+}
+
```
panarch, line 108: this one looks that primary_key planner is currently new to use this.
but could we have better idea not to add this kind of TryInto conversion trick?
it might better not to have this.

core/src/ast/mod.rs
```diff
@@ -206,20 +207,32 @@ impl ToSql for Statement {
...
-                            .join(", ");
+                            .collect::<Vec<_>>();
+
+                        // If we identified earlier that there is a primary key, we add it
+                        // at the end of the CREATE TABLE statement.
+                        if let Some(primary_key) = primary_key {
+                            body.push(primary_key);
+                        }
+
+                        let body = body.join(", ");
```
panarch, line 235: nit. Option implement IntoIterator so, we can have more simple version.
```rust
let body = columns
    .iter()
    .map(ToSql::to_sql)
    .chain(foreign_keys)
    .chain(primary_key)
    .collect::<Vec<_>>()
    .join(", ");
```

core/src/executor/fetch.rs
```diff
@@ -150,12 +149,26 @@ pub async fn fetch_relation_rows<'a, T: GStore>(
...
+                                )
+                                .await?,
+                            )?);
+                        }
+
+                        let key = if evaluated_primary_key_expressions.len() == 1 {
+                            Key::try_from(evaluated_primary_key_expressions.pop().unwrap())?
+                        } else {
+                            Key::try_from(evaluated_primary_key_expressions)?
+                        };
```
panarch, line 171: nit. using stream crate is quite useful here, like this below.

```rust
Some(IndexItem::PrimaryKey(exprs)) => {
    let key = stream::iter(exprs)
        .then(move |expr| {
            let filter_context = filter_context.as_ref().map(Rc::clone);

            async move {
                evaluate(storage, filter_context, None, expr)
                    .await
                    .and_then(Value::try_from)
                    .and_then(Key::try_from)
            }
        })
        .try_collect::<Vec<Key>>()
        .await
        .and_then(|keys| match keys.len() {
            1 => keys
                .into_iter()
                .next()
                .ok_or(FetchError::UnreachableEmptyPrimaryKey.into()),
            _ => Ok(Key::List(keys)),
        })?;
```

and we can consider adding an unreachable prefixed error to FetchError. 
```rust
    #[error("unreachable - primary key vector is empty")]
    UnreachableEmptyPrimaryKey,
```

core/src/data/value/selector.rs
LucaCappelletti94: Reverted

core/src/ast/ddl.rs
LucaCappelletti94: Cleaned up code as detailed.

core/src/ast/mod.rs
LucaCappelletti94: Cleaned up code as described

core/src/executor/fetch.rs
LucaCappelletti94: Refactored fetch code as described.

core/src/ast/expr.rs
LucaCappelletti94: Refactored code to remove try_into all together in a clean manner.

core/src/data/schema.rs
```diff
@@ -34,17 +35,128 @@ pub struct Schema {
...
+
+    /// Returns the table name associated with the schema.
+    pub fn table_name(&self) -> &str {
+        &self.table_name
+    }
+
+    /// Returns whether the schema has column definitions.
+    pub fn has_column_defs(&self) -> bool {
+        self.column_defs.is_some()
+    }
```
panarch, line 91: fields are public so could we just access by `self.primary_key`, `self.table_name` or `self.column_defs`?

core/src/data/schema.rs
```diff
@@ -34,17 +35,128 @@ pub struct Schema {
...
+    ) -> bool {
+        columns
+            .into_iter()
+            .any(|column| self.is_primary_key(column))
+    }
+
+    /// Returns the indices of the columns that compose the primary key.
+    pub fn get_primary_key_column_indices(&self) -> Option<&[usize]> {
+        self.primary_key.as_deref()
+    }
```
panarch, line 150: these, too.

core/src/data/schema.rs
LucaCappelletti94: Will remove shortly

core/src/data/schema.rs
LucaCappelletti94: Will remove shortly

core/src/data/schema.rs
LucaCappelletti94: Wait, I am not sure I understand - regarding table_name I agree, but the other two methods I would leave them be as they help with the code cleanliness and understandability.

LucaCappelletti94: I have now removed `table_name`, but I would prefer leaving the other two methods.

panarch: > Wait, I am not sure I understand - regarding table_name I agree, but the other two methods I would leave them be as they help with the code cleanliness and understandability.

if the `column_defs` and `primary_key` fields were private, then I am also for you. however, this is different.
I would think that this is against for code cleanliness, that we have two simple ways to achieve a single goal.
`has_column_defs()` or `self.column_defs.is_some()`.
someone may prefer `has_column_defs()` but others may not. and even someone thinks that it is better to use `self.column_defs.is_some()`.
without this kind of helper methods, we don't even need to care about this kind of selection issues.

if the helper method does a lot of roles so it is hard not to use that method, then that can be reasonable. but this case does not look like that.

LucaCappelletti94: Ok, I have removed them.

core/src/translate/mod.rs
```diff
@@ -97,30 +97,109 @@ pub fn translate(sql_statement: &SqlStatement) -> Result<Statement> {
             comment,
             ..
         } => {
-            let columns = columns
+            let mut primary_key: Vec<usize> = Vec::new();
```
panarch, line 100: how about checking the `constraints` first then columns?

if primary key constraint exists in the parsed schema, then we can assume that there will be no single column primary key declared in the list of sqlparser::ast::column_def.
then I think that the codes can be easier so we may not need comments for extra explanation.

panarch (commented):
this is a library project so we should not have any unwrap in the logic codes.
using panicable methods in test, surely its good to use that. but not in the logic.
could we remove all unwrap uses?
some of them can be simply removed but some may require code rearrangement, or some we might need to add unreachable or conflict errors.

if the error is not possible to make it due to that is caused by dependent library, then we can mark it `Unreachable-` prefix.
but in the case, that error is not possible to make it by normal logic but can be made by manipulating storage manually, then we can consider using `Conflict-` prefix.

core/src/translate/mod.rs
LucaCappelletti94: I do not think we can assume that, we would need to check and raise an adequate error if such a thing happens. I can clean up the code more though.

LucaCappelletti94: I have now cleaned up more the code and handled more corner cases that I had not covered earlier, such as multiple PRIMARY KEY definition in the column and mixed definitions in the columns and constraints. I have not swapped the columns and the constraints as the validation must happen for both, in any order.

LucaCappelletti94:
> this is a library project so we should not have any unwrap in the logic codes. using panicable methods in test, surely its good to use that. but not in the logic. could we remove all unwrap uses? some of them can be simply removed but some may require code rearrangement, or some we might need to add unreachable or conflict errors.
> 
> if the error is not possible to make it due to that is caused by dependent library, then we can mark it `Unreachable-` prefix. but in the case, that error is not possible to make it by normal logic but can be made by manipulating storage manually, then we can consider using `Conflict-` prefix.

I have now removed all unwrap statements relative to portions of the code that were not immediately checked within the context of the unwrap, i.e. all unwrap that remain are infallible but I do not know how to write the code in a cleaner manner.

core/src/ast/mod.rs
```diff
@@ -206,9 +207,24 @@ impl ToSql for Statement {
...
+
+                let primary_key = match (primary_key.as_ref(), columns.as_ref()) {
+                    (Some(indices), Some(columns)) => Some(format!(
+                        "PRIMARY KEY ({})",
+                        indices
+                            .iter()
+                            .copied()
+                            .map(|i| columns[i].name.clone())
+                            .collect::<Vec<_>>()
+                            .join(", ")
```
panarch, line 223: nit. clone uses here are not necessary.
```rust
columns
    .iter()
    .enumerate()
    .filter_map(|(i, column)| {
        indices.contains(&i).then_some(column.name.as_str())
    })
    .collect::<Vec<_>>()
    .join(", ")
```

core/src/ast_builder/create_table.rs
```diff
@@ -1,5 +1,9 @@
 use {
     super::Build,
+    crate::ast_builder::column_def::PrimaryKeyConstraintNode,
+    crate::error::TranslateError,
+    crate::parse_sql::parse_column_def,
+    crate::translate::translate_column_def,
     crate::{ast::Statement, ast_builder::ColumnDefNode, result::Result},
```
panarch, line 7: nit. let's merge this into a single `crate::{ .. }`

LucaCappelletti94: Is there a configuration of cargo fmt that can do that automatically or do I need to do that by hand?

panarch: as i know that this kind of merging use fmt option is only being supported in rust nightly. so.. I need to check it by myself.

LucaCappelletti94: I have no issue switching to the nightly build to run a fmt, do let me know if you can find the appropriate format option. It would surely massively speed up this process.

panarch: the last time I checked was may be about.. 3 years ago. so I dont remember that. could i get this from you if you ok..?

core/src/ast/mod.rs
LucaCappelletti94: Clone removed.

Any reason for switching from accessing by index to the contains? At this point in the code, we are guaranteed to have already validated the compatibility of columns and primary key indices, so if there is any out-of-bound access it would be relative to bugs in the code, and we would not want those to go unnoticed, right? A filter map may hide those.

core/src/data/key.rs
```diff
@@ -130,12 +132,24 @@ impl TryFrom<Value> for Key {
...
+    fn try_from(values: Vec<Value>) -> Result<Self> {
+        values
+            .into_iter()
+            .map(Key::try_from)
+            .collect::<Result<Vec<_>>>()
+            .map(Key::List)
+    }
+}
+
```
panarch, line 152: it would be not to have TryFrom. we already have too many From, TryFrom impls. it makes code harder to read but I didn't know that at the point I firstly add these.
currently, we are trying to use less From or TryFrom as possible.
this `TryFrom<Vec<Value>> for Key` is only being used in the above.
we can simply replace it.
```rust
List(values) => values
    .into_iter()
    .map(Key::try_from)
    .collect::<Result<Vec<_>>>()
    .map(Key::List),
```

core/src/ast/mod.rs
LucaCappelletti94: Resolved in commit #527f454e25f1e22146a96c762c3cff3b9870b068. Do let me know whether my approach is okay.

panarch: both indices first and columns first, they don't handle the unreachable validated error case of out-of-bound access.
only different is that the first one panics, the other does not panic.
if we dont have a plan to handle error here, i thought it would be better to remove panicable code.

core/src/data/key.rs
LucaCappelletti94: Resolved in commit b7f631b

core/src/translate/mod.rs
```diff
@@ -97,30 +97,72 @@ pub fn translate(sql_statement: &SqlStatement) -> Result<Statement> {
             comment,
             ..
         } => {
-            let columns = columns
+            let mut primary_key: Option<Vec<usize>> = None;
```
panarch, line 100: the code changes here, could you try writing the code without using `mut` keyword?
I think that then we can simplify the codes a lot.
if you stuck at something, then i will support on it.

core/src/translate/mod.rs
```diff
@@ -97,30 +97,72 @@ pub fn translate(sql_statement: &SqlStatement) -> Result<Statement> {
...
+                                            TranslateError::ColumnNotFoundInTable(
+                                                column.to_string(),
+                                            )
+                                        })?,
+                                );
+                            }
+                        }
+                    }
+                } else {
+                    foreign_keys.push(translate_foreign_key(constraint)?);
```
panarch, line 149: now we handle more than one constraint variant, so it would be good to change the method `translate_foreign_key` to `translate_contraint`.

core/src/ast_builder/create_table.rs
LucaCappelletti94: Ok found it, I have also created a `rustfmt.toml` defining this configuration so that it will be standardized.

LucaCappelletti94: Done in commit 9b75794

panarch:
formatting configuration should not be changed here. that is not related to this pr. we need to handle it separately.
and.. you should not change the toolchain from stable to nightly.
we dont have plan to drop rust stable support.

LucaCappelletti94:
> formatting configuration should not be changed here. that is not related to this pr. we need to handle it separately. and.. you should not change the toolchain from stable to nightly. we dont have plan to drop rust stable support.

Just switched back and forth from nightly to stable for this format operation. Will remove the formatting configuration shortly.

core/src/ast/mod.rs
LucaCappelletti94: While I agree that generally code that may panic is to be removed, here we are operating on an instantiated `CreateTable`, which carries several assumptions, one of which is that the primary key is compatible with the columns. If a non-self-consistent CreateTable were to be created, there should be a panic, as it is not an error relative to how the user is employing the library, but a library bug of which the panic would be an important signal.

core/src/translate/mod.rs
LucaCappelletti94: This would imply creating an enumeration of constraints. Should we do that? At this time, different constraints are handled separately.

core/src/translate/mod.rs
LucaCappelletti94: While I agree that this code (and other similar points) is rather ugly, I must admit that I am not sure how to clean it up. Do you have any suggestions?

core/src/translate/mod.rs
panarch: this is simply because the `translate_foreign_key` gets param as constraint. it is fully ok to separate it. but it would be simple to handle `translate_contraint` rather than destructuring enum variant and pass them to the seprate function.

core/src/ast/mod.rs
panarch: at least in the gluesql core workspace, we don't have plan to leave any kind of panicable codes. if you find it, that can be a bug and may be something that I missed during the review. we can fix it.
yes that can be some panicable case can happen. then, we can mark it using `Unreachable-` prefixed error rather than killing the program.
that has similar role but the only difference is that our library will not kill the application by itself.
of course we cannot make ideal, may be some dependent library has lots of unsafe codes and panicable codes, but at least we can control the code that we write.

panarch:
> > formatting configuration should not be changed here. that is not related to this pr. we need to handle it separately. and.. you should not change the toolchain from stable to nightly. we dont have plan to drop rust stable support.
> 
> Just switched back and forth from nightly to stable for this format operation. Will remove the formatting configuration shortly.

this change caused lots of unexpected code changes here. could you.. revert those, too?

core/src/translate/mod.rs
panarch: sure, for this part, I will provide suggestions on it. please give me some time. you can focus on other points.

LucaCappelletti94:
> > > formatting configuration should not be changed here. that is not related to this pr. we need to handle it separately. and.. you should not change the toolchain from stable to nightly. we dont have plan to drop rust stable support.
> > 
> > 
> > Just switched back and forth from nightly to stable for this format operation. Will remove the formatting configuration shortly.
> 
> this change caused lots of unexpected code changes here. could you.. revert those, too?

Ok, but if I revert to commit [b7f631b](https://github.com/gluesql/gluesql/pull/1515/commits/b7f631b573837758a0eb8956687cddf4b3b340c6) we still have to figure out how to uniform those import statements without the nightly refactor.

panarch:
%

> > > > formatting configuration should not be changed here. that is not related to this pr. we need to handle it separately. and.. you should not change the toolchain from stable to nightly. we dont have plan to drop rust stable support.
> > > 
> > > 
> > > Just switched back and forth from nightly to stable for this format operation. Will remove the formatting configuration shortly.
> > 
> > 
> > this change caused lots of unexpected code changes here. could you.. revert those, too?
> 
> Ok, but if I revert to commit [b7f631b](https://github.com/gluesql/gluesql/pull/1515/commits/b7f631b573837758a0eb8956687cddf4b3b340c6) we still have to figure out how to uniform those import statements without the nightly refactor.

for missing parts, I will be leaving comments. you can change it for something you find, remaining ones, I will

LucaCappelletti94:
> %
> 
> > > > > formatting configuration should not be changed here. that is not related to this pr. we need to handle it separately. and.. you should not change the toolchain from stable to nightly. we dont have plan to drop rust stable support.
> > > > 
> > > > 
> > > > Just switched back and forth from nightly to stable for this format operation. Will remove the formatting configuration shortly.
> > > 
> > > 
> > > this change caused lots of unexpected code changes here. could you.. revert those, too?
> > 
> > 
> > Ok, but if I revert to commit [b7f631b](https://github.com/gluesql/gluesql/pull/1515/commits/b7f631b573837758a0eb8956687cddf4b3b340c6) we still have to figure out how to uniform those import statements without the nightly refactor.
> 
> for missing parts, I will be leaving comments. you can change it for something you find, remaining ones, I will

I have formatted all of the imports by hand following the diffs against a version that I separately formatted using `cargo fmt` in nightly mode. Took some time but this way I got all of them.

core/src/ast_builder/create_table.rs
LucaCappelletti94: Done, manually, in commit [9c47a24](https://github.com/gluesql/gluesql/pull/1515/commits/9c47a24e02fb457ea12d679eaabf46bad2a60525) and 
9f24309

core/src/ast/mod.rs
LucaCappelletti94: I have refactored the code into the version you initially proposed in commit 14f1f7a1504a50fa766e55fc4919ff621f6db45c

core/src/translate/mod.rs
LucaCappelletti94: I have cleaned up more that area in commit 44f96e4bdbba05f211935f87685777c02f6c5b3e, I believe it should be a decent solution especially foreseeing how that match will expand in the UNIQUE and CHECK pull requests.

core/src/translate/mod.rs
LucaCappelletti94: Cleaned this up.

core/src/translate/mod.rs
LucaCappelletti94: Cleaned this up even more.

core/src/ast_builder/expr/function.rs
```diff
@@ -1,10 +1,7 @@
...
-    super::ExprNode,
-    crate::{
-        ast::{DateTimeField, Function},
-        ast_builder::{DataTypeNode, ExprList},
-        result::{Error, Result},
-    },
+use crate::{
+    ast::{DateTimeField, Function},
+    ast_builder::{DataTypeNode, ExprList, ExprNode},
+    result::{Error, Result},
```
panarch, line 4: could we deal these kind of changes in the separate PR?
if we can have these changes in another PR, then we can directly ship these to the main.
the changes themselves are very nice but these are not related to the composite primary key PR?

LucaCappelletti94: Ok - I will try to do a PR on which we can then re-align this PR. Hopefully, it will not be too much of a mess.

core/src/plan/primary_key.rs
```diff
@@ -1,3 +1,13 @@
+//! Submodule providing the planner for handling primary keys.
+//!
+//! # Primary Key Planner
+//! The primary goal of the primary key planner is to optimize the query by using the primary key
+//! whenever possible. This is done by checking if the primary key is used in the query, for instance
+//! as part of the WHERE clause. If the primary key is used, the planner will remove the primary key
+//! from the WHERE clause and move it into the index field. This will allow the query to be executed
+//! more efficiently by calling directly the Store::fetch_data instead of the Store::scan_data.
+//!
+
```
panarch, line 10: there looks that some edge cases still exist.
e.g.
```rust
let sql = "
    SELECT * FROM Player WHERE
        id = 100
        AND name = 'foo'
        AND (
            SELECT * FROM Player WHERE name = 'bar' AND Player.id = 1
        )
;";
```
in this case, the current planner cannot find primary key in subquery.

LucaCappelletti94: Ok, please add a couple of tests to this case.

core/src/ast_builder/expr/function.rs
LucaCappelletti94: Done, see PR #1527

core/src/plan/primary_key.rs
LucaCappelletti94: I have managed to add a few tests and reproducing, now fixing the thing.

LucaCappelletti94: Should now be handled by commit [71c4fc2](https://github.com/gluesql/gluesql/pull/1515/commits/71c4fc2db7d3f1a771a5bcff50764f5c291c82e7)

core/src/plan/primary_key.rs
```diff
@@ -101,113 +195,407 @@ impl<'a> PrimaryKeyPlanner<'a> {
...
     ) -> PrimaryKey {
-        let check_primary_key = |key: &Expr| {
-            let key = match key {
-                Expr::Identifier(ident) => ident,
-                Expr::CompoundIdentifier { ident, .. } => ident,
-                _ => return false,
+        enum PossibleResults {
+            Matched(PrimaryKey),
+            Retry(Box<Expr>, Box<Expr>),
+        }
```
panarch, line 204: ok, now I am beginning to find some points.
here, we currently add Partial variant to PrimaryKey enum.
rather than adding `PrimaryKey::Partial`, could we consider to add an extra param for partial primary key expression to `fn expr(..)`? then we might not need `PrimaryKey::Partial` and it may simplify the code logic.
I wonder your opinion. currently I'm not completely sure yet either.

LucaCappelletti94: The issue is that it is a result of the expr function, not a parameter of the function.

core/src/plan/primary_key.rs
```diff
@@ -101,113 +195,407 @@ impl<'a> PrimaryKeyPlanner<'a> {
...
+
+        // Returns the primary key variant associated to the provided key and value, if any.
+        let get_primary_key = |key: Box<Expr>, value: Box<Expr>| {
+            if !(check_evaluable(Some(current_context.clone()), &key)
+                && check_evaluable(None, &value))
+            {
+                return PossibleResults::Retry(key, value);
```
panarch, line 211: I am also not sure that we really need `PossibleResults::Retry` like repeatable logics.
the current state of primary key planner is very simple. just a logical planner.
if the planner find the first matched one, then it's enough. so there is no logic required for selecting the best options or something like that.

LucaCappelletti94: The Retry is needed to avoid some avoidable clones. If you prefer to clone, I can remove it.

LucaCappelletti94: See commit 373d427

panarch: ok, if that `Retry` is being used because of avoiding clones, then let's use clones. then we can find the way to remove clone without Retry logics.

core/src/plan/primary_key.rs
panarch: expr function is defined only for primary key planner. we don't need to be limited to the existing shape of the codes. then could we find something on it?

LucaCappelletti94: Since the method is recursive, as it is appropriate given that the task has a recursive definition, we need a way to uniquely represent when a step of the recursion yields a partial result. Other than adding an enum option that represents it, which is the current solution, I have no other ideas.

core/src/plan/primary_key.rs
LucaCappelletti94: Removed Retry. I do not think we can remove the clones without introducing the Retry variant.

core/src/plan/primary_key.rs
panarch: ok, lets keep talking on the comment below, there is the point I see. 
https://github.com/gluesql/gluesql/pull/1515#discussion_r1703113335

core/src/plan/primary_key.rs
```diff
@@ -101,113 +195,398 @@ impl<'a> PrimaryKeyPlanner<'a> {
...
+                    self.expr(
+                        outer_context.as_ref().map(Rc::clone),
+                        current_context,
+                        *left,
+                    ),
+                    self.expr(
+                        outer_context.as_ref().map(Rc::clone),
+                        current_context,
+                        *right.clone(),
+                    ),
```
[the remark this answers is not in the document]
LucaCappelletti94: Before processing either the left or right variants, we need both of the results otherwise the subqueries are not handled properly (that was the issue beforehand). I am not sure I understood what you mean by "parallel". I just re-read the code and tried to see if I could find further simplifications, and I do not see any.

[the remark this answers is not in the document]
panarch: ok, I will look deeper on it and leave comments. please let me have some more time on this.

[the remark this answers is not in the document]
LucaCappelletti94: Cleaned up implementation more in commits [5f89a1d](https://github.com/gluesql/gluesql/pull/1515/commits/5f89a1d0b941dc62a24b895d78bebffb3623a32a) and 
f81d8ea

panarch:
hi, sorry for the late review for this. I need to spend some more effort on reviewing primary key planner but there's some urgent tasks I need to finish until 28th, this month. I can review for small other PRs but for this, I will be back after 28th.

jeromegn:
I'd love this feature to land :) is there still more work required here?

Issue #1516, opened 2024-07-17
CASCADE on foreign key constraint is not supported.


LucaCappelletti94:
As per the title, at this time, having a CASCADE on a foreign key constraint will raise `Translate(UnsupportedConstraint("CASCADE"))`.

I will look into it.

PR #1517, opened 2024-07-17
Introducing support for foreign key with Restrict, Cascade, Set Null and Set Default

LucaCappelletti94:
This pull request solves part of issue #1516, introducing ON DELETE CASCADE, SET NULL and SET DEFAULT.
It DOES NOT add support for the ON UPDATE counter parts as, as, at this time, Unique keys cannot be used as FK nor primary keys can be updated.

This pull request is dependent on pull request #1515 being merged first.


core/src/ast/mod.rs
```diff
@@ -42,7 +42,59 @@ pub struct ForeignKey {
...
+    /// row(s) that reference it should have their referencing column(s) set to NULL.
+    SetNull,
+    #[strum(to_string = "SET DEFAULT")]
+    /// A SET DEFAULT constraint specifies that when a referenced row is deleted,
+    /// row(s) that reference it should have their referencing column(s) set to the column's default value.
+    SetDefault,
+    #[strum(to_string = "RESTRICT")]
+    /// A RESTRICT constraint specifies that when a referenced row is attempted to be deleted,
+    /// the deletion should be restricted i.e. prevented.
+    Restrict,
```
devgony, line 63: We decided merge `NO ACTION and RESTRICT` to NO ACTION so that it would be better to remove Restrict variant here.

core/src/ast/mod.rs
```diff
@@ -42,7 +42,59 @@ pub struct ForeignKey {
...
+    /// Returns whether the referential action is setting the referencing column to NULL.
+    pub fn is_set_null(&self) -> bool {
+        matches!(self, ReferentialAction::SetNull)
+    }
+
+    /// Returns whether the referential action is setting the referencing column to the column's default value.
+    pub fn is_set_default(&self) -> bool {
+        matches!(self, ReferentialAction::SetDefault)
+    }
+}
```
devgony, line 86: Seems like these methods were not used at all.

core/src/ast/mod.rs
LucaCappelletti94: Ok!

core/src/ast/mod.rs
LucaCappelletti94: I will remove those

LucaCappelletti94: Done!

core/src/ast/mod.rs
LucaCappelletti94: Done!

core/src/executor/delete.rs
```diff
@@ -14,27 +14,46 @@ use {
...
+use crate::prelude::Key;
+use futures::Future;
+use std::pin::Pin;
+
 #[derive(ThisError, Serialize, Debug, PartialEq, Eq)]
 pub enum DeleteError {
-    #[error("referencing column exists: {0}")]
-    ReferencingColumnExists(String),
+    #[error("Restrict reference column exists: {0}")]
+    RestrictingColumnExists(String),
```
devgony, line 25: What was the purpose of this name `RestrictingColumnExists`?
When I decided name `ReferencingColumnExists`, I focused on `Referencing` cuz FK has two directions `Referencing` and `Referenced` by which both of Rust developer and DB User can be confused.

LucaCappelletti94: That makes sense - I think I just meant to write `ReferencingColumnExists` and then I paused for a coffee and my brain reset. I will fix that shortly.

LucaCappelletti94: Ah yes, now looking at the code I recalled - The thing is that error is only raised when a Restricting constraint is in place, it is not about solely a Referencing column. This way, the error was more precise, detailing what was causing it.

devgony: Oh.. What i'm worried about is this part.
```rs
                        NoAction => {
                            return Err(DeleteError::RestrictingColumnExists(format!(
                                "{referencing_table_name}.{referencing_column_name}"
                            ))
                            .into());
                        }
```

https://github.com/gluesql/gluesql/pull/1517/files#diff-fb5db5309bc066a3ff406c630d67a7e4f0d9fcde46ab477e920a197a6f7c6cb4R106-R110

Cuz the actual arguments of Error are originated from `referencing~`.
And This is the only error case if there is Referencing column among 4 `ReferentialAction`s so that i thought `ReferencingColumnExists` would be enough.
And now even `ReferentialAction::Restrict` disappeared..

LucaCappelletti94: I am not sure I understood your comment - which part would you like me to change?

Also, Restrict was removed as in another note you stated it should be handled as NoAction, so now I removed that variant. Could you clarify what is the problem there?

test-suite/src/foreign_key.rs
```diff
@@ -87,42 +87,422 @@ test_case!(foreign_key, {
...
+        Ok(Payload::Select {
+            labels: vec![
+                "id".to_owned(),
+                "name".to_owned(),
+                "referenced_id".to_owned(),
+            ],
+            rows: vec![],
+        }),
+    )
+    .await;
```
devgony, line 133: These look wrong tests.
`ON UPDATE CASCADE` means If the refereced key is updated, referencing value should be updated cacadely.

Tests should be like this
```rs
    g.named_test(
        "Updating referenced row should update referencing row as well",
        "UPDATE ReferencedTableWithPK SET id = 3 WHERE id = 1;",
        Ok(Payload::Update(1)),
    )
    .await;

    g.named_test(
        "Referencing row should be deleted on update cascade",
        "SELECT id FROM ReferencingTableCascade WHERE referenced_id = 3;",
        Ok(Payload::Select {
            labels: vec!["id".to_owned()],
            rows: vec![vec![Value::I64(1), Value::I64(3)]],
        }),
    )
    .await;
```

But currently you will get `UpdateOnPrimaryKeyNotSupported`. Cuz we decided to prohibit updating PK.

So i suggest,
1. Implement FK on UNIQUE
2. Write test cases of `ON UPDATE CASCADE` of `ReferencedTableWithUnique`
3. Fix `ON UPDATE CASCADE` logic from deleting to updating

devgony (changes requested):

core/src/executor/delete.rs
devgony: I want to keep the name `ReferencingColumnExists`

LucaCappelletti94: Ok, reverting change.

LucaCappelletti94: Reverted in commit [56e7e9b](https://github.com/gluesql/gluesql/pull/1517/commits/56e7e9bb5a6bbc3d505ffc3ee357c188e9e736a5)

test-suite/src/foreign_key.rs
LucaCappelletti94: If `UpdateOnPrimaryKeyNotSupported`, then `ON UPDATE CASCADE` should be just illegal, right?

LucaCappelletti94: Ok so, I have now removed the wrong implementation of the on update X in the update method and rewrote the test suite regarding the use of the ON UPDATE x. Now the tests are correct. I have separated in two documents the tests regarding the ON UPDATE and the ON DELETE, since the expected behaviour is very different from what a fully working implementation should be. I have documented on the top of the ON UPDATE test suite the reason for this.

This was one in commit [b962010](https://github.com/gluesql/gluesql/pull/1517/commits/b96201060a34ffe9c16c98f2f5187b6908bae04e)

test-suite/src/foreign_key_on_update.rs
```diff
@@ -0,0 +1,221 @@
...
+    .await;
+
+    g.named_test(
+        "On update cascade should fail as primary key cannot be updated",
+        "CREATE TABLE Referencing (
+            id INTEGER PRIMARY KEY,
+            referenced_id INTEGER,
+            FOREIGN KEY (referenced_id) REFERENCES Referenced(id) ON UPDATE CASCADE
+        );",
+        Err(TranslateError::UnsupportedConstraint("ON UPDATE CASCADE".to_owned()).into()),
```
devgony, line 54: > What do you mean by Implement FK on UNIQUE ?

FK can be created on Column with Unique constraint as well.
And Unique column can be updated though PK can't.

So i want `cascade on unique` to be implemented.
1. Implement FK on UNIQUE
2. Write test cases of ON UPDATE CASCADE of ReferencedTableWithUnique
3. updating ReferencedTableWithUnique should update ReferencingTable as well.

devgony (changes requested):

test-suite/src/foreign_key_on_update.rs
LucaCappelletti94: Isn't this something better suited for another PR?

devgony: > Isn't this something better suited for another PR?

Yes, you'd better separate PR to implement FK on Unique !

PR #1529, opened 2024-08-07
FileStorage

panarch:
FileStorage is a custom storage implementation that utilizes the filesystem.

For each table name, the schema information is saved in a `.sql` format using a `CREATE TABLE` query.

The data is stored by creating a directory with the same name as the table and serializing the data using [RON](https://github.com/ron-rs/ron) format in the subdirectory.


.github/workflows/rust.yml
```diff
@@ -67,6 +67,7 @@ jobs:
           cd json-storage && cargo test --verbose && cd ..
           cd csv-storage && cargo test --verbose && cd ..
           cd parquet-storage && cargo test --verbose && cd ..
+          cs file-storage && cargo test --verbose && cd ..
```
devgony, line 70: is `cs` typro?

panarch: ooooh, thanks a lot!

storages/file-storage/src/store.rs
```diff
@@ -0,0 +1,95 @@
...
+
+#[async_trait(?Send)]
+impl Store for FileStorage {
+    async fn fetch_all_schemas(&self) -> Result<Vec<Schema>> {
+        let mut schemas = fs::read_dir(&self.path)
+            .map_storage_err()?
+            .map(|dir_entry| {
+                let dir_entry = dir_entry.map_storage_err()?;
+                let file_type = dir_entry.file_type().map_storage_err()?;
+                if file_type.is_dir() {
```
devgony, line 21:
```suggestion
                let path = dir_entry.path();
                let extension = path.extension().and_then(OsStr::to_str);
                if file_type.is_dir() || extension != Some("sql") {
```

Would it be better to validate if it's *.sql as well?

devgony: Or.. maybe we can reuse fetch_schema

storages/file-storage/src/store.rs
```diff
@@ -0,0 +1,88 @@
...
+        let mut entries = fs::read_dir(path)
+            .map_storage_err()?
+            .map(|dir_entry| {
+                let dir_entry = dir_entry.map_storage_err()?;
+                let file_type = dir_entry.file_type().map_storage_err()?;
+                if file_type.is_dir() {
+                    return Ok(None);
+                }
+
+                Ok(Some(dir_entry.path()))
```
devgony, line 71:
```suggestion
                let path = dir_entry.path();
                let extension = path.extension().and_then(OsStr::to_str);
                if file_type.is_dir() || extension != Some("ron") {
                    return Ok(None);
                }

                Ok(Some(path))
```

Similar pattern exists in scan_data !

devgony (approved):

PR #1550, opened 2024-08-13
Fix empty commit, Add command validation on GitStorage

devgony:
## Todo
- [x] extract fn GitStorage::git
- [x] validate std_err on executed command
- [x] fix empty commit with `early returning`

```sh
> cargo test  -p gluesql-git-storage arithmetic_error
[RUN] DELETE FROM Arith
#debug> git add .
#debug> git commit -m [GitStorage::delete_data] Arith - 0 rows
#debug> stdout: "On branch main\nnothing to commit, working tree clean\n"
#debug> status: exit status: 1
test arithmetic_error ... FAILED
```
- [ ] fix drop table consequently
```sh
CREATE TABLE DropTable (
    id INT,
    num INT,
    name TEXT
)
[RUN] DROP TABLE IF EXISTS DropTable;
[RUN] DROP TABLE IF EXISTS DropTable;
#debug>status: exit status: 1
#debug>stdout: On branch main
nothing to commit, working tree clean

test drop_table ... FAILED
```

PR #1551, opened 2024-08-15
Add the number of dropped tables to Payload::DropTable

devgony:
## Todo
- [x] Add the number of dropped tables to Payload::Droptable
- [x] early return if table does not exist
- should we use Arc? -> NO, use mut instead.
## Next
- Handle delete_schema bug in #1550

### DIFF
--- a/core/src/executor/alter/table.rs
+++ b/core/src/executor/alter/table.rs
@@ -216,7 +216,7 @@ pub async fn drop_table<T: GStore + GStoreMut>(
     table_names: &[String],
     if_exists: bool,
     cascade: bool,
-) -> Result<()> {
+) -> Result<usize> {
     for table_name in table_names {
         let schema = storage.fetch_schema(table_name).await?;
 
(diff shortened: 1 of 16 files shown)


core/src/executor/alter/table.rs
```diff
@@ -231,11 +235,13 @@ pub async fn drop_table<T: GStore + GStoreMut>(
         let referencings = storage.fetch_referencings(table_name).await?;
 
         if !referencings.is_empty() && !cascade {
-            return Err(AlterError::CannotDropTableWithReferencing {
-                referenced_table_name: table_name.into(),
-                referencings,
-            }
-            .into());
+            return Err::<usize, Error>(
```
panarch, line 238: these type can be inferenced.
as it was, `Err(..)` would be enough.

ever0de (approved):

panarch (approved):
Looks all nice! Thanks for the contribution 👍 👍

PR #1552, opened 2024-08-16
Fix drop_table to return early if schema not found with IF EXISTS

devgony:
## Todo
- [x] fix: early return if schema not found
## Next
- fix delete_schema bug in #1550 
- #1551


panarch (approved):
Thanks a lot!

PR #1570, opened 2024-09-06
Upgrading from SqlParser 0.47 to 0.48

LucaCappelletti94:
Most of the files updated are solely relative to the trailing commas present in many tests, which are not valid SQL syntax and as such, the SQL parser was not failing to process those.

### DIFF
--- a/core/src/translate/data_type.rs
+++ b/core/src/translate/data_type.rs
@@ -14,6 +14,17 @@ pub fn translate_data_type(sql_data_type: &SqlDataType) -> Result<DataType> {
             Ok(DataType::Int)
         }
         SqlDataType::Int8(None) => Ok(DataType::Int8),
+        SqlDataType::Int16 => Ok(DataType::Int16),
+        SqlDataType::Int32 => Ok(DataType::Int32),
+        SqlDataType::Int128 => Ok(DataType::Int128),
+        SqlDataType::UInt8 => Ok(DataType::Uint8),
+        SqlDataType::UInt16 => Ok(DataType::Uint16),
+        SqlDataType::UInt32 => Ok(DataType::Uint32),
+        SqlDataType::UInt64 => Ok(DataType::Uint64),
+        SqlDataType::UInt128 => Ok(DataType::Uint128),
+
+        SqlDataType::Float32 => Ok(DataType::Float32),
+        SqlDataType::Float64 => Ok(DataType::Float),
 
         SqlDataType::UnsignedInt(None) | SqlDataType::UnsignedInteger(None) => Ok(DataType::Uint64),
         SqlDataType::UnsignedInt8(None) => Ok(DataType::Uint8),
--- a/core/src/translate/data_type.rs
+++ b/core/src/translate/data_type.rs
@@ -72,10 +72,6 @@ mod tests {
 
         test!("INT" => SqlDataType::Int(None) => Ok(DataType::Int));
         test!("INTEGER" => SqlDataType::Integer(None) => Ok(DataType::Int));
-        test!("INT64" => SqlDataType::Int64 => Ok(DataType::Int));
-
-        test!("INT8" => SqlDataType::Int8(None) => Ok(DataType::Int8));
-
         test!("INT UNSIGNED" => SqlDataType::UnsignedInt(None) => Ok(DataType::Uint64));
         test!("INTEGER UNSIGNED" => SqlDataType::UnsignedInteger(None) => Ok(DataType::Uint64));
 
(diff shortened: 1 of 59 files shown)


core/src/translate/data_type.rs
```diff
@@ -14,6 +14,17 @@ pub fn translate_data_type(sql_data_type: &SqlDataType) -> Result<DataType> {
...
+        SqlDataType::Int32 => Ok(DataType::Int32),
+        SqlDataType::Int128 => Ok(DataType::Int128),
+        SqlDataType::UInt8 => Ok(DataType::Uint8),
+        SqlDataType::UInt16 => Ok(DataType::Uint16),
+        SqlDataType::UInt32 => Ok(DataType::Uint32),
+        SqlDataType::UInt64 => Ok(DataType::Uint64),
+        SqlDataType::UInt128 => Ok(DataType::Uint128),
+
+        SqlDataType::Float32 => Ok(DataType::Float32),
+        SqlDataType::Float64 => Ok(DataType::Float),
```
panarch, line 27: Nice, then we no longer need to keep `UINT8`, .. in `SqlDataType::Custom(_)` that are now unreachable codes.
we can remove those.

panarch: image

LucaCappelletti94: If I remove those, these tests fail:

```
test!("INT8" => Ok(DataType::Int8));
test!("INT16" => Ok(DataType::Int16));
test!("INT32" => Ok(DataType::Int32));
test!("INT128" => Ok(DataType::Int128));

test!("UINT8" => Ok(DataType::Uint8));
test!("UINT16" => Ok(DataType::Uint16));
test!("UINT32" => Ok(DataType::Uint32));
test!("UINT64" => Ok(DataType::Uint64));
test!("UINT128" => Ok(DataType::Uint128));
```

Should I remove these tests?

core/src/translate/mod.rs
```diff
@@ -281,20 +288,31 @@ pub fn translate(sql_statement: &SqlStatement) -> Result<Statement> {
...
 
     Ok(Assignment {
-        id: id
-            .first()
-            .ok_or(TranslateError::UnreachableEmptyIdent)?
-            .value
-            .to_owned(),
+        id: target.value.to_owned(),
         value: translate_expr(value)?,
     })
```
panarch, line 317: here, it would be good to add a new error for indicating tuple assignment is not supported.
// translate/error.rs - to add TupleAssignmentOnUpdate .. 
image


then,
```rust
    let SqlAssignment { target, value } = sql_assignment;
    let id = match target {
        SqlAssignmentTarget::Tuple(_) => {
            return Err(TranslateError::TupleAssignmentOnUpdateNotSupported(
                sql_assignment.to_string(),
            )
            .into());
        }
        SqlAssignmentTarget::ColumnName(SqlObjectName(id)) => id,
    };

    if id.len() > 1 {
        return Err(
            TranslateError::CompoundIdentOnUpdateNotSupported(sql_assignment.to_string()).into(),
        );
    }

    Ok(Assignment {
        id: id
            .first()
            .ok_or(TranslateError::UnreachableEmptyIdent)?
            .value
            .to_owned(),
        value: translate_expr(value)?,
    })
```

core/src/translate/mod.rs
```diff
@@ -399,4 +416,16 @@ mod tests {
...
+    fn test_tuple_assignment_on_update_not_supported() {
+        let sql = "UPDATE Foo SET (a, b) = (1, 2)";
+        let actual = parse(sql).and_then(|parsed| translate(&parsed[0]));
+        let expected = Err(TranslateError::TupleAssignmentOnUpdateNotSupported(
+            "(a, b) = (1, 2)".to_owned(),
+        )
+        .into());
+
+        assert_eq!(actual, expected);
+    }
```
panarch, line 430: Amazing, thanks a lot 👍

test-suite/src/update.rs
```diff
@@ -109,17 +109,17 @@ test_case!(update, {
...
-        ),
-        (
-            "UPDATE (SELECT * FROM ErrTestTable) SET 1 = 1",
-            Err(
-                TranslateError::UnsupportedTableFactor("(SELECT * FROM ErrTestTable)".to_owned())
-                    .into(),
-            ),
-        ),
+        // (
+        //     "UPDATE TableA INNER JOIN ErrTestTable ON 1 = 1 SET 1 = 1",
```
panarch, line 113: ok, here sqlparser become more accurate. so we need to change this to
```sql
UPDATE TableA INNER JOIN ErrTestTable ON 1 = 1 SET id = 1
```
then, it would be ok.

LucaCappelletti94: Done in commit 067a5af

panarch (approved):
Everything looks great. thanks a lot!

PR #1571, opened 2024-09-06
Upgrading from SQL Parser 0.48 to 0.49

LucaCappelletti94:


core/src/translate/query.rs
```diff
@@ -29,10 +29,15 @@ pub fn translate_query(sql_query: &SqlQuery) -> Result<Query> {
...
-        .collect::<Result<_>>()?;
+    let order_by = if let Some(order_by) = order_by {
+        order_by
+            .exprs
+            .iter()
+            .map(translate_order_by_expr)
+            .collect::<Result<_>>()?
+    } else {
+        vec![]
+    };
```
panarch, line 40: nit. it would be good to use flat_map here
```rust
    let order_by = order_by
        .iter()
        .flat_map(|order_by| order_by.exprs.iter().map(translate_order_by_expr))
        .collect::<Result<_>>()?;
```

LucaCappelletti94: I thought about that but I wasn't sure which one you'd prefer. Okay.

LucaCappelletti94: Done

panarch (approved):
wow, already 0.49. looks all nice! Thanks a lot 👍 👍

PR #1573, opened 2024-09-06
Implemented Check constraints

LucaCappelletti94:
A new version of the [older check constraints pull request](https://github.com/gluesql/gluesql/pull/1524). This pull request is considerably smaller than the earlier one.

### DIFF
--- a/core/src/ast/check_constraint.rs
+++ b/core/src/ast/check_constraint.rs
@@ -1,0 +1,108 @@
+//! Submodule implementing the CHECK constraint AST node.
+
+use serde::{Deserialize, Serialize};
+
+use super::{Expr, ToSql};
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Hash, Deserialize)]
+/// Struct representing a CHECK constraint.
+///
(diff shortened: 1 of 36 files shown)


LucaCappelletti94:
The reason for having that significant refactoring in implementing the `eq` and `cmp` operation as part of this pull request, which was surely needed as at this time comparing strings to integers was not raising an adequate error, is due to how the check constraints operate on expressions that evaluate to a Null.

Previously, statements comparing anything to a `Null`, instead of returning `Null`, returned `false`. This is fine for what entails a WHERE statement, as in that context expression that evaluates to `Null` are considered `false`, but in CHECK constraints an expression that returns `Null` is considered `true`.

For instance, suppose you have a table with an optional column `nickname` which you do not want to be longer than 500 characters, and you enforce this constraint with a CHECK. When you insert a new row without any nickname, you do not want the CHECK to block the insert by default, hence why `Null` are considered `true` in CHECK constraints.

LucaCappelletti94:
Moved commits relative to None cmp-eq fix tor PR #1574

PR #1574, opened 2024-09-08
Now comparing with None evaluates to None and not False

LucaCappelletti94:
The reason for having that significant refactoring in implementing the `eq` and `cmp` operation as part of this pull request, which was surely needed as at this time comparing strings to integers was not raising an adequate error, is due to how the check constraints operate on expressions that evaluate to a Null.

Previously, statements comparing anything to a Null returned false instead of returning Null. This is fine for what entails a WHERE statement, as in that context expression that evaluates to Null are considered false, but in CHECK constraints an expression that returns Null is considered true.

For instance, suppose you have a table with an optional column nickname that you do not want to be longer than 500 characters, and you enforce this constraint with a CHECK. When you insert a new row without any nickname, you do not want the CHECK to block the insert by default, hence why Null are considered true in CHECK constraints.

### DIFF
--- a/core/src/ast/data_type.rs
+++ b/core/src/ast/data_type.rs
@@ -31,5 +31,4 @@ pub enum DataType {
     List,
     Decimal,
     Point,
-    Null,
 }
--- a/core/src/ast/data_type.rs
+++ b/core/src/ast/data_type.rs
@@ -31,4 +31,5 @@ pub enum DataType {
     List,
     Decimal,
     Point,
+    Null,
 }
--- a/core/src/ast/operator.rs
+++ b/core/src/ast/operator.rs
@@ -1,6 +1,7 @@
 use {
     crate::ast::ToSql,
     serde::{Deserialize, Serialize},
+    std::fmt::Display,
 };
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
@@ -24,7 +25,7 @@ impl ToSql for UnaryOperator {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
 pub enum BinaryOperator {
     Plus,
     Minus,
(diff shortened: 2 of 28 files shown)


core/src/ast/data_type.rs
```diff
@@ -31,4 +31,5 @@ pub enum DataType {
     List,
     Decimal,
     Point,
+    Null,
```
panarch, line 34: Do we really need `DataType::Null`? `NULL` can be the value but it is hard to see the reason to have `NULL` as data type.

LucaCappelletti94: So the data type of a Null object would be? We can make it so the datatype() method returns an option, with None for that case.

panarch: ok, I thought that the this pr is for comparing between incompatible types to return error rather than false boolean value. could we remove other elements which are not directly related to the issue? then, it would be good to focus on the core

core/src/ast/operator.rs
```diff
@@ -46,31 +47,37 @@ pub enum BinaryOperator {
     BitwiseShiftRight,
 }
 
-impl ToSql for BinaryOperator {
-    fn to_sql(&self) -> String {
+impl Display for BinaryOperator {
```
panarch, line 50: we can use `ToSql`. I understand that it looks just ok to have `Display` impl, but basically we are trying to reduce the use of `From`, `TryFrom` and `Display` like implicit ones.

LucaCappelletti94: The implementation of `Display` was to implement `Debug`, and having both `ToSql` and `Display` was excessive code duplication.

core/src/ast/data_type.rs
LucaCappelletti94: The reason for the introduction of the [`datatype()` method](https://github.com/gluesql/gluesql/blob/3410f789b50e825b01e2e154dcf97608af47d9d6/core/src/data/value/literal.rs#L129-L158) is that it was needed to instantiate [the error when the comparison fails](https://github.com/gluesql/gluesql/blob/3410f789b50e825b01e2e154dcf97608af47d9d6/core/src/data/value/literal.rs#L120-L124)

core/src/ast/operator.rs
panarch: > The implementation of Display was to implement Debug

`Debug` is already implemented for `BinaryOperator`, then we may not need `Display`?

core/src/ast/data_type.rs
LucaCappelletti94: Removed the `Null` data type and introduced new Error messages that do not require `datatype()`

core/src/ast/operator.rs
LucaCappelletti94: Removed.

Issue #1219, opened 2023-05-25
Parameter binding support?


CryptoManiac:
Hi. Is there any plan to implement this feature? It's 2023 and nowadays having to deal with SQL injections is pretty much blocking issue for actual deployment. :(

panarch:
Sure, that's not hard to support it but currently there's no one working on it yet. I cannot ensure the shipping date but I think you can expect to see this in a few months.
For now, if you concern SQL injection issue, then we would recommend to use AST Builder for building queries.

geofmureithi:
@panarch any update on this? I could pick it up if need be.

panarch:
> @panarch any update on this? I could pick it up if need be.

sure 👍

geofmureithi:
Awesome!
My current approach would be use a parser, and then iterate through the tokens to generate the final output.
There is [sqlformat crate](https://github.com/shssoichiro/sqlformat-rs) which I found has similar functionality (plus formatting).
Let me know your feelings about the crate, and whether I can use it or port the specific parts that matter to us.

geofmureithi:
After a little more research, this crates uses https://github.com/sqlparser-rs/sqlparser-rs so I will work to integrate this feature at that point where I will look for [Value::Placeholder](https://docs.rs/sqlparser/0.39.0/sqlparser/ast/enum.Value.html#variant.Placeholder)

lostb1t:
what's the current suggested way todo this?

panarch:
> what's the current suggested way todo this?

currently there is no parameter binding support for SQL.
if you accept the param values from client-side, safe way is that you need to use AST builder.

smallfz:
Parameter binding is a great feature. Is there any updates? 
I dig a little into and there is a place looking good to adopt the Placeholder.

```patch
--- a/core/src/executor/evaluate/expr.rs
+++ b/core/src/executor/evaluate/expr.rs
@@ -9,7 +9,15 @@ use {
 };
 
 pub fn literal(ast_literal: &AstLiteral) -> Result<Evaluated<'_>> {
-    Literal::try_from(ast_literal).map(Evaluated::Literal)
+    match ast_literal {
+        AstLiteral::Placeholder(v) => {
+            // Ok(Evaluated::Value(Value::I64(3344)))
+            todo!()
+        }
+        _ => {
+            Literal::try_from(ast_literal).map(Evaluated::Literal)
+        }
+    }
 }
```

[...]

PR #1575, opened 2024-09-20
A proposal for parameters binding/resolving

smallfz:
I'm considering to create a DBMS(not in this pr nor this repo) based on gluesql. It serves on a tcp network and allows clients connect to and run sqls. The sql transmit through a network connection along side some possible parameters. I prefer not to use a single merged sql since that is considered not safe and a lot of works must carry out to avoid sql injection. So I need the feature of parameters resolving. #1219 

This is a skeleton of parameter mod that provides a function `resolve_parameters` to resolve placeholders in a statement.

```rust
#[cfg(test)]
mod tests {

    use gluesql::prelude::{Parameters, resolve_parameters};
    use gluesql::prelude::{execute, parse, plan, translate, Value};
    use gluesql_memory_storage::MemoryStorage;
    
    struct ParameterSet {
        values: Vec<Value>
    }

    impl Parameters for ParameterSet {
        fn get(&self, i: usize) -> Option<Vec<u8>> {
            if i>= 1 && i <= self.values.len() {
                Some((&self.values[i-1]).into())
            } else {
                None
            }
        }
    }

    #[tokio::test]
    async fn test_params_resolving() {
        let sql = "create table test(x int, y int);
insert into test (x, y) values($1, $2);
select $1 px, t.* from test t;";
        let rs = parse(sql).unwrap();
        let mut store = MemoryStorage::default();
        for stmt_ast in rs {
            let mut stmt = translate(&stmt_ast).unwrap();
            
            let ps = ParameterSet{
                values: vec![Value::I16(7), Value::I16(129)],
            };
            resolve_parameters(&ps, &mut stmt).unwrap();  // <---------- here it the function.

            let stmt = plan(&mut store, stmt).await.unwrap();

            let result = execute(&mut store, &stmt).await.unwrap();
            println!("result: {:?}", &result);
        }
    }

}
```


panarch:
GlueSQL's approach to AST (Abstract Syntax Tree) structure differs somewhat from other databases. To summarize:

Typically, databases have an AST for the parsing result, and after planning, they create another AST for execution. However, GlueSQL uses an external crate called `sqlparser` for parsing. Since `sqlparser` supports various SQL dialects beyond GlueSQL, we use the `translate` function to extract only the features used by GlueSQL.

In the `parse` step, SQL is converted into `sqlparser`'s AST using `sqlparser-rs`. Then, the `translate` function converts that AST into GlueSQL's AST. In this process, query planning in GlueSQL is optional. Once the AST is generated, it already has a basic execution plan, and the planner is used to devise better strategies. Therefore, there's effectively only one set of AST.

I noticed that you placed the parameter resolution here:

```rust
resolve_parameters(&ps, &mut stmt).unwrap();  // <---------- here is the function.
```

between `translate` and `plan`. After `translate`, the GlueSQL AST is produced, and modifying it would require changes throughout the planner and execution components.

Considering the features that query parameters can provide:

1. **Preventing SQL injection**
2. **Query plan caching**

The primary goal is preventing SQL injection; query plan caching is secondary and not yet supported in GlueSQL.

To satisfy the first requirement, I'd like to propose a simpler method. I'm not entirely certain this is the best approach, so I'd appreciate your feedback.

Instead of placing the parameter resolution between `translate` and `plan`, we can move it between `parse` and `translate`. `sqlparser` already supports parsing query parameters effectively. Instead of passing the parameters directly to the GlueSQL AST, we can process them by parsing only the literals of the query parameters in the `sqlparser` AST and modifying the AST accordingly.

Since `sqlparser` has implemented the visitor pattern, using `VisitorMut`, we can easily apply the query parameters. This way, from `translate` onwards, the code remains exactly the same.

In terms of performance, handling this within `translate` might be slightly better, but it would require passing a query parameter hashmap to all functions in `translate`, which would be a significant amount of work. By traversing the `sqlparser` AST with `VisitorMut` and replacing the placeholders, the process becomes much simpler.

Please let me know what you think of this approach!

smallfz:
Thanks for reviewing and the patient response! This give me a clearer picture about this awesome project. I was thinking to introduce a proposal that has minimum impact to the codebase but now I get that the current one is insufficient. 

> Instead of placing the parameter resolution between `translate` and `plan`, we can move it between `parse` and `translate`. `sqlparser` already supports parsing query parameters effectively. Instead of passing the parameters directly to the GlueSQL AST, we can process them by parsing only the literals of the query parameters in the `sqlparser` AST and modifying the AST accordingly.

I think this is clearly a better idea. It's neat enough. Parameters can fit directly into `sqlparser::ast::Value`. And less files need to be modified.

> Since `sqlparser` has implemented the visitor pattern, using `VisitorMut`, we can easily apply the query parameters. This way, from `translate` onwards, the code remains exactly the same.
> 
> In terms of performance, handling this within `translate` might be slightly better, but it would require passing a query parameter hashmap to all functions in `translate`, which would be a significant amount of work. By traversing the `sqlparser` AST with `VisitorMut` and replacing the placeholders, the process becomes much simpler.

I'll get into this `VisitorMut` trait. I learnt it providing functions like `visit_statements_mut`. Maybe this is the start point?

smallfz:
> Have you looked into this crate? https://github.com/sunng87/pgwire. 

Another fantastic crate! It looks powerful and complete. And postgresql is my favorite RDBMS.

My original thought is a simpler protocol built on top of websocket or http/2. That makes a browser web page can communicate with server directly without a backend involved. And also the server can be proxied by popular software and services, eg. nginx, cloudflare. There are already a lot of authentication methods to adopt, saving some time to implement. 
I think I'd better do some more research about this.

Thanks for the heads up and help! Really appreciate!

panarch:
> I'll get into this `VisitorMut` trait. I learnt it providing functions like `visit_statements_mut`. Maybe this is the start point?

That's right. I think `visit_expressions_mut` is probably the appropriate method to use.

ref. https://github.com/sqlparser-rs/sqlparser-rs/blob/fb42425d51162cd57e5a0cbd49eafc5743d60daf/src/ast/visitor.rs#L545-L552

In `sqlparser-rs`, placeholders are stored in `Value`, and that `Value` is included as a variant of `Expr`.

PR #1577, opened 2024-09-23
Introduced the Nullable enum and propagated Null evaluation

LucaCappelletti94:
As per title, this PR introduces the propagation of Null evaluations up to the root. In order to make the Null case semantically different from the None object, a new enum `Nullable` has been introduced.

### DIFF
--- a/core/src/data/nullable.rs
+++ b/core/src/data/nullable.rs
@@ -67,15 +67,3 @@ impl<T: Not<Output = T>> Not for Nullable<T> {
         }
     }
 }
-
-impl<T: BitOr<Output = T>> BitOr for Nullable<T> {
-    type Output = Self;
-
-    /// Bitwise ORs two nullable values, returning NULL if either value is NULL.
-    fn bitor(self, rhs: Self) -> Self::Output {
-        match (self, rhs) {
-            (Nullable::Null, _) | (_, Nullable::Null) => Nullable::Null,
-            (Nullable::Entry(lhs), Nullable::Entry(rhs)) => Nullable::Entry(lhs | rhs),
-        }
-    }
-}
--- a/core/src/data/nullable.rs
+++ b/core/src/data/nullable.rs
@@ -1,6 +1,9 @@
 //! Submodule providing the `Nullable` enum and associated traits.
 
 use core::ops::{Add, Div, Mul, Not, Sub};
+use std::ops::{
+    AddAssign, BitAnd, BitAndAssign, BitOr, BitOrAssign, DivAssign, MulAssign, SubAssign,
+};
 
 #[derive(Debug, Clone, Copy, Hash)]
 /// Enum representing a nullable value.
@@ -8,18 +11,18 @@ use core::ops::{Add, Div, Mul, Not, Sub};
 /// # Implementations details
 /// This enum differs from the `Option` enum in that it does not represent
 /// a `Option::None` value, but rather a `Nullable::Null` SQL `NULL` value.
-/// 
+///
 /// # Examples
-/// 
+///
 /// ```
 /// use gluesql_core::data::Nullable;
-/// 
+///
 /// let null: Nullable<i32> = Nullable::Null;
-/// 
+///
 /// assert!(null.is_null());
 /// assert!(!null.is_entry());
 /// ```
-/// 
+///
 pub enum Nullable<T> {
     /// Represents an SQL `NULL` value.
     Null,
@@ -29,19 +32,19 @@ pub enum Nullable<T> {
 
 impl<T> Nullable<T> {
     /// Returns `true` if the value is `Nullable::Null`.
-    /// 
+    ///
     /// # Examples
-    /// 
+    ///
     /// ```
     /// use gluesql_core::data::Nullable;
-    /// 
+    ///
     /// let null: Nullable<i32> = Nullable::Null;
-    /// 
+    ///
     /// assert!(null.is_null());
     /// assert!(!null.is_entry());
-    /// 
+    ///
     /// let entry: Nullable<i32> = 10.into();
-    /// 
+    ///
     /// assert!(!entry.is_null());
     /// assert!(entry.is_entry());
     /// ```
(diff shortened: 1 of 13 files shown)


panarch:
coverage report is supposed to be added as a comment but hmm.. it didn't work.
https://coveralls.io/builds/69942739
we can check the report in the above link.

for the first, we don't need to have unused methods.
https://coveralls.io/builds/69942739/source?filename=core%2Fsrc%2Fdata%2Fnullable.rs#L72
we can add the unused methods when those methods are required.

LucaCappelletti94:
I have added those methods to prevent a future wrong implementation, primarily in the case of the OPS. All of these methods are tested in the documentation, which does not seem included in the coverage. You may want to include the doctests in the coverage, as otherwise you get false negatives like in this case.

I agree that I can reasonably remove methods such as `as_entry`, although I believe it reasonable that these methods will be used in the near future as `Nullable` gets more integrated with other parts of the crate.

If I were not to include operations such as 'Add' for these objects, one may implement it in the future using some definition of NULL that is not correct.

core/src/executor/evaluate/expr.rs
```diff
@@ -23,6 +23,10 @@ pub fn binary_op<'a>(
     l: Evaluated<'a>,
     r: Evaluated<'a>,
 ) -> Result<Evaluated<'a>> {
+    if l.is_null() || r.is_null() {
+        return Ok(Evaluated::null());
+    }
```
panarch, line 28: it would be redundant to the below logic.
`evaluate_eq` is supposed to return `NULL` if the either left or right is `NULL` so we may not need this.

core/src/executor/evaluate/expr.rs
```diff
@@ -68,6 +72,10 @@ pub fn binary_op<'a>(
 }
 
 pub fn unary_op<'a>(op: &UnaryOperator, v: Evaluated<'a>) -> Result<Evaluated<'a>> {
+    if v.is_null() {
+        return Ok(Evaluated::null());
+    }
```
panarch, line 77: this is also the role of `Evaluated::{unary_plus, unary_minus}` to take. like the binary operators, this would be also redundant.

core/src/executor/evaluate/expr.rs
LucaCappelletti94: The same would need to be added to all the operations below, not only evaluate_eq. This way we avoid changing all of those operations. If you prefer, I can change all of the below operations to return `Nullable`, but I thought you preferred minimal changes.

panarch: this pr is targeting only `=` evaluate_eq operations. if something other also needs to be fixed, then we can do it in the following pr. let's focus only for the `evaluate_eq` for now.

LucaCappelletti94: This PR changes that operation by propagating the NULL. At this time, the implementation of the other operations, like all `binary_op` stops the propagation of the NULL. Either those operations also propagate the NULL, or it stops there. In order to propagate the NULL while making minimal changes, I have added that line.

panarch: ok, I understand the goal what we do for this and following PRs are for making the NULL propagation complete.
however, for us to do efficient review, the big single task needs to be splitted into small steps. that is why we focus only on `evaluate_eq` in this PR.
if we also want to handle other `binary_op` cases, this PR needs to be quite bigger than now.
could we limit the scope of this PR to make `NULL = {value}` or `{value} = NULL` return `NULL` which was previously `False`?

LucaCappelletti94: I will try to change that, though I suspect tests may fail. Hopefully, I am wrong on this and it is as easy as that.

LucaCappelletti94: I was mistaken - I suppose we do not have tests at this time that expect propagation across AND, OR and other such operations.  See commit [68d9f87](https://github.com/gluesql/gluesql/pull/1577/commits/68d9f872e73b1660e75a024e68eeb79f470ff546)

core/src/executor/evaluate/expr.rs
LucaCappelletti94: Removed in commit [68d9f87](https://github.com/gluesql/gluesql/pull/1577/commits/68d9f872e73b1660e75a024e68eeb79f470ff546)

core/src/data/nullable.rs
```diff
@@ -0,0 +1,109 @@
...
+    ///
+    /// assert!(null.is_null());
+    ///
+    /// let entry: Nullable<i32> = 10.into();
+    ///
+    /// assert!(!entry.is_null());
+    /// ```
+    pub fn is_null(&self) -> bool {
+        matches!(self, Nullable::Null)
+    }
```
panarch, line 52: good. we now use `Nullable::is_null` only in unit tests. could we also remove this?

LucaCappelletti94: You mean, adding a `#[cfg(test)]` on top of the method?

panarch: oh, I saw some remaining is_null cases. I will close this conversation and look more on it.

currently I am focusing on..
image

to handle negate evaluation, I think that we can find more simple way to handle this but not yet I catched the point.

core/src/executor/evaluate/mod.rs
```diff
@@ -48,6 +49,72 @@ pub async fn evaluate_stateless<'a, 'b: 'a>(
     evaluate_inner(storage, context, None, expr).await
 }
 
+async fn stream_contains_target<'life, 'stream, 'item>(
+    target: Evaluated<'life>,
+    negated: bool,
+    stream: impl Stream<Item = Result<Evaluated<'item>>> + 'stream,
+) -> Result<Nullable<bool>> {
```
panarch, line 56: I believe we can handle this part more simply. The key issue is the negate evaluation, but I haven't fully figured out how to approach it yet. I'll look into it further and leave a comment once I have a better idea.

core/src/data/nullable.rs
LucaCappelletti94: This is a different is_null method, not the Nullable one. It may be reasonable to use the `Nullable` enum within these other structs in the future so as to uniform the handling of NULLs, but it is not the same method at this time.

LucaCappelletti94: As of commit [a5e3c9b](https://github.com/gluesql/gluesql/pull/1577/commits/a5e3c9b68a06c4cb7c711b22682f5a60399cae48) `is_null` of the `Nullable` enum is a test-only method.

core/src/executor/evaluate/mod.rs
LucaCappelletti94: I have unified the branches of the method in commit [b0eaeff](https://github.com/gluesql/gluesql/pull/1577/commits/b0eaeffe300009f8e604e9560d2b74be6752e7b8). While this makes it less code, I am worried it severely reduces its readability. Let me know which one you prefer.

LucaCappelletti94:
Any updates?

panarch:
> Any updates?

I’ve been incredibly busy with other work lately, so it’s taken me much longer than expected. I’ll be back soon

PR #1582, opened 2024-11-09
fix: unable to query UUID primary key by literal value in evaluated

### LINKED ISSUE #1579: Cannot query UUID that it is primary key
I found this problem when I was writing test cases for `sea-orm` based on `gluesql`.

Steps to reproduce:

- Create some table that use UUID as the primary key's type.
- Insert an item.
- Try to select this item by primary key.

[Demo on gist](https://gist.github.com/langyo/5e2ac6fe0fde92650dce0cb3c4e5203a)

For example, this demo outputs:
```
DEBUG query first uuid: 0192b34f-f839-77d2-a8b2-68319c483254
DEBUG SQL: SELECT * from posts where id = 2090940224296650037553036752300028500
DEBUG sub query: [Select { labels: ["id", "title", "text", "id_2"], rows: [] }]
DEBUG SQL: SELECT * from posts where id = '0192b34f-f839-77d2-a8b2-68319c483254'
DEBUG sub query: [Select { labels: ["id", "title", "text", "id_2"], rows: [] }]
DEBUG SQL: SELECT * from posts where text = 'いいよ、来いよ'
DEBUG sub query: [Select { labels: ["id", "title", "text", "id_2"], rows: [[Uuid(2090940224296650037553036752300028500), Str("Homo"), Str("い いよ、来いよ"), Uuid(2090940224296650037553036752300028500)]] }]
DEBUG SQL: SELECT * from posts where id_2 = '0192b34f-f839-77d2-a8b2-68319c483254'
DEBUG sub query: [Select { labels: ["id", "title", "text", "id_2"], rows: [[Uuid(2090940224296650037553036752300028500), Str("Homo"), Str("い いよ、来いよ"), Uuid(2090940224296650037553036752300028500)]] }]
```

> I can solve this by setting the primary key type to a string or something, but this is obviously not a long-term solution.

cc @panarch

ever0de:
fixed #1579

This commit addresses the issue where select queries with a primary key of type UUID fail when using literal values in the SQL AST. The solution involves modifying the SQL AST evaluation to use `try_from_literal` instead of `Value::try_from` when the evaluated type is `literal`.

Steps to fix:
- Check if the evaluated type is `literal` in SQL AST.
- Use `try_from_literal` with `data_type` parsing to handle UUID values correctly.

- [ ] Currently, the implementation calls `fetch_schema` multiple times, leading to inefficiencies. This should be improved through refactoring.

### DIFF
--- a/core/src/data/value/literal.rs
+++ b/core/src/data/value/literal.rs
@@ -229,6 +229,7 @@ impl Value {
         }
     }
 
+    // sql like "CAST"
     pub fn try_cast_from_literal(data_type: &DataType, literal: &Literal<'_>) -> Result<Value> {
         match (data_type, literal) {
             (DataType::Boolean, Literal::Boolean(v)) => Ok(Value::Bool(*v)),
--- a/core/src/data/value/literal.rs
+++ b/core/src/data/value/literal.rs
@@ -229,7 +229,6 @@ impl Value {
         }
     }
 
-    // sql like "CAST"
     pub fn try_cast_from_literal(data_type: &DataType, literal: &Literal<'_>) -> Result<Value> {
         match (data_type, literal) {
             (DataType::Boolean, Literal::Boolean(v)) => Ok(Value::Bool(*v)),
--- a/core/src/executor/fetch.rs
+++ b/core/src/executor/fetch.rs
@@ -7,7 +7,10 @@ use {
             Values,
         },
         data::{get_alias, get_index, Key, Row, Value},
-        executor::{evaluate::evaluate, select::select},
+        executor::{
+            evaluate::{evaluate, Evaluated},
+            select::select,
+        },
         result::Result,
         store::{DataRow, GStore},
     },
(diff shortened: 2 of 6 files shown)


ever0de:
Key being searched: `Uuid(2092860986323542640217787511340351975)`
Key loaded from JSON storage: `Str("01931203-650f-7111-b1aa-15b4f5b541e7")`
As a result, the check at https://github.com/ever0de/gluesql/blob/99ddc88a8394d740e34fd5d25750deb9e80fe53f/storages/json-storage/src/store.rs#L54-L57 cannot pass.

ever0de:
There are currently missing test cases. Additional tests need to be conducted based on this PR.
Although saving is done by converting to a string, a test case is needed to verify whether the conversion follows the correct schema during the actual scan.

panarch (approved):
Looks all nice!

I will work on the following issue in https://github.com/gluesql/gluesql/pull/1599

PR #1585, opened 2024-11-29
Add redb storage

panarch:
# Key Changes

## New Redb Storage Backend
- **Added** `storages/redb-storage` with a full implementation of **`RedbStorage`**, including transaction management and error handling.  
  *Source: `storages/redb-storage/src/lib.rs`, `storages/redb-storage/src/core.rs`*
- **Introduced** `StorageError` variants tailored for Redb operations.  
  *Source: `storages/redb-storage/src/error.rs`*

## Workspace Updates
- **Registered** the new crate in **`Cargo.toml`**:  
```toml
gluesql-redb-storage = { path = "./storages/redb-storage", version = "0.16.3" }
```
- **Refreshed** the lockfile to include Redb dependencies.

## CI and Test Integration
- **Updated** GitHub Actions workflow to run tests for **`redb-storage`** alongside existing backends.  
  *Source: `.github/workflows/rust.yml`*
- **Added** tests to verify Redb behavior, including nested transaction handling and database-interface errors.  
  *Sources:  
  `storages/redb-storage/tests/nested_transaction.rs`,  
  `storages/redb-storage/tests/redb_interface_error.rs`*

## Impact
- GlueSQL can now operate with a **Redb backend**, using the same API as other storage modules.
- Continuous integration covers Redb storage to maintain stability across all supported backends.


panarch:
the lastest coverage: https://coveralls.io/builds/74046577
343 of 345 new or added lines in 8 files covered. (99.42%)

storages/redb-storage/src/core.rs
```diff
@@ -0,0 +1,247 @@
...
+const SCHEMA_TABLE_NAME: &str = "__SCHEMA__";
+const SCHEMA_TABLE: TableDefinition<&str, Vec<u8>> = TableDefinition::new(SCHEMA_TABLE_NAME);
+
+type Result<T> = std::result::Result<T, StorageError>;
+
+pub struct StorageCore {
+    db: Database,
+    txn: Option<redb::WriteTransaction>,
+    autocommit: bool,
+}
```
devgony, line 31: Does it need `autocommit` if `txn` is `None`?
If not, could we try it like this?
 
```rs
    pub enum TransactionState {
        None,
        Active {
            txn: redb::WriteTransaction,
            autocommit: bool,
        },
    }

    pub struct StorageCore {
        db: Database,
        state: TransactionState,
    }
```

panarch: wow, thanks a lot! this is certainly better to bind `txn` and `autocommit` to a single enum variant. There is no use case of `autocommit` without `txn`.

storages/redb-storage/src/core.rs
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    pub async fn append_data(&mut self, table_name: &str, rows: Vec<DataRow>) -> Result<()> {
+        let table_def = self.data_table_def(table_name)?;
+        let txn = self.txn_mut()?;
+        let mut table = txn.open_table(table_def)?;
+
+        for row in rows {
+            let key = Key::Uuid(Uuid::now_v7().as_u128());
+            let value = serialize(&(key.clone(), row))?;
```
devgony: We can remove clone
```rs
            let value = serialize(&(&key, row))?;
```

storages/redb-storage/src/core.rs
```diff
@@ -0,0 +1,274 @@
...
+        Ok(())
+    }
+
+    pub async fn insert_data(&mut self, table_name: &str, rows: Vec<(Key, DataRow)>) -> Result<()> {
+        let table_def = self.data_table_def(table_name)?;
+        let txn = self.txn_mut()?;
+        let mut table = txn.open_table(table_def)?;
+
+        for (key, row) in rows {
+            let value = serialize(&(key.clone(), row))?;
```
devgony: And here as well
```rs
            let value = serialize(&(&key, row))?;
```

storages/redb-storage/src/core.rs
panarch: oh! I totally missed this, thanks a lot!

devgony (approved):
👍👍👍

PR #1591, opened 2024-12-07
Improve CommandExt to return both `stderr` and `stdout`

devgony:
## Issue
- Currently CommandExt::execute returns only `stderr` but there are cases to return `stdout` even though `status is not success`

## Todo
- [x] merge `stdout` and `stderr`
- [x] add unit test cases


panarch (approved):
👍 👍 👍

Issue #1265, opened 2023-07-09
`Send`, `Sync` supertrait support for `GStore`, `GStoreMut`


ever0de:
Issues that need to be resolved to support this

1. `im_rc::HashMap` does not support `Send`, `Sync`
    - https://docs.rs/im-rc/latest/im_rc/struct.HashMap.html#impl-Send
2. Resolve `Rc` or Switching between `Rc` and `Arc` via features
3. `async_recursion` make the `?Send` option switchable via features
4. Adding with `Send + Sync` supertraits based on the `storage` parameter feature
5. Make it possible to change the `?Send` option of `GStore` supertraits in `store::*` to feature
6. ... 

There may be more features that need to be added or changed, but the above information is probably the cornerstone for supporting `Send` and `Sync`.
- https://github.com/ever0de/gluesql/commit/eb819de24d028d14276fe657306f3ec3fe0a75a7

TonyGiorgio:
Is anyone working on this atm or is gluesql not intended to actually run on servers?

enginefuture:
gluesql with poem-openapi has same issue
```rust
#[OpenApi]
impl Api {

    #[oai(path = "/getAsyncRoutes", method = "get")]
    async fn getAsyncRoutes(&self, authorization: MySecurityScheme) -> ApiLoginResponse<String> {
        // println!("username:::{:?}", username);
        let user_old = user::builder()
            .mobile(Some("sss".to_string()))
            .build_selectable();
        let sql = user_old.select();
        // let output = rt.block_on(async { execute_sql(sql).await.unwrap() });
        let output = execute_sql(sql).await.unwrap();
        ApiLoginResponse::Ok(Json(LoginResponse::ok("sucess".to_string())))
    }
}```
```shell
dyn std::future::Future<Output = SledRunner> cannot be sent between threads safely
the trait Send is not implemented for dyn std::future::Future<Output = SledRunner>
required for std::ptr::Unique<dyn std::future::Future<Output = SledRunner>> to implement Send
```

edezhic:
It looks like the linked commit has already switched a lot of Rc->Arc and im_rc crate can be replaced with im

Also, Sync is probably a lot less important since Send is enough for most cases, so maybe the scope of the initial work can be reduced

Don't know how to fit web storage into the picture thou. It seems to be the main issue and it can't support Send because things will go wrong if someone will attempt to send an instance into a web worker

ever0de:
[...]

PR #1593, opened 2024-12-14
impl Send futures

edezhic:
Resolution for https://github.com/gluesql/gluesql/issues/1265, based on https://github.com/ever0de/gluesql/tree/feat/send-feature

This is based on @ever0de 's work (kudos!), rebased on actual `main` and mostly proceeds in the same direction. `send` feature for the `gluesql-core` crate is complete but the original issue misses an important problem - how this feature interacts with other crates given the additive nature of rust/cargo features. For example - `SharedMemoryStorage` can work with send but it's dependency `MemoryStorage` can't. Similar problems with the composite storage and in other gluesql crates. This PR includes messed up `Cargo.toml` files and I'm not sure how to fix them - cargo doesn't support having both `send` and non-`send` versions of `gluesql-core` in the dependency graph so current project structure breaks completely. Maybe I'm missing some solution due to lack of experience working with complex workspaces, but maybe solution will require some restructuring decisions, idk. Need help from the maintainers here.

Also, in review you might notice a few weird things so I want to highlight them and provide some context:

- `core/src/executor/evaluate/mod.rs` (`// not using stream try_filter due to cryptic lifetimes issues`) - slight rewrite from a `try_filter` to a `while` loop with no change in logic afaik, still very much readable so I think should be fine

- `core/src/executor/aggregate/state.rs` (`// using iter_chunks because itertools::Chunk is !Send, and the while loop because iter_chunks:Chunks is not an Iterator`) - comment pretty much says it all, performance etc should be about the same

- `core/src/executor/aggregate/mod.rs` (`// not using iters and chaining due to cryptic lifetimes issues`) - updated code adds just a few allocations and pretty straightforward so I think it's fine

- `// these two same fns can be replaced with a impl type alias for the return type once its stabilized (https://rust-lang.github.io/impl-trait-initiative/explainer/tait.html)` in a few places, I've tried replacing with generics but in return positions they just create new problems. Since TAIT are already available in nightly I think it makes sense to await their stabilisation and just replace return types with `cfg`'d aliased types

- `test-suite/src/tester/mod.rs` (`// cfg(feature) shouldn't be inside macro because it will be expanded together with these directives`) - just a bit ugly and these duplicates won't be solved with TAIT, but can be minimised by splitting `cfg`'d part and the rest in different macros and combining them in another

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,22 +1,24 @@
 [workspace]
 resolver = "2"
 members = [
-  "cli",
+  #"cli",
   "core",
-  "pkg/rust",
-  "pkg/javascript",
-  "pkg/python",
-  "storages/*",
+  #"pkg/rust",
+  #"pkg/javascript",
+  #"pkg/python",
+  #"storages/*",
+  "storages/memory-storage",
+  #"storages/shared-memory-storage",
   "test-suite",
   "utils",
 ]
 default-members = [
-  "cli",
+  #"cli",
   "core",
-  "pkg/rust",
-  "pkg/javascript",
-  "pkg/python",
-  "storages/*",
+  #"pkg/rust",
+  #"pkg/javascript",
+  #"pkg/python",
+  #"storages/*",
   "test-suite",
   "utils",
 ]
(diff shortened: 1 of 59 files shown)


panarch:
resolved in #1688

PR #1599, opened 2025-03-19
Add data_type info to ast::IndexItem::PrimaryKey

panarch:
- [ ] Currently, the implementation calls fetch_schema multiple times, leading to inefficiencies. This should be improved through refactoring.

ref. https://github.com/gluesql/gluesql/pull/1582


panarch:
dup. https://github.com/gluesql/gluesql/pull/1624

PR #1674, opened 2025-07-19
Correct comparison between `interval` data type

moreal:
This pull request fixes the comparison between `interval` data type where `Interval::Month(...) > Interval::Microsecond(...)` was always treated as the month variable is larger than microsecond variable even if the month's inner value is zero.

I'm not sure about the microsecond value for 1 month, I supposed that it will be same as the `30 days`. Please let me know if you know or found the spec related with it 🙏🏻.

```rust
static MONTH_TO_MICROSECOND_MULTIPLIER: LazyLock<BigDecimal> =
    LazyLock::new(|| BigDecimal::from(30 * 24 * 60 * 60 * 1_000_000i64));
```


panarch:
Thanks for the PR!

I’d like to revisit whether `Interval` really needs to implement **`Ord`**.  
A meaningful total order across "year / month" and "day / second" fields is tricky, and the current *"1 month = 30 days"* assumption is essentially a workaround.

The only place that forces `Ord` right now is `core::data::key::Key`, where we define an internal ordering for index keys.  Maybe we can:

1. Keep that *storage‑layer* ordering inside `Key`; **but**
2. Expose only **`PartialOrd`** for `Value::Interval` (SQL layer), so application code isn’t tied to an arbitrary 30‑day rule.

Current ad‑hoc ordering for keys can be seen here:  
https://github.com/gluesql/gluesql/blob/bcd5c4754e5a5b296a3843f07e7f6e766ba571bc/core/src/data/key.rs#L56-L94

In short, instead of encoding a debatable calendar rule in `Ord`, could we drop `Ord` from `Interval` entirely (leaving `PartialOrd` only) and let `Key` keep its private total order?

--

> In short, instead of encoding a debatable calendar rule in `Ord`, could we drop `Ord` from `Interval` entirely (leaving `PartialOrd` only) and let `Key` keep its private total order?

Wow, I tried talking with ChatGPT o3 first about this PR and it additionally suggested this idea by itself. I like this one. that can make us start with small stuff.


Curious to hear your thoughts!

moreal:
If comparison between Interval data is undefined behavior, then supporting only partial comparison with `PartialOrd` seems like a good approach!

However, this raises a question: **is it correct behavior to use `Interval` as an index?** I'm wondering about this because something like BTree would require comparison operations[^1].

For now, I'll try implementing only `PartialOrd`.

[^1]: https://www.postgresql.org/docs/current/indexes-types.html#INDEXES-TYPES-BTREE

----

This is somewhat of a different topic, but while looking around the GlueSQL project, I was curious whether there are any correct reference implementations that serve as benchmarks or points of reference when implementing features. For instance, in cases like defining the comparison logic between Interval types as we're doing now, I wonder what the right way to implement it would be.  
When I personally look into it, there's the ANSI SQL standard as a specification, but it's quite expensive to purchase and read 😅 When browsing through the GlueSQL repository issues, I noticed something called [sqllogictest], and I'm hoping this might help test implementations, but if you know of anything else, sharing it would be of great help 🙏

[sqllogictest]: https://sqlite.org/sqllogictest/doc/trunk/about.wiki

moreal:
I made some modifications and pushed a commit related to this. While running queries in the GlueSQL CLI environment, I feel that comparison operations should be well-supported. For now, since it's undefined behavior, I've arbitrarily removed the malfunctioning code, but eventually I think we'll need to find some specification and implement it accordingly. The reference implementation or specification I mentioned above seems important. 😞 

```
gluesql> SELECT INTERVAL '0' month < INTERVAL '100' second;
| INTERVAL '0' MONTH < INTERVAL '100' SECOND |
|--------------------------------------------|
| FALSE                                      |
```

core/src/data/key.rs
```diff
@@ -75,7 +75,13 @@ impl Ord for Key {
             (Key::Date(l), Key::Date(r)) => l.cmp(r),
             (Key::Timestamp(l), Key::Timestamp(r)) => l.cmp(r),
             (Key::Time(l), Key::Time(r)) => l.cmp(r),
-            (Key::Interval(l), Key::Interval(r)) => l.cmp(r),
+            (Key::Interval(l), Key::Interval(r)) => {
+                l.partial_cmp(r).unwrap_or_else(|| match (l, r) {
+                    (Interval::Month(_), Interval::Microsecond(_)) => Ordering::Greater,
+                    (Interval::Microsecond(_), Interval::Month(_)) => Ordering::Less,
+                    _ => unreachable!("Covered by PartialOrd<Interval>::partial_cmp()."),
```
panarch: this is actually reachable.
Like other enum variants in `Key`, we need to set custom ordering between `Month` and `Microsecond`.
We can keep the same ordering that was in `Interval` module for now.
```rust
            (Interval::Month(_), Interval::Microsecond(_)) => Ordering::Greater,
            (Interval::Microsecond(_), Interval::Month(_)) => Ordering::Less,
```

moreal: I may misunderstand. `PartialOrd<Interval>::cmp` covers the following cases:

```rust
            (Interval::Month(l), Interval::Month(r)) => Some(l.cmp(r)),
            (Interval::Microsecond(l), Interval::Microsecond(r)) => Some(l.cmp(r)),
```

And the lines you referenced, covers the following cases:

```rust
            (Interval::Month(_), Interval::Microsecond(_)) => Ordering::Greater,
            (Interval::Microsecond(_), Interval::Month(_)) => Ordering::Less,
```

I couldn't find any case it doesn't catch. Did you mean it will be reachable if another variant (e.g., `Interval::Year`) is introduced?

panarch: Ah, now I see! that was my mistake.
I saw that the below line was reachable but ah yes. it is unreachable.
```rust
    _ => unreachable!("Covered by PartialOrd<Interval>::partial_cmp()."),
```

Then, could we do small rearrangement to the codes not to even have the above line?
It is a library project so we don't allow any kind of codes which can cause panic.
You can think that it is kind of.. project internal code convention.
When it is really unavoidable, then we add `Unreachable-` prefixed error variant and use it.

Here it would be too much to have error and it is even not possible in implementing `Ord`.

Returning `Ordering::Less` would be enough except for `(Interval::Month(_), Interval::Microsecond(_)) => Ordering::Greater`?

moreal: It is very reasonable that it must not cause any panics because GlueSQL is library!

As you suggested, I used `Ordering::Less` as default value!

```rust
(Key::Interval(l), Key::Interval(r)) => l.partial_cmp(r).unwrap_or(match (l, r) {
    (Interval::Month(_), Interval::Microsecond(_)) => Ordering::Greater,
    _ => Ordering::Less,
}),
```

moreal:
I pushed one more commit to write documentation on the `Interval` type to provide context on why it is made of `Month` and `Microsecond`.

panarch (approved):

PR #1675, opened 2025-07-19
Simplify conditional logic in `plan_query`

moreal:
This pull request include only cosmetic changes. In `plan_query` function at `core/src/plan/index.rs`, it used `match` keyword to detect the `index` variable is `Option::Some` and used the `index` variable, not inner value again.

So I though it can be simplified by using `if-else` syntax with `index.is_some()` condition and this pull request includes such changes.


ever0de (approved):

panarch (approved):
It's good to see having less indentation.
Thanks a lot 👍 👍

PR #1677, opened 2025-07-21
Add CURRENT_DATE, CURRENT_TIME, and CURRENT_TIMESTAMP functions

junghoon-vans:
## Summary
Implements three SQL standard date/time functions with parentheses support to improve SQL compatibility:
- `CURRENT_DATE` / `CURRENT_DATE()` - Returns current date (DATE type)
- `CURRENT_TIME` / `CURRENT_TIME()` - Returns current time (TIME type)
- `CURRENT_TIMESTAMP` / `CURRENT_TIMESTAMP()` - Returns current timestamp (TIMESTAMP type, equivalent to NOW())

## Implementation Details
- **AST Extension**: Added `CurrentDate()`, `CurrentTime()`, `CurrentTimestamp()` variants to `Function` enum
- **Parser Support**: Added function name recognition in `translate_function` with optional parentheses
- **AST Serialization**: Added `to_sql()` support for proper SQL generation with parentheses
- **Evaluation Logic**: Implemented UTC-based datetime extraction using chrono
- **Pattern Matching**: Updated expression planning to handle new functions

## Usage Examples
```sql
-- Basic usage (both syntaxes supported)
SELECT CURRENT_DATE, CURRENT_TIME(), CURRENT_TIMESTAMP;

-- In table creation with defaults
CREATE TABLE events (
    id INT,
    created_date DATE DEFAULT CURRENT_DATE,
    created_time TIME DEFAULT CURRENT_TIME(),
    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
);

-- In WHERE clauses and comparisons
SELECT * FROM events WHERE created_date = CURRENT_DATE();
```

## SQL Standard Compliance

These functions align with ANSI SQL:2003 and SQL:2008 standards, supporting both parenthesized and non-parenthesized syntax for improved compatibility with PostgreSQL, MySQL,
SQL Server, and Oracle.

## Technical Notes

- Functions support both CURRENT_DATE and CURRENT_DATE() syntax for maximum compatibility
- UTC-based implementation ensures consistent behavior across timezones

## Summary by CodeRabbit

* **New Features**
  * Added support for the SQL functions `CURRENT_DATE`, `CURRENT_TIME`, and `CURRENT_TIMESTAMP`.
  * These functions can now be used in queries, default column values, and expressions.

* **Tests**
  * Introduced new test cases to verify the correct behavior of `CURRENT_DATE`, `CURRENT_TIME`, and `CURRENT_TIMESTAMP` in various scenarios.


core/src/executor/evaluate.rs
```diff
@@ -589,6 +589,13 @@ async fn evaluate_function<'a, 'b: 'a, 'c: 'a, T: GStore>(
             return f::greatest(name, exprs);
         }
         Function::Now() => return Ok(Evaluated::Value(Value::Timestamp(Utc::now().naive_utc()))),
+        Function::CurrentDate() => {
+            return Ok(Evaluated::Value(Value::Date(Utc::now().date_naive())));
+        }
+        Function::CurrentTime() => return Ok(Evaluated::Value(Value::Time(Utc::now().time()))),
+        Function::CurrentTimestamp() => {
```
panarch: `NOW` and `CURRENT_TIMESTAMP` shares the same logic.
then we can use
```rust
Function::Now() | Function::CurrentTimestamp() => {
```

test-suite/src/function/current_datetime.rs
```diff
@@ -0,0 +1,59 @@
+use {
+    crate::*,
+    gluesql_core::prelude::{Payload, Value::*},
+};
+
+test_case!(current_datetime, {
```
panarch: ref. https://gluesql.org/docs/0.16.0/sql-syntax/functions/text/left

We aim to generate docs from the test-suite here.
This test itself looks good enough. but three different functions exist in a single test case `current_datetime`.
Then, it could be hard to make the automation flow.
Can we split this to have three test cases - `current_time`, `current_timestamp` and `current_date`.

panarch (changes requested):

test-suite/src/function/current_date.rs
```diff
@@ -0,0 +1,35 @@
...
+test_case!(current_date, {
+    let g = get_tester!();
+
+    macro_rules! date {
+        ($date: expr) => {
+            $date.parse().unwrap()
+        };
+    }
+
+    let test_cases = [
```
panarch: could we try
ref. https://github.com/gluesql/gluesql/blob/main/test-suite/src/function/slice.rs

```rust
g.named_test
g.test
```
patterns?

we are trying to replace legacy test templates to use `g.test` or `g.named_test`
(`g.test` for simple unnamed, and we can specify test case name using `g.named_test`)

junghoon-vans: I changed all test cases implemented with legacy test templates to the `g.named_test` method.

panarch (approved):
Looks all nice!

PR #1683, opened 2025-07-27
Add NULLIF functions

miinhho:
## Summary

Implements the SQL standard `NULLIF` function to enhance compatibility with standard SQL queries.

## Implementation Details

* **AST Extension**: Introduced a new `NullIf` variant in the `Function` enum to represent `NULLIF` expressions in the abstract syntax tree.
* **Parser Support**: Extended `translate_function` to recognize and correctly parse the `NULLIF` function.
* **AST Serialization**: Implemented `to_sql()` formatting support to correctly serialize `NULLIF` expressions with proper syntax and parentheses.
* **Evaluation Logic**: Added evaluation logic to execute the semantics of `NULLIF`, returning `NULL` when both arguments are equal, or the first argument otherwise.
* **Pattern Matching**: Updated expression planning to support pattern matching for the new `NULLIF` function variant.

## Usage Examples

```sql
-- Returns NULL because both arguments are equal
SELECT NULLIF('Hello', 'Hello');

-- Returns the first argument because they are not equal
SELECT NULLIF('Hello', 'World');

-- Practical use case: prevent division by zero
SELECT 
  value,
  divisor,
  value / NULLIF(divisor, 0) AS safe_division
FROM (
  VALUES 
    (100, 10),
    (50, 0),
    (30, 5)
) AS t(value, divisor);
```

## SQL Standard Compliance

The `NULLIF` function is defined in the **ISO/IEC 9075-2:2011 (SQL/Foundation)** standard, Section 6.27 “<case expression>”.
It is treated as a shorthand for a searched `CASE` expression:

```sql
NULLIF(a, b)
-- Equivalent to:
CASE WHEN a = b THEN NULL ELSE a END
```

## Summary by CodeRabbit

* **New Features**
  * Added support for the SQL function `NULLIF`, enabling expressions to return `NULL` if two values are equal, or the first value otherwise.

* **Bug Fixes**
  * Improved argument validation for `NULLIF` to enforce correct argument count with clear error messages.

* **Tests**
  * Added extensive tests for `NULLIF` covering different data types, usage scenarios, and error handling.


panarch (approved):

PR #1685, opened 2025-07-28
Evaluate expressions to NULL when any operand is NULL

moreal:
While working on sqllogictest-related tasks, I gained a better understanding of `NULL` behavior through failing test cases. Based on SQLite's behavior, I learned that operations like `NOT NULL` or `NULL = NULL` evaluate to `NULL`. While I cannot be certain without access to the ANSI SQL standard documentation, this appears to be the case based on manual testing (SQLite, PostgreSQL, MySQL).

This pull request also made `BETWEEN ... AND ...` and `... IN (...)`  handle `NULL`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of SQL NULL values in comparison, logical, and arithmetic operations to align with SQL standards.
  * Expressions involving NULL now consistently return NULL or false where appropriate, especially in BETWEEN, IN, and logical checks.

* **New Features**
  * Added comprehensive tests for SQL NULL behavior, including comparisons, BETWEEN, and IN list operations.
  * Expanded test coverage for edge cases involving NULL in SQL expressions.

* **Refactor**
  * Simplified and clarified internal handling of boolean and NULL logic in expression evaluation.
  * Enhanced expression evaluation with explicit null checks and streamlined boolean negation handling.


core/src/executor/evaluate/expr.rs
```diff
@@ -85,6 +93,10 @@ pub fn between<'a>(
     low: Evaluated<'a>,
     high: Evaluated<'a>,
 ) -> Result<Evaluated<'a>> {
+    if target.is_null() {
+        return Ok(target);
+    }
```
panarch: we may not need these `target.is_null()` check.
currently only the issue is from here.

```rust
   UnaryOperator::Not => v
            .try_into()
            .map(|v: bool| Evaluated::Value(Value::Bool(!v))),
```

for the other cases, null value is already being handled properly.
could we try change `Not` handler to return Null on null input?

panarch (changes requested):

core/src/executor/evaluate/expr.rs
moreal: Well... I'm not sure that I understood well but I think the `target.is_null()` check in `fn between` is still needed because `NULL BETWEEN <lhs> AND <rhs>` should return `NULL`.

moreal: I separated the part to `.unary_not()` function like other unary operators (e.g., `.unary_plus()`).

moreal:
I pushed a commit, 97eb5bc8bd481db8bea1549539401b1e5c0868da, which changes the return type of functions named `evaluate_eq` from `bool` to `Value`, `Literal` or `Evaluated<'>` to handle `Null` well.

I also consider such changes to `evaluate_cmp` but I'm not sure. Before starting to work on it, I want to know maintainer's opinion because it changes many lines. How do you think about these change? 🤔 @panarch

panarch:
> I pushed a commit, [97eb5bc](https://github.com/gluesql/gluesql/commit/97eb5bc8bd481db8bea1549539401b1e5c0868da), which changes the return type of functions named `evaluate_eq` from `bool` to `Value`, `Literal` or `Evaluated<'>` to handle `Null` well.
> 
> I also consider such changes to `evaluate_cmp` but I'm not sure. Before starting to work on it, I want to know maintainer's opinion because it changes many lines. How do you think about these change? 🤔 @panarch

Wow, this is great.
Thanks, changing `evaluate_eq` return type from `bool` to others can fix many existing unclear current NULL handlings.

Ok, I am full for this idea and I agree that is is necessary part of the current PR.
Then, could we split the task into two,
- changing `evaluate_eq` types
- the current PR: Evaluate expressions with NULL operand as NULL

Then I can review the `evaluate_eq` fastly, and then we can go to the next.

> I also consider such changes to evaluate_cmp

It sounds also good! yes that can include lots of code changes. it would be better to do this task by making a new PR.

moreal:
> Ok, I am full for this idea and I agree that is is necessary part of the current PR. Then, could we split the task into two,
> 
>     * changing `evaluate_eq` types
> 
>     * the current PR: Evaluate expressions with NULL operand as NULL
> 
> 
> Then I can review the `evaluate_eq` fastly, and then we can go to the next.

@panarch Thanks for your opinion! I excluded two commits related with changed the return type of `evaluate_eq` and I'll open a new PR including the two commits.

moreal:
@panarch Could you review this pull request when you have a moment? Please let me know if there's anything else I need to do! 🙏🏻

panarch:
> @panarch Could you review this pull request when you have a moment? Please let me know if there's anything else I need to do! 🙏🏻

Ah! I thought that this PR depends on https://github.com/gluesql/gluesql/pull/1705 but it wasn't and can be reviewed separately. Ok. I will review right now, thanks for noticing.

core/src/executor/filter.rs
```diff
@@ -59,5 +60,8 @@ pub async fn check_expr<'a, T: GStore>(
 ) -> Result<bool> {
     evaluate(storage, context, aggregated, expr)
         .await
-        .map(|evaluated| evaluated.try_into())?
+        .map(|evaluated| match evaluated {
+            Evaluated::Value(Value::Null) => Ok(false),
```
panarch: Evaluated also has `Literal` variant, then it would be better to use `evaluated.is_null()` then only using `Evaluated::Value(Value::Null)`?

core/src/executor/evaluate/expr.rs
```diff
@@ -39,6 +39,10 @@ pub fn binary_op<'a>(
         }};
     }
 
+    if l.is_null() || r.is_null() {
+        return Ok(Evaluated::Value(Value::Null));
+    }
```
panarch, line 44: Ok I see that this is necessary until we modify both `evaluate_eq` and `evaluate_cmp` return types.

core/src/executor/filter.rs
moreal: Ah it's my mistake. Thanks for catching it. I fixed it in [660405b](https://github.com/gluesql/gluesql/pull/1685/commits/660405b86c22d8473c312c7bdc1b5dee33e6cba7)

panarch (approved):
Looks all nice!! Thanks a lot for the contribution 👍 👍

PR #1724, opened 2025-08-10
Implement custom Eq and Hash for Value

zmrdltl:
- Add custom Hash implementation with consistent hashing for all Value types
- Implement Eq trait for Value to support HashSet/HashMap usage
- Handle float zero normalization (+0.0/-0.0) for consistent hashing
- Sort Map entries by key for deterministic hash values
- Use debug representation for complex types like Point

This enables using Value as keys in HashMap and elements in HashSet while maintaining proper equality semantics.


panarch (approved):
All looks great! Hash support to Value can make us add lots of new features.

PR #1725, opened 2025-08-10
Implement LEAST(...) function

JJong-Min:
- Implement LEAST(...) function

### DIFF
--- a/test-suite/src/function/least.rs
+++ b/test-suite/src/function/least.rs
@@ -1,0 +1,91 @@
+use {
+  crate::*,
+  gluesql_core::{
+      error::{EvaluateError, TranslateError},
+      prelude::Value::*,
+  },
+};
+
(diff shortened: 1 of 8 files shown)


test-suite/src/function/least.rs
```diff
@@ -0,0 +1,91 @@
...
+    .await;
+
+    g.test(
+        "SELECT LEAST(1, 2, 'bibibik') AS goat;",
+        Err(EvaluateError::NonComparableArgumentError("LEAST".to_owned()).into()),
+    )
+    .await;
+
+    g.test(
+        "SELECT LEAST(NULL, 'bibibik', 'babamba', 'melona') AS goat;",
```
panarch, line 73: Rather than error, for any kind of SQL functions and expressions, we can consider to return `NULL` when any operand comes with NULL

PR #1731, opened 2025-08-10
Added AST Builder tests for supported SQL functions

miinhho:
- Add ast builder test for generate_uuid, current_date, current_time, current_timestamp
- Fix ast builder test for nullif


test-suite/src/ast_builder/function/reference/current_time.rs
```diff
@@ -0,0 +1,25 @@
...
+    let expected = Ok(Payload::Create);
+    assert_eq!(actual, expected, "create table - Record");
+
+    let actual = table("Record")
+        .insert()
+        .values(vec!["1, CURRENT_TIME()"])
+        .execute(glue)
+        .await;
+    let expected = Ok(Payload::Insert(1));
+    assert_eq!(actual, expected, "insert - Record");
```
panarch, line 24: type_match used in generate_uuid test looks good.
It would be good to also use type_match here?

miinhho: Oh right, I'll try using type_match in the current_time test too!

miinhho:
Okay, I'll close this PR and split the changes into 3 PRs.

PR #1736, opened 2025-08-11
Fix nullif ast builder test

miinhho:
Improved and corrected the nullif function test case.


miinhho:
I missed moving test to reference directory. Fixing nullif ast builder test would be after that.

PR #1746, opened 2025-08-17
Remove ast::Statement ToSql trait implementation

panarch:
GlueSQL's AST includes some partial SQL structural information and execution plans.
ToSql assumes that our AST can be converted to SQL but it is actually not.
e.g.
non-clustered index, primary key, hash join, ...

This kind of partial support can make us hard to maintain the codes.

This PR starts removing ToSql impl from our AST nodes by removing ToSql impl in `ast::Statement`.
If there is need of restoring SQL from AST, then we can still use it from sqlparser-rs' AST, not gluesql's.

## Summary
- generate insert statements in CLI dump without building `Statement::Insert`
- remove `Statement`'s `ToSql` implementation and tests
- build schema DDL via string formatting instead of `Statement::CreateTable`

## Testing
- `cargo fmt --all`

------
https://chatgpt.com/codex/tasks/task_e_68a19365b804832ab8597857df0efd34

## Summary by CodeRabbit

* **Refactor**
  * Database dumps now emit consolidated multi-row INSERT statements for smaller, faster imports.
  * CREATE TABLE output is more consistent: table names are quoted; column definitions are omitted when empty; optional ENGINE and COMMENT clauses are included when present.
  * Index creation statements remain separate and follow the table DDL.

* **Chores**
  * Simplified SQL generation path for more predictable, readable dump and DDL output.


devgony (approved):
👍🏼 👍🏼 👍🏼

Issue #357, opened 2021-09-19
Add `TryInto<f64, i64 and String> for Evaluated`


panarch:
https://github.com/gluesql/gluesql/blob/86d5078f9e766a3465577c035b690e964712580f/src/executor/evaluate/evaluated.rs#L33-L42

There exists some inappropriate use case of `TryInto<Value> for Evaluated`.
`eval_to_str`, `eval_to_float` and `eval_to_integer` closures in `evaluate_function - executor/evaluate module` are not supposed to use `TryInto<Value>`.
They need direct conversion from `Evaluated` to `String`, `i64` and `f64`.

PR #1750, opened 2025-08-19
Add TryInto<f64, i64, String> for Evaluated to Support Direct Conversions

ljluestc:
This PR addresses [issue #357](https://github.com/gluesql/gluesql/issues/357), which requests direct conversions from the `Evaluated` type to `f64`, `i64`, and `String` by implementing the `TryInto` trait for these target types. The issue, opened by @panarch, notes that the existing `TryInto<Value>` implementation for `Evaluated` is inefficient or inappropriate for the `eval_to_str`, `eval_to_float`, and `eval_to_integer` closures in the `evaluate_function` module, as these require direct conversions without an intermediate `Value` type.

The changes introduce efficient `TryInto` implementations, update the relevant closures, and add comprehensive tests to ensure correctness and compatibility across GlueSQL's storage backends.

### Changes Made

1. **Added `TryInto` Implementations**:
   - Implemented `TryInto<f64>` for `Evaluated` in `gluesql/src/executor/evaluate/evaluated.rs`:
     - Converts `Evaluated::Literal` (`Number`, `Text`) and `Evaluated::Value` (`F64`, `I64`, `Str`) to `f64`.
     - Handles invalid types with `Error::InvalidType` and parsing errors with `Error::Parse`.
   - Implemented `TryInto<i64>` for `Evaluated`:
     - Converts `Evaluated::Literal` (`Number`, `Text`) and `Evaluated::Value` (`I64`, `F64`, `Str`) to `i64`.
     - Ensures `F64` values are whole numbers within `i64` bounds to prevent precision loss.
   - Implemented `TryInto<String>` for `Evaluated`:
     - Converts `Evaluated::Literal` (`Text`, `Number`, `Boolean`, `Null`) and `Evaluated::Value` (`Str`, `I64`, `F64`, `Bool`, `Null`) to `String`.
     - Provides straightforward string representations for supported types.

2. **Updated `evaluate_function` Closures**:
   - Modified the `eval_to_str`, `eval_to_float`, and `eval_to_integer` closures in `gluesql/src/executor/evaluate/mod.rs` to use the new `TryInto` implementations directly, replacing the previous `TryInto<Value>`-based logic.
   - Ensured proper error propagation using the `Result` type.

### DIFF
--- a/gluesql/src/executor/evaluate/mod.rs
+++ b/gluesql/src/executor/evaluate/mod.rs
@@ -1,0 +1,9 @@
+// ...existing code...
+
+// Replace the old closure definitions for eval_to_str, eval_to_float, eval_to_integer:
+let eval_to_str = |e: Evaluated| -> Result<String> { e.try_into() };
+let eval_to_float = |e: Evaluated| -> Result<f64> { e.try_into() };
+let eval_to_integer = |e: Evaluated| -> Result<i64> { e.try_into() };
+
+// ...existing code...
+
--- a/gluesql/src/executor/evaluate/evaluated.rs
+++ b/gluesql/src/executor/evaluate/evaluated.rs
@@ -1,0 +1,80 @@
+// ...existing code...
+
+use std::convert::TryInto;
+use crate::data::{Value, Literal};
+use crate::result::{Result, Error};
+
+// ...existing code...
+
+impl<'a> TryInto<f64> for Evaluated<'a> {
+    type Error = Error;
+
+    fn try_into(self) -> Result<f64> {
+        match self {
+            Evaluated::Literal(literal) => match literal {
+                Literal::Number(n) => n.parse::<f64>().map_err(|e| Error::Parse(e.to_string())),
+                Literal::Text(s) => s.parse::<f64>().map_err(|e| Error::Parse(e.to_string())),
+                _ => Err(Error::InvalidType("Cannot convert to f64".to_owned())),
+            },
+            Evaluated::Value(value) => match value {
+                Value::F64(f) => Ok(f),
+                Value::I64(i) => Ok(i as f64),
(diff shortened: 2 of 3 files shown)


gluesql/src/executor/evaluate/mod.rs
```diff
@@ -0,0 +1,9 @@
+// ...existing code...
+
+// Replace the old closure definitions for eval_to_str, eval_to_float, eval_to_integer:
+let eval_to_str = |e: Evaluated| -> Result<String> { e.try_into() };
+let eval_to_float = |e: Evaluated| -> Result<f64> { e.try_into() };
+let eval_to_integer = |e: Evaluated| -> Result<i64> { e.try_into() };
```
panarch, line 6: Thanks for the suggestion.
We may need to discuss about the decision.
Currently, existing `eval_to_str`, `eval_to_float`, .. functions in `evaluate` module are not ideal, not even ideal, it has certain needs of refactoring.

However, if we force to convert the `Evaluated` to `String`, `f64` and `i64`, then there can be some kind of questions.

one example to look around would be binary operators.
one of them, `MOD` or simply `1 + 2` like operators, etc.
`MOD(4, 3)`, `1 + 2`

Ok, Let's assume that `1 + 2` if operands are not from literal, if it was from `Value::I8` like some non-i64 ones.
Then the expected result of `1 + 2` is `3` with `Value::I8`, not `Value::I64`.
It looks that current suggestion to add try conversion to `Evaluted` cannot handle this kind of issues.

We still don't have the something simple and ideal answer for this. We partially use the binary_op macros and that is the best way for now.
Could we talk about this?

panarch: @codex I also need your opinion!

PR #1753, opened 2025-08-26
Remove unused ToSql impl for AlterTableOperation, OperateFunctionArg and Assignment

panarch:
## Summary
- remove unused ToSql for AlterTableOperation, OperateFunctionArg, and Assignment
- drop OperateFunctionArg to_sql test

## Testing
- `cargo fmt --all`
- `cargo test -p gluesql-core`

------
https://chatgpt.com/codex/tasks/task_e_68adbe10d4e4832a91a4c5c6a8202ed2

## Summary by CodeRabbit

* **Refactor**
  * Streamlined internal SQL generation for assignments and schema changes, reducing redundancy and maintenance overhead. Behavior and outputs remain unchanged for end-users.
* **Tests**
  * Removed obsolete tests tied to deprecated logic and updated related imports. Test suite adjusted accordingly and remains green.

### DIFF
--- a/core/src/ast.rs
+++ b/core/src/ast.rs
@@ -154,12 +154,6 @@ pub enum Variable {
     Version,
 }
 
-impl ToSql for Assignment {
-    fn to_sql(&self) -> String {
-        format!(r#""{}" = {}"#, self.id, self.value.to_sql())
-    }
-}
-
 impl ToSql for ForeignKey {
     fn to_sql(&self) -> String {
         let ForeignKey {
--- a/core/src/ast/ddl.rs
+++ b/core/src/ast/ddl.rs
@@ -47,30 +47,6 @@ pub struct OperateFunctionArg {
     pub default: Option<Expr>,
 }
 
-impl ToSql for AlterTableOperation {
-    fn to_sql(&self) -> String {
-        match self {
-            AlterTableOperation::AddColumn { column_def } => {
-                format!("ADD COLUMN {}", column_def.to_sql())
-            }
-            AlterTableOperation::DropColumn {
-                column_name,
-                if_exists,
-            } => match if_exists {
-                true => format!(r#"DROP COLUMN IF EXISTS "{column_name}""#),
-                false => format!(r#"DROP COLUMN "{column_name}""#),
-            },
-            AlterTableOperation::RenameColumn {
-                old_column_name,
-                new_column_name,
-            } => format!(r#"RENAME COLUMN "{old_column_name}" TO "{new_column_name}""#),
-            AlterTableOperation::RenameTable { table_name } => {
-                format!(r#"RENAME TO "{table_name}""#)
-            }
-        }
-    }
-}
-
 impl ToSql for ColumnDef {
     fn to_sql(&self) -> String {
         let ColumnDef {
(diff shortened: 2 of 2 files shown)


panarch:
@codex how do you think about this pr?

panarch:
> Codex Review: Here are some suggestions.
> 
> https://github.com/gluesql/gluesql/blob/98ba7204abcdb2622de4c02797bf99a07f8f1e79/core/src/ast.rs#L144-L148
> 
> 
> **[P1] Removing public ToSql impls breaks external callers**
> The commit removes the `ToSql` implementation for `Assignment` (and likewise for `AlterTableOperation` and `OperateFunctionArg` in `ast/ddl.rs`), but the structs and the `ToSql` trait remain `pub`. Any downstream crate that previously called `.to_sql()` on these AST nodes—for example when regenerating SQL for `UPDATE`, `ALTER TABLE`, or `CREATE FUNCTION` statements—will now fail to compile even though the behavior inside this repository is unaffected. Unless a major version bump is planned, consider keeping these impls or providing a migration path to avoid a semver-breaking API change.
> 
> Reply with `@codex fix comments` to fix any unresolved comments.
> 
> About Codex in GitHub

@codex That is the key point of this PR.
GlueSQL's AST does not aim to preserve SQL structural information. The essential purpose of GlueSQL AST is containing the execution plan. Previously, we thought that it would be convenient to let GlueSQL AST to support SQL printing using ToSql, but it has certain limitations; primary key, non-clustered index, join exeucution strategy,.. that cannot be matched to SQL.
Therefore, we now start to remove this kind of ambiguity. Removing all ToSql at once is too big and some internal and storage impl depends on ToSql impl. This PR starts removing some ToSql impl that are not related to any other codes in the gluesql workspace.
Does it make sense?

panarch:
> Your explanation makes sense—if GlueSQL’s AST is shifting toward representing execution plans rather than SQL text, removing `ToSql` impls is consistent with that design. The main concern is the public API: downstream crates that relied on `.to_sql()` will break. To avoid unexpected breakage, consider one of the following approaches:
> 
> * **Versioning:** Release a new major (or clearly documented minor) version, noting the removal of these impls as a breaking change.
> * **Migration path:** Provide guidance (e.g., “manually implement SQL emission” or “use a separate SQL formatter”) so downstream users know how to adapt.
> * **Optional stubs or feature flag (soft landing):** You could reintroduce `#[deprecated]` stub impls that panic at runtime, or gate the old behavior behind a feature flag disabled by default.
> 
> With clear release notes and a migration path, the PR’s intent is coherent and avoids confusion for anyone consuming the public API.
> 
> [View task →](https://chatgpt.com/codex/tasks/task_e_68afde085600832aa05b7b34f44c7f34)

@codex

- Versioning: we not yet release 1.0 so it would be enough to bump minor version.
- Migration path: using translate module, we convert the AST of sqlparser-rs that contains sql structural information to our AST, so someone who depends on `ToSql` can use sqlparser-rs' display. Would it be enough to include this info in the new version release note?
- Optional stubs or feature flag: not yet we are 1.0, so that might be too much for now.

ToSql support is an additional sub feature so it would be ok to drop.

PR #1789, opened 2025-09-19
Reject unsupported translate clauses

panarch:
## Summary
- treat WITH / FETCH / FOR UPDATE as unsupported query options during translation
- reject SELECT variants that introduce INTO, TOP, WINDOW, etc., via `UnsupportedSelectOption`
- add focused unit tests covering each new guard

## Testing
- cargo test -p gluesql-core translate::query::tests::query_options_rejected -- --exact
- cargo test -p gluesql-core translate::query::tests::select_options_rejected -- --exact
- cargo test -p gluesql-core translate::tests::delete_limit_not_supported -- --exact
- cargo clippy --all-targets -- -D warnings],

## Summary by CodeRabbit

* **New Features**
  * Earlier, clearer errors for unsupported SQL options across INSERT, UPDATE, DELETE, queries and SELECT, returning specific reasons for clauses like RETURNING, ON CONFLICT, DEFAULT VALUES, FROM, USING, ORDER BY, LIMIT, WITH, FETCH, LOCK, INTO, WINDOW.
  * Added new public error variants to represent those unsupported option cases.

* **Tests**
  * Expanded tests covering the new error paths and messages.
  * Fixed a test SQL statement to use a valid INSERT so expected rows are created.

### DIFF
--- a/core/src/translate.rs
+++ b/core/src/translate.rs
@@ -529,7 +529,6 @@ mod tests {
         ];
 
         for (sql, err) in cases {
-            #[allow(deprecated)]
             assert_translate_error(sql, err);
         }
     }
--- a/core/src/translate.rs
+++ b/core/src/translate.rs
@@ -93,18 +93,57 @@ pub fn translate(sql_statement: &SqlStatement) -> Result<Statement> {
             table,
             assignments,
             selection,
+            from,
+            returning,
             ..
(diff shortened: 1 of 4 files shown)


devgony (approved):
👍🏼 👍🏼 👍🏼

PR #1792, opened 2025-09-21
docs: document personal agent instructions

panarch:
## Summary
- document how collaborator-specific instructions should be stored locally
- ignore AGENTS.local.md so the personal notes stay untracked

## Testing
- cargo fmt --all


devgony (approved):
It looks so usefull 👍🏼👍🏼👍🏼

Issue #1786, opened 2025-09-14
Enable clippy pedantic warnings for code quality


zmrdltl:
## Issue Overview

- Apply clippy pedantic linting to improve code quality and consistency
- Ensure all code passes stricter Rust best practices

## Feature Details
- **Apply command**: `cargo clippy --all-features --tests -- -W clippy::pedantic`
- **Fix all warnings**: Update code to satisfy pedantic requirements
- **CI integration**: Add pedantic check to CI pipeline

PR #1800, opened 2025-09-26
Support only indexed placeholders for parameter binding

panarch:
## Summary
- remove question-mark placeholder normalization and require indexed forms
- introduce ParamLiteral conversions and tighten translation pipeline
- add example and unit tests for parameter binding and placeholder indexing

## Testing
- cargo fmt --all
- cargo test -p gluesql-core
- cargo test -p gluesql_memory_storage

### DIFF
--- a/core/src/glue.rs
+++ b/core/src/glue.rs
@@ -6,7 +6,7 @@ use {
         plan::plan,
         result::Result,
         store::{GStore, GStoreMut},
-        translate::translate,
+        translate::{ParamLiteral, translate_with_params},
     },
     futures::{
         TryStreamExt,
@@ -24,13 +24,28 @@ impl<T: GStore + GStoreMut> Glue<T> {
         Self { storage }
     }
 
-    pub async fn plan<Sql: AsRef<str>>(&mut self, sql: Sql) -> Result<Vec<Statement>> {
+    pub async fn plan_with_params<Sql, I, P>(
+        &mut self,
+        sql: Sql,
+        params: I,
+    ) -> Result<Vec<Statement>>
+    where
+        Sql: AsRef<str>,
+        I: IntoIterator<Item = P>,
+        P: Into<ParamLiteral>,
+    {
         let parsed = parse(sql)?;
+        let params: Vec<ParamLiteral> = params.into_iter().map(Into::into).collect();
         let storage = &self.storage;
         stream::iter(parsed)
-            .map(|p| translate(&p))
+            .map(|p| translate_with_params(&p, params.clone()))
             .then(|statement| async move { plan(storage, statement?).await })
             .try_collect()
+            .await
+    }
+
+    pub async fn plan<Sql: AsRef<str>>(&mut self, sql: Sql) -> Result<Vec<Statement>> {
+        self.plan_with_params(sql, std::iter::empty::<ParamLiteral>())
             .await
     }
 
@@ -38,8 +53,17 @@ impl<T: GStore + GStoreMut> Glue<T> {
         execute(&mut self.storage, statement).await
     }
 
-    pub async fn execute<Sql: AsRef<str>>(&mut self, sql: Sql) -> Result<Vec<Payload>> {
-        let statements = self.plan(sql).await?;
+    pub async fn execute_with_params<Sql, I, P>(
+        &mut self,
+        sql: Sql,
+        params: I,
+    ) -> Result<Vec<Payload>>
+    where
+        Sql: AsRef<str>,
+        I: IntoIterator<Item = P>,
+        P: Into<ParamLiteral>,
+    {
+        let statements = self.plan_with_params(sql, params).await?;
         let mut payloads = Vec::<Payload>::new();
         for statement in statements.iter() {
             let payload = self.execute_stmt(statement).await?;
@@ -47,5 +71,10 @@ impl<T: GStore + GStoreMut> Glue<T> {
         }
 
         Ok(payloads)
+    }
+
+    pub async fn execute<Sql: AsRef<str>>(&mut self, sql: Sql) -> Result<Vec<Payload>> {
+        self.execute_with_params(sql, std::iter::empty::<ParamLiteral>())
+            .await
     }
 }
--- a/core/src/translate.rs
+++ b/core/src/translate.rs
@@ -5,22 +5,31 @@ mod error;
 mod expr;
 mod function;
 mod operator;
+mod param;
 mod query;
 
 pub use self::{
     data_type::translate_data_type,
-    ddl::{translate_column_def, translate_operate_function_arg},
+    ddl::translate_column_def,
     error::TranslateError,
     expr::{translate_expr, translate_order_by_expr},
+    param::ParamLiteral,
     query::{alias_or_name, translate_query, translate_select_item},
 };
 
 use {
+    self::{
+        expr::{translate_expr_with_params, translate_order_by_expr_with_params},
+        query::translate_query_with_params,
+    },
     crate::{
-        ast::{Assignment, ForeignKey, ReferentialAction, Statement, Variable},
+        ast::{Assignment, Expr, ForeignKey, ReferentialAction, Statement, Variable},
         result::Result,
     },
-    ddl::translate_alter_table_operation,
+    ddl::{
+        translate_alter_table_operation_with_params, translate_column_def_with_params,
+        translate_operate_function_arg_with_params,
+    },
     sqlparser::ast::{
         Assignment as SqlAssignment, AssignmentTarget as SqlAssignmentTarget,
         CommentDef as SqlCommentDef, CreateFunctionBody as SqlCreateFunctionBody,
@@ -33,8 +42,24 @@ use {
 };
 
 pub fn translate(sql_statement: &SqlStatement) -> Result<Statement> {
+    translate_with_params(sql_statement, std::iter::empty::<ParamLiteral>())
+}
+
+pub fn translate_with_params<I, P>(sql_statement: &SqlStatement, params: I) -> Result<Statement>
+where
+    I: IntoIterator<Item = P>,
+    P: Into<ParamLiteral>,
+{
+    let params: Vec<ParamLiteral> = params.into_iter().map(Into::into).collect();
+
+    translate_internal(sql_statement, &params)
+}
+
+fn translate_internal(sql_statement: &SqlStatement, params: &[ParamLiteral]) -> Result<Statement> {
     match sql_statement {
-        SqlStatement::Query(query) => translate_query(query).map(Statement::Query),
+        SqlStatement::Query(query) => {
+            translate_query_with_params(query, params).map(Statement::Query)
+        }
         SqlStatement::Insert(SqlInsert {
             table_name,
             columns,
@@ -70,7 +95,7 @@ pub fn translate(sql_statement: &SqlStatement) -> Result<Statement> {
             let table_name = translate_object_name(table_name)?;
             let columns = translate_idents(columns);
             let source = match source.as_deref() {
-                Some(query) => translate_query(query)?,
+                Some(query) => translate_query_with_params(query, params)?,
                 None => {
                     return Err(TranslateError::DefaultValuesOnInsertNotSupported(
                         table_name.clone(),
@@ -109,9 +134,12 @@ pub fn translate(sql_statement: &SqlStatement) -> Result<Statement> {
                 table_name: translate_table_with_join(table)?,
                 assignments: assignments
                     .iter()
-                    .map(translate_assignment)
+                    .map(|assignment| translate_assignment_with_params(assignment, params))
                     .collect::<Result<_>>()?,
-                selection: selection.as_ref().map(translate_expr).transpose()?,
+                selection: selection
+                    .as_ref()
+                    .map(|expr| translate_expr_with_params(expr, params))
+                    .transpose()?,
             })
         }
         SqlStatement::Delete(SqlDelete {
@@ -153,7 +181,10 @@ pub fn translate(sql_statement: &SqlStatement) -> Result<Statement> {
 
             Ok(Statement::Delete {
                 table_name,
-                selection: selection.as_ref().map(translate_expr).transpose()?,
+                selection: selection
+                    .as_ref()
+                    .map(|expr| translate_expr_with_params(expr, params))
+                    .transpose()?,
             })
         }
         SqlStatement::CreateTable(SqlCreateTable {
@@ -168,7 +199,7 @@ pub fn translate(sql_statement: &SqlStatement) -> Result<Statement> {
         }) => {
             let columns = columns
                 .iter()
-                .map(translate_column_def)
+                .map(|column_def| translate_column_def_with_params(column_def, params))
                 .collect::<Result<Vec<_>>>()?;
 
             let columns = (!columns.is_empty()).then_some(columns);
@@ -185,7 +216,7 @@ pub fn translate(sql_statement: &SqlStatement) -> Result<Statement> {
                 name,
                 columns,
                 source: match query {
-                    Some(v) => Some(translate_query(v).map(Box::new)?),
+                    Some(v) => Some(translate_query_with_params(v, params).map(Box::new)?),
                     None => None,
                 },
                 engine: engine
@@ -213,7 +244,7 @@ pub fn translate(sql_statement: &SqlStatement) -> Result<Statement> {
 
             Ok(Statement::AlterTable {
                 name: translate_object_name(name)?,
-                operation: translate_alter_table_operation(operation)?,
+                operation: translate_alter_table_operation_with_params(operation, params)?,
             })
         }
         SqlStatement::Drop {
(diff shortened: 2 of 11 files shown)


panarch:
@codex review plz

core/src/translate/param.rs
```diff
@@ -0,0 +1,507 @@
...
+        })
+    }
+}
+
+impl IntoParamLiteral for NaiveDateTime {
+    fn into_param_literal(self) -> Result<ParamLiteral, TranslateError> {
+        Ok(ParamLiteral::TypedString {
+            data_type: DataType::Timestamp,
+            value: Utc.from_utc_datetime(&self).to_string(),
+        })
```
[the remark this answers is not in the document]
panarch: @codex thanks! updated, review again plz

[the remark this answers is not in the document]
panarch: @codex updated! thanks! review again plz

core/src/translate/param.rs
```diff
@@ -0,0 +1,529 @@
...
+    },
+    Interval {
+        expr: Box<ParamLiteral>,
+        leading_field: Option<DateTimeField>,
+        last_field: Option<DateTimeField>,
+    },
+}
+
+impl ParamLiteral {
+    pub const fn null() -> Self {
```
zmrdltl, line 30: It would be great to add `#[must_use]` here so the compiler warns when the return value of this function is ignored.

core/src/translate/param.rs
```diff
@@ -0,0 +1,529 @@
...
+        Self::Literal(AstLiteral::Null)
+    }
+
+    pub fn into_expr(self) -> Expr {
```
zmrdltl, line 35: same as above

core/src/translate/param.rs
```diff
@@ -0,0 +1,529 @@
...
+fn into_number_literal(value: impl ToString) -> Result<AstLiteral, TranslateError> {
+    let value_str = value.to_string();
+    match value_str.parse::<BigDecimal>() {
+        Ok(number) => Ok(AstLiteral::Number(number)),
+        Err(_) => Err(TranslateError::InvalidParamLiteral { value: value_str }),
+    }
+}
+
+pub trait IntoParamLiteral {
+    fn into_param_literal(self) -> Result<ParamLiteral, TranslateError>;
```
zmrdltl, line 73: Could you please add a brief doc comment here that includes a # Errors section explaining when `into_param_literal` may return a TranslateError? For example:

```rust
/// Converts the value into a [`ParamLiteral`] so it can be bound as a query parameter.
///
/// # Errors
///
/// Returns an error if the value cannot be represented as a [`ParamLiteral`],
/// for example when a numeric conversion or validation fails, or when downstream
/// constraints such as the `NonZeroUsize` placeholder index requirement reject it.
fn into_param_literal(self) -> Result<ParamLiteral, TranslateError>;
```

core/src/translate/param.rs
```diff
@@ -0,0 +1,529 @@
...
+            } => Expr::Interval {
+                expr: Box::new(expr.into_expr()),
+                leading_field,
+                last_field,
+            },
+        }
+    }
+}
+
+fn into_number_literal(value: impl ToString) -> Result<AstLiteral, TranslateError> {
```
zmrdltl: Would you consider taking this argument by reference instead? It’s only read (not consumed) in the body, so borrowing it would avoid unnecessary moves.

```rust
fn into_number_literal(value: &impl ToString)
```

core/src/translate/function.rs
```diff
@@ -18,16 +18,17 @@ use {
     },
 };
 
-pub fn translate_trim(
+pub(crate) fn translate_trim_with_params(
+    params: &[ParamLiteral],
     expr: &SqlExpr,
     trim_where: &Option<SqlTrimWhereField>,
     trim_what: &Option<Box<SqlExpr>>,
```
zmrdltl: Could we update the signature to take `Option<&SqlTrimWhereField>` and `Option<&SqlExpr>` rather than `&Option<_>`? Clippy’s ref_option lint (pedantic) warns about the extra layer of indirection, and switching to the `Option<&T>` form is the idiomatic fix. The call site can `simply pass trim_where.as_ref() / trim_what.as_deref()`, so no behaviour changes.

core/src/translate/expr.rs
```diff
@@ -207,13 +228,21 @@ pub fn translate_expr(sql_expr: &SqlExpr) -> Result<Expr> {
...
             kind,
-        } => translate_cast(kind, expr, data_type, format.as_ref()),
+        } => translate_cast_with_params(params, kind, expr, data_type, format.as_ref()),
 
         _ => Err(TranslateError::UnsupportedExpr(sql_expr.to_string()).into()),
     }
 }
 
-pub fn translate_order_by_expr(sql_order_by_expr: &SqlOrderByExpr) -> Result<OrderByExpr> {
+pub fn translate_expr(sql_expr: &SqlExpr) -> Result<Expr> {
```
zmrdltl: it would be nice to add a simple docs which function returns `Result` For example:
```rust
/// Translates a [`SqlExpr`] into `GlueSQL's` [`Expr`] without any external parameters.
///
/// # Errors
///
/// Returns an error when the SQL expression cannot be converted, for example if it
/// references unsupported syntax or literals.
```

core/src/translate/expr.rs
```diff
@@ -226,7 +255,12 @@ pub fn translate_order_by_expr(sql_order_by_expr: &SqlOrderByExpr) -> Result<Ord
     }
 
     Ok(OrderByExpr {
-        expr: translate_expr(expr)?,
+        expr: translate_expr_with_params(expr, params)?,
         asc: *asc,
     })
 }
+
+pub fn translate_order_by_expr(sql_order_by_expr: &SqlOrderByExpr) -> Result<OrderByExpr> {
```
zmrdltl: it would be nice to add a simple docs which function returns `Result` For example:
```rust
/// Translates a [`SqlOrderByExpr`] into `GlueSQL's` [`OrderByExpr`] without parameters.
///
/// # Errors
///
/// Returns an error when the order-by expression uses unsupported syntax (for example,
/// NULLS FIRST/LAST) or otherwise cannot be converted.
```

core/src/translate/query.rs
```diff
@@ -46,39 +50,59 @@ pub fn translate_query(sql_query: &SqlQuery) -> Result<Query> {
...
-    let body = translate_set_expr(body)?;
-    let order_by = order_by
-        .iter()
-        .flat_map(|order_by| order_by.exprs.iter().map(translate_order_by_expr))
-        .collect::<Result<_>>()?;
-    let limit = limit.as_ref().map(translate_expr).transpose()?;
+    let body = translate_set_expr_with_params(body, params)?;
+    let mut order_by_exprs = Vec::new();
+    for clause in order_by.iter() {
+        for expr in clause.exprs.iter() {
```
zmrdltl: You can simplify this by iterating over the containers directly—e.g. `for clause in order_by { ... } (or for clause in &order_by)`. That’s the pattern Clippy’s `explicit_iter_loop` prefers and keeps the loop a bit cleaner.

core/src/translate/query.rs
```diff
@@ -155,29 +191,39 @@ pub fn translate_select_item(sql_select_item: &SqlSelectItem) -> Result<SelectIt
...
+            }),
         SqlSelectItem::QualifiedWildcard(object_name, _) => Ok(SelectItem::QualifiedWildcard(
             translate_object_name(object_name)?,
         )),
         SqlSelectItem::Wildcard(_) => Ok(SelectItem::Wildcard),
     }
 }
 
-fn translate_table_with_joins(sql_table_with_joins: &SqlTableWithJoins) -> Result<TableWithJoins> {
+pub fn translate_select_item(sql_select_item: &SqlSelectItem) -> Result<SelectItem> {
```
zmrdltl: It would be great to add a short doc comment here with a `# Errors` section explaining when this Result can fail.

core/src/translate/query.rs
```diff
@@ -46,39 +50,59 @@ pub fn translate_query(sql_query: &SqlQuery) -> Result<Query> {
...
         body,
-        order_by,
+        order_by: order_by_exprs,
         limit,
         offset,
     })
 }
 
-fn translate_set_expr(sql_set_expr: &SqlSetExpr) -> Result<SetExpr> {
+pub fn translate_query(sql_query: &SqlQuery) -> Result<Query> {
```
zmrdltl: It would be great to add a short doc comment here with a `# Errors` section explaining when this Result can fail.

core/src/glue.rs
```diff
@@ -24,22 +24,49 @@ impl<T: GStore + GStoreMut> Glue<T> {
         Self { storage }
     }
 
-    pub async fn plan<Sql: AsRef<str>>(&mut self, sql: Sql) -> Result<Vec<Statement>> {
+    pub async fn plan_with_params<Sql, I, P>(
```
zmrdltl, line 33: It would be nice to add a simple docs include `# Errors` section. For example:
```rust
/// Plans all statements in the SQL string using the supplied parameters.
///
/// # Errors
///
/// Returns a translate error when parsing or parameter conversion fails,
/// or a planning error if building the execution plan does not succeed.
```

core/src/glue.rs
```diff
@@ -24,22 +24,49 @@ impl<T: GStore + GStoreMut> Glue<T> {
...
         let storage = &self.storage;
         stream::iter(parsed)
-            .map(|p| translate(&p))
+            .map(|p| translate_with_params(&p, params.clone()))
             .then(|statement| async move { plan(storage, statement?).await })
             .try_collect()
             .await
     }
 
+    pub async fn plan<Sql: AsRef<str>>(&mut self, sql: Sql) -> Result<Vec<Statement>> {
```
zmrdltl, line 62: It would be nice to add a simple docs include `# Errors` section. For example:
```rust
/// Plans all statements in the SQL string without parameters.
///
/// # Errors
///
/// Returns a translate or plan error when converting the SQL or creating
/// execution plans fails.
```

core/src/glue.rs
```diff
@@ -48,4 +75,9 @@ impl<T: GStore + GStoreMut> Glue<T> {
 
         Ok(payloads)
     }
+
+    pub async fn execute<Sql: AsRef<str>>(&mut self, sql: Sql) -> Result<Vec<Payload>> {
```
zmrdltl, line 102: It would be nice to add a simple docs include `# Errors` section. For example:
```rust
/// Executes the statements in the SQL string without parameters and
/// returns their payloads.
///
/// # Errors
///
/// Fails when planning or executing any statement fails during translation
/// or execution.
```

core/src/glue.rs
```diff
@@ -24,22 +24,49 @@ impl<T: GStore + GStoreMut> Glue<T> {
...
+            .await
+    }
+
     pub async fn execute_stmt(&mut self, statement: &Statement) -> Result<Payload> {
         execute(&mut self.storage, statement).await
     }
 
-    pub async fn execute<Sql: AsRef<str>>(&mut self, sql: Sql) -> Result<Vec<Payload>> {
-        let statements = self.plan(sql).await?;
+    pub async fn execute_with_params<Sql, I, P>(
```
zmrdltl, line 77: It would be nice to add a simple docs include `# Errors` section. For example:
```rust
/// Executes the statements in the SQL string using the supplied parameters
/// and returns their payloads.
///
/// # Errors
///
/// Fails when planning or executing any statement fails, including
/// translation errors while binding parameters.
```

core/src/glue.rs
```diff
@@ -24,22 +24,49 @@ impl<T: GStore + GStoreMut> Glue<T> {
...
+    ) -> Result<Vec<Statement>>
+    where
+        Sql: AsRef<str>,
+        I: IntoIterator<Item = P>,
+        P: IntoParamLiteral,
+    {
         let parsed = parse(sql)?;
+        let params: Vec<ParamLiteral> = params
+            .into_iter()
+            .map(|value| value.into_param_literal())
```
zmrdltl: It might read a little more naturally to call the method directly:
```rust
.map(IntoParamLiteral::into_param_literal)
```

core/src/translate/ddl.rs
```diff
@@ -94,14 +99,26 @@ pub fn translate_column_def(sql_column_def: &SqlColumnDef) -> Result<ColumnDef>
     })
 }
 
-pub fn translate_operate_function_arg(arg: &SqlOperateFunctionArg) -> Result<OperateFunctionArg> {
+pub fn translate_column_def(sql_column_def: &SqlColumnDef) -> Result<ColumnDef> {
```
zmrdltl: It would be nice to add a simple docs include `# Errors` section which returns `Result`

core/src/translate.rs
```diff
@@ -30,11 +39,31 @@ use {
...
         TableConstraint as SqlTableConstraint, TableFactor, TableWithJoins,
     },
+    std::num::NonZeroUsize,
 };
 
 pub fn translate(sql_statement: &SqlStatement) -> Result<Statement> {
+    translate_with_params(sql_statement, std::iter::empty::<ParamLiteral>())
+}
+
+pub fn translate_with_params<I, P>(sql_statement: &SqlStatement, params: I) -> Result<Statement>
```
zmrdltl: It would be great to add a short doc comment here with a `# Errors` section explaining when this `Result` can fail.

core/src/translate.rs
```diff
@@ -380,10 +441,15 @@ pub fn translate_assignment(sql_assignment: &SqlAssignment) -> Result<Assignment
             .ok_or(TranslateError::UnreachableEmptyIdent)?
             .value
             .to_owned(),
-        value: translate_expr(value)?,
+        value: translate_expr_with_params(value, params)?,
     })
 }
 
+pub fn translate_assignment(sql_assignment: &SqlAssignment) -> Result<Assignment> {
```
zmrdltl: It would be great to add a short doc comment here with a `# Errors` section explaining when this `Result` can fail.

core/src/translate/param.rs
panarch: Thanks for the comment! Applied 👍

core/src/translate/param.rs
panarch: Applied!

core/src/translate/param.rs
panarch: Also thanks for the good comment!
Now I agree that it would be better to care about doc comments.

core/src/translate/param.rs
panarch: Thanks!
changed to
```rust
fn into_number_literal<T>(value: &T) -> Result<AstLiteral, TranslateError>
where
    T: ToString + ?Sized
```

core/src/translate/function.rs
panarch: Thanks! now it becomes better :)
```rust
    trim_where: Option<&SqlTrimWhereField>,
    trim_what: Option<&SqlExpr>,
```

core/src/translate/expr.rs
panarch: Applied!

core/src/translate/expr.rs
panarch: Applied!

core/src/translate/query.rs
panarch: there was also other point that I missed here.
`order_by` is `Option` so it was not even proper to use `iter`
now,
```rust
    if let Some(order_by) = order_by {
        for expr in &order_by.exprs {
            order_by_exprs.push(translate_order_by_expr(expr, params)?);
        }
    }
```

core/src/translate/query.rs
panarch: Applied!

core/src/translate/query.rs
panarch: Applied!

core/src/glue.rs
panarch: Applied!

core/src/glue.rs
panarch: Applied!

core/src/glue.rs
panarch: Applied!

core/src/glue.rs
panarch: Applied!

core/src/glue.rs
panarch: Thanks! that looks better :) applied!

core/src/translate/ddl.rs
panarch: Applied!

core/src/translate.rs
panarch: Applied!

core/src/translate.rs
panarch: Applied!

panarch:
@zmrdltl Thanks a lot for the careful reviews!

I've applied all the comments you left.
also another changes from the latest,

I removed all the `_with_params` suffix except for `translate` & `translate_with_params`.
That was added in this PR - getting a bit more verbose but adding less changes to existing codes.
However, most of internal functions in `translate` module are not supposed to user-side uses so it would be better to change at this time rather than leaving for the following works.

panarch:
@zmrdltl 
> Clippy is reporting the doc_markdown lint because the documentation comment spells out GlueSQL without backticks. Would you please wrap it as `GlueSQL` across the new doc comments that were added? That should resolve the lint

Could we apply extra clippy rules in following PRs?
I removed the `doc_markdown` allow option then it generates not only related to the current changes but also other clippy errors. It would be better to work extra rules that is currently not applied at this point separately.

@coderabbitai make the issue of applying `doc_markdown` and add this https://github.com/gluesql/gluesql/issues/1786 issue to related.

PR #1801, opened 2025-09-26
Add runtime Value support to Expr

panarch:
* Introduce `Expr::Value` so planner and runtime can carry values without literal conversions
* Add `Value::to_sql()` helper and reuse it in CLI dump output
* Cover the new serialization pathways with focused unit tests

## Testing
- cargo test --package gluesql-core scalar_values_to_sql
- cargo test --package gluesql-core complex_values_to_sql
- cargo test --package gluesql-cli format_dump_row_values
- cargo test
- cargo clippy --package gluesql-core --package gluesql-cli --all-targets -- -D warnings

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -14,7 +14,7 @@ use {
         stream::{StreamExt, TryStreamExt},
     },
     gluesql_core::{
-        ast::{Expr, ToSql},
+        ast::ToSql,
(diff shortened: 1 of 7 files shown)


panarch:
Closing in favor of #1860, which builds on the same idea and additionally removes redundant Literal variants.

cc. @paulolieuthier

PR #1807, opened 2025-09-28
Add JSON -> operator support and extend usage example

### LINKED ISSUE #1793: PostgreSQL-style JSON Functions and Operators
## 📋 Summary

Add PostgreSQL-compatible JSON operators and functions to GlueSQL so JSON columns can be queried with familiar syntax. 
The existing Value::Map / Value::List internal representation remains, optionally exposed via a DataType::Json alias.

### ✅ Supported Today

- File-based JSON storage and schemaless data flow
- Value::Map / Value::List, Value::selector, and the UNWRAP() function
- List/Map helper functions (ENTRIES, KEYS, VALUES, LENGTH, SLICE, etc.)
- Conversions to/from serde_json::Value (core/src/data/value/json.rs)

### ❌ Missing

- PostgreSQL JSON operators: ->, ->>, #>, #>>
- JSON helper functions (json_array_length, json_object_keys, json_extract_path, …)
- A declarative JSON type in SQL (alias vs new DataType::Json)

### 💡Motivation

- Align behavior with mainstream databases for easier adoption
- Provide intuitive syntax when exploring JSON columns
- Strengthen hybrid scenarios combining SQL with schemaless storage

## Phase 1 Operators

Operator | Description | Example
-- | -- | --
-> | Extract JSON field/element (JSON) | profile -> 'name' → "Alice"
->> | Extract JSON field/element (text) | profile ->> 'name' → Alice
#> | Extract via JSON path array (JSON) | profile #> '{hobbies,0}' → "reading"
#>> | Extract via path array (text) | profile #>> '{hobbies,0}' → reading 

### 1. JSON Creation Functions

| Function | Description | Example | Result |
|----------|-------------|---------|--------|
| `array_to_json(anyarray)` | Convert array to JSON | `array_to_json('{{1,5},{99,100}}'::int[])` | `[[1,5],[99,100]]` |
| `row_to_json(record)` | Convert record to JSON object | `row_to_json(row(1,'foo'))` | `{"f1":1,"f2":"foo"}` |
| `to_json(anyelement)` | Convert value to JSON | `to_json('Hello'::text)` | `"Hello"` |

### 2. JSON Query Functions

| Function | Description | Example | Result |
|----------|-------------|---------|--------|
| `json_array_length(json)` | Get JSON array length | `json_array_length('[1,2,3]')` | `3` |
| `json_extract_path(json, text[])` | Extract JSON at path | `json_extract_path('{"a":{"b":1}}','a')` | `{"b":1}` |
| `json_extract_path_text(json, text[])` | Extract as text | `json_extract_path_text('{"a":{"b":1}}','a','b')` | `1` |

### 3. JSON Decomposition Functions

| Function | Description | Example | Result |
|----------|-------------|---------|--------|
| `json_each(json)` | Expand object to key-value pairs | `json_each('{"a":"foo","b":"bar"}')` | `a \| "foo"`<br>`b \| "bar"` |
| `json_each_text(json)` | Expand object to text pairs | `json_each_text('{"a":"foo"}')` | `a \| foo` |
| `json_object_keys(json)` | Get object keys | `json_object_keys('{"a":1,"b":2}')` | `a`<br>`b` |
| `json_array_elements(json)` | Expand array to elements | `json_array_elements('[1,2,3]')` | `1`<br>`2`<br>`3` |

### 4. JSON Record Functions

| Function | Description | Example |
|----------|-------------|---------|
| `json_populate_record(base, json)` | JSON object → single row | `json_populate_record(null::x, '{"a":1,"b":2}')` |
| `json_populate_recordset(base, json)` | JSON array → multiple rows | `json_populate_recordset(null::x, '[{"a":1},{"a":2}]')` |

### Phase 1: Implement Operators ⚙️
- [ ] Add JSON type as a SQL keyword
- [x] Implement -> operator (extract JSON field)
- [ ] Implement ->> operator (extract JSON field as text)
- [ ] Implement #> operator (extract nested JSON by path)
- [ ] Implement #>> operator (extract nested value as text by path)

### Phase 2: Implement Creation Functions ➕
- [ ] to_json(): Convert any value to JSON
- [ ] array_to_json(): Convert a SQL array to a JSON array
- [ ] row_to_json(): Convert a SQL row to a JSON object
Phase 3: Implement Query Functions 🔍
- [ ] json_array_length(): Return the length of a JSON array
- [ ] json_object_keys(): Return all keys from a JSON object
- [ ] json_extract_path(): Extract value by path (as JSON or text)
Phase 4: Implement Decomposition & Expansion Functions ✨
- [ ] json_array_elements(): Expand an array into a set of rows
- [ ] json_each(): Expand an object into a set of key-value rows
- [ ] json_populate_recordset(): Expand a JSON array into a set of table rows

SteelCrab:
## Summary
  - add the `GetJSONArrow` binary operator so `->` survives parsing and is evaluated against map/list values
  - handle string and integer selectors in `Evaluated::get_json_arrow`, including null fallbacks when the base is not JSON
  - expand `pkg/rust/examples/json_arrow.rs` to cover integer, text, float, boolean, map, and null results returned by the operator

  ## Testing
  - cargo fmt --all
  - cargo run --example json_arrow

  Fixes #1793


core/src/executor/evaluate/evaluated.rs
```diff
@@ -279,6 +279,60 @@ impl<'a> Evaluated<'a> {
...
+        match result {
+            Ok(value) => Ok(Evaluated::Value(value)),
+            Err(err) => match err {
+                Error::Value(value_err) => match *value_err {
+                    ValueError::SelectorRequiresMapOrListTypes => Ok(Evaluated::Value(Value::Null)),
+                    other => Err(Error::Value(Box::new(other))),
+                },
+                other => Err(other),
+            },
+        }
```
panarch: It would be unnecessary to convert from errors to non errors.
Sometimes it is hard to force this kind of rules, but at least we can make it in our projects.
When it comes `ValueError::SelectorRequiresMapOrListTypes` then it would be enough.

core/src/executor/evaluate/evaluated.rs
```diff
@@ -279,6 +279,60 @@ impl<'a> Evaluated<'a> {
...
+        .ok_or_else(|| EvaluateError::FunctionRequiresIntegerOrStringValue("->".to_owned()))?;
+
+        let result = match self {
+            Evaluated::Value(base) => {
+                if base.is_null() {
+                    return Ok(Evaluated::Value(Value::Null));
+                }
+
+                base.selector(&key)
```
panarch: `selector` function is currently being used in `UNWRAP` function that was added when we support neither array index expression and the current arrow operator.

ref. https://github.com/gluesql/gluesql/blob/main/core/src/data/value/selector.rs

`selector` can handle multiple accessors but now sqlparser-rs supports well-structured expression in their AST, we no more need to use `Value::selector` function.
And even after we support full set of json accessor functions, then we will be able to remove the existing `selector`.

Here arrow expression only access 1-step field, it would be better not to use `selector`. then we don't need to use unnecessary convertions from numeric types to String above.

core/src/executor/evaluate/evaluated.rs
SteelCrab: Thank you for the feedback. 👍
I agree that converting errors to non-errors is unnecessary in most cases.

core/src/executor/evaluate/evaluated.rs
SteelCrab: I agree that for the current arrow expression, using selector is unnecessary. ☺️

core/src/executor/evaluate/evaluated.rs
SteelCrab:
``` rust 
        match result {
            Ok(value) => Ok(Evaluated::Value(value)),
            Err(Error::Value(value_err)) => {
                if value_err.as_ref() == &ValueError::SelectorRequiresMapOrListTypes {
                    Ok(Evaluated::Value(Value::Null))
                } else {
                    Err(Error::Value(value_err))
                }
            }
            Err(err) => Err(err),
        } 
```
I'll update the code like below so that only `SelectorRequiresMapOrListTypes` returns `Null` and other errors are kept as they are

test-suite/src/function/trunc.rs
```diff
@@ -1,46 +1,62 @@
 use {
     crate::*,
-    gluesql_core::{executor::EvaluateError, prelude::Value::*},
+    gluesql_core::{error::TranslateError, executor::EvaluateError, prelude::Value::*},
```
panarch: This modification does not look related to the current PR - arrow operator support.
It would be good to refactor existing tests, but we can try this in other branches.

core/src/executor/evaluate/expr.rs
```diff
@@ -68,6 +68,7 @@ pub fn binary_op<'a>(
         BinaryOperator::BitwiseAnd => l.bitwise_and(&r),
         BinaryOperator::BitwiseShiftLeft => l.bitwise_shift_left(&r),
         BinaryOperator::BitwiseShiftRight => l.bitwise_shift_right(&r),
+        BinaryOperator::Arrow => l.get_json_arrow(&r),
```
panarch: `BinaryOperator::Array`, then it would be enough to use `l.arrow` rather than `get_json_arrow`?

core/src/executor/evaluate/evaluated.rs
```diff
@@ -279,6 +327,51 @@ impl<'a> Evaluated<'a> {
...
+            }
+        };
+
+        match value_result {
+            Ok(value) => Ok(Evaluated::Value(value)),
+            Err(Error::Value(err))
+                if err.as_ref() == &ValueError::SelectorRequiresMapOrListTypes =>
+            {
+                Ok(Evaluated::Value(Value::Null))
+            }
```
panarch: it looks that still converting error to non-error exist.
may be I don't catch the need of this. is there any special reason for us to keep this?

test-suite/src/function/trunc.rs
SteelCrab: Got it, I’ll keep that in mind.

core/src/executor/evaluate/expr.rs
SteelCrab: Yes, you’re right. I’ll change it to use l.arrow(&r).

core/src/executor/evaluate/evaluated.rs
SteelCrab: @panarch Thank you for pointing this out. You are correct that this still converts an error into a non-error by returning Ok(Null). The reason this was originally kept is to maintain compatibility with the existing behavior, where a missing path was treated as NULL rather than an error.

Also, in PostgreSQL when you query a non-existent key like:

```sql
SELECT '{"a":1}'::json->'b';
```

the result is not an error but simply NULL. Our current logic was designed to reflect this same semantics.

panarch: I agree that it would be good to return NULL than error for `SELECT '{"a":1}'::json->'b';`.

However,

```rust
Err(Error::Value(err))
    if err.as_ref() == &ValueError::SelectorRequiresMapOrListTypes =>
```

1. `ValueError::SelectorRequiresMapOrListTypes` does not look related to this case.
2. that error is also that we make it so could we find a better way not coverting from error to NULL?

SteelCrab: You are right.
Converting the `SelectorRequiresMapOrListTypes` error into `NULL` is unnecessary, and simplifying the `->` function to just propagate errors directly is the correct approach.

```rust
match value_result {
    Ok(value) => Ok(Evaluated::Value(value)),
    Err(err) => Err(err),
}
```

Now, for invalid types, a clear error is returned instead of `NULL`:

```sql
SELECT 1 -> 'a';
SELECT TRUE -> 'a';

[Error] ValueError: Selector requires MAP or LIST types
```

I’ve reimplemented it so that the error output appears as expected, and I’ll push the changes.

panarch:
- `core/src/executor/evaluate/evaluated.rs:111`
  At the moment both “base is not MAP/LIST” and “selector isn’t string/integer” flow into `ValueError::SelectorRequiresMapOrListTypes`. It would be helpful to split those cases into two arrow-specific `EvaluateError` variants (e.g. `ArrowBaseRequiresMapOrList` and `ArrowSelectorRequiresIntegerOrString`) so the messaging stays accurate and we avoid leaning on the legacy value-level error.
- `core/src/executor/evaluate/evaluated.rs:92`
  `select_arrow_value` currently lives at the top of `evaluated.rs`, while the rest of the helpers sit in `core/src/executor/evaluate/function.rs`. It would align better with the existing layout if the arrow helper lived there and `Evaluated::arrow` simply delegated to it.
- `test-suite/src/function/trunc.rs:1`
  The TRUNC test updates still appear out of scope for this PR. It would be great to move those into a separate PR so this one stays focused on the arrow operator.

SteelCrab:
Thank you for the review. I have read through all of your comments and will proceed with the updates. I will be sure to ask if anything is unclear.

core/src/executor/evaluate/evaluated.rs
```diff
@@ -279,6 +327,46 @@ impl<'a> Evaluated<'a> {
...
+                | Value::U16(_)
+                | Value::U32(_)
+                | Value::U64(_)
+                | Value::U128(_)
+        ) {
+            return Err(
+                EvaluateError::FunctionRequiresIntegerOrStringValue("->".to_owned()).into(),
+            );
+        }
+
```
SteelCrab, line 355: test

core/src/executor/evaluate/evaluated.rs
```diff
@@ -279,6 +327,46 @@ impl<'a> Evaluated<'a> {
         )
     }
 
+    pub fn arrow<'b>(&'a self, other: &Evaluated<'b>) -> Result<Evaluated<'b>> {
```
SteelCrab, line 330: fn arrow :

core/src/executor/evaluate/function.rs
```diff
@@ -1102,3 +1103,63 @@ pub fn dedup<'a>(list: Evaluated<'_>) -> ControlFlow<Evaluated<'a>> {
...
+                Value::I128(value) => Cow::Owned(value.to_string()),
+                Value::U8(value) => Cow::Owned(value.to_string()),
+                Value::U16(value) => Cow::Owned(value.to_string()),
+                Value::U32(value) => Cow::Owned(value.to_string()),
+                Value::U64(value) => Cow::Owned(value.to_string()),
+                Value::U128(value) => Cow::Owned(value.to_string()),
+                _ => {
+                    return Err(EvaluateError::ArrowSelectorRequiresIntegerOrString(format!(
+                        "{:?}",
+                        selector
```
zmrdltl: Could we simplify this to `"{selector:?}?"`

core/src/executor/evaluate/evaluated.rs
```diff
@@ -279,6 +279,48 @@ impl<'a> Evaluated<'a> {
...
+            .into());
+        }
+
+        let value_result = match self {
+            Evaluated::Value(base) => function::select_arrow_value(base, &selector),
+            _ => {
+                let base = Value::try_from(self.clone())?;
+                function::select_arrow_value(&base, &selector)
+            }
+        };
```
zmrdltl: It looks like this could use if let instead of a single-pattern match:
```rust
 let value_result = if let Evaluated::Value(base) = self {
    function::select_arrow_value(base, &selector)
} else {
    let base = Value::try_from(self.clone())?;
    function::select_arrow_value(&base, &selector)
};
```

core/src/executor/evaluate/function.rs
```diff
@@ -1102,3 +1103,63 @@ pub fn dedup<'a>(list: Evaluated<'_>) -> ControlFlow<Evaluated<'a>> {
...
+                Value::I128(value) => usize::try_from(*value).ok(),
+                Value::U8(value) => Some(*value as usize),
+                Value::U16(value) => Some(*value as usize),
+                Value::U32(value) => usize::try_from(*value).ok(),
+                Value::U64(value) => usize::try_from(*value).ok(),
+                Value::U128(value) => usize::try_from(*value).ok(),
+                _ => {
+                    return Err(EvaluateError::ArrowSelectorRequiresIntegerOrString(format!(
+                        "{:?}",
+                        selector
```
zmrdltl: same pattern as https://github.com/gluesql/gluesql/pull/1807#discussion_r2415346280

test-suite/src/expr/arrow.rs
```diff
@@ -0,0 +1,226 @@
...
+        ("UINT16", "CAST(3 AS UINT16)"),
+        ("UINT32", "CAST(3 AS UINT32)"),
+        ("UINT64", "CAST(3 AS UINT64)"),
+        ("UINT128", "CAST(3 AS UINT128)"),
+    ];
+
+    for (label, selector_expr) in typed_selectors {
+        let sql = format!(
+            "SELECT array->{} AS result FROM ArrowSample;",
+            selector_expr
```
zmrdltl: same pattern as https://github.com/gluesql/gluesql/pull/1807#discussion_r2415346280

test-suite/src/expr/arrow.rs
```diff
@@ -0,0 +1,226 @@
...
+        ("UINT16", "CAST(1 AS UINT16)"),
+        ("UINT32", "CAST(1 AS UINT32)"),
+        ("UINT64", "CAST(1 AS UINT64)"),
+        ("UINT128", "CAST(1 AS UINT128)"),
+    ];
+
+    for (label, selector_expr) in map_typed_selectors {
+        let sql = format!(
+            "SELECT object->{} AS result FROM ArrowSample;",
+            selector_expr
```
zmrdltl: same pattern as https://github.com/gluesql/gluesql/pull/1807#discussion_r2415346280

test-suite/src/expr/arrow.rs
```diff
@@ -0,0 +1,226 @@
...
+            '[1,"two",true,4.25,null]'
+        );
+        "#,
+    )
+    .await;
+
+    let nested_map = match Value::parse_json_map(r#"{"role":"admin"}"#).unwrap() {
+        Value::Map(map) => map,
+        _ => unreachable!(),
+    };
```
zmrdltl: To tighten things up a little, you could take advantage of let-else for the JSON parse:
```rust
let Value::Map(nested_map) = Value::parse_json_map(r#"{"role":"admin"}"#).unwrap() else {
    unreachable!()
};
```
Same behaviour, just a bit more direct for the “this must be a Map” case.

test-suite/src/expr/arrow.rs
```diff
@@ -0,0 +1,226 @@
+use {
+    crate::*,
```
zmrdltl: Could we replace the crate::* glob with an explicit list? Something like:
```rust
use crate::{row, select, select_with_null, stringify_label, test_case};
```
It keeps the imports clear in the review and avoids pulling in more than we actually use.

core/src/executor/evaluate/evaluated.rs
SteelCrab: Thanks for the suggestion! You're right, this is much more readable and concise. I'll apply it right away.  :)

core/src/executor/evaluate/function.rs
zmrdltl: This part does not appear to have been reflected yet.

core/src/executor/evaluate/evaluated.rs
```diff
@@ -279,6 +279,44 @@ impl<'a> Evaluated<'a> {
...
+                | Value::I128(_)
+                | Value::U8(_)
+                | Value::U16(_)
+                | Value::U32(_)
+                | Value::U64(_)
+                | Value::U128(_)
+        ) {
+            return Err(EvaluateError::ArrowSelectorRequiresIntegerOrString(format!(
+                "{:?}",
+                selector
```
zmrdltl: same pattern as https://github.com/gluesql/gluesql/pull/1807#discussion_r2415346280

core/src/executor/evaluate/function.rs
```diff
@@ -8,6 +8,7 @@ use {
     chrono::{Datelike, Duration, Months},
     md5::{Digest, Md5},
     rand::{Rng, SeedableRng, rngs::StdRng},
+    std::borrow::Cow,
```
zmrdltl: To stay consistent with the rest of the tree we typically collapse duplicate `std::…` imports into a single grouped statement (e.g. `use std::{borrow::Cow, …};`). Could we merge the new `std` line into the existing grouped import to match that style?

One small addition: if we address this spot and the nit here (https://github.com/gluesql/gluesql/pull/1807#discussion_r2415346280), that should wrap things up.

panarch (approved):
Looks also good to me! Thanks a lot for the contribution 👍 👍

PR #1811, opened 2025-09-29
Pass PR number via coverage artifact

panarch:
## Summary
Fork-originated Coverage runs expose an empty `pull_requests` array in the `workflow_run` payload, so Publish Coverage could not recover the PR number and failed when writing artifacts and comments. Persist the PR number inside the coverage artifact and let the publish job read it back so the workflow succeeds regardless of where the PR branch lives.

- record the pull request number in the coverage artifact when the Coverage workflow runs for PRs
- have the Publish Coverage workflow read that value from the downloaded artifact instead of assuming the payload contains PR metadata

## Testing
- cargo fmt --all

## Summary by CodeRabbit

* **Chores**
  * Improved coverage reporting pipeline in CI by recording the pull request number during PR runs and attaching it to the coverage artifact.
  * Updated the publish step to read the PR number from the artifact for workflow_run events, ensuring accurate association with the correct PR.
  * Now uploads the entire coverage directory as an artifact, enhancing reliability and diagnostics for coverage publishing.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -104,6 +104,9 @@ jobs:
         run: |
           xz -T0 -9e -c coverage/lcov.info > coverage/lcov.info.xz
           rm -f coverage/lcov.info
+      - name: Record PR number
+        if: github.event_name == 'pull_request'
+        run: echo "${{ github.event.pull_request.number }}" > coverage/pr-number
       - name: Set timestamp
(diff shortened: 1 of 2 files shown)


zmrdltl (approved):
I’ve looked over the coverage/publish workflow changes and they look good to me. Once this merges, we can nudge any older PRs to rebase (or push a fresh commit) so that their rerun Coverage job writes the new pr-number artifact; after that, publish-coverage will pick up the PR number and post the coverage report normally. From my side this is ready to go—thanks!

PR #1813, opened 2025-09-30
Add comprehensive unit tests for utils indexmap and vector modules

panarch:
## Summary
- add comprehensive unit tests for the utils indexmap wrapper covering creation, access, and iteration order
- expand the utils vector tests to exercise all mutation, query, and conversion APIs

## Testing
- cargo fmt --all
- cargo test -p gluesql-utils

------
https://chatgpt.com/codex/tasks/task_e_68db6159f53c832a880bdd8224def72c

### DIFF
--- a/utils/src/indexmap.rs
+++ b/utils/src/indexmap.rs
@@ -50,3 +50,49 @@ impl<K: Hash + Eq, V> IntoIterator for IndexMap<K, V> {
         self.0.into_iter()
     }
 }
+
+#[cfg(test)]
+mod tests {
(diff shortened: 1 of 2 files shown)


panarch:
> This utility will still panic on out-of-bounds indexes (inherited from Vec), though we don’t expose that through SQL. Optional nit: document or unit-test that behaviour if you think it helps future maintainers. Approved.

we might no longer need this utils after once we finished refactoring aggregate module.
but still this is being actively used, so..

PR #1821, opened 2025-10-14
Add swift bindings and SPM compatible prebuilt binaries.

autlyr21:
- Add basic swift bindings.
- Add tests for binding code.
- Add SPM compatible prebuilt binaries for IOS(v18+).

## Summary by CodeRabbit

- **New Features**
  - Adds a Swift package and iOS XCFramework for SwiftPM integration.
  - Exposes a public Swift API (GlueProtocol / Glue) to initialize DB backends, run async queries, and inspect storage kind.
  - Returns rich, strongly-typed results (Value, Payload, ColumnMeta, etc.) with helpers for Foundation interoperability.

- **Chores**
  - Adds scripts to build, generate bindings, and package the iOS XCFramework.
  - Expands workspace configuration to include the new Swift package.

### DIFF
--- a/pkg/swift/Cargo.toml
+++ b/pkg/swift/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "gluesql_swift"
+name = "gluesql-swift"
 version.workspace = true
 edition.workspace = true
 description.workspace = true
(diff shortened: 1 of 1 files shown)


pkg/swift/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "gluesql_swift"
```
panarch: let's use dash instead of underscore, both are allowed but we prefer `-` in our projects.
`gluesql-swift`

PR #1825, opened 2025-10-18
Allow storages to customize query planners

panarch:
## Summary
- introduce a `core::store::Planner` trait so each storage can own its query-planning pipeline
- replace the former global `plan` helper with direct calls into the storage-provided planner across Glue, bindings, and tests
- extract sled-specific planning logic into `storages/sled-storage::planner`, where it layers `plan_index` on top of the shared default passes

## Details
- `core/src/store/planner.rs` now hosts the default planner implementation (schema fetch → validation → primary-key + join passes); storages may override it to push down additional operations
- every built-in storage implements `Planner`; storages that only need the default behaviour simply derive it, while sled storage demonstrates how to extend the flow with index pushdown
- custom storage authors can now ship planner logic alongside `Store` / `StoreMut`, making storage-aware pushdown strategies (e.g. SQL engine translation) possible without touching the executor layer

## Summary by CodeRabbit

* **Refactor**
  * Query planning is now performed via a storage-backed planner interface (planning routed through each storage).
  * The public prelude no longer exposes the old standalone plan entrypoint.
* **New Features**
  * Many storage backends (memory, file, CSV, Sled, Parquet, Redis, Mongo, Web, Git, IDB, JSON, Redb, Composite, etc.) now implement the planner interface.
  * Mock and test tooling updated to support storage-backed planning.
* **API**
  * Schema fetching now accepts dynamically sized storage references, improving flexibility for custom/dynamic storages.

### DIFF
--- a/core/src/store/planner.rs
+++ b/core/src/store/planner.rs
@@ -9,7 +9,7 @@ use {
 };
 
 #[async_trait]
-pub trait Planner: Store + Sized {
+pub trait Planner: Store {
     async fn plan(&self, statement: Statement) -> Result<Statement> {
         let schema_map = fetch_schema_map(self, &statement).await?;
         validate(&schema_map, &statement)?;
--- a/core/src/plan/schema.rs
+++ b/core/src/plan/schema.rs
@@ -14,7 +14,7 @@ use {
     std::collections::HashMap,
 };
 
-pub async fn fetch_schema_map<T: Store>(
+pub async fn fetch_schema_map<T: Store + ?Sized>(
     storage: &T,
     statement: &Statement,
 ) -> Result<HashMap<String, Schema>> {
@@ -63,7 +63,10 @@ pub async fn fetch_schema_map<T: Store>(
     }
 }
 
-async fn scan_query<T: Store>(storage: &T, query: &Query) -> Result<HashMap<String, Schema>> {
+async fn scan_query<T: Store + ?Sized>(
+    storage: &T,
+    query: &Query,
+) -> Result<HashMap<String, Schema>> {
     let Query {
         body,
         limit,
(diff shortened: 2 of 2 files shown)


core/src/store/planner.rs
```diff
@@ -0,0 +1,22 @@
...
+pub trait Planner: Store + Sized {
+    async fn plan(&self, statement: Statement) -> Result<Statement> {
+        let schema_map = fetch_schema_map(self, &statement).await?;
+        validate(&schema_map, &statement)?;
+
+        let statement = plan_primary_key(&schema_map, statement);
+        let statement = plan_join(&schema_map, statement);
+
+        Ok(statement)
+    }
```
zmrdltl, line 21: Could we restore the `plan_index` call in the default planner implementation? Storages that simply `impl Planner for Foo {}`—CompositeStorage, FileStorage, JsonStorage, WebStorage, etc.—used to inherit index planning automatically, but with the current default they stop attaching secondary-index metadata, so the executor never calls `scan_indexed_data`. Only SledStorage overrides plan, which is why it still works. If we keep `plan_index` in the shared default (and let specialized planners override or extend it), every storage that relies on the default path will retain its existing index behaviour.

core/src/store/planner.rs
```diff
@@ -0,0 +1,22 @@
...
+    crate::{
+        ast::Statement,
+        plan::{fetch_schema_map, plan_join, plan_primary_key, validate},
+        result::Result,
+    },
+    async_trait::async_trait,
+};
+
+#[async_trait]
+pub trait Planner: Store + Sized {
```
zmrdltl: Would it be possible to keep `Planner` object-safe? The new `Self: Sized` bound means we can no longer use trait objects such as `Box<dyn GStore + GStoreMut>` when wiring storages into Glue. That pattern has worked up to now and is fairly common when GlueSQL sits behind a runtime abstraction layer, so the new constraint effectively breaks downstream consumers. If we can drop the `Sized` requirement (or perhaps add an object-safe shim), those existing integrations should continue to compile.

core/src/store/planner.rs
panarch: This is intentional. currently only SledStorage has Index & IndexMut supports so we don't need `plan_index` for all other storages.

PR #1831, opened 2025-10-26
Add LongArrow (->> ) operator for JSON text extraction

SteelCrab:
### Summary

- Add the LongArrow binary operator so ->> survives parsing and is evaluated against map/list values as text

- Implement Evaluated::long_arrow method that extracts JSON values and converts them to text representation

- Add select_long_arrow_value function in executor that wraps select_arrow_value and converts results to string

- Handle string and integer selectors, including null fallbacks when the base is not JSON or selector is missing

### Example
- Extract object field as text
`SELECT object->>'name' FROM table;  -- Returns "Han" (already string, unchanged)`

- Extract array element as text  
`SELECT array->>0 FROM table;  -- Returns "I64(1)" (number as text)`

- Missing keys/indices return NULL
`SELECT object->>'missing' FROM table;  -- Returns NULL`

### Testing
`cargo fmt --all`
`cargo clippy --all-targets -- -D warnings`
`cargo test -p gluesql-core`
`cargo test --test memory_storage`

## Summary by CodeRabbit

* **New Features**
  * Added the ->> operator to extract fields from JSON objects and elements from arrays, returning results as strings.
  * Preserves NULL propagation and yields stringified representations for non-string results.

* **Tests**
  * Comprehensive test coverage for object/array extraction, typed selectors, nulls, missing keys, and error cases.


SteelCrab:
@panarch Please review the PR. Thank you.

core/src/ast/operator.rs
```diff
@@ -327,6 +329,26 @@ mod tests {
...
+
+        assert_eq!(
+            r#""data" -> 'key'"#,
+            &Expr::BinaryOp {
+                left: Box::new(Expr::Identifier("data".to_owned())),
+                op: BinaryOperator::Arrow,
+                right: Box::new(Expr::Literal(AstLiteral::QuotedString("key".to_owned())))
+            }
+            .to_sql()
+        );
```
panarch: This does not look related to this PR. if `BinaryOperator::Arrow` unit test is missing then we can make a new PR for this.

core/src/executor/evaluate/evaluated.rs
```diff
@@ -296,6 +296,23 @@ impl<'a> Evaluated<'a> {
...
 
+    pub fn long_arrow<'b>(&'a self, other: &Evaluated<'b>) -> Result<Evaluated<'b>> {
+        let selector = Value::try_from(other.clone())?;
+
+        if selector.is_null() {
+            return Ok(Evaluated::Value(Value::Null));
+        }
+
+        let value_result = if let Evaluated::Value(base) = self {
+            function::select_long_arrow_value(base, &selector)
```
panarch: `select_long_arrow_value` looks so simple that it would be enough to be inside of the `Evaluated::long_arrow`. could we merge this?

test-suite/src/expr/long_arrow.rs
```diff
@@ -0,0 +1,224 @@
+use {
+    crate::{row, select, select_with_null, stringify_label, test_case},
+    gluesql_core::{
+        error::TranslateError,
+        executor::EvaluateError,
+        prelude::Value::{self, Str},
+    },
+};
+
+test_case!(long_arrow, {
```
panarch: test case in our `test-suite` has two roles
* for test
* doc for users - provide how to use the function or expression.

could we make the test case a bit better in the matter of documentation purpose?
it would be nice for this test to work as a guide doc for users to figure out how to use `->>`

PR #1837, opened 2025-11-03
Apply `clippy::needless_pass_by_value` clippy rule

moreal:
This pull request applies only `clippy::needless_pass_by_value`[^1] because the changes are already huge. Most changes convert parameter types to references when the value is not consumed in the body, but I made `ShowOption` in *cli/src/command.rs* implement `Copy` (and `Clone`) because the type's size seemed small enough to be a copy rather than a reference.

```rust
pub enum ShowOption {
    Tabular,
    Colsep,
    Colwrap,
    Heading,
    All,
}
```

[^1]: https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_value

## Summary by CodeRabbit

* **Tests**
  * Updated internal test infrastructure to use borrowed references for improved efficiency.

* **Chores**
  * Refactored function signatures to use borrowed references and string slices for memory optimization.
  * Updated compiler lint configuration.
  * Enhanced enum traits for internal operations.


cli/src/command.rs
```diff
@@ -83,7 +83,7 @@ impl SetOption {
     }
 }
 
-#[derive(Eq, Debug, PartialEq)]
+#[derive(Eq, Debug, PartialEq, Copy, Clone)]
```
panarch, line 86: I agree that for `ShowOption` it is better to add `Copy`. that's nice!

panarch (approved):
Wow that's so nice that clippy can even suggest like these.
Decision you made for adding `Copy` to `ShowOption` looks also very good.
Thanks for the contribution 👍 👍

PR #1858, opened 2025-12-07
Add long arrow operator (->>) support

SteelCrab:
## Summary
  Implements the long arrow operator (`->>`) for extracting JSON values as text, similar to
  PostgreSQL's `->>` operator.

  - Add `BinaryOperator::LongArrow` variant to AST
  - Implement `select_long_arrow_value` function that converts values to text
  - Add operator translation and evaluation dispatch
  - Add comprehensive test cases for Map/List text extraction

  ## Changes
  - `core/src/ast/operator.rs` - Add LongArrow operator variant
  - `core/src/executor/evaluate/` - Implement evaluation logic
  - `core/src/translate/operator.rs` - Add operator translation
  - `test-suite/src/expr/arrow.rs` - Add test cases

  ## Example
  ```sql
  SELECT object->>'name' FROM table;  -- Returns "value" as text
  SELECT array->>0 FROM table;        -- Returns first element as text

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added the long-arrow operator (`->>`) for extracting map and list values as text in queries.
  * Supports nested access, numeric and typed selectors, chaining, and null preservation.
  * Converts non-text results to text automatically.

* **Query Support**
  * Query translation now recognizes the long-arrow operator alongside existing arrow operators.

* **Tests**
  * Added coverage for extraction, chaining, null and invalid selections, typed selectors, and error handling.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


test-suite/src/expr/arrow.rs
```diff
@@ -216,4 +216,71 @@ test_case!(arrow, {
         g.named_test(&test_name, sql.as_str(), Ok(select!(result F64; 4.25_f64)))
             .await;
     }
+
+    // LongArrow (->>): Extract as text
```
panarch: it is different expression so it would be better to have different test case like `long_arrow` rather than adding these to `arrow`.

SteelCrab:
@coderabbitai review

SteelCrab:
i've met all the requirements for Codecov. 
Please review. :) @panarch

core/src/executor/evaluate/evaluated.rs
```diff
@@ -676,6 +693,38 @@ mod tests {
...
+    fn try_from_evaluated_value_to_bool() {
+        assert_eq!(
+            bool::try_from(Evaluated::Value(Cow::Owned(Value::Bool(true)))),
+            Ok(true),
+        );
+        assert_eq!(
+            bool::try_from(Evaluated::Value(Cow::Owned(Value::I64(42)))),
+            Err(EvaluateError::BooleanTypeRequired("42".to_owned()).into()),
+        );
+    }
```
panarch: `try_from_evaluated_value_to_bool` test seems unrelated to the long arrow operator feature.
Is there a specific reason it's included here? If not, could you move it to a separate PR to keep this PR focused on `->>` support?

core/src/executor/evaluate/evaluated.rs
```diff
@@ -168,6 +168,23 @@ impl<'a> Evaluated<'a> {
         value_result.map(|v| Evaluated::Value(Cow::Owned(v)))
     }
 
+    pub fn long_arrow<'b>(&'a self, other: &Evaluated<'b>) -> Result<Evaluated<'b>> {
+        let selector = Value::try_from(other.clone())?;
+
+        if selector.is_null() {
+            return Ok(Evaluated::Value(Cow::Owned(Value::Null)));
+        }
```
panarch: The null selector check here is unreachable. binary_op in `core/src/executor/evaluate/expr.rs` already handles null propagation for all binary operators before dispatching:

```rust
if l.is_null() || r.is_null() {
    return Ok(Evaluated::Value(Cow::Owned(Value::Null)));
}
```

So long_arrow is guaranteed to receive non-null operands. The existing arrow method has the same redundancy, but let's not carry it over - please remove the null check.

test-suite/src/expr/long_arrow.rs
```diff
@@ -0,0 +1,336 @@
+use {
```
panarch: The tests are too verbose and hard to scan as a feature reference. The `->>` operator is essentially `->` + convert to text, so tests should focus on that text conversion behavior rather than re-verifying logic already covered by select_arrow_value.

Some overlap with existing tests is fine if it helps readers understand the feature, but in that case it should be concise and readable - not exhaustive repetition like testing every integer type for CAST selectors.

A few suggestions:

1. Please consolidate into a single `test_case!`. There's no strong reason to split into 6 separate test cases for this feature.
2. Use `g.named_test` with descriptive names and leverage macros like `select!` / `select_with_null!` tabular format for readability:

```rust
  select_with_null!(
      id     | num    | name;
      Null     I64(2)   Str("Hello".to_owned());
      I64(1)   I64(9)   Str("World".to_owned())
  ),
```

3. One or two CAST cases are enough for typed selectors.

Tests should work as readable documentation for the feature. Right now the volume makes it hard to see what `->>` actually does at a glance.

panarch (changes requested):

panarch (approved):
Looks all nice!

PR #1862, opened 2025-12-13
Relax lifetime constraint on TableNameNode::delete

### LINKED ISSUE #1861: Overly restrictive lifetime bounds on `TableNameNode<'a>::delete()`
Thanks for building `gluesql`! I'm really enjoying hacking with it.

Today, `ast_builder::table("foo").delete()` is difficult to use, as it returns a `DeleteNode<'static>` -- this winds up requiring that all filter criteria passed in is also `'static`, making it very painful to work with.

I'm new to the `gluesql` codebase, but from what I can tell, this should be safe to relax to `DeleteNode<'a>`. I've done some testing on a project I'm working on via cargo patching, and it seems to work fine, as well.

If that sounds reasonable, I'm happy to send a patch over!

torrancew:
This commit relaxes the lifetime constraint on the value returned by TableNameNode::delete -- prior to this change, the return value was bound to `'static`, which makes using it very challenging/unergonomic.

After this change, the lifetime is bound to `'a`, similar to its peer functions.

This fixes #1861 

## Summary by CodeRabbit

* **Refactor**
  * Optimized internal data structure handling to improve memory management and type safety.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/core/src/ast_builder/table_name.rs
+++ b/core/src/ast_builder/table_name.rs
@@ -20,7 +20,7 @@ impl<'a> TableNameNode {
         SelectNode::new(table_factor)
     }
 
-    pub fn delete(self) -> DeleteNode<'static> {
+    pub fn delete(self) -> DeleteNode<'a> {
         DeleteNode::new(self.table_name)
     }
 


panarch (approved):
Looks nice! Thanks a lot for the catching this 👍

PR #1895, opened 2026-04-25
Bind aggregate slots during planning

panarch:
## Summary

This moves aggregate slot assignment from execution-time discovery into the planner.

Execution now evaluates aggregates through planner-assigned slots stored on aggregate expressions and the select-level aggregate slot table. This avoids runtime aggregate lookup by expression hash and makes aggregate binding an explicit planning requirement.

## Changes

- Add `plan_aggregate` to assign aggregate slots before execution.
- Bind aggregate slots for `SELECT`, `INSERT ... SELECT`, `CREATE TABLE ... AS SELECT`, and aggregate subqueries inside `UPDATE` / `DELETE` expressions.
- Update aggregate execution to read aggregate results by slot instead of runtime aggregate maps.
- Preserve AST semantic equality/hash behavior by ignoring aggregate planning metadata.
- Support empty global aggregate results such as `SELECT COUNT(*) FROM empty_table`.
- Add planner/unit and test-suite coverage for aggregate slot binding, CTAS/INSERT aggregate sources, and empty aggregate input.

## Summary by CodeRabbit

* **New Features**
  * Added planning for aggregate expressions so queries with aggregates are handled consistently.

* **Bug Fixes**
  * CREATE TABLE AS SELECT with aggregates now behaves correctly.
  * COUNT(*) on empty tables returns 0.
  * Aggregate expressions inside scalar functions, HAVING, and predicates evaluate correctly.
  * Query planning now preserves aggregate metadata across transformation steps.

* **Tests**
  * Expanded aggregate test coverage and new integration cases for CTAS, inserts, and edge conditions.


core/src/executor/alter/table.rs
```diff
@@ -196,18 +209,58 @@ pub async fn create_table<T: GStore + GStoreMut>(
...
+            };
 
             storage.append_data(target_table_name, rows).await
         }
         None => Ok(()),
     }
 }
 
+fn can_copy_source_schema(select: &Select) -> bool {
+    match &select.projection {
```
devgony, line 234: `can_copy_source_schema` needs to break if there are joins in SQL.

```text
gluesql> CREATE TABLE Authors (id INTEGER, name TEXT);
CREATE TABLE Books (id INTEGER, author_id INTEGER, title TEXT);
INSERT INTO Authors VALUES (1, 'Asimov'), (2, 'Clarke');
INSERT INTO Books VALUES (10, 1, 'Foundation'), (20, 2, 'Rendezvous');
Table created

Table created

2 rows inserted

2 rows inserted

gluesql> CREATE TABLE Report AS SELECT * FROM Authors JOIN Books ON Authors.id = Books.author_id;
Table created

gluesql> select * from Report;
```
- Wrong schema
| id | name   |
|----|--------|
| 1  | Asimov |
| 2  | Clarke |

```suggestion
fn can_copy_source_schema(select: &Select) -> bool {
    if !select.from.joins.is_empty() {
        return false;
    }
    match &select.projection {
```

devgony:
```text
gluesql> CREATE TABLE Empty (id INTEGER, name TEXT);
Table created

gluesql> SELECT COUNT(*), name FROM Empty;
[error] evaluate: context is required for identifier evaluation: "name"
```

Usually the error should be 
> SQL Error [42803]: ERROR: column "Empty.name" must appear in the GROUP BY clause or be used in an aggregate function
  Position: 18
Error position: line: 8 pos: 17
- PG example

panarch:
Opened #1896 to track aggregate semantic validation separately.

The current PR focuses on aggregate slot binding and execution cleanup. Invalid aggregate queries like:

```sql
CREATE TABLE Empty (id INTEGER, name TEXT);
SELECT COUNT(*), name FROM Empty;
```

currently fall through to an evaluator context error. That should be handled by planner/validation as a follow-up rather than folded into this slot-binding change.

devgony (approved):
👍🏼 👍🏼 👍🏼

PR #1899, opened 2026-04-25
feat: support `UNION`

### LINKED ISSUE #531: add ability to union two select statements together
Add the capability to union two select statements together:

```sql
gluesql> CREATE TABLE mytable (id INT);
gluesql> INSERT INTO mytable VALUES (1);
1 row inserted

gluesql> SELECT * FROM mytable UNION SELECT * FROM mytable;
[error] unsupported query set expr: SELECT * FROM mytable UNION SELECT * FROM mytable
```

ever0de:
Introduce SetExpr::Union to the GlueSQL AST and wire it through every layer
that inspects or builds query bodies.

Translation layer parses sqlparser SetOperation nodes with Union operator into
SetExpr::Union, carrying the all flag to distinguish UNION from UNION ALL.
Unsupported set operators (EXCEPT, INTERSECT) still return
UnsupportedQuerySetExpr.

Executor resolves UNION early in select_with_labels by recursively evaluating
both sides, collecting results into an in-memory Vec, applying deduplication for
UNION (distinct) or skipping it for UNION ALL, relabeling all result rows with
the left-side column labels so ORDER BY references resolve correctly, and then
applying sort and limit before returning a unified stream.

Planner passes (aggregate, index, join, primary-key, schema, schemaless, and
validate) now handle SetExpr::Union by visiting both branches so planning
transforms apply recursively to each sub-query.

INSERT and CTAS paths treat a Union body the same as a Select body, delegating
to select_with_labels for row and column-definition resolution.

## Summary by CodeRabbit

* **New Features**
  * Full support for UNION and UNION ALL in queries, subqueries, and CREATE TABLE AS SELECT; UNION deduplicates, UNION ALL preserves duplicates.
  * ORDER BY, LIMIT, and OFFSET are applied correctly to UNION results; mismatched column counts produce a clear validation error.

* **Bug Fixes**
  * UNION queries now behave correctly across schema inference, inserts, and fetches (including schemaless data).

* **Tests**
  * Added comprehensive unit and integration tests covering UNION/UNION ALL, ordering, CTAS, schemaless cases, and error paths.


panarch:
I think the core AST shape is a good fit here: `SetExpr::Union { left: Box<SetExpr>, right: Box<SetExpr>, all }` matches the supported SQL model, where UNION operands are set expressions and `ORDER BY` / `LIMIT` / `OFFSET` belong to the outer `Query`.

The mismatch seems to be in the AST builder layer. `union()` currently accepts `QueryNode`, so each operand is first built as a full `Query`, and then only `query.body` is moved into `SetExpr::Union`. Since `QueryNode` can also represent `OrderByNode`, `LimitNode`, and `OffsetNode`, invalid UNION operands are representable in the typed builder and then rejected later with `UnionBranchWithClause`.

Could we model the same boundary in the builder, e.g. with a `SetExprNode` or a `Prebuild<SetExpr>`-style layer? SELECT / VALUES / UNION-capable nodes would build set expressions, while `ORDER BY` / `LIMIT` / `OFFSET` remain query-level wrappers around the completed set expression. That would make invalid branch modifiers unrepresentable in the typed builder, instead of accepting them and rejecting them during `Query::try_from`.

devgony (changes requested):
```sql
select 1, 'a'
union 
select 2, 3;
```
should get type error

```sql
-- example from pg
ERROR:  22P02: invalid input syntax for type integer: "a"
LINE 1: select 1, 'a'
                  ^
```
but currently it doesn't
```sql
gluesql> select 1, 'a'
union select 2, 3;
| 1 | 'a' |
|---|-----|
| 1 | a   |
| 2 | 3   |
```

test-suite/src/union.rs
```diff
@@ -0,0 +1,190 @@
...
+
+    // UNION with VALUES
+    g.named_test(
+        "UNION with VALUES",
+        "SELECT id FROM A WHERE id = 1 UNION VALUES (10), (20) ORDER BY id",
+        Ok(select!(id; I64; 1; 10; 20)),
+    )
+    .await;
+
+    // UNION as a derived subquery (covers fetch.rs SetExpr::Union branch)
```
panarch: Nit: this and the similar `covers ...` comments below describe the internal coverage target rather than the behavior being tested. The `named_test` labels and SQL already make the scenario clear, so could we drop the implementation-specific parts or rephrase them around user-visible behavior? Otherwise these comments can become stale when the implementation is refactored.

core/src/plan/union.rs
```diff
@@ -0,0 +1,352 @@
...
+                "SELECT id, name FROM T UNION SELECT id, label FROM S"
+            )
+            .is_ok()
+        );
+    }
+
+    #[test]
+    fn unknown_expression_type_is_skipped() {
+        let storage = run("");
+        // Complex expression — type not statically known, should not error.
```
panarch: Nit: for this kind of test-only explanation, could we move the reason into the assertion message instead of a standalone comment? For example, `assert!(..., "complex expression type is not statically known, so plan-time validation should skip it")`. That keeps the context attached to the failure output and avoids comments that only restate why the assertion should pass. Same idea could apply to the similar skip/explanation comments below.

core/src/plan/union.rs
```diff
@@ -0,0 +1,358 @@
...
+
+    if let Some(set_expr) = set_expr {
+        validate_set_expr(schema_map, set_expr)?;
+    }
+
+    Ok(())
+}
+
+fn validate_set_expr(schema_map: &SchemaMap, set_expr: &SetExpr) -> Result<()> {
+    let SetExpr::Union { left, right, .. } = set_expr else {
```
panarch: Could we make this traversal cover nested queries as well? Right now this returns `Ok(())` for any non-`Union` `SetExpr`, so the type check only runs when the statement/source body itself is a UNION.

This top-level case is rejected:

```sql
SELECT 1, 'a' UNION SELECT 2, 3;
```

But the same UNION is skipped once it appears under a `Select`, for example:

```sql
SELECT * FROM (SELECT 1, 'a' UNION SELECT 2, 3) AS t;
```

That query currently succeeds and returns mixed column types. Since `select_union` only checks column counts at execution time, `validate_union` needs to walk `Select` bodies too, including derived tables and subquery expressions, so nested UNIONs get the same type compatibility check.

ever0de:
**Why UNION ORDER BY goes through `sort_stateless`, not `Sort::apply`**

GlueSQL has two distinct sort implementations:

| | `Sort::apply` | `sort_stateless` |
|---|---|---|
| Used by | plain `SELECT` | `UNION`, `VALUES` |
| Row type | `RowContext` (table + schema aware) | flat `Vec<Value>` with labels only |
| Positional index (`ORDER BY 1`) | ✅ resolved via column position | ✅ resolved via `positional_index()` helper (added in this PR) |
| Expression evaluation | full context-aware evaluate | `evaluate_stateless` (no table context) |

`Sort::apply` re-evaluates each ORDER BY expression against a `RowContext` that carries the originating table's schema.  A UNION result row has no such context — it is a flat tuple produced by merging two independent streams.  Passing it to `Sort::apply` would require fabricating a context, which is both wrong and fragile.

`sort_stateless` evaluates expressions purely against the row's column values (positional index or stateless expression evaluation), which is exactly what the SQL standard requires for compound-query ORDER BY:

  *"In a compound SELECT, all ORDER BY expressions are handled as aliases for one of the result columns."* — SQLite docs §4
 
**Why ORDER BY forces materialisation**
 
Streaming sort is not possible in the general case: you cannot emit the first output row until you have seen all input rows, because any unseen row might sort before it.  This is why the `select_union` function takes two paths:
 
- **No ORDER BY** → stream lazily; `LIMIT` can terminate the stream early without ever reading the full result set.  UNION DISTINCT uses `TryStreamExt::try_filter` with a `HashSet` to deduplicate on-the-fly, so `LIMIT` short-circuits even for the DISTINCT case.
- **ORDER BY present** → materialise both sides, optionally deduplicate, sort with `sort_stateless`, then apply `LIMIT`/`OFFSET` over the sorted `Vec`.


sqlparser result
```
Query {
    body: SetOperation {
        op: Union,
        set_quantifier: None,          // default DISTINCT value
        left: Select { projection: [UnnamedExpr(Value(Number(BigDecimal("3e0"), false)))], from: [], ... },
        right: Select { projection: [UnnamedExpr(Value(Number(BigDecimal("1e0"), false)))], from: [], ... },
    },
    order_by: Some(OrderBy {
        exprs: [OrderByExpr {
            expr: Value(Number(BigDecimal("1e0"), false)),  // literal number, not column index
            asc: Some(true),
            nulls_first: None,
        }],
    }),
    limit: None,
    offset: None,
}
```
----

gluesql result:
```
Query {
    body: Union {
        left: Select {
            distinct: false,
            projection: SelectItems([
                Expr { expr: Literal(Number(BigDecimal("3e0"))), label: "3" }
            ]),
            from: TableWithJoins {
                relation: Series { alias: TableAlias { name: "Series" }, size: Literal(Number("1e0")) }
            },
            selection: None, group_by: [], having: None,
        },
        right: Select {
            distinct: false,
            projection: SelectItems([
                Expr { expr: Literal(Number(BigDecimal("1e0"))), label: "1" }
            ]),
            from: TableWithJoins {
                relation: Series { alias: TableAlias { name: "Series" }, size: Literal(Number("1e0")) }
            },
            ...
        },
        all: false,   ← DISTINCT (default)
    },
    order_by: [
        OrderByExpr {
            expr: Literal(Number(BigDecimal("1e0"))),  ← literal number 1 
            asc: Some(true),
        }
    ],
    limit: None,
    offset: None,
}
```

## reference

sqlite: https://www.sqlite.org/lang_select.html#compound_select
> only the last or right-most simple SELECT may have an ORDER BY clause. That ORDER BY clause will apply across all elements of the compound

psql: https://www.postgresql.org/docs/current/queries-union.html
> You can also surround an individual query with parentheses. This is important if the query needs to use any of the clauses discussed in following sections, such as LIMIT. Without parentheses, you'll get a syntax error, or else the clause will be understood as applying to the output of the set operation rather than one of its inputs. For example,
> 
> SELECT a FROM b UNION SELECT x FROM y LIMIT 10
> is accepted, but it means
> 
> (SELECT a FROM b UNION SELECT x FROM y) LIMIT 10
> not
>
> SELECT a FROM b UNION (SELECT x FROM y LIMIT 10)

mysql: https://dev.mysql.com/doc/refman/8.4/en/union.html

core/src/plan/union.rs
```diff
@@ -0,0 +1,498 @@
...
+fn validate_table_factor(schema_map: &SchemaMap, tf: &TableFactor) -> Result<()> {
+    if let TableFactor::Derived { subquery, .. } = tf {
+        validate_set_expr(schema_map, &subquery.body)?;
+    }
+    Ok(())
+}
+
+/// Recursively walk an expression, validating any embedded subquery bodies.
+fn validate_expr(schema_map: &SchemaMap, expr: &Expr) -> Result<()> {
+    match expr {
```
panarch: Could we reuse the existing read-only expression visitor here instead of manually matching a subset of `Expr` variants? This traversal still misses subqueries nested under expression containers that `try_visit_expr` already covers, such as `CASE`, function/aggregate arguments, `BETWEEN`, `IN (...)` lists, `IS NULL`, arrays, and array indexes.

For example, this currently bypasses UNION type validation even though the direct `EXISTS (...)` case is rejected:

```sql
SELECT CASE WHEN EXISTS (SELECT 1 UNION SELECT 'a') THEN 1 ELSE 0 END;
```

`schemaless/validate.rs` already uses `try_visit_expr` for this pattern, so this validator could share the same traversal and only handle `Subquery` / `InSubquery` / `Exists` in the callback.

ever0de:
https://www.postgresql.org/docs/current/typeconv-union-case.html

docs/union-fix-plan.md
```diff
@@ -0,0 +1,233 @@
+# UNION 구현 버그 수정 계획
```
panarch: It would be better to remove this from the PR, along with `docs/union-implementation.md`. Both look like internal implementation notes that were included by accident.

ever0de: thanks!

PR #1900, opened 2026-05-03
Introduce statement execution plans

panarch:
## Summary

This PR introduces `StatementPlan` as the execution-facing representation and starts separating execution metadata from AST.

The main boundary this PR establishes is:

- AST represents parsed SQL syntax and storage-facing definitions.
- `StatementPlan` represents the structure consumed by planning and execution.
- `ast_builder` remains a fluent query builder, but its primary `build()` output is now `StatementPlan`.
- Builder nodes may temporarily hold plan-only metadata, such as hash join executor selection and index selection, but that metadata is emitted only into plan types and is rejected when constructing AST.

The scope is intentionally conservative. This PR keeps most plan structs close to the existing AST shape so the transition stays reviewable. Deeper plan redesigns, such as pipeline/operator nodes, table scan access paths, and projection expansion, are left for follow-up work.

## Current Changes

- [x] Add `StatementPlan` and plan-side query/expression/table/join structures.
- [x] Make execution paths consume planned statements.
- [x] Move aggregate slot binding metadata to plan types.
- [x] Move hash join executor metadata out of AST and into `JoinExecutorPlan`.
- [x] Make `ast_builder` produce `StatementPlan` as its primary `build()` output.
- [x] Keep explicit AST build paths only where AST data is still needed.
- [x] Reject AST construction for plan-only hash join and index builder nodes.
- [x] Narrow builder-internal constructors and helper methods.
- [x] Remove `Projection::SchemalessMap` from AST and keep it plan-only.
- [x] Remove `TableFactor::Table.index` and `ast::IndexItem` from AST.
- [x] Keep index planning metadata only in `TableFactorPlan` / `IndexItemPlan`.
- [x] Update affected translation, planner, executor, and tests.

## Follow-up Issues

- #1904: Reject unsupported table function arguments during SQL translation.
- #1905: Apply schemaless rewrites to `CREATE TABLE AS SELECT` sources.
- #1906: Include `ORDER BY` expressions in schema dependency scanning.
- #1907: Validate derived tables using projected output scope.
- #1908: Resolve custom functions during planning and execute prepared function bodies.
- #1909: Rename `ast_builder` to `query_builder`.
- #1910: Evolve `StatementPlan` into a pipeline/operator execution plan.

## Out Of Scope

- Full pipeline/operator plan graph.
- Table scan/access-path plan redesign.
- Projection wildcard expansion at plan time.
- Broader executor simplification.
- Public API naming cleanup around direct planned execution.
- Renaming `ast_builder` / builder API naming in this PR.

## Summary by CodeRabbit

* **New Features**
  * Added a planner intermediate representation to improve query planning, execution, and SQL-to-plan translation.

* **Bug Fixes**
  * More specific error messages for invalid index/default expressions and stateless subquery misuse.
  * Safer default and function-argument evaluation to prevent ambiguous failures.

* **Refactor**
  * Builders, planners, and executors now consistently use plan-layer representations for simpler, more reliable query handling.


core/src/plan/validate.rs
```diff
@@ -117,35 +117,35 @@ fn get_labels(schema: &Schema) -> Option<Vec<&str>> {
...
-        TableFactor::Table { name, .. } => {
+        TableFactorPlan::Table { name, .. } => {
             let schema = schema_map.get(name);
             schema.map(|schema| Arc::from(Context::new(get_labels(schema), None)))
         }
-        TableFactor::Derived { subquery, .. } => contextualize_query(schema_map, subquery),
-        TableFactor::Series { .. } | TableFactor::Dictionary { .. } => None,
+        TableFactorPlan::Derived { subquery, .. } => contextualize_query(schema_map, subquery),
+        TableFactorPlan::Series { .. } | TableFactorPlan::Dictionary { .. } => None,
     }
```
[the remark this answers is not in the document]
panarch: Thanks for pointing this out. I created #1907 to track it separately.

This looks like an existing validation-scope issue rather than something introduced by this execution-plan split: derived tables should expose their projected output labels to the outer query instead of leaking their inner source scope. I’ll handle it as a follow-up issue rather than expanding the scope of this PR.

core/src/executor/evaluate.rs
```diff
@@ -376,57 +376,72 @@ async fn evaluate_function<'a, 'b: 'a, T: GStore>(
...
+            let row = Cow::Owned(Row {
+                columns: columns.into(),
+                values,
+            });
+            let context = RowContext::new(name, row, None);
+            let context = Some(Arc::new(context));
+
+            let body = plan_scalar_expr(body.clone());
+            let evaluated = evaluate_inner(storage, context, None, &body).await?;
+            let value = evaluated.try_into()?;
```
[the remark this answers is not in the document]
panarch: Thanks for the review. I agree this area has a scope-boundary problem, but I’m going to handle it as part of a broader custom-function planning issue rather than patching this branch locally.

I created #1908 to track the underlying problem. Custom functions are currently fetched and planned inside expression evaluation, and the storage trait returns borrowed function definitions, which makes the feature effectively memory-storage-oriented. This also means function bodies are planned too late and miss planner passes such as aggregate slot binding.

A concrete case is:

```sql
CREATE FUNCTION f() RETURN (SELECT COUNT(*) FROM Users);
SELECT f();
```

This currently reaches execution with an unplanned aggregate inside the function body.

The direction in #1908 is to resolve custom function definitions during planning/preparation, build a prepared function registry, and evaluate prepared bodies during execution. That should also give us the right place to define and enforce function default/body scope rules consistently.

storages/sled-storage/src/index_sync.rs
```diff
@@ -278,7 +279,8 @@ async fn evaluate_index_key(
         columns: columns.unwrap_or(&[]),
         values: row,
     });
-    let evaluated = evaluate_stateless(context, index_expr)
+    let index_expr = plan_scalar_expr(index_expr.clone());
```
devgony: `let index_expr = plan_scalar_expr(index_expr.clone());` // re-allocs the whole plan tree per row

Convert each index expression to ExprPlan once (cache it alongside the SchemaIndex, or convert in the caller before the row loop) and pass &ExprPlan into evaluate_index_key

https://github.com/gluesql/gluesql/compare/add-execution-plan...devgony:5be97d7f

panarch: Thanks!

I moved the `plan_scalar_expr(index_expr.clone())` call out of `evaluate_index_key` and precompute a `PlannedIndex` when entering `IndexSync` / create-drop index sync paths. The row loop now reuses `&ExprPlan`, so we no longer clone and rebuild the plan tree per row.

core/src/plan/statement/expr.rs
```diff
@@ -0,0 +1,718 @@
...
+        values: Option<ExprPlan>,
+    },
+    Dedup(ExprPlan),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct AggregatePlan {
+    pub func: AggregateFunctionPlan,
+    pub distinct: bool,
+    pub slot: Option<usize>,
```
devgony, line 310: Planner-only fields (slot / aggregate_slots) now participate in derived Eq/Hash/serde
  
  AggregatePlan (core/src/plan/statement/expr.rs:306) and SelectPlan (core/src/plan/statement/query.rs:21) now #[derive(PartialEq, Eq, Hash, Serialize, Deserialize)], which pulls the planner-only slot / aggregate_slots fields into equality, hashing, and serialization.
The old ast::Aggregate / ast::Select deliberately excluded these via hand-written impls and #[serde(default, skip)] (the now-removed select_eq_and_hash_ignore_aggregate_slots test guarded exactly this invariant).

ever0de:
NIT:

`plan_scalar_expr` recursively walks the entire `ast::Expr` tree and constructs a new `ExprPlan` tree.
Every node incurs a `Box::new` heap allocation, and collection nodes like `InList`, `Case`, and `ArrayIndex` allocate a fresh `Vec`.
The cost is O(N) proportional to the number of AST nodes — each call is equivalent to cloning the entire expression tree into a new representation.

Currently, `Indexes::find` and `find_ordered` call this conversion inside an iterator closure,
meaning with M schema indexes and K calls to `find`/`find_ordered`, up to **M × K** conversions occur.
(K > 1 is common because `plan_index_expr` recurses through `AND` branches, calling `find` repeatedly with the same `indexes`.)

Pre-converting all schema index expressions when constructing `Indexes` makes every subsequent `find`/`find_ordered` call a plain `==` comparison.

```diff
-struct Indexes<'a>(&'a [SchemaIndex]);
+struct Indexes<'a>(Vec<(ExprPlan, &'a SchemaIndex)>);

 impl Indexes<'_> {
     fn find(&self, target: &ExprPlan) -> Option<String> {
         self.0
             .iter()
-            .find(|SchemaIndex { expr, .. }| plan_scalar_expr(expr.clone()) == *target)
-            .map(|SchemaIndex { name, .. }| name.to_owned())
+            .find(|(planned_expr, _)| planned_expr == target)
+            .map(|(_, SchemaIndex { name, .. })| name.to_owned())
     }

     fn find_ordered(&self, target: &OrderByExprPlan) -> Option<String> {
         self.0
             .iter()
-            .find(|SchemaIndex { expr, order, .. }| {
-                if plan_scalar_expr(expr.clone()) != target.expr {
+            .find(|(planned_expr, SchemaIndex { order, .. })| {
+                if planned_expr != &target.expr {
                     return false;
                 }
                 // ...
             })
-            .map(|SchemaIndex { name, .. }| name.to_owned())
+            .map(|(_, SchemaIndex { name, .. })| name.to_owned())
     }
 }

 fn indexes(&self, relation: &TableFactorPlan) -> Option<Indexes<'a>> {
     match relation {
         TableFactorPlan::Table { name, .. } => self
             .schema_map
             .get(name)
-            .map(|schema| Indexes(&schema.indexes)),
+            .map(|schema| Indexes(
+                schema.indexes.iter()
+                    .map(|si| (plan_scalar_expr(si.expr.clone()), si))
+                    .collect(),
+            )),
         _ => None,
     }
 }
```

Trade-off: `Indexes` construction now pays 1 `Vec` allocation + M conversions up front,
but this replaces K × M conversions with M — a win whenever K ≥ 2 (any query with `AND` conditions).
Since `SchemaIndex.expr` is stored as `ast::Expr` due to storage serialization constraints,
changing the field type itself is out of scope here; this is the lightest viable improvement within the current architecture.

core/src/plan/statement/expr.rs
panarch: I think this is intentional for the plan layer.

The old AST fields had to be ignored because `slot` / `aggregate_slots` were execution metadata attached to syntax/catalog data. In this PR those fields moved into `StatementPlan`, where they are part of the executable plan state.

For serialization, we should keep them. A serialized/deserialized `StatementPlan` needs to preserve aggregate bindings, otherwise aggregate evaluation can no longer read/write the correct slots.

For Eq/Hash, I think including them is also reasonable for plan identity. When we need logical aggregate identity independent of slot binding, we already use a separate `AggregateKey { func, distinct }` in `plan/aggregate.rs` instead of hashing `AggregatePlan` directly.

panarch:
> NIT:
> 
> `plan_scalar_expr` recursively walks the entire `ast::Expr` tree and constructs a new `ExprPlan` tree. Every node incurs a `Box::new` heap allocation, and collection nodes like `InList`, `Case`, and `ArrayIndex` allocate a fresh `Vec`. The cost is O(N) proportional to the number of AST nodes — each call is equivalent to cloning the entire expression tree into a new representation.
> 
> Currently, `Indexes::find` and `find_ordered` call this conversion inside an iterator closure, meaning with M schema indexes and K calls to `find`/`find_ordered`, up to **M × K** conversions occur. (K > 1 is common because `plan_index_expr` recurses through `AND` branches, calling `find` repeatedly with the same `indexes`.)
> 
> Pre-converting all schema index expressions when constructing `Indexes` makes every subsequent `find`/`find_ordered` call a plain `==` comparison.
> 
> ```diff
> -struct Indexes<'a>(&'a [SchemaIndex]);
> +struct Indexes<'a>(Vec<(ExprPlan, &'a SchemaIndex)>);
> 
>  impl Indexes<'_> {
>      fn find(&self, target: &ExprPlan) -> Option<String> {
>          self.0
>              .iter()
> -            .find(|SchemaIndex { expr, .. }| plan_scalar_expr(expr.clone()) == *target)
> -            .map(|SchemaIndex { name, .. }| name.to_owned())
> +            .find(|(planned_expr, _)| planned_expr == target)
> +            .map(|(_, SchemaIndex { name, .. })| name.to_owned())
>      }
> 
>      fn find_ordered(&self, target: &OrderByExprPlan) -> Option<String> {
>          self.0
>              .iter()
> -            .find(|SchemaIndex { expr, order, .. }| {
> -                if plan_scalar_expr(expr.clone()) != target.expr {
> +            .find(|(planned_expr, SchemaIndex { order, .. })| {
> +                if planned_expr != &target.expr {
>                      return false;
>                  }
>                  // ...
>              })
> -            .map(|SchemaIndex { name, .. }| name.to_owned())
> +            .map(|(_, SchemaIndex { name, .. })| name.to_owned())
>      }
>  }
> 
>  fn indexes(&self, relation: &TableFactorPlan) -> Option<Indexes<'a>> {
>      match relation {
>          TableFactorPlan::Table { name, .. } => self
>              .schema_map
>              .get(name)
> -            .map(|schema| Indexes(&schema.indexes)),
> +            .map(|schema| Indexes(
> +                schema.indexes.iter()
> +                    .map(|si| (plan_scalar_expr(si.expr.clone()), si))
> +                    .collect(),
> +            )),
>          _ => None,
>      }
>  }
> ```
> 
> Trade-off: `Indexes` construction now pays 1 `Vec` allocation + M conversions up front, but this replaces K × M conversions with M — a win whenever K ≥ 2 (any query with `AND` conditions). Since `SchemaIndex.expr` is stored as `ast::Expr` due to storage serialization constraints, changing the field type itself is out of scope here; this is the lightest viable improvement within the current architecture.

Thanks, fixed in https://github.com/gluesql/gluesql/pull/1900/commits/108e05ba28e83ca3329c30984d8414e89787a100.

I added a private `PlannedSchemaIndex` inside the index planner and now pre-convert each `SchemaIndex.expr` once when constructing `Indexes`. `find` and `find_ordered` now compare against the cached `ExprPlan` instead of cloning and converting the schema expression on each lookup.

ever0de (approved):

devgony (approved):
👍🏼 👍🏼 👍🏼

PR #1901, opened 2026-05-03
feat(redb): Add new Redb v3 storage format

edomaur:
Draft only for testing the v3 file format and adding indexes.
* root idea was to replicate in redb v4 all the sled backend features, but rework of this PR focused on redb update only, no new features except the new storage format.

## Summary by CodeRabbit

* **New Features**
  * New redb databases now use the latest file format.
  * Added migration support for existing redb storage files from older formats.
  * Added CLI and library guidance for upgrading storage files.

* **Bug Fixes**
  * Preserved table data, schema, and schemaless records during migration.
  * Added clear errors when storage migration is required.

* **Documentation**
  * Documented migration steps, compatibility considerations, and backup recommendations.
  * Clarified storage fixture formats and version history.


panarch:
Thanks for working on redb 4 support.

I think we should first align on the direction here. Since GlueSQL already has `gluesql-redb-storage`, and this PR appears to target the same storage concept and use case, I'm not sure adding a separate `redb4-storage` crate is the right shape for this change.

The existing redb storage already has storage format versioning, migration support, fixtures, and CLI upgrade flow. If the goal is to support newer redb versions, I think it would be better to evolve the existing `gluesql-redb-storage` crate and extend its migration path, rather than introduce a parallel storage backend.

Index support also seems like a separate concern from the redb version upgrade. Splitting this into two smaller PRs would make the review much easier:

1. Upgrade existing `gluesql-redb-storage` to support redb 3/4 with the required format compatibility or migration path.
2. Add index support on top of that storage after the version upgrade is settled.

So my preference is not to add `redb4-storage` as a new crate unless there is a specific compatibility or maintenance reason that requires keeping it separate. If such a reason exists, it would be helpful to document the intended relationship between `redb-storage` and `redb4-storage`.

edomaur:
Hello !

> 
>     1. Upgrade existing `gluesql-redb-storage` to support redb 3/4 with the required format compatibility or migration path.
> 
>     2. Add index support on top of that storage after the version upgrade is settled.
> 

Yes, you're right, I'll do that. On point 1 I need to understand the migration path, I created the redb4-storage to get a better idea of what is needed, but if someone has a redb2 with v2 file format, it needs to be migrated to the new file format before bumping the redb crate to 3.X.X or 4.X.X (file format migration seems to be in 2.6.X only)

Is there already a migration path existing in another backend that I should use as a model ?

panarch:
Thanks, that makes sense.

For the migration model, I think the closest reference is the existing `gluesql-redb-storage` migration flow itself:

- `storages/redb-storage/src/migration.rs`
- `storages/redb-storage/src/migration/v1_to_v2.rs`
- `storages/redb-storage/tests/migration_v1_to_v2.rs`
- `cli/src/upgrade.rs`

One thing that may help clarify the direction is separating the version layers:

- redb crate version: currently `2.6`, target can be `4.x`
- redb on-disk file format: old format `v2`, newer format `v3`
- GlueSQL redb storage format: currently `REDB_STORAGE_FORMAT_VERSION = 2`, next one could be `3`

So I don’t think this necessarily needs to be split into “redb crate 3 first, then redb crate 4” as separate PRs. The final storage implementation can target redb crate `4.x`.

The important part is that the existing `gluesql-redb-storage` upgrade flow should handle the bridge before the redb 4.x implementation opens old databases. For example, a new GlueSQL-side `v2_to_v3` migration could use redb `2.6`’s `Database::upgrade()` to migrate the underlying redb file format from `v2` to `v3`, then update GlueSQL’s own storage format metadata to `3`. After that, the normal storage implementation can use redb crate `4.x`.

That would keep the compatibility path inside the existing `gluesql-redb-storage` crate without introducing a parallel `redb4-storage` backend.

edomaur:
> For the migration model, I think the closest reference is the existing `gluesql-redb-storage` migration flow itself:
> 
>     * `storages/redb-storage/src/migration.rs`
> 
>     * `storages/redb-storage/src/migration/v1_to_v2.rs`
> 
>     * `storages/redb-storage/tests/migration_v1_to_v2.rs`
> 
>     * `cli/src/upgrade.rs`
> 

Oh, wow, I missed that... :roll_eyes: 

Ok, I'll do it the same way, the path should be straigthforward.

edomaur:
@panarch okay, here is a 1st draft of the Redb update, still need to be thoroughly tested before making this PR ready to review.

panarch (approved):

PR #1912, opened 2026-05-10
Fix aggregate functions to skip NULL values per SQL standard

### LINKED ISSUE #1897: Aggregate functions do not skip NULL values (SUM, MIN, MAX, AVG, VARIANCE, STDEV)
## Summary

Aggregate functions `SUM`, `MIN`, `MAX`, `AVG`, `VARIANCE`, and `STDEV` do not ignore `NULL` values as required by the SQL standard. `NULL` values poison the accumulator, producing incorrect results.

`COUNT` is **not** affected — it correctly handles `NULL` via an explicit `is_null()` check.

## Affected functions and behavior

| Function | NULL position | Expected | Actual |
|----------|-------------|----------|--------|
| `SUM` | anywhere | ignore NULL | `NULL` poisons result |
| `AVG` | anywhere | ignore NULL | `NULL` poisons sum **and** inflates count |
| `VARIANCE` | anywhere | ignore NULL | `NULL` poisons sum/sum_square **and** inflates count |
| `STDEV` | anywhere | ignore NULL | `NULL` poisons sum/sum_square **and** inflates count |
| `MIN` | first row only | ignore NULL | `NULL` becomes sticky initial value |
| `MAX` | first row only | ignore NULL | `NULL` becomes sticky initial value |

## Reproducing

```sql
CREATE TABLE T (val INTEGER NULL);
INSERT INTO T VALUES (NULL), (5), (3);

-- SUM: NULL anywhere poisons the result
SELECT SUM(val) FROM T;
-- Expected: 8
-- Actual:   NULL  (Null.add(5) -> Null, Null.add(3) -> Null)

-- AVG: NULL anywhere poisons sum AND inflates count
SELECT AVG(val) FROM T;
-- Expected: 4  (8 / 2, only non-null rows counted)
-- Actual:   NULL  (sum=Null, count=3)

-- MIN: first-row NULL is sticky
SELECT MIN(val) FROM T;
-- Expected: 3
-- Actual:   NULL  (Null.evaluate_cmp(5) -> None -> no update)

-- MAX: first-row NULL is sticky
SELECT MAX(val) FROM T;
-- Expected: 5
-- Actual:   NULL  (Null.evaluate_cmp(5) -> None -> no update)

-- MIN/MAX work correctly when NULL is not the first row:
CREATE TABLE T2 (val INTEGER NULL);
INSERT INTO T2 VALUES (5), (NULL), (3);
SELECT MIN(val) FROM T2;
-- Returns: 3  (correct, because first value was non-NULL)
```

## Root cause

`AggrValue::new` and `AggrValue::accumulate` in `core/src/executor/aggregate/state.rs` have no NULL-skipping logic for non-COUNT aggregates.

### `AggrValue::new` (first row)

When the first evaluated value is `NULL`, it is used as the initial accumulator value without any check:

```rust
// state.rs - AggrValue::new
AggregateFunction::Sum(_) => {
    // ...
    Self::Sum {
        value,  // <- NULL is stored directly as initial sum
        distinct_values,
    }
}
```

Compare with `COUNT`, which correctly handles NULL:

```rust
AggregateFunction::Count(CountArgExpr::Expr(_)) => {
    Self::Count {
        wildcard: false,
        count: i64::from(!value.is_null()),  // <- NULL -> count 0
        distinct_values,
    }
}
```

### `AggrValue::accumulate` (subsequent rows)

No NULL check before arithmetic. `Value::add` and `Value::multiply` propagate NULL per SQL expression semantics (`NULL + 5 = NULL`), but aggregate semantics require ignoring NULL:

```rust
// state.rs - AggrValue::accumulate
Self::Sum { value, distinct_values } => {
    // no null check on new_value
    *value = value.add(new_value)?;  // <- I64(8).add(Null) -> Null
    Ok(true)
}
```

### `evaluate_cmp` returns `None` for NULL

For MIN/MAX, `Value::Null.evaluate_cmp(x)` falls through to the `_ => None` catch-all, so the comparison never matches and the initial NULL is never replaced:

```rust
// value.rs - evaluate_cmp
match (self, other) {
    (Value::I64(l), _) => l.partial_cmp(other),
    // ...
    _ => None,  // <- (Null, I64(5)) lands here
}
```

## Suggested fix

Two changes in `State::accumulate` and `AggrValue::accumulate`:

1. **`State::accumulate`**: When the slot is `None` (first value) and the value is `NULL`, do not create an `AggrValue`. Leave the slot as `None`. The `export()` method already maps `None` to `empty_value()` which returns `I64(0)` for COUNT and `Null` for everything else — so an all-NULL column correctly returns NULL for SUM/MIN/MAX/AVG.

2. **`AggrValue::accumulate`**: Add a NULL guard at the top of each non-COUNT arm:

```rust
Self::Sum { value, distinct_values } => {
    if new_value.is_null() {
        return Ok(false);
    }
    // ... existing logic
}
```

This mirrors how `COUNT(expr)` already handles NULLs and aligns with the SQL standard (Section 10.9, General Rule 4: "the null value is effectively eliminated before the set function is applied").

## Impact

Any query using `SUM`, `AVG`, `MIN`, `MAX`, `VARIANCE`, or `STDEV` on a column containing `NULL` values may produce incorrect results. The severity depends on NULL position:

- **SUM/AVG/VARIANCE/STDEV**: A single NULL **anywhere** in the data makes the entire result `NULL`
- **MIN/MAX**: Only broken when the first encountered row (per group) has a `NULL` value

kimjune01:
## Summary

Aggregate functions `SUM`, `MIN`, `MAX`, `AVG`, `VARIANCE`, and `STDEV` now correctly ignore `NULL` values during accumulation, matching the SQL standard (Section 10.9, General Rule 4: "the null value is effectively eliminated before the set function is applied").

## Problem

NULL values poisoned aggregate accumulators, producing incorrect results:
- **SUM/AVG/VARIANCE/STDEV**: A single NULL anywhere made the entire result `NULL`
- **MIN/MAX**: NULL as first row became sticky (comparisons returned `None`, preventing updates)

`COUNT` was unaffected — it already had explicit `is_null()` checks.

## Changes

- Add NULL skip logic to `AggrValue::accumulate` for all non-COUNT aggregates
- In `State::accumulate`, skip `AggrValue::new` when the first value is NULL for non-COUNT aggregates (leaves the slot as `None` so the next non-NULL value becomes the initial accumulator)
- Add `debug_assert!` to `AggrValue::new` to enforce the NULL-skipping invariant
- Update existing tests that expected incorrect NULL-poisoned results
- Add comprehensive NULL-handling test suite (`test-suite/src/aggregate/null_handling.rs`):
  - NULL as first row for SUM, MIN, MAX, AVG, VARIANCE, STDEV
  - All-NULL column behavior (returns `NULL` for all aggregates except `COUNT(val)` which returns `0`)

## Test coverage

\`\`\`sql
-- Before: NULL poisons the result
CREATE TABLE T (val INTEGER NULL);
INSERT INTO T VALUES (NULL), (5), (3);
SELECT SUM(val) FROM T;
-- Before: NULL
-- After:  8

SELECT MIN(val) FROM T;
-- Before: NULL  (sticky from first row)
-- After:  3

SELECT AVG(val) FROM T;
-- Before: NULL  (poisoned sum, inflated count)
-- After:  4.0  (only non-NULL values counted)
\`\`\`

New tests verify:
- NULL-first-row handling for all affected aggregates
- All-NULL columns return `NULL` (not `0`) for SUM/MIN/MAX/AVG/VARIANCE/STDEV
- All-NULL columns return `0` for `COUNT(val)` (not `COUNT(*)`)

Fixes #1897

## Summary by CodeRabbit

* **Bug Fixes**
  * Aggregate functions (SUM, MIN, MAX, AVG, VARIANCE, STDEV) now ignore leading NULL inputs so results reflect subsequent non-NULL values; COUNT behavior unchanged.
* **Tests**
  * Added and updated tests to verify aggregates when the first value is NULL and when all inputs are NULL, confirming numeric results or NULL as appropriate.


ever0de (approved):

core/src/executor/aggregate/state.rs
```diff
@@ -402,13 +428,15 @@ impl<'a, T: GStore> State<'a, T> {
...
-                aggr_value.accumulate(&value)?;
-            }
-            None => {
-                group.values[slot] = Some(AggrValue::new(aggregate, &value)?);
+        if let Some(aggr_value) = group.values[slot].as_mut() {
+            aggr_value.accumulate(&value)?;
+        } else {
+            // For non-COUNT aggregates, skip NULL values when initializing
+            // Leave the slot as None so the next non-NULL value becomes the first
+            if value.is_null() && !matches!(aggregate.func, AggregateFunctionPlan::Count(_)) {
```
panarch, line 436: Could we move this NULL-elimination guard above the `group.values[slot]` branch?

Right now first NULLs are skipped here, but later NULLs are skipped again inside each non-COUNT `AggrValue::accumulate` arm. That splits the invariant across two layers. Since `State::accumulate` already has both the evaluated `value` and the `aggregate.func`, it can return early for every non-COUNT NULL before touching the aggregate slot:

```rust
if value.is_null() && !matches!(aggregate.func, AggregateFunctionPlan::Count(_)) {
    return Ok(());
}
```

That keeps the invariant in one place: `AggrValue` never receives NULL for non-COUNT aggregates. All-NULL inputs still work because the slot remains `None` and export falls back to `empty_value(aggregate)`. The repeated `new_value.is_null()` guards in `Sum`, `Min`, `Max`, `Avg`, `Variance`, and `Stdev` can then be removed.

test-suite/src/aggregate/null_handling.rs
```diff
@@ -0,0 +1,113 @@
+use {crate::*, gluesql_core::prelude::Value::*};
+
+test_case!(null_first_row_sum, {
+    let g = get_tester!();
+
+    g.run("CREATE TABLE T (val INTEGER NULL);").await;
+    g.run("INSERT INTO T VALUES (NULL), (5), (3);").await;
+
+    // SUM should ignore NULL values
```
panarch, line 9: Could we use `g.named_test(...)` for these cases instead of comments?

The comments describe the behavior being tested, but `named_test` would put that description in the failure message as well. The aggregate tests already use this pattern in places like `aggregate/expr.rs`.

kimjune01:
Thanks @panarch and @ever0de for the careful review here.

Since this opened, main has moved substantially past the branch: core execution was converted async → sync (#1928), and the SQL test cases migrated from Rust modules to `.sql` fixtures (#1941), which removes the `test-suite/src/aggregate/null_handling.rs` file this PR's tests live in. Rather than force a 7-week-stale branch through a full rewrite, I'm closing this.

If NULL-skipping still diverges from the SQL standard on current main, I'll reopen the fix fresh against the new fixture layout and fold in the single-layer guard @panarch suggested (hoisting the non-COUNT NULL skip above the slot branch so the invariant lives in one place). Appreciate the time.

PR #1925, opened 2026-06-01
Remove Python package from the workspace

panarch:
## Summary

This PR removes the Python package from the GlueSQL workspace.

The Python package is being split out into its own repository, [`gluesql/gluesql-py`](https://github.com/gluesql/gluesql-py), so this repository no longer needs to keep the Python package source, Python-specific CI, or Python publish workflow in the main workspace.

## Changes

- Remove `pkg/python`
- Remove Python CI and publish workflows
- Remove `pkg/python` from Cargo workspace members
- Remove the now-unnecessary `gluesql-py` exclude from the Rust publish workflow
- Prune Python-only dependencies from `Cargo.lock`
- Remove the Python package cache entry from `.gitignore`

## Context

GlueSQL has historically kept Rust, JavaScript, and Python packaging in one repository. As the language packages have grown separate release and maintenance concerns, keeping Python in its own repository should make packaging, CI, and publishing easier to manage independently.

The Rust publish workflow now only excludes `gluesql-js`, because the Python package is no longer part of this workspace.

## Summary by CodeRabbit

* **Removals**
  * Removed Python language bindings and associated API/type hints.
  * Removed Python package configuration, packaging metadata, and PyPI publish workflow.
  * Removed Python examples, tests, development docs, and local build/test configurations.
  * Removed Python-related CI workflows and workspace membership so Python distribution is no longer built or published.


devgony (approved):
👍🏼 👍🏼 👍🏼

PR #1926, opened 2026-06-01
Remove JavaScript package from the workspace

panarch:
## Summary

This PR removes the JavaScript package from the GlueSQL workspace.

The JavaScript package has been split out into its own repository, [`gluesql/gluesql-js`](https://github.com/gluesql/gluesql-js), so this repository no longer needs to keep the JavaScript package source, JavaScript-specific CI, or npm publish workflow in the main workspace.

## Changes

- Remove `pkg/javascript`
- Remove JavaScript CI and npm publish workflows
- Remove `pkg/javascript` from Cargo workspace members
- Remove the now-unnecessary `gluesql-js` exclude from the Rust publish workflow
- Remove JavaScript package version bumping from the prepare-release workflow
- Prune JavaScript-package-only dependencies from `Cargo.lock`
- Keep JavaScript and Node.js getting started docs, with links to the [`gluesql/gluesql-js`](https://github.com/gluesql/gluesql-js) repository

## Context

GlueSQL has historically kept Rust, JavaScript, and Python packaging in one repository. As the language packages have grown separate release and maintenance concerns, keeping JavaScript in its own repository should make package maintenance, CI, and publishing easier to manage independently.

This repository remains the main GlueSQL project and documentation hub, while [`gluesql/gluesql-js`](https://github.com/gluesql/gluesql-js) owns the JavaScript package source and release workflow.

## Release Notes

* **Chores**
  * Moved JavaScript package to separate repository; users should reference `gluesql-js` for JavaScript/Node.js implementations.
  * Removed Web Storage and IndexedDB browser storage support; focus now on core reference storage implementations (in-memory, sled, JSON).
  * Updated documentation to reflect architectural changes and storage scope adjustments.

PR #1928, opened 2026-06-02
Convert core and native storages to sync execution

panarch:
## Motivation

GlueSQL's async support has mostly been driven by browser-facing storage constraints such as IndexedDB, rather than by the core execution model itself.

That async boundary ended up spreading through the planner, executor, storage traits, CLI, examples, tests, and native storage implementations. For Rust users and native storages, this added futures/stream plumbing and async dependencies even when the underlying work was synchronous or backed by synchronous APIs.

With the JavaScript and Python packages being split out of this workspace, this PR makes the Rust core synchronous again and keeps browser/package-specific async concerns out of the core storage contract.

## Changes

- Convert `gluesql-core` planning and execution APIs from async to sync
- Replace storage scan streams with lazy iterator-based scan APIs
- Convert native storage implementations to the sync `Store` / `StoreMut` contracts
- Remove async trait, futures, Tokio, and stream utility dependencies where no longer needed
- Relax storage trait `Send + Sync` requirements so thread-safety is an implementation choice instead of a core contract
- Use `Rc` in core execution contexts where thread-safe sharing is no longer required
- Update CLI, Rust package examples, macros tests, and the shared test suite for sync execution

## Out of scope

JavaScript/wasm and Python package updates are intentionally out of scope for this PR.

These should be merged first:

- #1925 Remove Python package from the workspace
- #1926 Remove JavaScript package from the workspace


core/src/executor/delete.rs
```diff
@@ -24,62 +23,56 @@ pub enum DeleteError {
...
+            let expr = &ExprPlan::BinaryOp {
+                left: Box::new(ExprPlan::Identifier(referencing_column_name.clone())),
+                op: BinaryOperator::Eq,
+                right: Box::new(ExprPlan::Value(value)),
+            };
 
-                let columns = Arc::from(Vec::new());
-                let referencing_rows =
-                    fetch(storage, referencing_table_name, columns, Some(expr)).await?;
+            let columns = Rc::from(fetch_columns(storage, referencing_table_name)?);
```
devgony: ### 1. 
Worth calling out that this line also changed behavior vs main, which passed Arc::from(Vec::new()) (empty columns) here. With empty columns the identifier
referencing_column_name couldn't resolve, the predicate erred, and .next().is_some() counted that error as "row exists" — so the old NO ACTION check blocked the delete
whenever the referencing table was non-empty, regardless of whether the value was actually referenced. 

Therefore, add a regression test. test-suite/src/foreign_key.rs only covers the block case (DELETE … WHERE id = 1, where id = 1 is referenced). Please add a case where the
referencing table is non-empty but the value being deleted is not referenced, so the delete should succeed

### 2. 
Line 59 calls fetch_columns(storage, referencing_table_name)? — a fetch_schema storage read — inside the deleted-row × foreign-key loop. The result depends only on
referencing_table_name, which is fixed before the loop, so a DELETE matching N rows on a table with M foreign keys issues N × M schema reads. On a disk-backed store
(sled/file) each is real I/O (e.g. deleting 10k rows from a table with one FK ⇒ 10k fetch_schema calls).

Hoist the lookup. Pair each referencing with its columns once, right after fetch_referencings, so fetch_schema runs ≤ M times instead of N × M. 

```diff
     let columns = Rc::from(fetch_columns(storage, table_name)?);
-    let referencings = storage.fetch_referencings(table_name)?;
+    let referencings = storage
+        .fetch_referencings(table_name)?
+        .into_iter()
+        .map(|referencing| {
+            let columns = Rc::from(fetch_columns(storage, &referencing.table_name)?);
+            Ok((referencing, columns))
+        })
+        .collect::<Result<Vec<(Referencing, Rc<[String]>)>>>()?;
+
     let mut keys = Vec::new();
     for item in fetch(storage, table_name, columns, selection)? {
         let (key, row) = item?;

-        for Referencing {
-            table_name: referencing_table_name,
-            foreign_key:
-                ForeignKey {
-                    referencing_column_name,
-                    referenced_column_name,
-                    on_delete,
-                    ..
-                },
-        } in &referencings
+        for (
+            Referencing {
+                table_name: referencing_table_name,
+                foreign_key:
+                    ForeignKey {
+                        referencing_column_name,
+                        referenced_column_name,
+                        on_delete,
+                        ..
+                    },
+            },
+            columns,
+        ) in &referencings
         {
             let value = row
                 .get_value(referenced_column_name)
                 .ok_or(DeleteError::ValueNotFound(referenced_column_name.clone()))?
                 .clone();

             let expr = &ExprPlan::BinaryOp {
                 left: Box::new(ExprPlan::Identifier(referencing_column_name.clone())),
                 op: BinaryOperator::Eq,
                 right: Box::new(ExprPlan::Value(value)),
             };

-            let columns = Rc::from(fetch_columns(storage, referencing_table_name)?);
-            let mut referencing_rows = fetch(storage, referencing_table_name, columns, Some(expr))?;
+            let mut referencing_rows =
+                fetch(storage, referencing_table_name, Rc::clone(columns), Some(expr))?;
```

panarch: Thanks, addressed.

I split this into two commits:
- Added a regression test for deleting an unreferenced referenced-table row while the referencing table is non-empty.
- Preloaded referencing table columns before the deleted-row loop, so `fetch_columns` runs once per referencing FK instead of once per deleted row.

`Rc::clone` is only used to pass the already-loaded column list into each lazy `fetch` iterator; it does not re-read the schema.

devgony (approved):
👍🏼 👍🏼 👍🏼

ever0de (approved):

PR #1936, opened 2026-07-02
Reject multi-column IN subqueries

ssseft:
## Summary

  `IN (subquery)` now rejects subqueries that return more than one column instead of
  silently comparing only the first column.

  This matches existing scalar subquery behavior by returning `MoreThanOneColumnReturned`.

  ## Problem

  `IN (subquery)` previously accepted multi-column subquery results.

  ```sql
  CREATE TABLE T (id INTEGER, name TEXT);

  SELECT id FROM T WHERE id IN (SELECT id, name FROM T);
  ```

  The subquery returns both `id` and `name`, but GlueSQL only used the first column and
  ignored the rest.

  ## Changes

  - Use `select_with_labels` in `ExprPlan::InSubquery` evaluation to inspect the projected
  subquery column count.
  - Return `EvaluateError::MoreThanOneColumnReturned` when an `IN` subquery projects more
  than one column.
  - Reuse the selected row iterator from `select_with_labels` for the actual `IN`
  comparison.
  - Add regression coverage for multi-column `IN` subqueries.

  ## Test coverage

  ```sql
  -- Before: silently compares only the first column
  SELECT id FROM Player WHERE id IN (SELECT id, name FROM Player);

  -- After: returns MoreThanOneColumnReturned
  ```

  New tests verify:

  - Multi-column `IN` subqueries with returned rows are rejected
  - Multi-column `IN` subqueries with empty result sets are also rejected

  Fixes #1880

  ## Testing

  ```sh
  cargo fmt --all
  cargo test -p gluesql_memory_storage nested_select --verbose
  ```

 result : All passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved evaluation of nested `IN (subquery)` queries by validating the subquery’s projected columns before performing membership checks.
  * Subqueries that don’t return exactly one column now fail with a clear, dedicated error: “IN (subquery) must return one column”.
* **Tests**
  * Added coverage for invalid nested-select cases to confirm the new error is returned.
  * Updated nested `IN` test queries to select only the required single column.

### DIFF
--- a/core/src/executor/evaluate.rs
+++ b/core/src/executor/evaluate.rs
@@ -7,7 +7,7 @@ use {
     self::function::BreakCase,
     super::{
         context::{AggregateValues, RowContext},
-        select::select,
+        select::{select, select_with_labels},
     },
     crate::{
         data::{CustomFunction, Interval, Row, Value},
@@ -175,9 +175,14 @@ where
                 return Err(EvaluateError::SchemalessProjectionForInSubQuery.into());
             }
             let target = eval(target_expr)?;
+            let (labels, rows) = select_with_labels(storage, subquery, context.cloned())?;
+
+            if labels.len() > 1 {
+                return Err(EvaluateError::MoreThanOneColumnReturned.into());
+            }
 
             let mut matched = false;
-            for row in select(storage, subquery, context.cloned())? {
+            for row in rows {
                 let value = row?.into_values().into_iter().next().unwrap_or(Value::Null);
                 let evaluated = Evaluated::Value(Cow::Owned(value));
 
--- a/core/src/executor/evaluate.rs
+++ b/core/src/executor/evaluate.rs
@@ -178,7 +178,7 @@ where
             let (labels, rows) = select_with_labels(storage, subquery, context.cloned())?;
 
             if labels.len() > 1 {
-                return Err(EvaluateError::MoreThanOneColumnReturned.into());
+                return Err(EvaluateError::InSubqueryMustReturnOneColumn.into());
             }
 
             let mut matched = false;
--- a/test-suite/src/nested_select.rs
+++ b/test-suite/src/nested_select.rs
@@ -1,6 +1,9 @@
 use {
     crate::*,
-    gluesql_core::{executor::Payload, prelude::Value},
+    gluesql_core::{
+        executor::{EvaluateError, Payload},
+        prelude::Value,
+    },
 };
 
 test_case!(nested_select, {
@@ -110,5 +113,14 @@ test_case!(nested_select, {
 
     for (sql, expected) in test_cases {
         g.test(sql, expected);
+    }
+
+    let error_cases = [
+        "SELECT id FROM Player WHERE id IN (SELECT id, name FROM Player)",
+        "SELECT id FROM Player WHERE id IN (SELECT id, name FROM Player WHERE id = 0)",
+    ];
+
+    for sql in error_cases {
+        g.test(sql, Err(EvaluateError::MoreThanOneColumnReturned.into()));
     }
 });
--- a/test-suite/src/nested_select.rs
+++ b/test-suite/src/nested_select.rs
@@ -121,6 +121,9 @@ test_case!(nested_select, {
     ];
 
     for sql in error_cases {
-        g.test(sql, Err(EvaluateError::MoreThanOneColumnReturned.into()));
+        g.test(
+            sql,
+            Err(EvaluateError::InSubqueryMustReturnOneColumn.into()),
+        );
     }
 });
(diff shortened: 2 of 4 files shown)


test-suite/src/nested_select.rs
```diff
@@ -111,4 +114,13 @@ test_case!(nested_select, {
     for (sql, expected) in test_cases {
         g.test(sql, expected);
     }
+
+    let error_cases = [
+        "SELECT id FROM Player WHERE id IN (SELECT id, name FROM Player)",
```
panarch, line 119: The existing `index_nested` fixture looks a bit too permissive for what it is trying to test. Its purpose seems to be verifying that an index can still be picked from inside a nested `IN` subquery, but it uses `id IN (SELECT * FROM User WHERE id = 1)`, which depends on the old behavior where GlueSQL silently used only the first projected column.

With this PR, that query now correctly returns `MoreThanOneColumnReturned`, so the index test should probably make the selected column explicit:

```sql
SELECT id FROM User WHERE id = 1
```

That keeps the index-usage coverage intact while avoiding reliance on the previous multi-column subquery behavior. Currently `cargo test -p gluesql_sled_storage --test sled_storage index_nested` fails on this branch.

panarch (commented):
One related fixture still needs updating.

core/src/executor/evaluate.rs
```diff
@@ -175,9 +175,14 @@ where
                 return Err(EvaluateError::SchemalessProjectionForInSubQuery.into());
             }
             let target = eval(target_expr)?;
+            let (labels, rows) = select_with_labels(storage, subquery, context.cloned())?;
+
+            if labels.len() > 1 {
+                return Err(EvaluateError::MoreThanOneColumnReturned.into());
```
panarch: Given that GlueSQL already has context-specific errors for regular subqueries vs `IN (subquery)` (`SchemalessProjectionForSubQuery` and `SchemalessProjectionForInSubQuery`), I wonder if this should use a dedicated error variant instead of reusing `MoreThanOneColumnReturned`.

`MoreThanOneColumnReturned` is currently used for scalar subqueries such as `SELECT (SELECT 1, 2)`, where the expression must collapse to a single value. For `IN (subquery)`, the rule is slightly different: the subquery may return many rows, but it must project exactly one comparison column. A dedicated error like `InSubqueryMustReturnOneColumn` would make the user-facing diagnostic and tests more precise.

panarch (commented):
One more API/diagnostic suggestion.

panarch:
Just a small note for future updates: since GlueSQL usually squash-merges PRs, it is totally fine to add follow-up commits instead of force-pushing amended commits during review. That tends to make review follow-up easier because previous comments and incremental changes remain easier to track, and the final commit history will still be cleaned up by the squash merge.

panarch:
Could you sync this branch with the latest `main`? Codecov currently reports `Report is 1 commits behind head on main`, so the coverage diff is being computed against a stale base.

The current coverage result looks fine, but I’d like to review the final coverage diff against the current base before merging. A follow-up sync commit is fine; the coverage workflow should run again automatically after the branch is updated.

test-suite/src/nested_select.rs
ssseft: I misunderstood the purpose of this fixture. Thank you for the explanation.

I updated it to use `SELECT id FROM User WHERE id = 1`.

core/src/executor/evaluate.rs
ssseft: Agreed, thanks for the suggestion.

  I added `EvaluateError::InSubqueryMustReturnOneColumn` and updated the `IN (subquery)`
  validation/tests to use the dedicated error.

ssseft:
> Just a small note for future updates: since GlueSQL usually squash-merges PRs, it is totally fine to add follow-up commits instead of force-pushing amended commits during review. That tends to make review follow-up easier because previous comments and incremental changes remain easier to track, and the final commit history will still be cleaned up by the squash merge.

Thanks for letting me know. I’ll avoid force-pushing during review.

ssseft:
> Could you sync this branch with the latest `main`? Codecov currently reports `Report is 1 commits behind head on main`, so the coverage diff is being computed against a stale base.
> 
> The current coverage result looks fine, but I’d like to review the final coverage diff against the current base before merging. A follow-up sync commit is fine; the coverage workflow should run again automatically after the branch is updated.

I checked locally, and it looks like the branch is already up to date. Could you what still needs to be synced?

panarch:
> > Could you sync this branch with the latest `main`? Codecov currently reports `Report is 1 commits behind head on main`, so the coverage diff is being computed against a stale base.
> > The current coverage result looks fine, but I’d like to review the final coverage diff against the current base before merging. A follow-up sync commit is fine; the coverage workflow should run again automatically after the branch is updated.
> 
> I checked locally, and it looks like the branch is already up to date. Could you what still needs to be synced?

Ah, this was caused by the last commit was docs only so that was ignored by coverage tools.
that's interesting. this all looks great. thanks a lot for the contribution 👍 👍 👍

panarch (approved):

PR #1938, opened 2026-07-12
Replace grcov with cargo-llvm-cov for coverage

panarch:
## Summary

Replace grcov with cargo-llvm-cov to make the coverage setup easier for contributors to use locally.

The current workflow relies on manually configuring coverage instrumentation, collecting raw profiles, merging them with `llvm-profdata`, and passing the result to grcov. Reproducing the same setup locally requires contributors to understand and assemble those individual steps.

cargo-llvm-cov provides a Cargo-native interface for running tests and generating coverage reports, allowing contributors to use the same coverage tool locally that CI uses. It also simplifies the GitHub Actions workflow by handling instrumentation and profile management internally.

Branch coverage is no longer required, and the existing `--excl-line` patterns were workarounds for grcov-specific behavior, so they are not carried over.

## Changes

- Replace the pinned grcov installation with cargo-llvm-cov
- Run workspace coverage with `--all-features`, preserving Redis, Mongo, and Git storage coverage
- Preserve the existing coverage scope, including test sources and the current CLI/example exclusions
- Keep the existing disk-space cleanup
- Continue generating LCOV for Codecov, Coveralls, artifacts, and gluesql.org publishing

## Summary by CodeRabbit

* **Chores**
  * Updated the automated code coverage workflow to generate LCOV coverage using a `cargo llvm-cov`-based process instead of the previous approach.
  * Improved coverage report generation with updated filename-ignore behavior and consistent output to `coverage/lcov.info`.
  * Enhanced cleanup of temporary coverage/build artifacts and reduced leftover cached data after the workflow completes.


devgony (approved):
👍🏼 👍🏼 👍🏼

PR #1944, opened 2026-07-14
Wait for sled lock release when opening storage

panarch:
Sled may keep the database file locked briefly after a flush while its I/O worker finishes. Wait up to one second for the lock to clear instead of failing immediate reopen operations.

Apply the same behavior to regular and migration opens, reduce repeated opens in migration tests, and cover lock release, timeout, and unrelated errors deterministically.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database startup reliability by adding lock-aware retries when the storage is temporarily locked.
  * Migration now uses the same lock-aware database opening behavior for consistent results.
* **Tests**
  * Expanded coverage for lock contention, retry timing, timeout behavior (including capped waiting), and ensuring non-lock errors fail immediately.

### DIFF
--- a/storages/sled-storage/tests/migration_v1_to_v2.rs
+++ b/storages/sled-storage/tests/migration_v1_to_v2.rs
@@ -13,12 +13,16 @@ use {
     std::{
         collections::BTreeMap,
         fs::{create_dir_all, remove_dir_all, remove_file, write},
+        io::ErrorKind,
         path::{Path, PathBuf},
-        time::{SystemTime, UNIX_EPOCH},
+        thread,
+        time::{Duration, Instant, SystemTime, UNIX_EPOCH},
     },
 };
 
 const STORAGE_FORMAT_VERSION_KEY: &str = "__GLUESQL_STORAGE_FORMAT_VERSION__";
+const LOCK_WAIT_TIMEOUT: Duration = Duration::from_secs(1);
+const LOCK_WAIT_INTERVAL: Duration = Duration::from_millis(10);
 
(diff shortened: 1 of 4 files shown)


storages/sled-storage/tests/migration_v1_to_v2.rs
```diff
@@ -80,11 +84,106 @@ fn data_key(table_name: &str, key: Vec<u8>) -> Vec<u8> {
...
+    open_sled_with_lock_wait_by(
+        || config.open(),
+        Instant::now,
+        thread::sleep,
+        LOCK_WAIT_TIMEOUT,
+        LOCK_WAIT_INTERVAL,
+    )
+}
+
+fn open_sled_with_lock_wait_by<T>(
```
devgony: This duplicates the lock-wait retry logic from `src/open.rs`, and the three tests
below validate this copied helper rather than the production implementation.

The end-to-end migration tests that reopen sled are useful and should stay, but the
mocked retry/timeout/non-lock-error cases are already covered by the unit tests in
`src/open.rs`. Could we remove this duplicated helper and those three tests, keeping
only the integration-level `open_sled` fixture?

panarch: Thanks for pointing this out. Those mocked cases were only exercising the copied fixture helper, not the production implementation.

I removed the duplicated helper and tests. Raw sled is now opened directly only when creating a fresh fixture path, while reopens after migration go through `SledStorage::try_from` and use the production lock-wait path. The end-to-end migration tests remain.

devgony (approved):
👍👍👍

PR #1963, opened 2026-07-25
Fix CompositeStorage rollback delegation

### LINKED ISSUE #1962: CompositeStorage::rollback() incorrectly calls commit() on inner storages
## Description

While working on #1959, I found that `CompositeStorage::rollback()` calls `commit()` on each inner storage instead of `rollback()`.

The current implementation is:

```rust
fn rollback(&mut self) -> Result<()> {
    for storage in self.storages.values_mut() {
        storage.commit()?;  
    }

    Ok(())
}
```

As a result, when the executor attempts to roll back a transaction after an error, transactional inner storages may commit their changes instead of reverting them.

## Expected behavior

`CompositeStorage::rollback()` should delegate the rollback operation to every inner storage:

```rust
fn rollback(&mut self) -> Result<()> {
    for storage in self.storages.values_mut() {
        storage.rollback()?;  
    }

    Ok(())
}
```

seosangwon:
## Summary

- Delegate `CompositeStorage::rollback()` to `rollback()` on each inner storage instead of `commit()`.

Closes #1962

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed transaction rollback so underlying storage operations are correctly rolled back instead of committed.

### DIFF
--- a/storages/composite-storage/src/transaction.rs
+++ b/storages/composite-storage/src/transaction.rs
@@ -23,7 +23,7 @@ impl Transaction for CompositeStorage {
 
     fn rollback(&mut self) -> Result<()> {
         for storage in self.storages.values_mut() {
-            storage.commit()?;
+            storage.rollback()?;
         }
 
         Ok(())


panarch (approved):
Good catch on this long-standing rollback bug.
The fix looks correct and is appropriately scoped.

PR #1968, opened 2026-07-26
Stream Parquet rows instead of loading them all at once

### LINKED ISSUE #1967: Stream Parquet scans instead of materializing the full table
## Background

`ParquetStorage::scan_data` currently creates a Parquet row iterator, decodes every record into GlueSQL values, collects all decoded rows into a `Vec`, and only then returns a `RowIter`.

This defeats the lazy `Store::scan_data` interface:

```text
Parquet file
  -> decode all rows
  -> collect all GlueSQL rows into Vec
  -> return RowIter
  -> executor applies query operators
```

Consequently, `LIMIT` and early primary-key lookups cannot stop file decoding early, and peak memory grows with the total table size.

## Proposed change

Keep the Parquet reader alive inside the returned iterator by constructing an owning Parquet `RowIter` with `RowIter::from_file_into(Box<dyn FileReader>)`.

The storage iterator should convert each Parquet record to `(Key, Vec<Value>)` only when GlueSQL consumes its next item:

```text
Parquet row iterator
  -> decode one record on demand
  -> convert it to a GlueSQL row
  -> return it to the executor
```

The change is internal to `gluesql-parquet-storage`; it does not require a `Store` trait, planner, or executor API change.

A local prototype confirms that avoiding full-table materialization substantially improves early-stop queries. Detailed benchmark methodology and results will be included in the implementing PR.

## Acceptance criteria

- [ ] `ParquetStorage::scan_data` does not collect all rows into a `Vec` before returning.
- [ ] The returned iterator owns the Parquet reader for its full lifetime.
- [ ] Structured-table scans preserve row values and primary-key behavior.
- [ ] Schemaless-table scans preserve existing behavior.
- [ ] `fetch_data` can stop consuming rows after finding a requested key.
- [ ] Existing Parquet storage tests continue to pass.
- [ ] The implementing PR documents repeatable performance measurements for early-stop and full-scan queries.

## Out of scope

- Column projection pushdown.
- Predicate/row-group pushdown using Parquet statistics.
- Arrow batch-based execution.
- Changing Parquet storage's full-file rewrite behavior for mutations.
- Changes to the core planner or the public `Store` API.

## Related issues

- #242 — `Store::scan_data` iterator implementation discussion
- #131 — partial-column fetching discussion
- #1910 — operator-oriented execution plans and scan access paths
- #1957 — benchmark and performance-regression infrastructure

dgg1dbg:
## Summary

- Stream Parquet records directly into GlueSQL's `RowIter` instead of materializing the full table in a `Vec`.
- Keep the Parquet reader alive for the iterator lifetime with the owning Parquet row iterator API.
- Determine the primary-key column once per scan and avoid the previous extra `Value` clone.

This preserves structured and schemaless scan behavior while allowing consumers to stop reading early.

Closes #1967

## Benchmark

Release-mode warm-cache measurements used a 250,000-row, 5.8 MiB Parquet file. Each result is the median of repeated queries that open storage and execute the query.

| Query | Before | After | Change |
| --- | ---: | ---: | ---: |
| `SELECT * FROM Foo LIMIT 10` | 71.0 ms | 1.11 ms | 64.0x faster |
| `SELECT * FROM Foo WHERE id = 1` | 71.0 ms | 1.14 ms | 62.3x faster |
| `SELECT COUNT(*) FROM Foo` | 83.9 ms | 74.9 ms | 10.7% faster |
| `SELECT * FROM Foo WHERE id = 249999` | 71.3 ms | 60.0 ms | 15.9% faster |

## Validation

- `cargo fmt --all`
- `cargo test -p gluesql-parquet-storage`

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved Parquet data scanning to process records incrementally, avoiding upfront loading of all rows into memory.
  * Reduced memory usage for large datasets.

* **Bug Fixes**
  * Improved row and key handling when a primary-key column is present.
  * Ensured consistent scanning behavior for data without predefined column definitions.

* **Tests**
  * Added coverage to verify that schemaless value conversion issues are surfaced correctly during scanning.

### DIFF
--- a/storages/parquet-storage/src/lib.rs
+++ b/storages/parquet-storage/src/lib.rs
@@ -8,8 +8,8 @@ use {
         store::{Metadata, Planner, Store},
     },
     parquet::{
-        file::{reader::FileReader, serialized_reader::SerializedFileReader},
-        record::Row,
+        file::serialized_reader::SerializedFileReader,
+        record::{Row, reader::RowIter as ParquetRowIter},
     },
     std::{
         collections::BTreeMap,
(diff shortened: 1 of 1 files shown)


panarch (approved):
Nice improvement. Both structured and schemaless scans now avoid materializing the entire file while preserving the existing behavior.
Looks all nice!

Issue #1684, opened 2025-07-27
Tracking Unimplemented SQL Functions and Features


ever0de:
This list can be used as a roadmap for enhancing our SQL feature set. Community contributions for any of these features are welcome. Each feature could potentially be broken out into its own dedicated issue.

# A Note on Contributions:
The list below is currently based on SQLite compatibility. However, we are not strictly limited to functions supported by SQLite. We welcome contributions for commonly used functions from other major databases like PostgreSQL, MySQL, etc.
Furthermore, you are welcome to propose or implement useful functions that are not on this list.

You can refer to the [test-suite](https://www.google.com/url?sa=E&q=https%3A%2F%2Fgithub.com%2Fgluesql%2Fgluesql%2Ftree%2Fmain%2Ftest-suite%2Fsrc) to easily check if a function is already implemented.

### Legend
- ⚠️ Partial Support
- ❌ Not Supported

---

### Expressions

| Syntax | Status | Notes |
| --- | :---: | --- |
| `CAST (expr AS type)` | ⚠️ | `TRY_CAST` and `SAFE_CAST` are not supported and will cause a translation error. Syntax for specifying a particular format during a `CAST` is also not supported. |
| `COLLATE` | ⚠️ | The `COLLATE` clause itself is parsed, but user-defined `COLLATION` rules are not supported. |
| `(NOT) REGEXP` | ❌ | The `REGEXP` operator is not supported. |
| `agg() FILTER (WHERE ...)` | ❌ | The `FILTER` clause within aggregate functions is not supported. |
| `... OVER (...)` | ❌ | Window functions are not supported. |
| `RAISE()` | ❌ | The `RAISE()` function is not supported. |

### SQL Statements

| Syntax | Status | Notes |
| --- | :---: | --- |
| `SELECT ... DISTINCT` | ❌ | The `SELECT DISTINCT` syntax is not supported and will cause a translation error. |

#### Scalar Functions

| Function | Status | Notes |
| --- | :---: | --- |
| `changes()` | ❌ | |
| `char(X1,X2,...,XN)` | ⚠️ | The `CHR` function only supports a single integer argument. Using multiple arguments will cause an error. |
| `format(FORMAT,...)` | ❌ | C-style `printf` formatting is not supported. For date/time formatting, use the `FORMAT` function. |
| `glob(X,Y)` | ❌ | |
| `hex(X)` | ❌ | |
[...]

PR #1969, opened 2026-07-27
Support window functions in SELECT queries

asish-singh:
This adds window function support, the `... OVER (...)` row listed as not yet supported in the tracking issue #1684.

A window function computes a value for every row of the result by looking at a set of related rows chosen with an `OVER` clause, without collapsing those rows the way `GROUP BY` does.

## What is supported

Ranking functions `ROW_NUMBER()`, `RANK()` and `DENSE_RANK()`.

Offset functions `LAG(expr [, offset [, default]])` and `LEAD(expr [, offset [, default]])`. The offset defaults to one, and the optional default value is returned when the target row falls outside the partition, otherwise such rows yield `NULL`.

The aggregates `COUNT`, `SUM`, `MIN`, `MAX` and `AVG` used as window functions. `NULL` handling is the same as the existing aggregates, since both now share one accumulator.

`OVER` takes an optional `PARTITION BY` expression list and an optional `ORDER BY` list. `PARTITION BY` splits the rows into independent groups, rows whose partition expressions are all equal belong together, and rows with `NULL` partition values group the way `GROUP BY` already groups them. Omitting it treats the whole result as one partition. `ORDER BY` inside `OVER` uses the same `ASC` and `DESC` behavior as query level `ORDER BY`, and rows that tie on every ordering expression are peers.

Window functions are accepted in the `SELECT` projection, including inside larger expressions and under column aliases.

## Running aggregate semantics

An aggregate with no `ORDER BY` in its `OVER` clause computes over the whole partition. With `ORDER BY` it computes a running value, covering the partition from its first row through the current row and all of the current row's peers, so peers always share the same result. With no `ORDER BY` every row of a partition is a peer of every other, which is what makes those two rules consistent with each other.

## Errors

The following are reported as errors rather than silently accepted.

- A window function combined with `GROUP BY`, `HAVING` or `SELECT DISTINCT` in the same query.
- A window function nested inside another window function or inside a regular aggregate.
- A window function outside the `SELECT` projection.
- A `LAG` or `LEAD` offset that is negative, fractional, or not a literal.
- `DISTINCT` inside a window aggregate.
- An explicit frame clause such as `ROWS` or `RANGE` inside `OVER`, which is out of scope here.

## Implementation notes

The work sits in the ordinary query pipeline, so `WHERE` filtering, joins, projections mixing window and plain columns, several window functions with different `OVER` clauses in one query, and query level `ORDER BY` over the finished result all behave as expected.

The first commit is a pure refactor with no behavior change. It moves the `AggrValue` accumulator out of the `GROUP BY` state module so window aggregates and grouped aggregates share one implementation and cannot drift apart on `NULL` handling.

## Tests

Eight new SQL fixtures under `test-suite/fixtures/window` cover `ROW_NUMBER`, `RANK` and `DENSE_RANK` including ties, aggregates over a whole partition, running aggregates and peer sharing, `LAG` and `LEAD` with offsets and defaults, partitioning including `NULL` partition keys and the single partition case, window functions mixed with plain columns and joins and query level `ORDER BY`, and window functions inside larger expressions and under aliases.

A rust case in `test-suite/src/window_error.rs` covers the error paths above and first asserts that a well formed window query succeeds, so a failure there cannot be mistaken for the feature simply being unsupported.

A documentation page was added under the querying section of the SQL syntax guide.

`cargo test -p gluesql-core`, `cargo test -p gluesql_memory_storage --test memory_storage`, `cargo clippy --all-targets -- -D warnings` and `cargo fmt --all` are all clean.

## Summary by CodeRabbit

* **New Features**
  * Added support for SQL window functions, including `ROW_NUMBER`, `RANK`, `DENSE_RANK`, `LAG`, `LEAD`, and windowed aggregates.
  * Supports `PARTITION BY`, `ORDER BY`, running aggregates, ranking, offsets, and default values.
  * Added validation for unsupported or invalid window-function usage.
* **Documentation**
  * Added SQL window-function documentation with syntax, examples, and usage restrictions.
* **Tests**
  * Added comprehensive coverage for ranking, partitions, aggregates, offsets, ordering, joins, and validation errors.


panarch:
Thank you for putting together such a substantial implementation. This PR is valuable in showing that the window-function behavior works and in establishing the relevant test cases.
After reviewing the structure more closely, I think we need to complete some related planning and execution-layer groundwork before we can merge window functions. In particular, we are working toward binding column references during planning (#1978) and representing query execution as explicit typed stages (#1937). These are not window features themselves, but they should give us the right place to represent window evaluation without adding more special-case responsibilities to the current execution path.
I’m not asking you to take on those prerequisite changes. Would you be willing to keep this PR open and wait for that groundwork to land? Once it is ready, we can revisit this implementation and work out how to adapt it to the new structure. Some restructuring may be necessary, but the implementation and tests here should remain a useful foundation.

asish-singh:
Thank you for the kind words and for taking the time to review such a large PR so carefully. I really appreciate that you would rather keep this open than close it.

Waiting for #1978 and #1937 sounds like the right plan, and I am glad to do so. Reading both, I can see why you want them in place first. The window implementation currently has to reach into projection and sorting from the outside, and the structure you are building would give window evaluation a proper home.

While it waits, I will keep the branch rebased on main so it stays mergeable and the tests keep passing. Whenever the new structure is ready, I would be happy to adapt this implementation to it, and please feel free to reshape or borrow anything here in the meantime.

One question on the groundwork itself. I can see #1937 is already well underway on your side. If #1978 is not something you have already started, I would genuinely enjoy taking a run at it. The motivation you wrote is clear about the destination, resolving column references during planning so the executor stops interpreting names. If you are open to that, I would post a short design sketch in the issue first and only start once the direction has your nod, so nothing collides with anything you may already have in flight. And if you would rather keep it with the rest of the roadmap, that is completely fine too, everything above stands either way.

asish-singh:
Quick follow-up now that #1937 has merged. #1978 remains open and unassigned, and I'm still happy to take it and post a short design sketch first.

I also see that this branch now conflicts with `main`. Would you prefer that I rebase and adapt #1969 now, or wait until the direction for #1978 is settled so I don't restructure the window work twice?

panarch:
Thanks for following up, and sorry for the delay.

When I asked you to wait, I listed #1978 alongside #1937. Looking at it again, #1937 was the real blocker: window evaluation needed an explicit place in the execution pipeline, and that is now on `main`. I thought it would be good to have #1978 (and #1977, which it builds on) in place first, but that is going to take longer than I expected, and window functions don't really depend on it.

So rather than starting with #1978, please go ahead and rebase and adapt #1969 onto the current `main`. Thanks for offering to take #1978, but let's set it aside for now and focus on getting window functions in.

Once it's updated, I'll review it on its own so it can be merged independently of #1977 and #1978. Whichever of those lands after this PR will take care of updating the window function code at that point, so you don't need to worry about restructuring the window work twice.

PR #1970, opened 2026-07-27
Fix ambiguous column resolution across query scopes

### LINKED ISSUE #1965: Relation alias lookup falls back across shadowed scopes and silently accepts duplicate aliases
Hello! While working on #1407, I found a broader relation-alias correctness issue involving nested query scopes and duplicate aliases.

When the same alias exists in an inner and outer query, a missing column on the inner relation can fall back to the outer relation. Duplicate relation aliases within the same `FROM` scope are also accepted and silently resolved to one side.

If the issue and proposed direction below make sense, would you be open to me submitting a PR to address it?

My plan would be to add regression tests and update alias lookup so that:

- an alias in the current scope shadows aliases from outer scopes completely;
- duplicate relation aliases in the same scope are rejected;
- correlated references continue to work when the alias is not defined in the current scope.

## Reproduction

Tested with GlueSQL `0.19.0` using `MemoryStorage`.

```sql
CREATE TABLE Foo (
    id INTEGER PRIMARY KEY,
    name TEXT
);

CREATE TABLE Bar (
    id INTEGER PRIMARY KEY,
    title TEXT
);

INSERT INTO Foo VALUES
    (1, 'foo-one'),
    (2, 'foo-two');

INSERT INTO Bar VALUES
    (1, 'bar-one'),
    (2, 'bar-two');
```

### 1. Inner alias falls back to the outer alias

```sql
SELECT (
    SELECT A.name
    FROM Bar AS A
    LIMIT 1
)
FROM Foo AS A;
```

The inner `A` refers to `Bar`, which does not have a `name` column.

Expected:

```text
identifier not found: A.name
```

Actual:

```text
foo-one
foo-two
```

`A.name` is resolved from the outer `Foo AS A`, even though the inner scope already defines `A`.

A correlated reference should only search the outer scope when the alias itself is not bound in the current scope. Once the inner scope defines `A`, it should shadow the outer `A` regardless of whether the requested column exists.

### 2. Duplicate aliases in a JOIN are accepted

```sql
SELECT A.id
FROM Foo AS A
JOIN Bar AS A ON A.id = A.id;
```

Expected:

```text
error: duplicate relation alias A
```

Actual:

```text
id
1
2
```

Both relations use the same alias, but the query succeeds and `A.id` is silently resolved to one of them.

### 3. Alias-less self JOIN is accepted

```sql
SELECT Foo.id
FROM Foo
JOIN Foo ON Foo.id = Foo.id;
```

Expected:

```text
error: duplicate or ambiguous relation identifier Foo
```

Actual:

```text
id
1
2
```

Because both table factors expose the same relation identifier, `Foo.id` cannot uniquely identify one side of the JOIN.

For comparison, GlueSQL already detects an ambiguous unqualified column when the relation names are distinct:

```sql
SELECT id
FROM Foo
JOIN Bar ON Foo.id = Bar.id;
```

This correctly fails during planning with:

```text
column reference id is ambiguous, please specify the table name
```

## Cause

The executor's `RowContext::get_alias_value` continues searching the parent context when it finds a matching alias but cannot find the requested column:

```rust
if *table_alias == target_table_alias {
    let value = row.get_value(target);

    if value.is_some() {
        value
    } else {
        next.as_ref()
            .and_then(|context| {
                context.get_alias_value(target_table_alias, target)
            })
    }
}
```

This does not distinguish between:

1. the alias was not found in the current scope; and
2. the alias was found, but the column does not exist on that relation.

Only the first case should search an outer query scope.

For JOIN contexts, alias lookup across `RowContext::Bridge` uses `left.or_else(right)`. Therefore, when both relations expose the same alias, the left relation wins instead of reporting ambiguity.

The planning context has similar boolean lookup behavior: it records whether an aliased column exists, but not whether an alias is shadowed or appears more than once in the current scope.

This was discovered while extending `check_table_factor` for #1407, but the problem is broader than table-factor evaluability because it affects runtime identifier lookup and relation validation as well.

## Proposed direction

- Represent alias lookup with enough information to distinguish:
  - alias not found;
  - alias found with the requested column;
  - alias found without the requested column;
  - alias is ambiguous within the current scope.
- Stop searching parent query scopes as soon as an alias binding is found.
- Reject duplicate explicit aliases within one `FROM` scope.
- Reject alias-less self JOINs whose relation identifiers cannot be distinguished.
- Preserve valid correlated subqueries when the referenced alias is absent from the inner scope.

Regression tests should cover:

- valid correlated subqueries;
- valid inner alias shadowing;
- missing columns on a shadowing alias;
- duplicate explicit JOIN aliases;
- alias-less self JOINs;
- ambiguous unqualified JOIN columns.

red-sprout:
## Summary

- detect ambiguous unqualified column references within the same query scope
- reject duplicate relation identifiers in joins
- preserve duplicate output column names from derived tables and positional aliases
- enforce alias shadowing and correlated subquery scope resolution
- resolve simple `ORDER BY` names against output aliases first
- retain input-column resolution for `ORDER BY` expressions
- prevent index ordering optimization from consuming output-alias ordering
- detect schemaless column ambiguity at runtime

## Testing

- `cargo fmt --all`
- `cargo test -p gluesql-core --lib`
- `cargo test -p gluesql_memory_storage --test memory_storage sql_join`
- `cargo test -p gluesql_sled_storage --test sled_storage sql_index`
- `cargo test -p gluesql_sled_storage --test sled_storage sql_join`
- changed-line coverage: 100%
- changed-branch coverage: 100%

Closes #1965

## Summary by CodeRabbit

* **New Features**
  * Enhanced SQL identifier/alias resolution for joins, nested queries, and schemaless data.
  * Added clearer errors for ambiguous identifiers and duplicate relation identifiers.
* **Bug Fixes**
  * Correct handling of missing vs ambiguous columns (including better NULL vs “not found/ambiguous” behavior).
  * Fixed `ORDER BY` behavior when ordering by projected output aliases, including preventing incorrect index selection.
  * Improved sorting behavior and correlated subquery alias scoping.
* **Tests**
  * Expanded coverage for ambiguity, alias shadowing, and schemaless JOIN scenarios.

### DIFF
--- a/core/src/executor/context.rs
+++ b/core/src/executor/context.rs
@@ -3,5 +3,5 @@ mod row_context;
 
 pub use {
     aggregate_context::{AggregateContext, AggregateValues},
-    row_context::RowContext,
+    row_context::{RowContext, ValueLookup},
 };
--- a/core/src/executor/context/row_context.rs
+++ b/core/src/executor/context/row_context.rs
@@ -124,32 +124,37 @@ impl<'a> RowContext<'a> {
 
     pub fn get_alias_value(&'a self, target_table_alias: &str, target: &str) -> Option<&'a Value> {
         match self.lookup_alias_value(target_table_alias, target) {
-            AliasLookup::Bound(value) => value,
-            AliasLookup::Unbound => None,
+            ValueLookup::Found(value) => Some(value),
+            ValueLookup::Unbound | ValueLookup::Missing | ValueLookup::Ambiguous => None,
         }
     }
 
-    fn lookup_alias_value(
+    pub fn lookup_alias_value(&'a self, target_table_alias: &str, target: &str) -> ValueLookup<'a> {
+        match self.lookup_bound_alias_value(target_table_alias, target) {
+            AliasLookup::Bound(value) => value,
+            AliasLookup::Unbound => ValueLookup::Unbound,
+        }
+    }
+
+    fn lookup_bound_alias_value(
         &'a self,
         target_table_alias: &str,
         target: &str,
-    ) -> AliasLookup<Option<&'a Value>> {
+    ) -> AliasLookup<ValueLookup<'a>> {
         match self {
             Self::Data {
                 table_alias, row, ..
             } if *table_alias == target_table_alias => {
-                let value = match lookup_row_value(row, target) {
-                    ValueLookup::Found(value) => Some(value),
-                    ValueLookup::Unbound | ValueLookup::Missing | ValueLookup::Ambiguous => None,
-                };
-                AliasLookup::Bound(value)
+                AliasLookup::Bound(lookup_row_value(row, target))
             }
             Self::Data {
                 next: Some(next), ..
-            } => next.lookup_alias_value(target_table_alias, target),
+            } => next.lookup_bound_alias_value(target_table_alias, target),
             Self::Bridge { left, right } => {
-                match left.lookup_alias_value(target_table_alias, target) {
-                    AliasLookup::Unbound => right.lookup_alias_value(target_table_alias, target),
+                match left.lookup_bound_alias_value(target_table_alias, target) {
+                    AliasLookup::Unbound => {
+                        right.lookup_bound_alias_value(target_table_alias, target)
+                    }
                     bound @ AliasLookup::Bound(_) => bound,
                 }
             }
@@ -304,6 +309,20 @@ mod tests {
         assert!(matches!(
             scope.lookup_value("id"),
             ValueLookup::Found(Value::I64(0))
+        ));
+    }
+
+    #[test]
+    fn qualified_lookup_preserves_missing_document_key() {
+        let context = document_context("A", &[], None);
+
+        assert!(matches!(
+            context.lookup_alias_value("A", "id"),
+            ValueLookup::Missing
+        ));
+        assert!(matches!(
+            context.lookup_alias_value("B", "id"),
+            ValueLookup::Unbound
         ));
     }
 
--- a/core/src/executor/context/row_context.rs
+++ b/core/src/executor/context/row_context.rs
@@ -1,5 +1,5 @@
 use {
-    crate::data::{Row, Value},
+    crate::data::{Row, SCHEMALESS_DOC_COLUMN, Value},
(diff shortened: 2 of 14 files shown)


panarch (changes requested):
Thanks for the thorough work on this. The issue cases and regression tests are valuable, and I agree that alias shadowing, duplicate relation identifiers, and correlated references need to be handled correctly.

While reviewing the implementation, I started to think that we may be addressing part of this at the wrong layer.

`plan::validate` now understands query scopes well enough to detect ambiguity and shadowing, but that resolution result is discarded because validation only returns `Result<()>`. The executor then receives unresolved `ExprPlan::Identifier` values and performs name resolution again through `RowContext` and `ValueLookup`.

This means the planner and executor each maintain their own scope-resolution rules, which makes it easy for them to disagree and pushes more planning responsibility into the execution layer.

Now that we have separate `StatementPlan` and `ExprPlan` structures, I think we have a good opportunity to do a small prerequisite improvement before merging this behavior:

1. Add a resolved column-reference representation to `ExprPlan`.
2. Change the name-validation step into a binding step that records which relation and column each schemaful identifier refers to.
3. Make the executor consume that binding instead of searching query scopes again.
4. Keep runtime lookup only for genuinely dynamic cases such as schemaless columns.

I am not suggesting that we redesign the whole planner in this PR. The first prerequisite PR could cover only schemaful identifier binding while preserving existing behavior. The current `Scope` work and most of the regression tests here should still be reusable.

After that, this PR could be split or rebased on top of the prerequisite work and focus on the actual behavior changes: duplicate relation rejection, alias shadowing, correlated references, and dynamic schemaless ambiguity.

Would you be open to treating this PR as a design discussion for now and working through those prerequisite PRs first?

red-sprout:
@panarch Thanks for the thoughtful review. I agree with this direction.

While working on this PR, I also felt that resolving identifiers separately in the planner and executor was an intermediate design, and I ended up reworking that area several times. As you pointed out, `plan::validate` already determines much of the resolution information but discards it, while the executor reconstructs it through `RowContext`.

I opened #1976 to discuss and track the prerequisite work for preserving schemaful identifier bindings in the execution plan while maintaining existing behavior.

I’m happy to keep this PR as a design discussion for now and work through the prerequisite first.

PR #1972, opened 2026-07-28
Reject TEMPORARY/LIKE/CLONE options on CREATE TABLE

### LINKED ISSUE #1973: CREATE TABLE silently ignores TEMPORARY/LIKE/CLONE options
**Description**

`CREATE TABLE` silently ignores the `TEMPORARY`, `LIKE`, and `CLONE` options. The parser accepts them, but the translate step drops the fields via a catch-all `..` in the `Statement::CreateTable` match arm, so the statement executes as if the options were never specified — no error, no actual TEMPORARY/LIKE/CLONE semantics.

Since GlueSQL doesn't implement TEMPORARY tables, `LIKE`, or `CLONE`, users end up with a table created silently, without any indication that the clause they specified was ignored.

**Steps to Reproduce**

```sql
CREATE TEMPORARY TABLE Foo (id INTEGER);
-- creates a regular (non-temporary) table, no error

CREATE TABLE Foo (id INTEGER);
CREATE TABLE Bar LIKE Foo;
-- creates an empty, unrelated table instead of applying LIKE semantics, no error

CREATE TABLE Baz CLONE Foo;
-- same issue, no error
```

**Expected Behavior**
Since these options aren't supported, GlueSQL should return an explicit error at translate time — consistent with the existing UnsupportedInsertOption / UnsupportedUpdateOption pattern — instead of silently ignoring the clause.

**Actual Behavior**
The table is created successfully with none of the TEMPORARY / LIKE / CLONE semantics applied, and no error or warning is surfaced to the user.

**Additional Context**
Relevant code: core/src/translate.rs, in the Statement::CreateTable translate arm — the temporary, like, clone fields are currently discarded by .. 
A fix is up in [#1972](https://github.com/gluesql/gluesql/pull/1972).

sweetpark:
**Summary**

* Fixed an issue where `TEMPORARY` / `LIKE` / `CLONE` options on `CREATE TABLE` were silently ignored
* If any of the three options is specified, the translate step now returns `TranslateError::UnsupportedCreateTableOption`
* Since GlueSQL doesn't support these options, we now surface a clear error instead of silently ignoring them, following the same pattern as the existing `UnsupportedInsertOption` / `UnsupportedUpdateOption`

**Changes**

* `core/src/translate.rs`: In the `CreateTable` translate logic, explicitly check the `temporary` / `like` / `clone` fields and return early on violation. Added a related unit test (`create_table_options_not_supported`)
* `core/src/translate/error.rs`: Added the `UnsupportedCreateTableOption(&'static str)` error variant
* `test-suite/fixtures/alter/create_table.sql`: Added 3 fixture test cases, one each for `TEMPORARY TABLE`, `LIKE`, and `CLONE`

**Test plan**

* [x] `cargo test -p gluesql-core translate` (unit tests)
* [x] Confirmed core-related tests pass
* [x] `cargo clippy --all-targets -- -D warnings`
* [x] `cargo fmt --all` check

Fixes #1973


panarch (approved):
Looks all good to me. This correctly rejects `TEMPORARY`, `LIKE`, and `CLONE`
instead of silently discarding them, with coverage in both the translation
tests and SQL fixtures.

The broader stringly typed unsupported-option error pattern is tracked
separately in #1975 and does not block this change.

PR #1980, opened 2026-07-31
Use typed enums for unsupported SQL options in translate errors

### LINKED ISSUE #1975: Use typed enums for unsupported SQL options in translate errors
## Description

Several `TranslateError` variants represent unsupported SQL options using `&'static str` payloads:

```rust
UnsupportedInsertOption(&'static str),
UnsupportedUpdateOption(&'static str),
UnsupportedDeleteOption(&'static str),
UnsupportedQueryOption(&'static str),
UnsupportedSelectOption(&'static str),
```

The values passed to these variants are known, finite sets of SQL options. However, representing them as strings means the compiler cannot validate the available cases.

## Motivation

The current representation has several limitations:

- option-name typos are not detected by the compiler;
- downstream users cannot exhaustively match individual options;
- adding or removing an unsupported option is not reflected in the type definition;
- human-readable messages and serialized error values are coupled together.

Although `&'static str` is safe for the current string-literal values, typed enums would express the domain more explicitly.

## Proposed Direction

Introduce a typed enum for each statement category and derive its human-readable representation using the existing `strum_macros::Display` dependency.

For example:

```rust
use {
    serde::Serialize,
    strum_macros::Display,
};

#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Display)]
pub enum InsertOption {
    #[strum(to_string = "RETURNING clause")]
    Returning,

    #[strum(to_string = "ON CONFLICT clause")]
    OnConflict,

    #[strum(to_string = "table alias")]
    TableAlias,

    #[strum(to_string = "PARTITION clause")]
    Partition,

    #[strum(to_string = "OVERWRITE clause")]
    Overwrite,

    #[strum(to_string = "TABLE keyword")]
    TableKeyword,
}
```

The corresponding error variant could carry the typed value:

```rust
#[error("unsupported INSERT option: {0}")]
UnsupportedInsertOption(InsertOption),
```

The same approach can be applied consistently to unsupported options for `UPDATE`, `DELETE`, queries, `SELECT`, and other statement types.

## Acceptance Criteria

- Unsupported SQL options are represented by typed enums rather than strings.
- Human-readable error messages use the existing `strum_macros::Display` derive pattern.
- Unsupported-option errors across the translate layer are migrated consistently.
- Relevant translation tests and SQL fixtures are updated.

sweetpark:
**Summary**

`TranslateError`'s "unsupported option" variants used `&'static str`/
`String` to describe the rejected clause, which meant typos in the
reason weren't caught by the compiler and there was no single place
listing every valid value. This replaces those fields with dedicated
enums deriving `strum::Display`, so the compiler now checks the option
names and error messages stay auto-derived instead of hand-typed at
every call site.

Converted 7 variants total:

- `UnsupportedInsertOption` → `InsertOption`
- `UnsupportedUpdateOption` → `UpdateOption`
- `UnsupportedDeleteOption` → `DeleteOption`
- `UnsupportedQueryOption` → `QueryOption`
- `UnsupportedSelectOption` → `SelectOption`
- `UnsupportedCreateTableOption` → `CreateTableOption`
- `UnsupportedJoinConstraint` → `JoinConstraintReason`

The issue only listed the first 5 as examples. `UnsupportedCreateTableOption`
(added in #1972) and `UnsupportedJoinConstraint` have the exact same
closed-set-of-strings pattern, so I included them too rather than
leaving inconsistent holdouts.

`strum_macros` was already a dependency, so no new crates were added.
`Display` output is byte-for-byte unchanged.

`Serialize` is implemented by delegating to `Display` via
`serializer.collect_str(self)` (through a small module-private
`macro_rules!` in `error.rs`), instead of deriving `Serialize` and
repeating each variant's string a second time in `#[serde(rename)]`.
This keeps JSON output identical while avoiding the exact kind of
duplication that caused a regression during development (see test
plan below) — one string per variant instead of two that can drift
out of sync.

**Test plan**

- [x] `cargo test -p gluesql-core --lib` — all 515 tests pass
- [x] `cargo test -p gluesql_memory_storage --test memory_storage` — all
      196 fixture-based integration tests pass. All 13 storage crates
      run the exact same fixtures via `generate_store_tests!`, and
      `translate()` runs before any storage is touched, so this one
      backend's pass is representative of all of them.
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [x] `cargo fmt --all`
- [x] `cargo build --all-features` — clean
- [x] Manually verified via `gluesql-cli` and `serde_json::to_string`
      that each of the 7 option categories produces the same Display
      message and JSON output as before the refactor (e.g.
      `unsupported INSERT option: RETURNING clause`,
      `{"UnsupportedCreateTableOption":"TEMPORARY clause"}`)

Closes #1975

## Summary by CodeRabbit

* **Improvements**
  * SQL translation errors now provide structured, consistently categorized reasons for unsupported options.
  * Enhanced diagnostics cover unsupported INSERT, UPDATE, DELETE, SELECT, query, CREATE TABLE, and join options.
  * Error messages remain clear and readable while offering more reliable values for integrations.
* **Bug Fixes**
  * Improved reporting for unsupported `USING` and `NATURAL` join constraints.
* **Tests**
  * Expanded coverage for structured error details across SQL translation scenarios.

### DIFF
--- a/core/src/translate.rs
+++ b/core/src/translate.rs
@@ -209,7 +209,7 @@ pub fn translate_with_params(
             } else if like.is_some() {
                 Some(CreateTableOption::Like)
             } else if clone.is_some() {
-                Some(CreateTableOption::Clone)
+                Some(CreateTableOption::CloneTable)
             } else {
                 None
             };
@@ -686,7 +686,7 @@ mod tests {
             ),
             (
                 "CREATE TABLE Foo CLONE Bar",
-                TranslateError::UnsupportedCreateTableOption(CreateTableOption::Clone),
+                TranslateError::UnsupportedCreateTableOption(CreateTableOption::CloneTable),
             ),
         ];
 
--- a/core/src/translate.rs
+++ b/core/src/translate.rs
@@ -12,8 +12,8 @@ pub use self::{
     data_type::translate_data_type,
     ddl::translate_column_def,
     error::{
-        CreateTableOption, DeleteOption, InsertOption, QueryOption, SelectOption, TranslateError,
-        UpdateOption,
+        CreateTableOption, DeleteOption, InsertOption, JoinConstraintReason, QueryOption,
+        SelectOption, TranslateError, UpdateOption,
     },
     expr::{translate_expr, translate_order_by_expr},
     param::{IntoParamLiteral, ParamLiteral},
--- a/core/src/translate.rs
+++ b/core/src/translate.rs
@@ -11,7 +11,10 @@ mod query;
 pub use self::{
     data_type::translate_data_type,
     ddl::translate_column_def,
-    error::TranslateError,
+    error::{
+        CreateTableOption, DeleteOption, InsertOption, QueryOption, SelectOption, TranslateError,
+        UpdateOption,
+    },
     expr::{translate_expr, translate_order_by_expr},
     param::{IntoParamLiteral, ParamLiteral},
     query::{alias_or_name, translate_query, translate_select_item},
@@ -74,17 +77,17 @@ pub fn translate_with_params(
             ..
         }) => {
             let violation = if returning.is_some() {
-                Some("RETURNING clause")
+                Some(InsertOption::Returning)
             } else if on.is_some() {
-                Some("ON CONFLICT clause")
+                Some(InsertOption::OnConflict)
             } else if table_alias.is_some() {
-                Some("table alias")
+                Some(InsertOption::TableAlias)
             } else if partitioned.is_some() {
-                Some("PARTITION clause")
+                Some(InsertOption::Partition)
             } else if *overwrite {
-                Some("OVERWRITE clause")
+                Some(InsertOption::Overwrite)
             } else if *table {
-                Some("TABLE keyword")
+                Some(InsertOption::TableKeyword)
             } else {
                 None
             };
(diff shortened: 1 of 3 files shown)


sweetpark:
re check

panarch:
The failing macro snapshot does not indicate a problem with this PR’s CREATE TABLE change.

The new `CreateTableOption::Clone` name only changes how rustc renders the unrelated `ToGlueRow` trait-bound diagnostic, from `Clone` to `std::clone::Clone`. Since trybuild compares the entire stderr output, that formatting difference causes the check to fail.

I opened #1981 to remove the derive macro’s incidental `Clone` requirement entirely. Once that lands, merging the latest main into this branch should resolve the failure without updating an unrelated snapshot here.

sweetpark:
Thanks for the pointer — I confirmed the same `Clone` naming collision on my end and pushed a fix.

I renamed `CreateTableOption::Clone` to `CreateTableOption::CloneTable` (`d2d7c116`). This follows the same naming style already used in `InsertOption` (e.g. `TableAlias`, `TableKeyword`, which describe the clause rather than reusing a bare keyword), so it's consistent with the rest of this PR rather than a one-off workaround. `Display`/JSON output is unchanged (`#[strum(to_string = "CLONE clause")]` stays the same), and all 12 checks are now passing on that commit, including `Run tests` and `coverage`.

Happy to revisit the naming once #1981 lands if you'd prefer something else.

panarch (approved):
Looks all nice!
Thanks a lot 👍 👍 👍

PR #1982, opened 2026-08-01
Fix schema dependency scanning for ORDER BY expressions

### LINKED ISSUE #1906: Include ORDER BY expressions in schema dependency scanning
`fetch_schema_map` currently scans query body, `LIMIT`, and `OFFSET`, but does not scan `ORDER BY` expressions.

This is a dependency scan gap because GlueSQL allows scalar subqueries inside `ORDER BY` expressions.

Example:

```sql
SELECT *
FROM Foo
ORDER BY (SELECT id FROM Bar LIMIT 1);
```

In this query, `Bar` is referenced only inside `ORDER BY`. The executor can evaluate this as a scalar subquery, but schema dependency scanning does not include `Bar`.

Expected behavior:

- `fetch_schema_map` should include schemas referenced from `QueryPlan.order_by`.
- The scanner should traverse each `OrderByExprPlan.expr` with the same expression scanner used for projection, selection, group by, having, `LIMIT`, and `OFFSET`.
- Add regression coverage for `ORDER BY` scalar subquery dependency scanning.

seosangwon:
## Summary

* Include schema dependencies referenced by `QueryPlan.order_by`.
* Scan every `OrderByExprPlan.expr` using the existing `scan_expr()` logic.
* Add regression tests for scalar subqueries in `ORDER BY`, including queries with multiple ordering expressions.

Closes #1906

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved query schema detection for schemas referenced in `ORDER BY` clauses.
  * Added support for subqueries and compound expressions in ordering logic.
  * Ensures schemas from `LIMIT` and `OFFSET` expressions are also recognized.

### DIFF
--- a/core/src/plan/schema.rs
+++ b/core/src/plan/schema.rs
@@ -97,9 +97,9 @@ fn scan_query<T: Store + ?Sized>(
 ) -> Result<HashMap<String, Schema>> {
     let QueryPlan {
         body,
+        order_by,
         limit,
         offset,
-        ..
     } = query;
 
     let schema_list = match body {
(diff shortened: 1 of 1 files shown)


panarch (approved):
Looks all nice!
Thanks a lot 👍 👍 👍

PR #1983, opened 2026-08-01
Reject unsupported table function arguments

### LINKED ISSUE #1904: Reject unsupported table function arguments during SQL translation
## Summary

`translate_table_factor` currently accepts table factors with function arguments in cases where GlueSQL does not support them, silently dropping the arguments.

For example, dictionary tables such as `GLUE_TABLES` do not accept arguments, but this can currently be translated as if the arguments were not present:

```sql
SELECT * FROM GLUE_TABLES(1);
```

This should be rejected instead of being treated like:

```sql
SELECT * FROM GLUE_TABLES;
```

The same concern applies to unknown table functions such as:

```sql
SELECT * FROM Foo(1);
```

If `sqlparser` represents this as `SqlTableFactor::Table { args: Some(_), .. }`, GlueSQL should not fall back to a plain table reference and discard the arguments.

## Expected Behavior

- `SERIES(...)` remains supported.
- Dictionary tables such as `GLUE_OBJECTS`, `GLUE_TABLES`, `GLUE_INDEXES`, and `GLUE_TABLE_COLUMNS` should only be accepted without arguments.
- Any unsupported table factor with `args: Some(_)` should return `TranslateError::UnsupportedQueryTableFactor`.

## Current Code

In `core/src/translate/query.rs`, dictionary table matches use `_` for `args`:

```rust
("GLUE_TABLES", _) => Ok(TableFactor::Dictionary { ... })
```

The fallback branch also accepts any remaining table factor as a plain table:

```rust
_ => Ok(TableFactor::Table { ... })
```

This can silently discard unsupported table arguments.

## Suggested Fix

Update the table factor match so dictionary tables match only `args == None`, and add an explicit `(_, Some(_))` rejection branch before the plain table fallback.

## Tests

Add translation or test-suite coverage for cases such as:

```sql
SELECT * FROM GLUE_TABLES(1);
SELECT * FROM GLUE_INDEXES(1);
SELECT * FROM Foo(1);
```

kwondo1017:
**Summary**

- Reject table factors with unsupported function arguments during translation.
- Allow dictionary tables only when no arguments are supplied.
- Add regression tests for dictionary and unknown table factors.

**Test plan**

- [x] `cargo fmt -- --check`
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `cargo build --all-features`
- [x] `cargo test -p gluesql-core`

Closes #1904

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unsupported arguments from being accepted for non-series table sources.
  * Preserved dictionary table support when used without arguments.
  * Enforced the requirement that `SERIES` receive exactly one argument.
  * Added clearer errors for invalid table-factor arguments, including dictionary tables, ordinary tables, and `SERIES`.

### DIFF
--- a/core/src/translate/query.rs
+++ b/core/src/translate/query.rs
@@ -261,10 +261,18 @@ fn translate_table_factor(
             let alias = translate_table_alias(alias.as_ref());
 
             match (object_name.as_str(), args) {
-                ("SERIES", Some(SqlTableFunctionArgs { args, .. })) => Ok(TableFactor::Series {
-                    alias: alias_or_name(alias, object_name),
-                    size: translate_table_args(args)?,
-                }),
+                ("SERIES", Some(SqlTableFunctionArgs { args, .. })) => {
+                    if args.len() > 1 {
+                        return Err(TranslateError::UnsupportedQueryTableFactor(
+                            sql_table_factor.to_string(),
+                        )
+                        .into());
+                    }
+                    Ok(TableFactor::Series {
+                        alias: alias_or_name(alias, object_name),
+                        size: translate_table_args(args)?,
+                    })
+                }
                 ("GLUE_OBJECTS", None) => Ok(TableFactor::Dictionary {
                     dict: Dictionary::GlueObjects,
                     alias: alias_or_name(alias, object_name),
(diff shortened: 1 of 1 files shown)


panarch (approved):
Looks all nice!
Thanks a lot 👍 👍 👍

PR #1987, opened 2026-08-03
Apply schemaless rewrites to CREATE TABLE AS SELECT sources

### LINKED ISSUE #1905: Apply schemaless rewrites to CREATE TABLE AS SELECT sources
## Summary

The schemaless planner validates `CREATE TABLE ... AS SELECT` source queries, but does not apply the schemaless transform to them.

This means CTAS queries reading from schemaless tables can pass validation but execute with unrewritten identifiers.

## Current Behavior

In `core/src/plan/schemaless/validate.rs`, `Statement::CreateTable { source, .. }` is validated:

```rust
Statement::CreateTable { source, .. } => source
    .as_ref()
    .map_or(Ok(()), |query| validate_query(schema_map, query)),
```

But in `core/src/plan/schemaless.rs`, `transform_statement` handles `Query`, `Insert`, `Update`, and `Delete`, while `CreateTable` falls through unchanged.

## Example

Given a schemaless table:

```sql
CREATE TABLE Item;
INSERT INTO Item VALUES ('{"id":100,"name":"Test"}');
```

This CTAS query should rewrite `name` to `_doc['name']` in the source query:

```sql
CREATE TABLE ItemName AS SELECT name FROM Item;
```

Without the rewrite, the CTAS source query keeps `name` as a normal identifier.

## Expected Behavior

`CREATE TABLE ... AS SELECT` should receive the same schemaless query rewrite as `SELECT` and `INSERT ... SELECT`.

## Suggested Fix

Add a `CreateTable` branch to `transform_statement` and call `transform_query` on `source` when present.

Add a regression test covering CTAS from a schemaless table.

jinwooky:
## Summary

- apply the schemaless transform to `CREATE TABLE ... AS SELECT` source queries
- add plan-level tests for CTAS rewrite, wildcard projection mode, and schemaful no-op
- add an end-to-end fixture covering both CTAS paths from a schemaless table

## Why

`plan::schemaless` validates CTAS source queries in `validate_statement`, but
`transform_statement` handled only `Query`, `Insert`, `Update`, and `Delete` 
`CreateTable` fell through to `_ => statement`.

As a result, a CTAS reading from a schemaless table passed validation and then
reached the executor with unrewritten identifiers, failing with
`EvaluateError::IdentifierNotFound`.

`CREATE TABLE ItemName AS SELECT name FROM Player`, before:

```
SelectItems([Expr { expr: Identifier("name"), label: "name" }])
```

after:

```
SelectItems([Expr { expr: ArrayIndex {
    obj: Identifier("_doc"),
    indexes: [Literal(QuotedString("name"))],
}, label: "name" }])
```

This also makes the `ProjectionPlan::SchemalessMap` arm of
`can_copy_source_schema` reachable from SQL for the first time, so
`CREATE TABLE X AS SELECT * FROM <schemaless>` now copies the source's
`column_defs: None` and produces a schemaless target table.

The new branch follows the same shape `plan::aggregate` already uses for
CTAS sources.

## Tests

- `cargo test -p gluesql-core`: 515 passed
- `cargo test -p gluesql_memory_storage`: 196 passed
- `cargo test -p gluesql-json-storage`: 191 passed
- `cargo test -p gluesql-test-suite`: 16 passed
- `cargo clippy --all-targets -- -D warnings`: clean
- `cargo fmt --all`

Closes #1905

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved support for creating tables from query results when using schemaless data.
  - Queries selecting specific columns or all available columns now produce correctly structured tables.
  - Ensured created tables contain the latest source data after updates.
  - Added coverage for schema-aware and schemaless table creation scenarios.

### DIFF
--- a/core/src/planner/schemaless.rs
+++ b/core/src/planner/schemaless.rs
@@ -109,6 +109,28 @@ fn transform_statement<S: BuildHasher>(
                 selection,
             }
         }
+        StatementPlan::CreateTable {
+            if_not_exists,
+            name,
+            columns,
+            mut source,
+            engine,
+            foreign_keys,
+            comment,
+        } => {
+            if let Some(source) = source.as_mut() {
+                transform_query(schema_map, source);
+            }
+            StatementPlan::CreateTable {
+                if_not_exists,
+                name,
+                columns,
+                source,
+                engine,
+                foreign_keys,
+                comment,
+            }
+        }
         _ => statement,
     }
 }
@@ -493,6 +515,19 @@ mod tests {
         let actual = "DELETE FROM Item WHERE Item.id = 1";
         let expected = "DELETE FROM Item WHERE Item.id = 1";
         test(actual, expected, "delete schemaful compound no-op");
+
+        let actual = "CREATE TABLE ItemName AS SELECT name FROM Player";
+        let expected = "CREATE TABLE ItemName AS SELECT _doc['name'] as name FROM Player";
+        test(actual, expected, "create table as select");
+
+        let actual = "CREATE TABLE ItemName AS SELECT id FROM Player WHERE id = 1";
+        let expected =
+            "CREATE TABLE ItemName AS SELECT _doc['id'] as id FROM Player WHERE _doc['id'] = 1";
+        test(actual, expected, "create table as select with filter");
+
+        let actual = "CREATE TABLE ItemId AS SELECT id FROM Item";
+        let expected = "CREATE TABLE ItemId AS SELECT id FROM Item";
+        test(actual, expected, "create table as select schemaful no-op");
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


panarch (changes requested):
Could we extend the existing `test` helper to support both regular queries and CTAS instead of adding a nearly identical `test_ctas` closure?

```rust
let test = |sql: &str, expected_schemaless_map: bool| {
    let parsed = parse(sql).expect(sql).into_iter().next().unwrap();
    let statement = StatementPlan::from(translate(&parsed).unwrap());
    let schema_map = fetch_schema_map(&storage, &statement).unwrap();
    let planned = plan_schemaless(&schema_map, statement).unwrap();

    let body = match planned {
        StatementPlan::Query(query) => query.body,
        StatementPlan::CreateTable {
            source: Some(source),
            ..
        } => source.body,
        _ => panic!("expected query or create table as select"),
    };
    let SetExprPlan::Select(select) = body else {
        panic!("expected select query");
    };

    assert_eq!(
        matches!(select.projection, ProjectionPlan::SchemalessMap),
        expected_schemaless_map,
        "{sql}"
    );
};
```

This would let the new CTAS cases reuse the existing projection-mode assertion directly.

jinwooky:
Merged `test_ctas` into the existing `test` helper: the closure now matches on both
`StatementPlan::Query` and `StatementPlan::CreateTable { source: Some(..) }`, so the
CTAS cases go through the same projection assertion.

Also rebased onto main. Since the module moved to `core/src/planner/` and
`SetExprPlan` was replaced by `QueryPlan::Project`, the extraction differs slightly
from your snippet:
```rust
let query = match planned {
    StatementPlan::Query(query) => query,
    StatementPlan::CreateTable { source: Some(source), .. } => *source,
    _ => panic!("expected query or create table as select"),
};
let QueryPlan::Project(project) = query else {
    panic!("expected project query");
};
```
The shape you suggested is unchanged (one helper, one assertion).

panarch (approved):
Looks all nice!
Thanks a lot 👍 👍 👍 

fyi. Just a small workflow note for future updates: GlueSQL normally squash-merges a PR into a single commit, so unless a rebase is genuinely necessary, there’s no need to rebase the branch while addressing review feedback. Feel free to add and push follow-up commits instead—we’ll squash them when merging. Keeping the updates incremental also makes it easier for reviewers to see what changed since the previous review.

PR #1988, opened 2026-08-04
Rename the join planner to the hash join planner

### LINKED ISSUE #1966: Rename the join planner to the hash join planner
## Description

The planner in `core/src/plan/join.rs` is currently named as a general join planner, but its responsibility is specifically to detect hash join candidates and produce `JoinExecutorPlan::Hash` execution plans. When no hash join candidate is found, it preserves the nested-loop plan.

Rename the planner and its public entry point so that their names accurately describe this responsibility.

## Changes

- Rename `core/src/plan/join.rs` to `core/src/plan/hash_join.rs`.
- Rename `JoinPlanner` to `HashJoinPlanner`.
- Rename `plan_join` to `plan_hash_join`.
- Update the core and Sled planner pipelines to use `plan_hash_join`.
- Update related imports, test helpers, and test names.

## Acceptance criteria

- The generated `StatementPlan` remains unchanged for both hash and nested-loop joins.
- Existing hash join and nested-loop join planner tests continue to pass.
- No join semantics or execution behavior is changed.
- No new planner hierarchy or module nesting is introduced as part of this issue.

MinhyukWoo:
## Summary

- Rename `core/src/plan/join.rs` to `core/src/plan/hash_join.rs`.
- Rename `JoinPlanner` to `HashJoinPlanner`.
- Rename the public entry point `plan_join` to `plan_hash_join`.
- Update the core and Sled planner pipelines to call `plan_hash_join`.
- Rename the test helper to `plan_hash_join`.

No join semantics or execution behavior is changed, and no new planner
hierarchy or module nesting is introduced.

## Why

The planner reads as a general join planner, but its only responsibility is to
detect hash join candidates and produce `JoinExecutorPlan::Hash`. When no
candidate is found it leaves the nested-loop plan untouched, so it is an
optional optimization pass rather than the planner that decides how joins are
executed. Renaming it makes that scope visible at the call site in both
pipelines, next to its sibling passes `plan_index` and `plan_aggregate`.

It also removes a name collision: `plan_join` previously referred both to this
pass and to a private traversal step in `core/src/plan/aggregate.rs` that walks
`JoinPlan` nodes. Those live at different levels of abstraction — the aggregate
one makes no planning decision and only recurses into nested expressions,
including the `key_expr`/`value_expr`/`where_clause` this pass produces. It is
left as is, since it follows the `plan_<node type>` convention shared by its
siblings `plan_query`, `plan_select`, and `plan_expr`.

## Notes for reviewers

- `gluesql_core::plan::plan_join` was a public export, so third-party storages
  that implement a custom `Planner` need to update the import. Nothing inside
  the workspace depends on the old name.
- `HashJoinPlanner::plan_join_condition` keeps its name. It is not the planner
  or its public entry point, so renaming it looked out of scope for this issue.
  Happy to include it if you would prefer the private method renamed too.
- The commits are split one per rename so the diff is easy to follow; the file
  move is a content-identical rename (`core/src/plan/{join.rs => hash_join.rs}`).

## Tests

- `cargo fmt --all --check`
- `cargo test -p gluesql-core` — 515 passed
- `cargo test -p gluesql_sled_storage` — 246 passed
- `cargo test -p gluesql_memory_storage` — 196 passed

Both planner pipelines are covered end to end: `gluesql_memory_storage` exercises
the default pipeline in `core/src/store/planner.rs` and `gluesql_sled_storage`
exercises the Sled pipeline, which is the only one that also runs `plan_index`.

Closes #1966

## Summary by CodeRabbit

* **Refactor**
  * Updated query planning to use hash joins consistently.
  * Renamed internal join-planning terminology for clarity.
  * Updated core and storage planners to use the revised planning interface.
  * Planning behavior and test expectations remain unchanged.


panarch (approved):
Looks all nice!
Thanks a lot 👍 👍 👍

Issue #1910, opened 2026-05-10
Evolve StatementPlan into a pipeline/operator execution plan


panarch:
`StatementPlan` was introduced as the execution-facing representation, but its current shape intentionally stays close to the existing AST structure.

That was a conservative first step: #1900 separates execution metadata from AST and moves plan-only fields such as aggregate slots, hash join executor selection, schemaless projection, and index selection into plan types. However, the plan structure is still mostly a mirror of SQL syntax.

The next step is to evolve `StatementPlan` into a real execution plan organized around operators/pipelines rather than AST-shaped clauses.

Current shape:

```text
QueryPlan
  body
  order_by
  limit
  offset

SelectPlan
  projection
  from
  selection
  group_by
  having
  aggregate_slots
```

This still reflects SQL syntax order. Execution, however, is closer to a pipeline:

```text
Source/TableScan
  -> Join
  -> Filter
  -> Aggregate
  -> Project
  -> Sort
  -> Limit
  -> Offset
```

Expected direction:

- Introduce operator-oriented plan nodes incrementally.
- Represent sources explicitly, e.g. table scans, derived subqueries, series, dictionary/system sources, and values.
- Move `ORDER BY`, `LIMIT`, and `OFFSET` from query syntax fields toward pipeline operators.
- Represent projection as explicit output items with labels and possibly inferred metadata.
- Represent aggregate evaluation as bound slots/operators instead of runtime registry construction.
- Represent table access paths, including primary key and non-clustered index selection, as scan/access-path metadata.
- Keep AST as syntax/storage-facing definitions and avoid reintroducing execution metadata into AST.
- Preserve behavior while replacing AST-shaped plan pieces one area at a time.

Initial candidates:

- Split `TableFactorPlan` into explicit source plan nodes.
- Convert `ORDER BY`, `LIMIT`, and `OFFSET` into pipeline operators.
- Move projection wildcard expansion into planning.
- Make aggregate slot registry/binding a plan artifact.
- Move index access selection into table scan/access path plans.

Non-goals for the first step:

- Rewriting the whole executor in one PR.
- Introducing an overly generic graph abstraction before the concrete operator needs are clear.
- Changing SQL semantics while restructuring plan shape.

Rationale:

A pipeline/operator plan should make execution order explicit, reduce planner/executor coupling, and avoid AST-shaped special cases leaking into execution code. It also gives future planner work a clearer place to attach decisions such as access paths, join strategy, aggregate binding, and output schema calculation.

PR #1989, opened 2026-08-06
Add tests for untested TranslateError paths

kwondo1017:
## What this PR does

Of the 62 `TranslateError` variants, **14 variants (15 construction sites)** were not exercised by any test — neither by the SQL fixtures in `test-suite/fixtures/` nor by the inline test modules in `core/src/translate*`. Any refactoring (e.g. the upcoming StatementPlan redesign, #1910) could silently drop or change one of these error paths without a single test failing.

This PR is **tests only** — no variant is renamed or removed, and no parsing behavior changes.

Each untested path was classified by reading the sqlparser 0.52 sources directly (GlueSQL pins `PostgreSqlDialect`, `core/src/parse_sql.rs`):

### Reachable through plain SQL — pinned with unit tests (7 variants)

| Variant | Reproduction |
|---|---|
| `UnreachableSubqueryFunctionArgNotSupported` | `SELECT ARRAY(SELECT 1)` — sqlparser turns `ARRAY(<query>)` into `FunctionArguments::Subquery`, so this is reachable despite the name |
| `UnreachableUnaryOperator` | `SELECT \|/ 25` — PG-only prefix operators (`\|/`, `\|\|/`, `!!`, `@`) parse but are not handled; also reachable despite the name |
| `UnsupportedGroupByAll` | `SELECT a FROM Foo GROUP BY ALL` |
| `UnsupportedLiteral` | `SELECT $$abc$$` (dollar-quoted string) |
| `UnsupportedMultipleAlterTableOperations` | `ALTER TABLE Foo DROP COLUMN a, DROP COLUMN b` |
| `UnsupportedQueryTableFactor` | `SELECT * FROM (Foo JOIN Bar ON 1 = 1)` (nested join) |
| `ParameterIndexOutOfRange` | `SELECT $2` with one bound parameter (`bind_placeholder` unit test + `translate_with_params` end-to-end) |

Tests live in the module that owns each construction site, reusing the existing helpers (`assert_translate_error`, `assert_query_error`, `parse_expr` + `translate_expr`). New `#[cfg(test)]` modules were added to `operator.rs` and `literal.rs`.

### Unreachable by parser contract — recorded as contract tests (7 variants / 8 sites)

For each defensive branch, `parser_rejects_inputs_guarded_by_defensive_variants` asserts that the nearest-miss SQL fails to **parse** with sqlparser 0.52 + `PostgreSqlDialect`, with a comment naming the guarded variant:

| Variant | Contract evidence |
|---|---|
| `UnreachableEmptyAlterTableOperation` | operations list is `parse_comma_separated` (≥ 1) — `ALTER TABLE Foo` is a parse error |
| `UnreachableEmptyIdent` | assignment target needs ≥ 1 identifier — `UPDATE Foo SET = 1` |
| `UnreachableEmptyObject` | `parse_object_name` needs ≥ 1 identifier — `DROP TABLE` |
| `UnreachableEmptyTable` | DELETE's FROM list is ≥ 1 — `DELETE FROM` |
| `UnreachableForeignKeyColumns` (2 sites) | both column lists are `parse_parenthesized_column_list(Mandatory, allow_empty = false)` — `FOREIGN KEY ()`, `REFERENCES B ()`, and `REFERENCES B` all fail to parse |
| `UnreachableOmittingFromInDelete` | `FromTable::WithoutKeyword` is only produced for BigQuery/Generic dialects; a positive-control test pins that `DELETE FROM Foo` parses as `WithFromKeyword` |
| `UnsupportedTrimChars` | `trim_characters` comes only from the comma syntax, which is Snowflake/BigQuery/Generic-only — `SELECT TRIM('x', 'y')` fails to parse |

If a future sqlparser upgrade changes any of these contracts, the tests break — turning a silent "unreachable became reachable" into a visible failure.

## Follow-up candidates (out of scope here)

- `UnreachableSubqueryFunctionArgNotSupported` and `UnreachableUnaryOperator` are misnamed — they are reachable through plain SQL. Renaming is left for a separate PR.
- Found while auditing: `DELETE Foo FROM Bar` (MySQL multi-table syntax) parses under `PostgreSqlDialect` and translate silently ignores the leading `tables` list. Possibly worth its own issue.

## Verification

- `cargo test -p gluesql-core` — 522 passed
- `cargo clippy -p gluesql-core --all-targets` — clean
- `cargo fmt -p gluesql-core -- --check` — clean
- Coverage re-audit: every one of the 62 variant names now appears in fixture or inline test code

## Summary by CodeRabbit

- **Tests**
  - Added coverage for unsupported SQL syntax and translation scenarios, including:
    - Out-of-range parameter placeholders
    - Multiple `ALTER TABLE` operations
    - Subqueries in function arguments
    - Dollar-quoted string literals
    - PostgreSQL-only unary operators
    - `GROUP BY ALL`
    - Parenthesized nested joins
  - Added parser validation ensuring `DELETE` statements include `FROM`.

### DIFF
--- a/core/src/translate.rs
+++ b/core/src/translate.rs
@@ -711,4 +711,71 @@ mod tests {
         let params = [1_i64.into_param_literal()];
         assert_invalid_placeholder(&params, "$foo");
     }
+
+    #[test]
+    fn reject_out_of_range_placeholder_index() {
+        let params = [1_i64.into_param_literal()];
+        let err = bind_placeholder(&params, "$2").unwrap_err();
+        assert_eq!(
+            err,
+            TranslateError::ParameterIndexOutOfRange { index: 2, len: 1 }.into()
+        );
+
+        let parsed = parse("SELECT $2").expect("parse");
+        let actual = translate_with_params(&parsed[0], &params);
+        let expected = Err(TranslateError::ParameterIndexOutOfRange { index: 2, len: 1 }.into());
+        assert_eq!(actual, expected);
+    }
+
+    #[test]
+    fn multiple_alter_table_operations_not_supported() {
+        assert_translate_error(
+            "ALTER TABLE Foo DROP COLUMN a, DROP COLUMN b",
+            TranslateError::UnsupportedMultipleAlterTableOperations,
+        );
+    }
+
+    #[test]
+    fn parser_rejects_inputs_guarded_by_defensive_variants() {
+        // Each SQL below is the nearest miss for a defensive TranslateError
+        // branch; sqlparser (0.52, PostgreSqlDialect) rejects it at parse
+        // time, so the branch is unreachable through the public SQL API. If a
+        // sqlparser upgrade makes one of these parse, the guarded branch may
+        // have become reachable and deserves a real translate test instead.
+        let cases = [
+            // TranslateError::UnreachableEmptyAlterTableOperation
+            "ALTER TABLE Foo",
+            // TranslateError::UnreachableEmptyIdent
+            "UPDATE Foo SET = 1",
+            // TranslateError::UnreachableEmptyObject
+            "DROP TABLE",
+            // TranslateError::UnreachableEmptyTable
+            "DELETE FROM",
+            // TranslateError::UnreachableForeignKeyColumns (referencing side)
+            "CREATE TABLE T (a INT, FOREIGN KEY () REFERENCES B (id))",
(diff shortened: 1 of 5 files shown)


kwondo1017:
@panarch 
It looks like the reviewer and assignee weren't set on this PR, likely due to the GitHub incident when it was opened — CI initially failed to run as well, but all checks have since passed after a re-run. Could you add me as the assignee and set up the review when you get a chance?

PR #1991, opened 2026-08-09
Reject unsupported transaction statement clauses

kwondo1017:
**Summary**

- `START TRANSACTION`, `COMMIT`, and `ROLLBACK` were translated with `..` patterns that silently discarded every clause GlueSQL does not support. The worst case was `ROLLBACK TO SAVEPOINT sp1`: the whole transaction was rolled back while the response claimed success.
- Destructure the three statements explicitly and reject transaction modes (`READ ONLY`, `ISOLATION LEVEL ...`), the SQLite `BEGIN` modifier, `AND CHAIN`, and `TO SAVEPOINT` with a new `TranslateError::UnsupportedTransactionOption(TransactionOption)`, following the existing `InsertOption`-style option enums.
- Add regression tests for each rejected clause (including the `SAVEPOINT`-keyword-omitted `ROLLBACK TO sp1` form and a direct-AST test for the parser-unreachable SQLite modifier) plus positive controls for the plain statement forms.

**Test plan**

- [x] `cargo fmt -- --check`
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `cargo check --workspace`
- [x] `cargo test -p gluesql-core`

## Summary by CodeRabbit

* **Bug Fixes**
  * Transaction statements now reject unsupported modes, modifiers, chaining options, and savepoint clauses.
  * Standard `BEGIN`, `START TRANSACTION`, `COMMIT`, and `ROLLBACK` statements remain supported.
  * Error messages now identify the unsupported transaction option.

### DIFF
--- a/core/src/translate.rs
+++ b/core/src/translate.rs
@@ -345,6 +345,7 @@ pub fn translate_with_params(
         SqlStatement::StartTransaction {
             modes,
             modifier,
+            // `begin` only records the `BEGIN` vs `START TRANSACTION` spelling.
             begin: _,
         } => {
             let violation = if !modes.is_empty() {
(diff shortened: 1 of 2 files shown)


panarch (approved):
Looks all nice!
Thanks a lot 👍 👍 👍

PR #1993, opened 2026-08-09
Add EXPLAIN query support

panarch:
## Summary

Add `EXPLAIN` support for query statements so users can inspect GlueSQL's planned typed execution pipeline without executing the query.

## Changes

- Translate `EXPLAIN <query>` into dedicated AST and statement-plan variants.
- Render query plans through a semantic tree that describes scans, joins, filters, aggregation, projection, ordering, pagination, expressions, and subqueries.
- Report selected table access paths, including primary-key lookups, secondary indexes, and full scans.
- Return a dedicated explain payload and display it in both tabular and plain CLI output.
- Add readable SQL fixtures covering primary-key access, join and aggregation pipelines, and expression subqueries.

## Impact

Users can see how `SELECT` and `VALUES` queries will be executed, including planner-selected access paths, execution stages, and subquery modes. This also gives contributors a stable representation for understanding planner transformations.

## Summary by CodeRabbit

* **New Features**
  * Added support for SQL `EXPLAIN` queries.
  * View detailed plans for scans, joins, filters, aggregations, sorting, pagination, and subqueries.
  * Query plans are available in tabular and plain-text CLI formats.
* **Bug Fixes**
  * `EXPLAIN` queries now pass through validation and planning consistently.
* **Tests**
  * Added coverage for query-plan output across common query patterns and CLI formats.

### DIFF
--- a/core/src/plan/explain.rs
+++ b/core/src/plan/explain.rs
@@ -4,23 +4,19 @@ pub fn explain(query: &QueryPlan) -> Vec<String> {
     explain_lines(query)
 }
 
+fn explain_lines(explainable: &impl Explain<Output = ExplainNode>) -> Vec<String> {
+    let mut context = ExplainContext::default();
+    let root = explainable.explain(&mut context);
+    let root = context.with_subqueries(root);
+    let mut lines = Vec::new();
+    render_root(&root, &mut lines);
+    lines
+}
+
 pub(crate) trait Explain {
     type Output;
 
     fn explain(&self, context: &mut ExplainContext) -> Self::Output;
-}
-
-#[derive(Default)]
-pub(crate) struct ExplainContext {
-    next_subquery_id: usize,
-    subqueries: Vec<ExplainNode>,
-}
-
-#[derive(Clone, Copy)]
-pub(crate) enum ExplainSubqueryMode {
-    OneRow,
-    AllRows,
-    Exists,
 }
 
 #[derive(Debug, PartialEq, Eq)]
@@ -80,6 +76,36 @@ impl ExplainNode {
         self.children.extend(children);
         self
     }
+
+    fn title(&self) -> String {
+        match &self.annotation {
+            Some(annotation) => format!("{} ({annotation})", self.name),
+            None => self.name.clone(),
+        }
+    }
+}
+
+#[derive(Clone, Copy)]
+pub(crate) enum ExplainSubqueryMode {
+    OneRow,
+    AllRows,
+    Exists,
+}
+
+impl Display for ExplainSubqueryMode {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(match self {
+            Self::OneRow => "one row",
+            Self::AllRows => "all rows",
+            Self::Exists => "exists",
+        })
+    }
+}
+
+#[derive(Default)]
+pub(crate) struct ExplainContext {
+    next_subquery_id: usize,
+    subqueries: Vec<ExplainNode>,
 }
 
 impl ExplainContext {
@@ -115,25 +141,6 @@ impl ExplainContext {
 
         ExplainNode::new("root").with_children(std::iter::once(main).chain(self.subqueries))
     }
-}
-
-impl Display for ExplainSubqueryMode {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.write_str(match self {
-            Self::OneRow => "one row",
-            Self::AllRows => "all rows",
-            Self::Exists => "exists",
-        })
-    }
-}
-
-fn explain_lines(explainable: &impl Explain<Output = ExplainNode>) -> Vec<String> {
-    let mut context = ExplainContext::default();
-    let root = explainable.explain(&mut context);
-    let root = context.with_subqueries(root);
-    let mut lines = Vec::new();
-    render_root(&root, &mut lines);
-    lines
 }
 
 fn render_root(node: &ExplainNode, lines: &mut Vec<String>) {
@@ -198,15 +205,6 @@ fn render_child(node: &ExplainNode, prefix: &str, last: bool, lines: &mut Vec<St
         lines.push(format!("{child_prefix}│"));
     }
     render_children(node, &child_prefix, lines);
-}
-
-impl ExplainNode {
-    fn title(&self) -> String {
-        match &self.annotation {
-            Some(annotation) => format!("{} ({annotation})", self.name),
-            None => self.name.clone(),
-        }
-    }
 }
 
 #[cfg(test)]
@@ -352,77 +350,6 @@ LIMIT 10 OFFSET 5
     }
 
     #[test]
-    fn explains_expression_subqueries_as_referenced_plans() {
-        assert_eq!(
-            explain_sql(
-                r"
-CREATE TABLE Player (id INT);
-CREATE TABLE Badge (player_id INT);
-",
-                r"
-EXPLAIN
-SELECT id, (SELECT COUNT(*) AS total FROM Badge) AS badge_count
-FROM Player
-WHERE id IN (SELECT player_id FROM Badge)
-AND EXISTS (
-    SELECT *
-    FROM Badge
-    WHERE Badge.player_id = Player.id
-)
-",
-            ),
-            r"
-• root
-├── • project
-│   │ columns: id, @S1 AS badge_count
-│   │
-│   └── • filter
-│       │ expression: id IN (@S2) AND EXISTS (@S3)
-│       │
-│       └── • scan Player
-│             access: full scan
-│
-├── • subquery
-│   │ id: @S1
-│   │ exec mode: one row
-│   │
-│   └── • project
-│       │ columns: COUNT(*) AS total
-│       │
-│       └── • aggregate
-│           │ aggregates: COUNT(*)
-│           │
-│           └── • scan Badge
-│                 access: full scan
-│
-├── • subquery
-│   │ id: @S2
-│   │ exec mode: all rows
-│   │
-│   └── • project
-│       │ columns: player_id
-│       │
-│       └── • scan Badge
-│             access: full scan
-│
-└── • subquery
-    │ id: @S3
-    │ exec mode: exists
-    │
-    └── • project
-        │ columns: *
-        │
-        └── • filter
-            │ expression: Badge.player_id = Player.id
-            │
-            └── • scan Badge
-                  access: full scan
-"
-            .trim()
-        );
-    }
-
-    #[test]
     fn explains_distinct_grouping_and_having() {
         assert_eq!(
             explain_sql(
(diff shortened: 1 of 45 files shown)


devgony:
Reference: https://www.postgresql.org/docs/current/using-explain.html

core/src/translate.rs
```diff
@@ -63,6 +63,18 @@ pub fn translate_with_params(
...
+            analyze: false,
+            verbose: false,
+            statement,
+            format: None,
+            options: None,
+            ..
+        } => match statement.as_ref() {
+            SqlStatement::Query(query) => translate_query(query, params).map(Statement::Explain),
+            _ => Err(TranslateError::UnsupportedStatement(sql_statement.to_string()).into()),
+        },
```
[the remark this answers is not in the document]
panarch: Good catch. Fixed in 331cd72e by matching `query_plan: false` explicitly and removing the wildcard pattern, so future `Statement::Explain` fields require an intentional decision. I also added `EXPLAIN QUERY PLAN SELECT * FROM Foo` to the unsupported EXPLAIN regression cases.

core/src/plan/statement/expr/function.rs
```diff
@@ -0,0 +1,1355 @@
...
+            then: ExprPlan::Identifier("fallback".to_owned()),
+        };
+        let expected = "IF_NULL(value, fallback)";
+        test(&actual, expected);
+
+        let actual = FunctionExprPlan::NullIf {
+            expr1: ExprPlan::Identifier("left".to_owned()),
+            expr2: ExprPlan::Identifier("right".to_owned()),
+        };
+        let expected = "NULL_IF(left, right)";
```
devgony: Looks like SCREAMING_SNAKE_CASE made wrong function name.
IF_NULL should be IFNULL.
NULL_IF should be NULLIF.

core/src/plan/explain.rs
```diff
@@ -0,0 +1,576 @@
...
+    pub(crate) fn subquery_count(&self) -> usize {
+        self.subqueries.len()
+    }
+
+    pub(crate) fn register_subquery(
+        &mut self,
+        query: &QueryPlan,
+        mode: ExplainSubqueryMode,
+    ) -> String {
+        self.next_subquery_id += 1;
```
devgony, line 123: No identity memo here: the same subquery reached via projection and `aggregate_slots` is registered twice (`@S1`/`@S2`, identical subtrees).
Key a memo on `(QueryPlan, ExplainSubqueryMode)` and reuse the id.
```
gluesql> EXPLAIN SELECT SUM(a + (SELECT b FROM T LIMIT 1)) FROM T;
| QUERY PLAN                                                        |
|-------------------------------------------------------------------|
| • root                                                            |
| ├── • project                                                     |
| │   │ columns: SUM(a + @S1) AS SUM(a + (SELECT b FROM T LIMIT 1)) |
| │   │                                                             |
| │   └── • aggregate                                               |
| │       │ aggregates: SUM(a + @S2)                                |
| │       │                                                         |
| │       └── • scan T                                              |
| │             access: full scan                                   |
| │                                                                 |
| ├── • subquery                                                    |
| │   │ id: @S1                                                     |
| │   │ exec mode: one row                                          |
| │   │                                                             |
| │   └── • limit                                                   |
| │       │ count: 1                                                |
| │       │                                                         |
| │       └── • project                                             |
| │           │ columns: b                                          |
| │           │                                                     |
| │           └── • scan T                                          |
| │                 access: full scan                               |
| │                                                                 |
| └── • subquery                                                    |
|     │ id: @S2                                                     |
|     │ exec mode: one row                                          |
|     │                                                             |
|     └── • limit                                                   |
|         │ count: 1                                                |
|         │                                                         |
|         └── • project                                             |
|             │ columns: b                                          |
|             │                                                     |
|             └── • scan T                                          |
|                   access: full scan                               |
```

core/src/plan/statement/expr/function.rs
panarch: Thank you for pointing this out. You’re right—these functions should retain their canonical SQL names. Fixed in [50956aef](https://github.com/gluesql/gluesql/pull/1993/commits/50956aef331d6f5968dafbedd3eaad9f7b617fc7) so EXPLAIN now renders `IFNULL` and `NULLIF` correctly.

core/src/plan/explain.rs
panarch: Thank you for the detailed report and the concrete example. The duplicate registration was caused by rendering the same planned aggregate expression from both the projection and aggregate_slots.

Fixed in [fde3819f](https://github.com/gluesql/gluesql/pull/1993/commits/fde3819f27df307f45595b3d229137efce14e3d1). I took a slightly different approach from structural memoization because that could merge separately written but structurally identical scalar subqueries that represent distinct executions. `EXPLAIN` now uses the planner’s aggregate slot as identity: consumers reference `@A1`, while the aggregate node defines `@A1 = SUM(a + @S1)`, so the nested subquery is registered only once.

With the same query, the output is now:
```
• root
├── • project
│   │ columns: @A1 AS SUM(a + (SELECT b FROM T LIMIT 1))
│   │
│   └── • aggregate
│       │ aggregates: @A1 = SUM(a + @S1)
│       │
│       └── • scan T
│             access: full scan
│
└── • subquery
    │ id: @S1
    │ exec mode: one row
    │
    └── • limit
        │ count: 1
        │
        └── • project
            │ columns: b
            │
            └── • scan T
                  access: full scan
```
This keeps the aggregate definition in one place and registers the nested subquery only once, while separately written scalar subqueries can still retain distinct IDs.

core/src/plan/statement/expr/function.rs
```diff
@@ -0,0 +1,1355 @@
...
+            Self::ToDate { expr, format } => {
+                fmt_call("TO_DATE", [expr, format], context, &mut output);
+            }
+            Self::ToTimestamp { expr, format } => {
+                fmt_call("TO_TIMESTAMP", [expr, format], context, &mut output);
+            }
+            Self::ToTime { expr, format } => {
+                fmt_call("TO_TIME", [expr, format], context, &mut output);
+            }
+            Self::Position {
```
devgony, line 628: The rename fixes IFNULL/NULLIF, but the failure mode remains: each arm hardcodes its SQL-surface name and the unit test asserts whatever the arm emits, so parser/renderer divergence is self-consistent and undetectable.
POSITION still shows it — line 631 renders the comma form (`POSITION(substring, value)`, asserted at 1206), which the parser only accepts as `POSITION(x IN y)`; pasted back it becomes `Function::Custom` and fails at evaluation.

Suggest a round-trip regression test: for each variant's rendered expression, parse `SELECT {expr}`, translate, and assert it maps back to the same function variant. That pins the renderer to what the parser accepts and would have caught IF_NULL, NULL_IF, and POSITION at once.

panarch: Thank you for the detailed report and the concrete `POSITION` reproduction. You're right that the expected strings could previously move together with the formatter and leave parser/renderer divergence undetected.

`POSITION` now emits its SQL surface syntax as `POSITION(needle IN value)` instead of the comma form. The fixture uses `needle` because `SUBSTRING` is parsed as a keyword by the PostgreSQL dialect.

I also extended the existing exhaustive `FunctionExprPlan` EXPLAIN test so that each rendered expression is parsed and translated back, then compared with the original function plan structure. This verifies the function variant, arguments, and argument order. As a result, the comma-form `POSITION` would translate back as `Custom` and fail even if its expected output were changed to match the renderer.

The round-trip cases use SQL-compatible representative arguments. Plan-only references such as `@S1` and `@A1` remain covered by their dedicated EXPLAIN tests.

core/src/plan/statement/expr.rs
```diff
@@ -456,284 +222,537 @@ impl From<ast::Expr> for ExprPlan {
...
+            output.push(' ');
+            fmt_expr(pattern, context, output);
+        }
+        ExprPlan::BinaryOp { left, op, right } => {
+            fmt_expr(left, context, output);
+            output.push(' ');
+            output.push_str(&op.to_sql());
+            output.push(' ');
+            fmt_expr(right, context, output);
+        }
```
devgony: '||' was converted to '+' which currently we don't support
```sql
EXPLAIN SELECT name || 'x'
;
| QUERY PLAN                           |
|--------------------------------------|
| • project                            |
| │ columns: name + 'x' AS name || 'x' |
| │                                    |
| └── • series Series                  |
|       size: 1                        |
```

devgony:
An EXPLAIN submitted in the same batch as its DDL is planned against the pre-batch catalog, so it prints an unoptimized plan (no index/PK rewrite).
This is truthful — the sibling queries in that batch execute the same unoptimized plans — but differs from PostgreSQL/MySQL/SQLite, which analyze+plan each statement at its own execution time, and it will surprise users pasting scripts.

1) Execute queries at once
```sql
gluesql> CREATE TABLE Item (id INTEGER, flag BOOLEAN NULL);
  CREATE INDEX idx_flag ON Item (flag);
  INSERT INTO Item VALUES (1, TRUE), (2, NULL), (3, FALSE);

  EXPLAIN SELECT id FROM Item WHERE flag IS NULL;
  SELECT id FROM Item WHERE flag IS NULL;
  EXPLAIN SELECT id FROM Item WHERE flag IS NOT NULL;
  SELECT id FROM Item WHERE flag IS NOT NULL;



;
Table created

Index created

3 rows inserted
```
| QUERY PLAN                     |
|--------------------------------|
| • project                      |
| │ columns: id                  |
| │                              |
| └── • filter                   |
|     │ expression: flag IS NULL |
|     │                          |
|     └── • scan Item            |
|           access: full scan    |

| id |
|----|
| 2  |

| QUERY PLAN                         |
|------------------------------------|
| • project                          |
| │ columns: id                      |
| │                                  |
| └── • filter                       |
|     │ expression: flag IS NOT NULL |
|     │                              |
|     └── • scan Item                |
|           access: full scan        |

| id |
|----|
| 1  |
| 3  |


2) Execute each queries 1 by 1

```sql
gluesql> CREATE TABLE Item (id INTEGER, flag BOOLEAN NULL);
Table created

gluesql> 
CREATE INDEX idx_flag ON Item (flag);
Index created

gluesql> INSERT INTO Item VALUES (1, TRUE), (2, NULL), (3, FALSE);
3 rows inserted

gluesql> EXPLAIN SELECT id FROM Item WHERE flag IS NULL;
```
| QUERY PLAN                   |
|------------------------------|
| • project                    |
| │ columns: id                |
| │                            |
| └── • scan Item              |
|       access: index idx_flag |
|       predicate: = NULL      |

```sql
gluesql> EXPLAIN SELECT id FROM Item WHERE flag IS NULL;
  SELECT id FROM Item WHERE flag IS NULL;
  EXPLAIN SELECT id FROM Item WHERE flag IS NOT NULL;
  SELECT id FROM Item WHERE flag IS NOT NULL;
```
| QUERY PLAN                   |
|------------------------------|
| • project                    |
| │ columns: id                |
| │                            |
| └── • scan Item              |
|       access: index idx_flag |
|       predicate: = NULL      |

| id |
|----|
| 2  |

| QUERY PLAN                   |
|------------------------------|
| • project                    |
| │ columns: id                |
| │                            |
| └── • scan Item              |
|       access: index idx_flag |
|       predicate: < NULL      |

| id |
|----|
| 3  |
| 1  |

panarch:
ref. https://github.com/gluesql/gluesql/issues/2009

PR #1998, opened 2026-08-17
Reject unsupported CREATE INDEX options during translation

jun02160:
## Summary

This PR makes `CREATE INDEX` translation reject unsupported index options explicitly instead of silently discarding them.

Previously, options parsed by `sqlparser` such as `UNIQUE`, `CONCURRENTLY`, `IF NOT EXISTS`, `USING`, `INCLUDE`, `NULLS [NOT] DISTINCT`, `WITH`, and `WHERE` were ignored and translated into a regular `Statement::CreateIndex`.

## Problem

For example:

```sql
CREATE TABLE Test (id INTEGER);
INSERT INTO Test VALUES (1);

CREATE UNIQUE INDEX ux_id ON Test (id);

INSERT INTO Test VALUES (1);
SELECT id FROM Test WHERE id = 1;

```

Before this change, `CREATE UNIQUE INDEX` succeeded as a regular index, and duplicate values could still be inserted. **This means the** `UNIQUE` **semantics were silently lost.**

## Changes
- Add `CreateIndexOption` to represent unsupported `CREATE INDEX` options
- Add `TranslateError::UnsupportedCreateIndexOption`
- Explicitly inspect parsed `CREATE INDEX` option fields during translation
- Add regression coverage for each unsupported option
- Preserve existing behavior for plain `CREATE INDEX`

## Verification

Ran:

```bash
cargo test -p gluesql-core create_index_options_not_supported
cargo test -p gluesql-core translate::tests
cargo test -p gluesql-core
cargo test -p gluesql_sled_storage sql_index_basic
cargo fmt --all
git diff --check

```

Also manually verified the same SQL scenarios with a small SledStorage-based program that executes SQL against the current upstream behavior and this branch through the real `Glue::execute` path.

| Scenario                                     | Before   | After                                      |
| -------------------------------------------- | -------- | ------------------------------------------ |
| Plain `CREATE INDEX`                         | succeeds | succeeds                                   |
| `CREATE UNIQUE INDEX`                        | succeeds | rejected                                   |
| Duplicate insert after `CREATE UNIQUE INDEX` | succeeds | skipped because index creation is rejected |
| `CREATE INDEX CONCURRENTLY`                  | succeeds | rejected                                   |
| `CREATE INDEX IF NOT EXISTS`                 | succeeds | rejected                                   |
| `CREATE INDEX ... USING`                     | succeeds | rejected                                   |
| `CREATE INDEX ... INCLUDE`                   | succeeds | rejected                                   |
| `CREATE INDEX ... NULLS NOT DISTINCT`        | succeeds | rejected                                   |
| `CREATE INDEX ... WITH`                      | succeeds | rejected                                   |
| `CREATE INDEX ... WHERE`                     | succeeds | rejected                                   |

Plain `CREATE INDEX` still succeeds in both cases.

## Summary by CodeRabbit

* **Bug Fixes**
  * `CREATE INDEX` statements now clearly reject unsupported options, including unique, concurrent, conditional, access method, included columns, null-distinctness, storage, and predicate clauses.
  * Improved error reporting identifies the specific unsupported index option.
* **Tests**
  * Added coverage for each unsupported `CREATE INDEX` option.

### DIFF
--- a/core/src/translate.rs
+++ b/core/src/translate.rs
@@ -12,8 +12,8 @@ pub use self::{
     data_type::translate_data_type,
     ddl::translate_column_def,
     error::{
-        CreateTableOption, DeleteOption, InsertOption, JoinConstraintReason, QueryOption,
-        SelectOption, TransactionOption, TranslateError, UpdateOption,
+        CreateIndexOption, CreateTableOption, DeleteOption, InsertOption, JoinConstraintReason,
+        QueryOption, SelectOption, TransactionOption, TranslateError, UpdateOption,
     },
     expr::{translate_expr, translate_order_by_expr},
     param::{IntoParamLiteral, ParamLiteral},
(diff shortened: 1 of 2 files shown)


panarch (approved):
Looks all nice!

Issue #1999, opened 2026-08-17
A derived table on the right of a join can reference left columns without LATERAL


MinhyukWoo:
### What happens

```sql
CREATE TABLE Player (id INTEGER);                  -- 1, 2
CREATE TABLE Item (id INTEGER, player_id INTEGER); -- (10, 1), (11, 2)

SELECT * FROM Player
JOIN (SELECT * FROM Item WHERE Item.player_id = Player.id) i ON TRUE;
```

This succeeds:

```
labels: ["id", "id", "player_id"]
rows:   [[1, 10, 1], [2, 11, 2]]
```

The derived table is re-evaluated per left row with `Player.id` bound, which is `LATERAL` behaviour. `LEFT JOIN` does the same.

Standard SQL requires `LATERAL` for a derived table to reference preceding items in the same `FROM` clause. GlueSQL has no `LATERAL` keyword, so this is an implicit extension rather than an explicitly supported feature.

### Why it matters

It falls out of the nested loop executor re-preparing the right source per left row, rather than from a deliberate decision, so the behaviour is not stated anywhere and nothing pins it. Two consequences:

- It is not obvious to a user which correlations resolve and which do not.
- Any executor that reads the right side once instead of per left row cannot support it. RIGHT JOIN support (#63) hits this: it reads the right relation once to give each row a stable position for match tracking, so the same query fails with `CompoundIdentifierNotFound`. Blocking it there is intentional — a preserved right relation is undefined if it depends on the left row — but it does mean the behaviour now differs by join type.

### Suggested direction

Since this falls out of the executor path rather than a design decision, making the behaviour uniform across join types looks preferable to pinning it as-is:

- Reject a derived source that references the left relation, so every join type behaves the same, and
- add explicit `LATERAL` if the correlation is worth supporting, requiring it rather than inferring it.

Happy to work on either once the direction is settled.

PR #2000, opened 2026-08-17
Support RIGHT OUTER JOIN

### LINKED ISSUE #63: RIGHT [OUTER] JOIN support

MinhyukWoo:
Closes #63.

`RIGHT JOIN` / `RIGHT OUTER JOIN` now parses, plans, and executes. An unmatched right row is emitted once with every accumulated left relation NULL-extended.

## Approach

Following the model agreed in #63, a RIGHT JOIN is not executable as translated. It passes through three states:

1. **Translation** produces `UnplannedRightOuterJoinPlan` — a temporary state that carries no NULL-extension decision and therefore cannot execute.
2. **`plan_right_outer_join`** (new pass, runs right after `plan_schemaless`) consumes every temporary node and decides two things:
   - `null_extend` — which accumulated left relations an unmatched right row pads out, as an alias list ordered base-first.
   - `TableAccessPlan::FullScanRequired` on the left base source — the unmatched pass needs the complete left input, so no planner may narrow it.
3. **`RightOuterJoinPlan`** is what every later planner and the executor see. Its `null_extend` field is not an `Option`.

Later planners pass the lowered node through and never re-interpret RIGHT JOIN syntax. The primary key and index planners needed **no logic changes at all**: they already gate on `access == FullScan`, so the new marker excludes them automatically.

The executor drives the join itself rather than reusing the left-driven `JoinCandidates` pipeline, because "was this right row matched by *any* left row?" cannot be answered inside a single left-keyed group. The right relation is read once and each row's **position** is its identity, so duplicate rows keep their multiplicity and rows excluded from a hash candidate index (NULL keys, right-side predicates) still reach the unmatched pass.

## Why not rewrite to LEFT JOIN

Two reasons it is not a translation-level swap:

- `SELECT *` column order would change.
- In `A JOIN B RIGHT JOIN C` the left side is a composite, not a table. Swapping would require putting a join tree on the right of a left-deep plan.

## Limitations

These are intentional in this PR. Happy to fold any of them in if you prefer.

- **The right relation is buffered in memory.** Position-based match tracking needs the right rows available for a second pass, and the storage traits have no rewind. For hash plans this is not new — `hash::build_rows` already materializes the right side — but for nested loop plans it replaces per-left-row rescanning with `O(right)` memory. I could not judge what limit is acceptable here.
- **A RIGHT JOIN's own `ON` is not promoted to a hash mechanism**, so `A RIGHT JOIN B ON A.id = B.k` runs as a nested loop while the LEFT equivalent runs as a hash join. Correctness is unaffected. Doing it properly means generalizing the two existing 104-line `plan_*_nested_loop_condition` functions rather than adding a third near-identical copy, which felt out of scope. Happy to open a follow-up issue, or to fold it into this PR if you prefer.
- **ORDER BY index scans are disabled on the pinned base source.** That optimization only reorders rows and is safe under RIGHT JOIN, but it shares the `access == FullScan` gate with the unsafe ones. Re-enabling is a one-line condition widening; kept conservative here and pinned by a fixture.
- **`query_builder` has no RIGHT JOIN support yet**, per the scope agreed in #63.
- **A right source correlated to the left errors** (`CompoundIdentifierNotFound`). A preserved right relation is undefined if it depends on the left row. Note that INNER/LEFT joins currently *allow* this, which looks unintended rather than RIGHT JOIN specific — raised separately in #1999.

## Notes for review

- `TableAccessPlan` gains a variant and is `Serialize`/`Deserialize`, so this is a public type change. Execution is identical to `FullScan`; the variant only tells planners to leave the source alone. If you would rather not widen that enum, the fallback is a flag on the lowered node plus tree walks in the consumers — which is the "planners re-interpret RIGHT JOIN" shape you asked to avoid, so I went with the variant.
- `null_extend` is an alias list, not aliases plus column names. Column names are computable at plan time, but that would duplicate `output_labels` + `labels::resolve` into the planner and give column naming two sources of truth. The executor pairs the list with its runtime sources **by position**, so a mismatch surfaces as an error rather than mis-named NULL columns. Pairing by alias instead would break when the left prefix reuses one — `Player a JOIN Item a` makes the list `["a", "a"]` — which a fixture now pins.
- **The marker goes on the left base source only.** That is complete today because the only source any planner narrows is the base source — there are three assignments to `.access` in the workspace and all three reach it via `base_source_mut()`. It is an implicit invariant rather than an enforced one though: predicate pushdown to a joined source would bypass the marker and produce phantom NULL-extended rows. Note this is not reachable on `main` today, since LEFT JOIN's base source is its *preserved* side. I can mark every left source instead if you would rather not rely on the invariant — it needs a mutable counterpart to `joined_sources()`, so roughly 40 lines.
- `RightOuterJoinInputPlan::Hash` is not produced by any planner yet, but the executor must handle a hash mechanism regardless: a RIGHT JOIN's `ON` is a `JoinConditionPlan`, whose input enum includes `Hash`. Rather than leave that arm as a landmine for the follow-up, it is implemented and covered by a test that builds the plan directly.

## Tests

- Planner unit tests, including a sweep asserting no temporary node survives in **any** of 19 expression positions — a missed position is an execution error here, not a lost optimization.
- Executor tests for the unmatched pass, duplicate-row multiplicity, and the hash lookup path.
- SQL fixtures: three-table chains, chained RIGHT JOINs, empty left source, aggregate over NULL-extended rows, `WHERE` after the join, derived-source and subquery nesting, schemaless tables, an alias reused across the left prefix, and index/primary-key regressions.
- `cargo fmt --check` clean, `clippy --all-targets --all-features` warning-free, 3174 tests across 95 binaries passing.

## Summary by CodeRabbit

* **New Features**
  * Added support for SQL `RIGHT JOIN` and `RIGHT OUTER JOIN`.
  * Preserves unmatched right-side rows with `NULL` values for missing left-side data.
  * Supports right joins with conditions, filtering, aggregation, subqueries, aliases, chaining, and indexed queries.
* **Bug Fixes**
  * Corrected full-scan handling and index selection for right joins.
* **Documentation**
  * Added RIGHT JOIN syntax guidance, examples, and NULL-propagation details.
* **Tests**
  * Added comprehensive coverage for execution, planning, indexing, schemaless queries, and error handling.

### DIFF
--- a/core/src/ast/query.rs
+++ b/core/src/ast/query.rs
@@ -97,6 +97,7 @@ pub struct Join {
 pub enum JoinOperator {
     Inner(JoinConstraint),
     LeftOuter(JoinConstraint),
+    RightOuter(JoinConstraint),
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
@@ -444,6 +445,7 @@ impl Join {
         let (join_operator, join_constraint) = match join_operator {
             JoinOperator::Inner(join_constraint) => ("INNER JOIN", join_constraint),
             JoinOperator::LeftOuter(join_constraint) => ("LEFT OUTER JOIN", join_constraint),
+            JoinOperator::RightOuter(join_constraint) => ("RIGHT OUTER JOIN", join_constraint),
         };
 
         let join_constraint = if quoted {
@@ -1134,6 +1136,30 @@ mod tests {
         }
         .to_sql();
         assert_eq!(actual, expected);
+
+        let actual = r#"RIGHT OUTER JOIN "PlayerItem""#;
+        let expected = Join {
+            relation: TableFactor::Table {
+                name: "PlayerItem".to_owned(),
+                alias: None,
+            },
+            join_operator: JoinOperator::RightOuter(JoinConstraint::None),
+        }
+        .to_sql();
+        assert_eq!(actual, expected);
+
+        let actual = r#"RIGHT OUTER JOIN "PlayerItem" ON "PlayerItem"."user_id" = "Player"."id""#;
+        let expected = Join {
+            relation: TableFactor::Table {
+                name: "PlayerItem".to_owned(),
+                alias: None,
+            },
+            join_operator: JoinOperator::RightOuter(JoinConstraint::On(expr(
+                r#""PlayerItem"."user_id" = "Player"."id""#,
+            ))),
+        }
+        .to_sql();
+        assert_eq!(actual, expected);
     }
 
     #[test]
@@ -1194,6 +1220,30 @@ mod tests {
             },
             join_operator: JoinOperator::LeftOuter(JoinConstraint::On(expr(
                 "PlayerItem.age > Player.age AND PlayerItem.user_id = Player.id AND PlayerItem.amount > 10 AND PlayerItem.amount * 3 <= 2",
+            ))),
+        }
+        .to_sql_unquoted();
+        assert_eq!(actual, expected);
+
+        let actual = "RIGHT OUTER JOIN PlayerItem";
+        let expected = Join {
+            relation: TableFactor::Table {
+                name: "PlayerItem".to_owned(),
+                alias: None,
+            },
+            join_operator: JoinOperator::RightOuter(JoinConstraint::None),
+        }
+        .to_sql_unquoted();
+        assert_eq!(actual, expected);
+
+        let actual = "RIGHT OUTER JOIN PlayerItem ON PlayerItem.user_id = Player.id";
+        let expected = Join {
+            relation: TableFactor::Table {
+                name: "PlayerItem".to_owned(),
+                alias: None,
+            },
+            join_operator: JoinOperator::RightOuter(JoinConstraint::On(expr(
+                "PlayerItem.user_id = Player.id",
             ))),
         }
         .to_sql_unquoted();
--- a/core/src/executor/alter/table.rs
+++ b/core/src/executor/alter/table.rs
@@ -225,10 +225,15 @@ fn source_for_schema_copy(project: &ProjectPlan) -> Option<&SourcePlan> {
         ProjectInputPlan::Source(relation) => relation,
         ProjectInputPlan::Filter(filter) => match &filter.input {
             FilterInputPlan::Source(relation) => relation,
-            FilterInputPlan::InnerJoin(_) | FilterInputPlan::LeftOuterJoin(_) => return None,
+            FilterInputPlan::InnerJoin(_)
+            | FilterInputPlan::LeftOuterJoin(_)
+            | FilterInputPlan::UnplannedRightOuterJoin(_)
+            | FilterInputPlan::RightOuterJoin(_) => return None,
         },
         ProjectInputPlan::InnerJoin(_)
         | ProjectInputPlan::LeftOuterJoin(_)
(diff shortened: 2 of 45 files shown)


MinhyukWoo:
Closing this in favour of a different approach, following the suggestion to normalize RIGHT JOIN into LEFT JOIN in the planner rather than adding a dedicated executor. I agree that is the better endpoint — it is what MySQL does, and it would leave GlueSQL with no right-join-specific executor at all.

Two things in GlueSQL block that today. Both are independent of RIGHT JOIN, so I will raise them as separate issues rather than smuggling them into a feature PR.

**1. Output column order is decided at execution, not in the plan.** `SELECT *` survives planning as `projection: SelectItems([Wildcard])` and is expanded by `project::labels::resolve` from the runtime `SelectedSources`. So swapping join inputs changes the output column order, and several consumers read columns by position:

```sql
CREATE TABLE Dest (a INTEGER, b TEXT);
INSERT INTO Dest SELECT * FROM A RIGHT JOIN B ON TRUE;   -- ok
INSERT INTO Dest SELECT * FROM B LEFT JOIN A ON TRUE;    -- Err(IncompatibleDataType)
```

The two queries are logically equivalent. Here the types differ so it errors; with matching types it would store swapped data silently. `CREATE TABLE AS SELECT *` and `ORDER BY <ordinal>` behave the same way.

**2. Join plans are left-deep.** `NestedLoopJoinPlan.right` is a `SourcePlan`, not a join, so `C LEFT JOIN (A ⋈ B)` cannot be represented. The grouping cannot be flattened away — with `A = {1,2}`, `B = {1}`, `C = {2}`:

```sql
SELECT A.a, B.b, C.c FROM A JOIN B ON A.a = B.b RIGHT JOIN C ON C.c = A.a;
→ [[NULL, NULL, 2]]     -- c=2 matches no row of A⋈B, so A and B are both NULL

SELECT A.a, B.b, C.c FROM C LEFT JOIN A ON C.c = A.a LEFT JOIN B ON A.a = B.b;
→ [[2, NULL, 2]]        -- wrong: a=2 survives because C meets A before B is consulted
```

Wrapping the left side in a derived table gives the right rows but hides the inner aliases, so `ON C.c = A.a` then fails with `CompoundIdentifierNotFound`.

Once both are in place, RIGHT JOIN reduces to swapping the children during translation, and everything this PR added becomes unnecessary — the plan types, the ~320-line executor with its buffering and position tracking, and `TableAccessPlan::FullScanRequired`.

That last one is worth noting, because normalization removes the problem rather than working around it. The marker existed because RIGHT JOIN puts the **non-preserved** side at the base source, which is exactly where the primary key and index planners narrow. After normalization the base is the **preserved** side, and narrowing the preserved side by a predicate on itself is equivalent to filtering after the join. Equi-joins also stop being quadratic, since a normalized RIGHT JOIN is an ordinary LEFT JOIN and picks up hash promotion for free.

#63 stays open. The branch remains at `MinhyukWoo:feature/right-outer-join` if any of it is useful later — the SQL fixtures in particular should carry over, since they encode the semantics regardless of how the join is executed.

Thanks for steering this before it landed.

PR #2001, opened 2026-08-17
Add PostgreSQL-style regex operators

### LINKED ISSUE #1997: Decide the initial regular-expression matching interface
## Motivation

GlueSQL supports `LIKE` and `ILIKE`, but it does not support regular-expression
predicates or functions.

The feature tracker #1684 records `(NOT) REGEXP`, `REGEXP_LIKE`, and
`REGEXP_REPLACE` as unsupported. Before implementing one of them, we should
choose the initial public interface and compatibility contract.

The current parser uses `PostgreSqlDialect`. Its upstream AST already represents
PostgreSQL regex operators (`~`, `~*`, `!~`, `!~*`), but GlueSQL currently
rejects them during translation.

## Decision requested

Which interface should GlueSQL support first?

1. PostgreSQL predicate operators: `~`, `~*`, `!~`, `!~*`
2. SQL-standard `SIMILAR TO`
3. Function APIs such as `REGEXP_LIKE`
4. MySQL-compatible `REGEXP` / `RLIKE`

## Proposed initial scope

If predicate operators are preferred, keep the first PR limited to matching.

```sql
name ~ '^[A-Z]+$'
name ~* '^alice$'
name !~ 'admin'
name !~* 'admin'
```

- Support string operands only.
- Preserve the existing NULL-expression policy.
- Return a clear error for an invalid pattern.
- Document the supported regex dialect explicitly.
- Add fixtures for positive, negative, case-insensitive, NULL, non-string,
  and invalid-pattern cases.

## Out of scope

- Capture/extraction APIs such as `regexp_match`
- Replacement/count/position APIs such as `regexp_replace`, `regexp_count`,
  and `regexp_substr`
- `SIMILAR TO` and `LIKE ... ESCAPE` semantics
- MySQL aliases unless they are selected as the initial compatibility target

## Expected outcome

The first contribution establishes one unambiguous regex contract without
claiming cross-database compatibility that the chosen regex engine cannot
provide. Follow-up issues can add functions or additional dialect syntax.

## Context

- Feature tracker: #1684
- Existing pattern-matching precedent: #251, #314
- Open `LIKE ... ESCAPE` work: #1011

red-sprout:
## Summary

- Add PostgreSQL-style regular-expression operators: `~`, `~*`, `!~`, and `!~*`
- Translate parser-specific PG operators into a dedicated GlueSQL regex expression without extending `BinaryOperator`
- Support regex expressions through AST, plan, planner traversal, executor, and query builder APIs
- Use Rust `regex` syntax; preserve `NULL`, report invalid patterns, and reject non-string operands
- Add SQL fixtures and unit/query-builder coverage for matching, negation, case-insensitivity, errors, and `NULL`

## Testing

- `cargo fmt --all -- --check`
- `git diff --check`
- `cargo test -p gluesql-core --lib`
- `cargo test -p gluesql_memory_storage regex_operators --all-features`
- changed-line coverage: 100%

`cargo clippy --all-targets -- -D warnings` is blocked by pre-existing, unrelated lint failures in `macros/src/lib.rs`, `core/src/executor/evaluate/evaluated/eq.rs`, and `core/src/planner/expr/plan_expr/function.rs`.

Closes #1997

## Summary by CodeRabbit

* **New Features**
  * Added regular-expression filtering with case-sensitive and case-insensitive matching.
  * Added positive and negated regex operators.
  * Added query-builder methods for regex filters.
  * Regex matching supports `NULL` values and provides clear errors for invalid patterns or non-text operands.

* **Bug Fixes**
  * Improved handling and diagnostics for invalid regular-expression patterns.

### DIFF
--- a/test-suite/fixtures/regex.sql
+++ b/test-suite/fixtures/regex.sql
@@ -27,13 +27,41 @@ SELECT name FROM RegexItem WHERE name ~* '^a|g';
 
 VALUES (1 ~ 'x');
 -- @expect: error Value.RegexOnNonString
+-- @json:
+-- {
+--   "base": {
+--     "I64": 1
+--   },
+--   "operator": "~",
+--   "pattern": {
+--     "Str": "x"
+--   }
+-- }
 
 VALUES (1 !~* 'x');
 -- @expect: error Value.RegexOnNonString
+-- @json:
+-- {
+--   "base": {
+--     "I64": 1
+--   },
+--   "operator": "!~*",
+--   "pattern": {
+--     "Str": "x"
+--   }
+-- }
 
 -- a non-text operand is rejected even when the other side is NULL
 VALUES (1 ~ NULL);
 -- @expect: error Value.RegexOnNonString
+-- @json:
+-- {
+--   "base": {
+--     "I64": 1
+--   },
+--   "operator": "~",
+--   "pattern": "Null"
+-- }
 
 VALUES ('x' ~ '[');
 -- @expect: error StringExt.InvalidRegexPattern
--- a/test-suite/src/lib.rs
+++ b/test-suite/src/lib.rs
@@ -191,6 +191,7 @@ macro_rules! generate_store_tests {
         sql_case!(default);
         sql_case!(limit);
         sql_case!(like_ilike);
+        sql_case!(regex_operators);
         sql_case!(filter);
         sql_case!(inline_view);
         sql_case!(values);
--- a/test-suite/src/lib.rs
+++ b/test-suite/src/lib.rs
@@ -191,7 +191,7 @@ macro_rules! generate_store_tests {
         sql_case!(default);
         sql_case!(limit);
         sql_case!(like_ilike);
-        sql_case!(regex_operators);
+        sql_case!(regex);
         sql_case!(filter);
         sql_case!(inline_view);
         sql_case!(values);
--- a/core/src/ast/expr.rs
+++ b/core/src/ast/expr.rs
@@ -1,7 +1,7 @@
 use {
     super::{
         Aggregate, BinaryOperator, DataType, DateTimeField, Function, Literal, Query, ToSql,
-        ToSqlUnquoted, UnaryOperator,
+        ToSqlUnquoted, UnaryOperator, regex_operator,
     },
     crate::data::Value,
     serde::{Deserialize, Serialize},
@@ -42,6 +42,12 @@ pub enum Expr {
         expr: Box<Expr>,
         negated: bool,
         pattern: Box<Expr>,
+    },
+    Regex {
+        expr: Box<Expr>,
+        negated: bool,
+        pattern: Box<Expr>,
+        case_sensitive: bool,
     },
     BinaryOp {
         left: Box<Expr>,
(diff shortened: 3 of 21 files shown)


test-suite/src/lib.rs
```diff
@@ -192,6 +191,7 @@ macro_rules! generate_store_tests {
         sql_case!(default);
         sql_case!(limit);
         sql_case!(like_ilike);
+        sql_case!(regex_operators);
```
panarch: Could we name this just `regex`?

Fixture naming here goes by the thing itself, without a suffix — `expr/between.sql`,
`expr/concat.sql`, `expr/bitwise_and.sql`, `expr/in_list.sql`, `expr/arrow.sql`, and
`like_ilike.sql` right next to this one. `regex_operators.sql` is the only fixture in
the tree carrying an `_operators` suffix.

`expr/unary_operator.sql` is the one exception, but that name exists because `-`, `+`,
`!`, `~` have nothing in common but "unary". Here all four operators *are* regex, so
`regex` names them exactly.

There's also nothing to disambiguate against: there are no regex functions in the
codebase today, and if `REGEXP_REPLACE` and friends land later they'd live under
`function/`, so they wouldn't collide with a top-level `regex`.

    test-suite/fixtures/regex_operators.sql -> test-suite/fixtures/regex.sql
    sql_case!(regex_operators);             -> sql_case!(regex);

test-suite/fixtures/regex.sql
```diff
@@ -0,0 +1,42 @@
...
+
+VALUES (1 ~ 'x');
+-- @expect: error Value.RegexOnNonString
+
+VALUES (1 !~* 'x');
+-- @expect: error Value.RegexOnNonString
+
+-- a non-text operand is rejected even when the other side is NULL
+VALUES (1 ~ NULL);
+-- @expect: error Value.RegexOnNonString
```
panarch, line 56: Could the three `Value.RegexOnNonString` cases pin their payload with `@json`?

Without a `@json` body, `assert_serialized` (test-suite/src/fixture.rs:323) only calls
`json_has_path` — it checks that the variant path exists and nothing else, so `base`,
`pattern`, and `operator` all go unverified. With a body it does a full `pretty_assert_eq!`
on the serialized error.

`like_ilike.sql` pins all 4 of its error cases this way, and across the fixtures there are
522 `@expect: error` lines against 554 `@json:` blocks, so this is the norm.

I tried the three blocks below locally and `cargo test -p gluesql_memory_storage` stays
green at 198 passed:

    VALUES (1 ~ 'x');
    -- @expect: error Value.RegexOnNonString
    -- @json:
    -- {
    --   "base": {
    --     "I64": 1
    --   },
    --   "operator": "~",
    --   "pattern": {
    --     "Str": "x"
    --   }
    -- }

    VALUES (1 !~* 'x');
    -- @expect: error Value.RegexOnNonString
    -- @json:
    -- {
    --   "base": {
    --     "I64": 1
    --   },
    --   "operator": "!~*",
    --   "pattern": {
    --     "Str": "x"
    --   }
    -- }

    VALUES (1 ~ NULL);
    -- @expect: error Value.RegexOnNonString
    -- @json:
    -- {
    --   "base": {
    --     "I64": 1
    --   },
    --   "operator": "~",
    --   "pattern": "Null"
    -- }

Note `Value::Null` serializes as the bare string `"Null"`, not an object.

The two `StringExt.InvalidRegexPattern` cases are better left as they are — their `error`
field carries the `regex` crate's multi-line parse message, which would tie the fixture to
that crate's wording.

red-sprout:
I opened #2005 to track compiled-regex caching as a follow-up, so this issue can stay focused on the initial matching interface.

Could you let me know whether the current implementation is ready to merge as-is, or if there is anything you would like me to address first?

panarch (changes requested):

panarch (approved):
Looks all nice!
Thanks a lot 👍 👍 👍

PR #2013, opened 2026-08-24
Deprecate Sled storage

panarch:
## Summary
- mark `SledStorage` deprecated in v0.20.0 and announce its removal in v0.21.0
- guide new persistent-storage users and general examples toward `RedbStorage`
- switch the CLI SQL dump path and composite-storage coverage from Sled to Redb
- retain Sled-specific examples and tests for the deprecation period with targeted warning allowances

## Summary by CodeRabbit

* **Deprecation**
  * Sled Storage is deprecated as of v0.20.0 and scheduled for removal in v0.21.0.
  * Existing Sled deployments receive warnings; new persistent deployments should use Redb Storage.

* **Improvements**
  * Database dumps continue to support Sled Storage with deprecation warnings.
  * Migration guidance now recommends importing data into Redb Storage.

* **Documentation**
  * CLI, Rust, storage, and migration guidance now include Redb examples and updated Sled information.

* **Examples**
  * Standard Rust examples and integration coverage now demonstrate Redb Storage.


cli/src/lib.rs
```diff
@@ -94,6 +102,11 @@ pub fn run() -> Result<()> {
     let path = path.as_deref();
 
     match (path, storage, dump) {
+        (Some(path), None, Some(dump_path)) => {
+            let mut storage = RedbStorage::new(path).expect("failed to load redb-storage");
+
+            dump_database(&mut storage, dump_path)?;
+        }
```
devgony, line 110: --dump can no longer dump a sled database — documented command silently changed semantics                                           
 
gluesql --path <sled_dir> --dump out.sql (the documented invocation) now runs Database::open on a sled directory → panic via .expect("failed to load             
   redb-storage"). If the path doesn't exist, redb creates an empty DB → writes an empty dump.                                                                      
 
--storage sled --dump falls to the REPL arm at line 118, ignoring --dump. No invocation can dump sled anymore — during the exact deprecation window whose stated 
   purpose ("existing deployments can continue using it") is migrating off sled. Dump→import is the sled→redb migration route.                                      
 
Fix: genericize dump_database<T: Store + Transaction> (both traits already imported at line 16), add a sled dump arm with warn_sled_storage_deprecated(), widen  
   the redb arm to None | Some(Storage::Redb).

cli/src/lib.rs
```diff
@@ -163,10 +172,19 @@ pub fn run() -> Result<()> {
     Ok(())
 }
 
-pub fn dump_database(storage: &mut SledStorage, dump_path: PathBuf) -> Result<()> {
-    let file = File::create(dump_path)?;
+pub fn dump_database(storage: &mut RedbStorage, dump_path: PathBuf) -> Result<()> {
+    storage.begin(false)?;
```
devgony: `begin(false)` breaks dump streaming by making `scan_data` eagerly collect the entire table into memory. Use `begin(true)` to preserve lazy iteration and bounded memory usage.

pkg/rust/tests/glue.rs
```diff
@@ -59,16 +60,19 @@ fn basic<T: GStore + GStoreMut + Planner>(mut glue: Glue<T>) {
     );
 }
 
-#[cfg(feature = "gluesql_sled_storage")]
+#[cfg(feature = "gluesql-redb-storage")]
```
devgony, line 63: This feature isn't in the fast CI gate: .github/workflows/rust.yml:51 still passes `--features "gluesql_memory_storage gluesql_sled_storage`", so redb_basic and the redb-gated examples are silently cfg'd out there. 
Please add gluesql-redb-storage to that list.

devgony (approved):
👍🏼 👍🏼 👍🏼

PR #2014, opened 2026-08-30
Stage the file-storage migration instead of rewriting in place

### LINKED ISSUE #2004: file-storage migration applies partially and leaves no trace when it is interrupted
## Summary
`gluesql_file_storage::migrate_to_latest` rewrites row files while it is still scanning the table.

In `storages/file-storage/src/migration/v1_to_v2.rs`, `migrate_rows` reads each row file, converts it if needed, and immediately writes it back with `write_file_atomically`. The table schema file is updated only **after all row files for that table have been processed**.

This means a migration can **fail after some rows have already been rewritten**, while the table schema still has the old format marker state. There is also no marker on disk saying that a migration was in progress, so an interrupted migration looks the same as a storage that simply still needs migration.

This is not necessarily data loss, because re-running the migration can converge. But it makes the failure mode harder to reason about and leaves no clear trace of what happened.

## Current behavior
The relevant flow looks like this.
```rust
// storages/file-storage/src/migration/v1_to_v2.rs

for row_path in row_paths {
    let data = fs::read_to_string(&row_path)?;

    if ron::from_str::<FileRow>(&data).is_ok() {
        continue;
    }

    let row = /* parse v1 shape, or return an error */;
    let serialized = to_string_pretty(&row, PrettyConfig::default())?;

    write_file_atomically(&row_path, &serialized)?;
    report.rewritten_rows += 1;
}
```

For example, suppose a v1 table has three row files and the second one cannot be decoded.

Because row files are processed in sorted path order, the first row may already be rewritten before the second row fails.
```bash
>>> migrate_to_latest(...)
    FAILED: [FileStorage] failed to parse v1 row file '<path>/Foo/02.ron'

  Foo.sql : v1 / no latest format marker
  01.ron  : v2      <-- already rewritten
  02.ron  : invalid
  03.ron  : v1
```

A write-phase interruption can lead to a similar observability problem. For example, a full disk, permission error, or crash can happen after some files have already been replaced. Since there is no migration-in-progress marker, a later open still reports a normal `migration required` state rather than an interrupted migration state.

## Expected behavior
- For failures that can be detected before writing, such as an invalid v1 row file, migration should fail before modifying any row files.
- For failures or interruptions that happen during the write phase, the storage should leave an explicit trace that migration was in progress. Normal `FileStorage::new` should not open a half-migrated storage as if it were just an ordinary old-format storage, and `migrate_to_latest` should be able to resume or finish the migration.

## Why this matters
File storage is made of multiple files, so `write_file_atomically` can only replace one file atomically at a time. It cannot make the whole storage migration atomic across many row files and schema files.

That makes the migration protocol important. Without a planning phase or an in-progress marker, the storage can end up in a mixed state that is technically resumable but not clearly identifiable.

`redb-storage` has a stronger migration boundary here because row rewrites are performed inside a `WriteTransaction` and committed once. `file-storage` does not have that kind of transaction boundary, so it needs a file-level protocol to make interrupted migration states explicit.

## Possible direction
One possible direction is to split migration into two phases.

1. **plan before writing**
    - read all row files that need migration
    - decode them as v1 or recognize them as already-v2
    - serialize the new v2 shape in memory
    - fail before writing if any row cannot be decoded
    
    This would make decode/parse failures leave the storage untouched.
    
2. **add a migration-in-progress marker at the storage root**
    - create the marker before the first row write
    - keep it while row files and schema files are being replaced
    - remove it only after migration finishes
    - make `FileStorage::new` refuse to open while the marker exists
    - make `migrate_to_latest` treat the marker as a resumable migration state
    
    This would not make the whole file-storage migration truly atomic, but it would make interrupted migrations identifiable and resumable.
    

## Scope
file-storage only. sled is excluded given it is planned for removal, and redb already has per-phase
atomicity through its write transaction. No format change to schema or row files.

storage | how it rewrites rows today | in scope
-- | -- | --
file | writes each row file inside the scan loop (storages/file-storage/src/migration/v1_to_v2.rs, migrate_rows) | yes
redb | all row rewrites go into one WriteTransaction, committed once (storages/redb-storage/src/migration.rs, migrate_database_to_latest) | no
sled | every row is decoded and re-encoded into a Vec before the first tree.insert (storages/sled-storage/src/migration/v1_to_v2.rs, migrate_tree) | no, and it is planned for removal

jun02160:
## Summary
`migrate_to_latest` rewrote row files in place while it was still scanning them. 
In `migration/v1_to_v2.rs`, `migrate_rows` read each row file, converted it, and wrote it straight back; the schema marker was updated only after every row of that table had been processed.

So a parse failure part-way through left the table half converted, and nothing on disk said a migration had run. 
An interruption inside `write_file_atomically` could also leave the canonical file missing with only a `.tmp-`/`.bak-` file beside it, which the next run silently skipped.

This replaces the in-place rewrite with a whole-storage staging approach based on the discussion in #2004.

This scenario shows the same invalid v1 row case before and after the change:
gluesql-migration-scenario-a

## How it works now

For a storage at `db/`, migration uses three siblings of the storage root — siblings because the root itself is absent between the two cutover renames, and because a sibling keeps every rename on one filesystem:

1. **Lock** — claim `db.migration-lock` before any change.
2. **Build** — create an empty latest-format `FileStorage` at `db.migrating/` and stream every table into it, one row at a time. `db/` is only read.
3. **Cut over** — rename `db/` to `db.backup/`, then `db.migrating/` to `db/`.
4. **Clean up** — remove `db.backup/`, then release the lock last.

`FileStorage::new` checks the lock **before** `create_dir_all`, so an open during cutover cannot paper over the absent root with a fresh empty directory.

The intended recovery property is that **every expected interruption state keeps one complete authoritative copy on disk**. 
`db/` holds it until the new copy is finished, `db.backup/` holds it across the two-rename window where `db/` does not exist, and the new `db/` holds it afterwards. No step leaves a partially written table visible under the canonical path.
gluesql-migration-protocol
_*the same directory at each step, and which copy is authoritative_

## Recovery

The lock record and the observed layout are read once into a single `MigrationState`. Combinations the protocol cannot produce are rejected rather than assigned a meaning by whichever branch happens to match. With no lock present the sibling paths belong to whoever made them, so a `db.backup/` the user created by hand is refused rather than mistaken for migration debris.

The lock also carries a nonce. Resuming it writes a fresh one, and every rename and removal goes through `lock.rename` / `lock.discard_backup`, which re-check ownership first — so a migration that was taken over stops instead of acting on directories another run is now responsible for. There is no heartbeat and no staleness window; nothing depends on the clock.

lock | on disk | state | action
-- | -- | -- | --
absent | db/ only | Idle | nothing to recover; siblings present without a lock are refused
building | db/ ( + db.migrating/ ) | Building | operator — refuse, naming every path; nothing is removed
ready | db/  • db.migrating/ | ReadyToCutover | operator — refuse, naming every path; nothing is removed
ready | db.backup/ ( + db.migrating/ ), no db/ | SourceMoved | roll a completed staging copy forward, else restore the backup
ready | db/  • db.backup/ | Published | remove the backup, then the lock
ready | db/ only | CleanupPending | release the lock
anything else | any other combination | rejected | operator — refuse; never guess which directory is data

## Pre-cutover states remain manual
The issue comment (https://github.com/gluesql/gluesql/issues/2004#issuecomment-5321887391) lists four interruption states. The ones after cutover are repaired automatically. The two before it — `Building` and `ReadyToCutover`, covering *"Before cutover, `db/` is authoritative and an incomplete `db.migrating/` can be discarded"* — are **reported instead of repaired**.

Discarding that staging copy requires knowing its owner is dead. The standard library has no advisory file lock and no portable process liveness check, and this change adds no dependency. An earlier draft approximated liveness with a heartbeat and a staleness window; review showed that is not merely imprecise, it creates a data-loss path:

1. the verbatim-copy phase has no heartbeat, so a live build's lock goes stale
2. a second process takes over and deletes the live build's staging directory
3. both rebuild into the same path and both reach cutover; the second one renames the already-published new `db/` to `db.backup/` and deletes it

So in those two states migration refuses, prints the full directory state, and names the exact paths to remove:
```bash
[FileStorage] an interrupted migration was found for 'db': storage exists=true,
backup 'db.backup' exists=false, staging 'db.migrating' exists=true. Nothing was
removed and the storage was not modified. Another migration may still be running;
once you have confirmed that none is, remove the staging and backup directories
left by that migration and then 'db.migration-lock', and run the migration again.
```

`db/` is intact and untouched throughout, so the cost is an explicit operator step rather than any risk to data. The two states stay separate even though both currently refuse, since they mark different points in the protocol and could take different explicit recovery policies later. Two ways to lift the restriction, each relaxing a different constraint — happy to take either if preferred:

- add an advisory-lock dependency (`fs2` / `fd-lock`), which gives real liveness
- add an explicit `--recover` flag or `recover_migration()` entry point

## Trade-offs

- Not atomically visible. A reader that opens during cutover is refused rather than served a half-state. The lock is checked at open time only, so a handle opened before the migration started keeps writing, and whatever it writes after its table has been copied is discarded with the backup — the migration assumes no open handles, and the docs say so.
- One extra full traversal, and roughly one extra copy of the storage on disk while it runs.
- Every expected interruption state keeps one complete authoritative copy and a defined next step.

## Preserved

- Same `MigrationReport` meaning. `rewritten_rows` still counts only rows that were actually in the v1 shape, so rows already in the latest shape inside a v1 table are not counted.
- A storage already in the latest format takes the cheap path: no lock, no copy, no cutover, and the same report as before.
- Entries that migration does not interpret — such as a `.git` directory from git-storage or stray user files — survive the cutover.
- A leftover `.tmp-`/`.bak-` file is now a hard error. The canonical file it replaced may be missing.
- Staged files and their directories are synced before cutover, since `StoreMut::insert_data` does not fsync and cutover removes the backup. The directory holding the lock and the two sibling names is flushed after each namespace transition, so a returned `Ok` is durable. Symlinks are recreated, so nothing outside the staged copy is touched.
- Directory permissions are carried over to the staged copy, which `create_dir_all` would otherwise widen to the default mode.

New restrictions worth calling out, each of which refuses and names the problem:

- A symlinked storage root, because cutover renames the root itself and would move the data off its target.
- A `.sql` file whose DDL names a different table than its file name, which would otherwise be re-emitted over that table and drop the file.
- A row whose stored key does not name its file, since staging derives the name from the key and the row would take another row's place.
- A symlinked `.sql` or `.ron`, which the storage reads as data but the migration would copy unconverted.

## Layout of the change

`migration.rs` held the whole thing before. It is now a public surface plus smaller modules, so the recovery decisions are readable on their own.

module | job
-- | --
migration.rs | public surface and the four-phase sequence
migration/paths.rs | sibling path derivation, pre-migration validation, and the directory flush
migration/lock.rs | persisted phase, ownership nonce, and guarded namespace transitions
migration/recovery.rs | derive MigrationState, and execute one transition per state
migration/staging.rs | rejecting a source the migration cannot stage, then building the copy
migration/schema_file.rs | the format marker and version classification
migration/atomic_file.rs | single-file atomic replace, and recognising what it leaves behind
migration/v1_to_v2.rs | the v1 row shapes and their decoder, nothing else

`staging.rs` takes the decoder as an argument rather than calling `v1_to_v2` directly, so the next format step supplies its own and reuses the rest unchanged.

## Tests

`storages/file-storage/tests/migration_v1_to_v2.rs`, plus unit tests in the new modules. The interruption layouts are produced by the real build and rename steps rather than constructed by hand, so recovery is tested against what the code actually leaves behind.

- a later row fails to parse → error, and every source byte is unchanged
- a leftover `.tmp-`/`.bak-` file is detected even when no table needs migrating
- `FileStorage::new` is refused before it creates the canonical path
- both cutover renames, and a leftover lock, recover automatically
- a staging directory beside an intact root is refused and preserved
- a `db.backup/` the user made by hand is never removed
- every reachable (lock, layout) pair maps to one state, and every other pair is rejected
- a `.sql` file whose DDL names another table is refused, and so is a row whose key does not name its file
- a symlinked schema or row is refused rather than copied unconverted
- a backup that cannot be removed still releases the lock, so the migrated storage opens
- directory permissions survive the staged copy
- v1 and v2 tables mixed in one storage, and v2-shaped rows inside a v1 table
- concurrent migrations converge with no row lost
- re-running after a successful migration stays a no-op

## Verification

command | result
-- | --
`cargo clippy --all-targets -- -D warnings` | clean
`cargo fmt -- --check` | clean
`cargo build --all-features` | ok
`cargo test -p gluesql-file-storage` | 252 passed
`cargo test -p gluesql-cli` | 14 passed
`cargo test -p gluesql-git-storage` | 583 passed
whole workspace (excluding mongo / redis, which need servers) | 3220 passed, 0 failed

## Docs

`docs/docs/storages/supported-storages/file-storage.md` now has a `## Storage format migration` section, following the shape of the existing `redb-storage.md` migration docs. ([98c676d](https://github.com/gluesql/gluesql/pull/2014/commits/98c676d4f81ea40a7649507e304156fcdc430fc3))

The section documents both the existing v1 → v2 upgrade path, which was already available but undocumented, and the new recovery behavior added here: the sibling paths created during migration, the lock that blocks opens while migration or recovery is in progress, and what to do if a run is interrupted.

Please review the docs change as well, especially the wording around the two states that are intentionally left as a manual operator step.

Closes #2004

## New Features
- Added migration support for legacy file-storage formats.
- Migrations now use staged processing and atomic replacement to preserve data until completion.
- Migration progress and outcomes support recovery after interruptions.
- Existing handles can continue writing during migration, with post-copy writes retained according to migration rules.

## Bug Fixes
- Improved recovery and cleanup of interrupted migrations.
- Prevented storage access while migration is in progress.
- Added validation for unsupported formats, inconsistent schemas, symbolic links, and unsafe layouts.

## Documentation
- Added migration guidance covering CLI and library workflows, requirements, recovery, and cleanup.


panarch (changes requested):
First, thank you for putting so much care into this migration. The overall safety model is clear: the original storage remains untouched while the replacement is built, cutover happens through two renames, and each expected interruption point retains a complete authoritative copy. The tests around those interruption points are also especially helpful.

While reading the implementation, however, I found the migration state itself harder to follow than the underlying protocol. I think this comes from representing one migration state through two mostly independent models:

- `Layout` describes which of the storage, staging, and backup directories currently exist.
- `LockState` separately describes `LockPhase` and the `owns_staging` / `owns_backup` flags.

Each model is understandable on its own, but recovery has to interpret their Cartesian product. A reader therefore needs to combine the directory layout, phase, and ownership flags mentally before knowing what point of the migration has been reached.

Ignoring the nonce, `LockState` alone can express eight combinations:

```text
phase:          Building | Staged
owns_staging:  false | true
owns_backup:   false | true
```

The normal migration path appears to produce only three of them:

```text
Building, owns_staging=false, owns_backup=false
Building, owns_staging=true,  owns_backup=false
Staged,   owns_staging=true,  owns_backup=true
```

For example, combinations such as these are representable by the types even though the normal protocol does not seem able to produce them:

```text
Building + owns_backup=true
Staged + owns_staging=false
Staged + owns_backup=false
```

The directory layout adds another independent dimension. As far as I can follow the current implementation, the reachable combinations during a normal run are approximately:

| Point in the migration | `Layout` | Persisted lock state |
|---|---|---|
| Lock created | `Settled` | `Building`, owns neither sibling |
| Staging claimed, before its directory exists | `Settled` | `Building`, owns staging |
| Staging is being built | `StagingBesideStorage` | `Building`, owns staging |
| Staging completed, before the first rename | `StagingBesideStorage` | `Staged`, owns staging and backup |
| Original storage moved to backup | `StorageRenamedAway { staging_present: true }` | `Staged`, owns staging and backup |
| Staging published at the canonical path | `BackupBesideStorage` | `Staged`, owns staging and backup |
| Backup removed, before lock release | `Settled` | `Staged`, owns staging and backup |

This also means that the same `Layout` variant can describe very different migration states. For example:

- `Settled + Building` can mean that migration has only just started.
- `Settled + Staged` can mean that migration has completed and only lock cleanup remains.
- `StagingBesideStorage + Building` means that the staging copy may be incomplete.
- `StagingBesideStorage + Staged` means that the staging copy is complete and ready for cutover.

`finish_interrupted` currently reconstructs these meanings through guards over `Layout`, `LockPhase`, and the ownership flags. That works for the combinations it recognizes, but it makes the actual state machine implicit, and it leaves the types able to express many combinations whose meaning is unclear.

I wonder whether the migration could instead expose one semantic state after inspecting the lock record and the filesystem layout. The exact names are only illustrative, but the state space appears to be close to:

| Migration state | Authoritative copy | Meaning |
|---|---|---|
| `Idle` | `db/` | No migration is in progress |
| `Building` | original `db/` | Staging is absent or incomplete |
| `ReadyToCutover` | original `db/` | Staging is complete |
| `SourceMoved` | `db.backup/` | First rename completed; canonical path is absent |
| `Published` | new `db/` | Second rename completed; old backup remains |
| `CleanupPending` | new `db/` | Backup is gone; migration marker remains |

The normal transitions would then be a small, linear sequence:

```text
Idle
  ── create marker and claim staging ──▶ Building
  ── finish and sync staging ──────────▶ ReadyToCutover
  ── rename storage to backup ─────────▶ SourceMoved
  ── rename staging to storage ────────▶ Published
  ── remove backup ────────────────────▶ CleanupPending
  ── release marker ───────────────────▶ Idle
```

Recovery would add only a few explicit edges:

```text
Building / ReadyToCutover
  The original storage is still authoritative.
  While another migration may still be alive, preserve everything and refuse.
  Once an abandoned migration has been confirmed, staging can be discarded
  and the state can return to the original Idle state. An explicit recovery
  policy could also choose to continue from ReadyToCutover.

SourceMoved
  If the completed staging copy is available, publish it and move to Published.
  Otherwise restore the authoritative backup and return to the original Idle
  state.

Published
  Remove the old backup and move to CleanupPending.

CleanupPending
  Release the marker and return to the new Idle state.
```

With that model, ownership also seems derivable from the migration state rather than being an independent pair of booleans:

- `Building` claims staging.
- `ReadyToCutover` and the cutover states claim staging and backup.
- `SourceMoved` identifies backup as authoritative.
- `Published` and `CleanupPending` identify the canonical storage as authoritative.

I am not suggesting that the lock file or directory inspection should disappear. They are necessary persisted evidence after a process interruption. The distinction I am suggesting is that they could be treated as inputs to one inspection step:

```text
lock record + observed filesystem layout
                    ↓
             MigrationState
                    ↓
       normal or recovery transition
```

Combinations that cannot be produced by the protocol could then be rejected as inconsistent instead of being implicitly assigned behavior by whichever `match` guard happens to accept them.

This does not necessarily require changing the current module layout. Even a single combined `MigrationState` at the boundary between inspection and execution could make the implementation easier to enter: normal execution and recovery would use the same states, the authoritative copy would be explicit in each state, and tests could describe the permitted states and edges directly.

Would you consider consolidating the current `Layout × LockState` interpretation in this direction? The proposed names and exact state boundaries are not requirements; the main suggestion is to make the small set of reachable migration states and transitions explicit, rather than representing them as independent layout, phase, and ownership dimensions.

jun02160:
@coderabbitai full review

jun02160:
@panarch Thank you, and sorry for the slow turnaround.
I implemented the state-model refactor in [33b0ebe](https://github.com/gluesql/gluesql/pull/2014/commits/33b0ebee56f9aae13a0ef686b1372cbf98116664).

`Layout` and `LockState` are gone. The lock record and the observed layout are now read once into a single `MigrationState`:

```
lock record + observed layout → MigrationState → normal or recovery transition
```

Once the phase and layout were combined, `owns_staging` and `owns_backup` became redundant. The lock now stores only a phase (building or ready) and the nonce; the phase together with the observed layout identifies the six reachable states:

lock | layout | state
-- | -- | --
absent | db/ only | Idle
building | db/ (staging present or not) | Building
ready | db/ + staging | ReadyToCutover
ready | backup (+ staging), no db/ | SourceMoved
ready | db/ + backup | Published
ready | db/ only | CleanupPending
anything else |   | rejected

`finish_interrupted` now has one arm per state with no match guards. A table-driven test covers every reachable state and representative invalid combinations.

The refactor also exposed an existing test for `building` with the storage root absent. That state cannot be produced by the protocol: the first rename only runs after the staging copy is marked ready. The previous guards accepted it accidentally, so the test now covers a reachable interruption state instead.

`Building` and `ReadyToCutover` currently take the same action —preserve everything and refuse because another migration may still be alive— following the recovery policy you outlined. I kept them separate because they represent different protocol points, and a future explicit recovery policy could safely continue from `ReadyToCutover`.

The state classification, inspection, and recovery transitions remain together in `recovery.rs`; this did not require another module.

Since that refactor, follow-up commits through [76aaa5f](https://github.com/gluesql/gluesql/pull/2014/commits/76aaa5f40cb039261ae6bd1bd5457dadf3267fa4) have addressed staged-copy integrity issues found in a later review pass. Two are behaviour changes worth flagging: a schema file or a row whose name disagrees with what it holds used to overwrite another table or row while reporting success, and a symlinked entry used to be staged as the link rather than as data. Both are now refused. The rest cover directory permissions and the durability of the lock record and the sibling namespace. The PR description has been updated accordingly.

PR #2020, opened 2026-08-30
Add connection pooling to RedisStorage

### LINKED ISSUE #2019: RedisStorage serializes all queries through a single connection
## Summary

`RedisStorage` opens exactly one Redis connection and guards it with a `Mutex`. Every query — from every thread — has to acquire that one lock before it can talk to Redis, so concurrent callers are serialized at the adapter layer even though Redis itself handles concurrent clients fine.

https://github.com/gluesql/gluesql/blob/1e1b15c9/storages/redis-storage/src/lib.rs#L21-L38

```rust
pub struct RedisStorage {
    pub namespace: String,
    pub conn: Mutex<Connection>,
}

impl RedisStorage {
    pub fn new(namespace: &str, url: &str, port: u16) -> Self {
        let redis_url = format!("redis://{url}:{port}");
        let conn = redis::Client::open(redis_url)
            .expect("Invalid connection URL")
            .get_connection()
            .expect("failed to connect to Redis");

        RedisStorage {
            namespace: namespace.to_owned(),
            conn: Mutex::new(conn),
        }
    }
    // ...
}
```

Every `Store`/`StoreMut` method call locks `self.conn` for the duration of its Redis round trip. There is no connection pool and no multiplexed client — just one TCP connection shared behind a `std::sync::Mutex`.

## Current architecture

```mermaid
flowchart TB
    subgraph App["Application process"]
        T1["Thread 1<br/>glue.execute(...)"]
        T2["Thread 2<br/>glue.execute(...)"]
        T3["Thread 3<br/>glue.execute(...)"]
        M["Mutex&lt;Connection&gt;<br/>(RedisStorage.conn)"]
        T1 -->|lock| M
        T2 -->|lock, waits| M
        T3 -->|lock, waits| M
    end
    M -->|"single TCP connection<br/>(one request in flight at a time)"| R[("redis-server")]

    style M fill:#f66,stroke:#900,color:#fff
```

Threads 2 and 3 block on the mutex for the full round trip of thread 1's query, even though `redis-server` itself is perfectly capable of servicing all three concurrently.

## Why this matters

- **Throughput ceiling**: no matter how many threads issue queries against the same `RedisStorage`, Redis I/O never overlaps — the adapter caps concurrency at 1, independent of server-side capacity.
- **Latency amplification**: a slow query (large `SCAN`, big value) blocks every other thread's Redis access for its entire duration, not just its own.
- **Mismatch with the backend**: this is the one persistent-storage backend where the underlying system is explicitly designed for many concurrent lightweight connections, and the adapter throws that away.

## Possible direction

```mermaid
flowchart TB
    subgraph App["Application process"]
        T1["Thread 1<br/>glue.execute(...)"]
        T2["Thread 2<br/>glue.execute(...)"]
        T3["Thread 3<br/>glue.execute(...)"]
        P["ConnectionPool<br/>(e.g. redis::aio ConnectionManager or r2d2-redis)"]
        T1 -->|checkout| P
        T2 -->|checkout| P
        T3 -->|checkout| P
    end
    P -->|conn A| R[("redis-server")]
    P -->|conn B| R
    P -->|conn C| R

    style P fill:#6c6,stroke:#060,color:#fff
```

Replace the single `Mutex<Connection>` with a pool (`r2d2` + `r2d2-redis`, or the `redis` crate's own `ConnectionManager`, which multiplexes over one auto-reconnecting connection). Each call checks out a connection instead of blocking on a single shared one.

Open questions worth resolving in the PR itself:
- Pool sizing / configuration surface on `RedisStorage::new` (fixed pool size vs. configurable).
- Whether `ConnectionManager`'s multiplexing is sufficient on its own, or whether a true multi-connection pool (`r2d2`) is needed for the transaction/pipeline usage this storage already does.
- Backward compatibility of the public `pub conn` field, if it stays public.

## Scope

`redis-storage` only. No format or protocol changes — this is purely about how the adapter manages its connection(s) to `redis-server`.

SteelCrab:
Fixes #2019

## Summary

- `RedisStorage` held a single `Mutex<Connection>`, so every `Store`/`StoreMut` call serialized on that one lock before it could reach `redis-server`, independent of how much concurrency `redis-server` itself could handle
- Replace it with a small self-rolled `ConnectionPool` (`Mutex<Vec<Connection>>` + `Condvar`, no new crate dependency) in a new `pool` module, sized from `std::thread::available_parallelism()`
- `RedisStorage::new(namespace, url, port)`'s public signature is unchanged; only the `pub conn: Mutex<Connection>` field becomes `pub pool: ConnectionPool`
- `RedisStorage` has no explicit multi-statement transactions (`begin` with `autocommit = false` already errors), so each `Store`/`StoreMut` call can safely check out and return its own connection without needing to pin one across calls

## Problem

```rust
pub struct RedisStorage {
    pub namespace: String,
    pub conn: Mutex<Connection>,
}
```

With N threads sharing one `RedisStorage`, Redis I/O never overlapped: every method locked `self.conn` for the full round trip of its Redis call, so a slow query blocked every other thread's Redis access for its entire duration, even though `redis-server` handles many concurrent clients without issue.

## Changes

- New `pool` module: `ConnectionPool::new(&client, size)` pre-connects `size` connections; `checkout()` blocks on a `Condvar` until one is idle and hands back a `PooledConnection` (`Deref`/`DerefMut` to `redis::Connection`) that returns itself to the pool on `Drop`
- `RedisStorage::new` sizes the pool from `std::thread::available_parallelism()` (falls back to 1)
- All 18 `self.conn.lock_err()` call sites across `lib.rs`, `alter_table.rs`, and `metadata.rs` now go through `self.pool.checkout()`
- Removed the now-unused `mutex` module (`MutexExt`)
- Updated the 7 test call sites that reached into `storage.conn` directly (`redis_store.rs`, `redis_alter_table_errors.rs`) to use `storage.pool.checkout()`

## Test plan

- [x] `cargo test -p gluesql-redis-storage --features test-redis` (new `tests/pool.rs` + full existing suite, against a local `redis-server`)
- [x] `cargo clippy -p gluesql-redis-storage --features test-redis --all-targets -- -D warnings`
- [x] `cargo fmt -p gluesql-redis-storage`
- [x] `cargo build --workspace`

## Summary by CodeRabbit

- **New Features**
  - Redis storage now supports concurrent operations through a connection pool.
  - Connections are automatically reused and returned after each operation.
  - Pool capacity adjusts to available system parallelism.

- **Bug Fixes**
  - Improved handling of concurrent Redis requests and temporary connection unavailability.

- **Tests**
  - Added coverage for concurrent connection usage and waiting behavior when all connections are occupied.


panarch (changes requested):
A pool owned by a single `RedisStorage` does not add concurrency inside GlueSQL, and connections closed by the server stay in it. A cloneable `RedisStorage` whose clones share one r2d2 pool would fit better.

### Why the current pool doesn't help

- `Glue::execute` takes `&mut self`, so one `Glue` runs one Redis command at a time and never needs more than one connection at once. Even so, this PR opens `available_parallelism()` connections (14 on a 14-core machine) every time `RedisStorage::new` is called.
- When the server closes a connection, for example because of Redis's `timeout` setting, the pool keeps it and hands it out again. redis-rs does not reconnect, so every command on that connection fails.

Measured with `CONFIG SET timeout 1` and a 50-row table: 4 s of sequential `scan_data`, then concurrent calls on an `Arc<RedisStorage>`.

| Case | Before this PR | This PR |
|---|---|---|
| Connections opened by `RedisStorage::new` | 1 | 14 |
| 8 threads × 20 `scan_data` | 160 ok | 20 ok / 140 errors |
| 8 threads × 50 `fetch_data` on existing rows | 400 found | 22 found / 378 `Ok(None)` |

`fetch_data` returns `Ok(None)` for rows that exist because a failed GET is treated as a missing key (see the last section).

### Suggestion: cloneable `RedisStorage` backed by r2d2

`RedisStorage` could be cloneable, with all clones sharing one connection pool. It would be created once, and each `Glue` would get a clone, the same way `SharedMemoryStorage` is used:

```rust
// default options
let storage = RedisStorage::new("app", "localhost", 6379);

// custom options
let storage = RedisStorage::with_config(
    "app",
    "localhost",
    6379,
    RedisPoolConfig { max_size: 32, ..Default::default() },
)?;

// one Glue per thread or request, all sharing the same connections
let mut glue = Glue::new(storage.clone());
```

This helps when many `Glue` instances are created, such as one per request, or when several namespaces use the same Redis server.

The `redis` crate already supports r2d2 through its `r2d2` feature. r2d2 opens connections only when they are needed, keeps their number within `max_size`, and drops connections that stopped working, so this crate would not need its own pool code:

```toml
redis = { version = "0.32", features = ["r2d2"] }
r2d2 = "0.8" # redis does not re-export r2d2
```

`RedisPoolConfig` would be defined in this crate rather than exposing r2d2 types, so upgrading r2d2 or redis would not change the public API. Some of r2d2's own defaults don't fit here, so the defaults would be:

| Option | Default | Reason |
|---|---|---|
| `max_size` | 10 | same as r2d2 |
| `min_idle` | 1 | r2d2's default equals `max_size`, so `new` would open 10 connections and wait for all of them |
| `connection_timeout` | 5 s | r2d2's default is 30 s, so each call would wait up to 30 s while Redis is down |
| `idle_timeout` | 5 min | r2d2's default is 10 min; idle connections should be closed before the server or a load balancer closes them |

r2d2's PING check before every use (`test_on_check_out`) would be turned off inside the storage. This storage takes a connection for every Redis command, so a PING before each one would double the number of round trips. A connection that stopped working is still dropped after its first failed command.

Each pool also starts 3 background threads. Creating the storage once and cloning it is cheap, but calling `new` for every request would cost more than it does now, which would be worth a note in the docs.

### Before using multiple connections: GET error handling

GET results are read as `String`. For a missing key Redis returns nil, which can't be read as a `String`, so the code treats every error as "key missing". Connection errors are treated the same way:

```rust
let value = redis::cmd("GET").arg(&key).query::<String>(&mut *conn);
if let Ok(value) = value { /* ... */ }
Ok(None) // a connection error also ends up here
```

With one connection this rarely shows up, because an earlier command on the same broken connection fails first. With several connections, a GET can hit a broken connection while earlier commands worked. An INSERT with an existing primary key would then not see the existing row, and would overwrite it instead of failing with a duplicate-key error.

Reading the result as `Option<String>` keeps the two cases apart:

```rust
let value = redis::cmd("GET")
    .arg(&key)
    .query::<Option<String>>(&mut *conn) // missing key -> Ok(None)
    .map_err(/* ... */)?;                 // connection error -> Err
```

The same pattern is in `fetch_all_schemas`, `fetch_schema`, `scan_data`, `scan_table_meta` and `redis_delete_schema`. The pool change depends on this fix, so it would be easier to review as a separate PR that is merged first.

If the intended use case is different, for example sharing one `Arc<RedisStorage>` and calling `Store` methods directly, please share it.

PR #2021, opened 2026-08-30
Add JSON_BUILD_ARRAY support

mors119:
## Summary

Add PostgreSQL-style `JSON_BUILD_ARRAY` support to GlueSQL.

This follows the JSON support direction discussed in #1793 and uses GlueSQL's existing `Value::List` representation instead of introducing a separate JSON data type.

`JSON_BUILD_ARRAY` accepts a variadic list of values, preserves their order, and supports heterogeneous and nested values.

```sql
SELECT JSON_BUILD_ARRAY(1, 'GlueSQL', TRUE, NULL);
```

The result is represented internally as a GlueSQL `LIST`:

```text
[1, "GlueSQL", true, null]
```

## Context

Issue #1793 proposed extending GlueSQL with PostgreSQL-style JSON functions and operators while continuing to use the existing `Value::Map` / `Value::List` representation.

The JSON operator work in #1807 and #1858 follows the same direction for accessing MAP/LIST values.

This PR extends that work to JSON creation by adding `JSON_BUILD_ARRAY` as a small, self-contained first step.

Related to #1793.

## PostgreSQL behavior

The implementation is based on PostgreSQL's `json_build_array` behavior.

PostgreSQL defines:

```text
json_build_array(VARIADIC "any") → json
```

as a function that builds a possibly heterogeneous JSON array from a variadic argument list.

Reference:

https://www.postgresql.org/docs/19/functions-json.html

In PostgreSQL:

```sql
SELECT json_build_array(1, 2, 'foo', 4, 5);
```

produces:

```json
[1, 2, "foo", 4, 5]
```

GlueSQL does not currently have a separate `JSON` / `JSONB` value type, so this PR represents the result using the existing `Value::List`.

This means the function follows PostgreSQL's calling style and value-preservation behavior while using GlueSQL's native value model.

## Changes

- Add `JSON_BUILD_ARRAY` to the GlueSQL function pipeline
- Translate variadic SQL arguments into the corresponding function representation
- Plan and evaluate all arguments in their original order
- Return the evaluated values as `Value::List`
- Preserve `NULL` values as list elements
- Support heterogeneous argument types
- Support nested `LIST` and `MAP` values
- Add SQL fixture coverage
- Add function documentation

## Supported behavior

Empty argument list:

```sql
SELECT JSON_BUILD_ARRAY();
```

returns an empty `LIST`.

Mixed values:

```sql
SELECT JSON_BUILD_ARRAY(1, 'GlueSQL', TRUE, NULL);
```

returns:

```text
[1, "GlueSQL", true, null]
```

Nested values are preserved using GlueSQL's existing `Value` representation rather than being serialized to text.

## Scope

This PR intentionally implements only `JSON_BUILD_ARRAY`.

The following are not included:

- `JSON_BUILD_OBJECT`
- `TO_JSON`
- `ARRAY_TO_JSON`
- a dedicated JSON / JSONB data type
- additional JSON operators
- general planner or storage changes
- Query Builder support; this PR targets SQL syntax only

The planner changes are limited to the plumbing required to represent and evaluate the new function.

A follow-up PR can add `JSON_BUILD_OBJECT` after the function shape and behavior introduced here have been reviewed.

## Tests

Validated with:

```bash
cargo test -p gluesql-core
cargo test -p gluesql_memory_storage --test memory_storage sql_function_json_build_array
cargo fmt --all
```

## Summary by CodeRabbit

* **New Features**
  * Added support for the `JSON_BUILD_ARRAY` SQL function.
  * Supports zero or more heterogeneous values, including `NULL`, nested lists, and maps.
  * Preserves input values and returns a JSON-style array.

* **Documentation**
  * Added syntax, examples, and compatibility guidance for `JSON_BUILD_ARRAY`.

* **Tests**
  * Added coverage for empty arrays, mixed types, null handling, and nested values.


panarch:
Hi @mors119, thanks for the careful work on this. The wiring, tests, and docs are all well done. Sorry your question on #1793 went unanswered before you started.

Strictly speaking, GlueSQL doesn't have a JSON type. What it has are the semi-structured `MAP` and `LIST` types, and a `LIST` can already be built at the expression level with the array literal, giving the same result:

```sql
SELECT JSON_BUILD_ARRAY(1, 'GlueSQL', TRUE, NULL);  -- [1,"GlueSQL",true,null]
SELECT [1, 'GlueSQL', TRUE, NULL];                  -- [1,"GlueSQL",true,null]
```

Given that, the function would mainly serve PostgreSQL name compatibility, which isn't a priority at the moment. That's why this one is being closed.

The lack of a JSON type is also what separates this from `->` and `->>`. Those are expression-level operators for accessing `MAP` and `LIST` values and don't assume a JSON type, whereas a `JSON_*` function implies a JSON value that GlueSQL doesn't have. The same reasoning would apply to `JSON_BUILD_OBJECT` as well.

Thanks again for the contribution!

